PR #6, opened 2025-06-04
Create RequestScheme type to handle registry interactions

adityaramani:
Its a bit weird to have to pass `--http` when pulling / pushing images if we are trying to do so from a local registry

This PR adds a `RequestScheme` enum type which tries to detect if the connection to the registry should be over http or https. This auto detection is the default behavior and can be overridden by passing the `--scheme http/https` flag to the `run` ,`pull`, `push` `login` and `registry default` command

The auto detection works by looking at the hostname / IP. HTTP is used in the following cases
 - If the hostname starts with `localhost` or  is `127.x.x.x`
 - if the host looks like an IP from the private address space
 - If the host ends with the current default dns domain name

Also in this PR:
- Removing some used CLI flag groups and TODOs
- Better error messages when registry credentials are not found for a host

Signed-off-by: Aditya Ramani <a_ramani@apple.com>

### DIFF
--- a/Sources/ContainerClient/RequestScheme.swift
+++ b/Sources/ContainerClient/RequestScheme.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © none Apple Inc. and the container project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the container project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/Tests/ContainerClientTests/RequestSchemeTests.swift
+++ b/Tests/ContainerClientTests/RequestSchemeTests.swift
@@ -1,5 +1,5 @@
 //===----------------------------------------------------------------------===//
-// Copyright © none Apple Inc. and the container project authors. All rights reserved.
+// Copyright © 2025 Apple Inc. and the container project authors. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


egernst:
@adityaramani - can we include the text you have in this PR in the commit itself? I feel like it'd be more useful to see it when reviewing git history (its a good writeup)

adityaramani:
@egernst Yes - thats the default behavior on merge. The PR title + description get added as part of the commit that goes to main

Sources/ContainerClient/Flags.swift
```diff
@@ -33,10 +33,6 @@ public struct Flags {
             help: "Current working directory for the container")
         public var cwd: String?
 
-        // FIXME: Implement.
```
jglogan, line 36: create issue?

Sources/CLI/RunCommand.swift
```diff
@@ -33,18 +33,14 @@ extension Application {
         @OptionGroup
         var processFlags: Flags.Process
 
-        // FIXME: Add in detach keys support.
```
jglogan, line 36: create issue

egernst:
Ah, got it - great.  Thanks Aditya.

Sources/CLI/Container/ContainerStart.swift
```diff
@@ -32,10 +32,6 @@ extension Application {
         @Flag(name: .shortAndLong, help: "Attach container's STDIN")
         var interactive = false
 
-        // FIXME: Add in detach keys support.
```
jglogan, line 35: create issue?

Sources/CLI/Container/ContainerExec.swift
```diff
@@ -29,10 +29,6 @@ extension Application {
         @OptionGroup
         var processFlags: Flags.Process
 
-        // FIXME: Add in detach keys support.
```
jglogan, line 32: create issue?

Sources/ContainerClient/RequestScheme.swift
```diff
@@ -0,0 +1,67 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerizationError
+
+public enum RequestScheme: String, Sendable {
```
jglogan, line 20: missing docc?

Sources/ContainerClient/RequestScheme.swift
```diff
@@ -0,0 +1,67 @@
...
+            self = .https
+        case RequestScheme.auto.rawValue:
+            self = .auto
+        default:
+            throw ContainerizationError(.invalidArgument, message: "Unsupported scheme \(rawValue)")
+        }
+    }
+
+    package func schemeFor(host: String) -> Self {
+        guard host.count > 0 else {
```
jglogan, line 43: is empty host an error?

Sources/ContainerClient/Flags.swift
adityaramani: Done

Sources/ContainerClient/RequestScheme.swift
adityaramani: Added

Sources/ContainerClient/RequestScheme.swift
adityaramani: Fixed

PR #9, opened 2025-06-04
Publish docs github action

katiewasnothere:


.gitignore
```diff
@@ -24,3 +24,6 @@ test_results/
 *.ext4
 *.pkg
 *.swp
+
+# API docs for local preview only.
+_site/
```
dkovba: This was already added in https://github.com/apple/container/pull/9.

.github/workflows/docs-release.yml
```diff
@@ -0,0 +1,37 @@
+# Manual workflow for releasing docs ad-hoc. Workflow can only be run for main or release branches. 
+# Workflow does NOT publish a release of container.
+name: Deploy application website
+on: 
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+  pages: write
```
madrob: can we move the job-specific ones to jobs instead of being global?

katiewasnothere: Updated. I'll need to check that the permissions are correct for the deployment job with a test run after we merge.

PR #23, opened 2025-06-06
Allow printed output to flush to prevent a race between progress updates and other output

dkovba:
This PR also improves clearing the last progress update after moving to a new task using `setDescription()`.

### DIFF
--- a/Sources/TerminalProgress/ProgressBar+Terminal.swift
+++ b/Sources/TerminalProgress/ProgressBar+Terminal.swift
@@ -72,7 +72,7 @@ extension ProgressBar {
 
         // Clears previously printed lines.
         var lines = ""
-        if terminalWidth > 0 {
+        if terminating.hasSuffix("\r") && terminalWidth > 0 {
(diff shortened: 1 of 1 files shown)


Sources/TerminalProgress/ProgressConfig.swift
```diff
@@ -86,6 +89,7 @@ public struct ProgressConfig: Sendable {
     ///   - clearOnFinish: The flag indicating whether to clear the progress bar before reseting the cursor. The default is `true`.
     ///   - disableProgressUpdates: The flag indicating whether to update the progress bar. The default is `false`.
     public init(
+        terminal: FileHandle = .standardError,
```
adityaramani, line 92: Change to `.standardOutput` ? Thats what the doc string mentions

dkovba: Let's update the doc string then! Thank you for catching.

adityaramani:
Thank you for the fixes!

PR #24, opened 2025-06-06
Refine user documentation.

jglogan:
- Extract README sections into BUILDING and CONTRIBUTING documents. This should work better than sections in README as it keeps the README short, and the BUILDING and CONTRIBUTING files will stand out in the file list.
- Merged localdev document into BUILDING.
- Fixed broken links.
- Changed DNS setup so later DNS-dependent steps should work.
- Added h1 and link to SECURITY document.


README.md
```diff
@@ -1,7 +1,7 @@
-# `container`
-
 ![introductory movie showing some basic commands](./docs/assets/landing-movie.gif)
 
+# `container`
```
adityaramani, line 3: Is moving this bellow the gif intentional?

jglogan: yep, sort of an A/B test, we'll see what folks like better

PR #33, opened 2025-06-07
Add make target to fetch the default kernel

adityaramani:
Replaces the `--install-dependencies` flag with the `(enable/disable)-kernel-install` flag. If the flag is not passed - the default behavior is prompting the user for a response.

Also adds a make target `install-kernel`

```
➜ bin/container system start
Installing base container filesystem...
No default kernel configured.
Install the recommended default kernel from [https://github.com/kata-containers/kata-containers/releases/download/3.17.0/kata-static-3.17.0-arm64.tar.xz]? [Y/n]: n
Please use the `container system kernel set --recommended` command to configure the default kernel

➜ bin/container system start --disable-kernel-install

➜ bin/container system start --enable-kernel-install
Installing kernel...

➜ bin/container system start --help

USAGE: container system start ... [--enable-kernel-install] [--disable-kernel-install]
...
  --enable-kernel-install/--disable-kernel-install
                          Specify if the default kernel should be installed or not.

```

### DIFF
--- a/Sources/CLI/System/SystemStart.swift
+++ b/Sources/CLI/System/SystemStart.swift
@@ -34,7 +34,9 @@ extension Application {
         @Flag(name: .long, help: "Enable debug logging for the runtime daemon.")
         var debug = false
 
-        @Flag(name: .long, inversion: .prefixedEnableDisable, help: "Specify if the default kernel should be installed or not.")
+        @Flag(
+            name: .long, inversion: .prefixedEnableDisable,
+            help: "Specify whether the default kernel should be installed or not. The default behavior is to prompt the user for a response.")
         var kernelInstall: Bool?
 
         func run() async throws {


egernst:
I feel like if the user specifies the fetch kernel flag they are effectively saying yes. Do you think it’s necessary to take prompt?

adityaramani:
Well, today there was no way for a headless user to say that they didnt want the default kernel to be configured on system start. And using of the same `--fetch-default-kernel` option for this made sense. 

Since we need to provide 3 possible options, we couldnt treat the parameter as a boolean flag - which is the reason for needing the 'yes' or 'no'

adityaramani:
Updated the behavior to switch back to using a flag that defaults to a nil value - which uses the interactive mode. The description reflects the new behavior

Sources/CLI/System/SystemStart.swift
```diff
@@ -34,8 +34,8 @@ extension Application {
         @Flag(name: .long, help: "Enable debug logging for the runtime daemon.")
         var debug = false
 
-        @Flag(name: .long, help: "Do not prompt for confirmation before installing runtime dependencies")
-        var installDependencies: Bool = false
+        @Flag(name: .long, inversion: .prefixedEnableDisable, help: "Specify if the default kernel should be installed or not.")
```
jglogan: grammar nit: Specify whether

Sources/CLI/System/SystemStart.swift
```diff
@@ -110,16 +110,23 @@ extension Application {
             let defaultKernelURL = kernelDependency.source
             let defaultKernelBinaryPath = ClientDefaults.get(key: .defaultKernelBinaryPath)
 
-            print("No default kernel configured.")
-            print("Install the recommended default kernel from [\(kernelDependency.source)]? [Y/n]: ", terminator: "")
-            if !installDependencies {
+            var shouldInstallKernel = false
```
jglogan, line 115: a discussion for a follow-up maybe:

Better to assume `--enable-kernel-install` or `--disable-kernel-install` if `system start` isn't run from an interactive terminal?

My feeling is that that is too magical. It's simpler to have it just as it is now, with yes/no/default to interactive, and that the following sort of usage is accepted (though flags would be the recommended approach):

```bash
echo y | container system start
```

Sources/CLI/System/SystemStart.swift
jglogan: also should we mention that the default is interactive/prompt?

adityaramani: Updated

PR #39, opened 2025-06-08
Use bookworm in tutorial, disable console block copy.

jglogan:
- Switched to use bookworm Python image as `curl` is not in slim.
- Use `pre` tags for console blocks. We lose syntax highlighting, but get rid of the copy button.
- Fix `dockerfile` tag in tutorial.
- Fix GitHub typo in README.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -231,7 +231,11 @@ You will often see these two options abbreviated and specified together as `-ti`
 
 ### Access the web server from another container
 
-Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `wget` command to retrieve the `index.html` content from the first container.
+Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `curl` command to retrieve the `index.html` content from the first container.
+
+```bash
+container run -it --rm web-test curl http://192.168.64.3
+```
 
 <pre>
 % container run -it --rm web-test curl http://192.168.64.3


mavenugo:
Please refer to https://github.com/apple/container/pull/40 where I used the \<pre\> tag to replace the code blocks to avoid the copy-to-clipboard... 

I still see the copy-to-clipboard option with console tag.

docs/tutorial.md
```diff
@@ -204,43 +204,43 @@ open http://my-web-server.test
...
 ### Access the web server from another container
 
-Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `curl` command to retrieve the `index.html` content from the first container.
+Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `wget` command to retrieve the `index.html` content from the first container.
 
-```shellsession
-% container run -it --rm web-test curl http://192.168.64.3
-<!DOCTYPE html><html><head><title>Hello</title></head><body><p><img src="logo.jpg" style="width: 2rem; height: 2rem;">Hello, world!</p></body></html>
+<pre>
+% container run -it --rm web-test wget -q -O - http://192.168.64.3
```
mavenugo: when I tried the web-test image, I dont see `wget` either.

egernst:
...testing now...

docs/tutorial.md
```diff
@@ -271,11 +273,12 @@ container images push registry.example.com/fido/web-test:latest
 
 ### Pull and run your image
 
-To validate your published image, remove your existing web server image, and then run using the remote image:
+To validate your published image, stop your current web server container, remove the image that you built, and then run using the remote image:
 
 ```bash
```
mavenugo, line 282: This should be changed to \<pre\> as well. since "registry.example.com" and "fido" are all sample ones, there is no use for the users to copy them in clipboard. Also, this is a multi line command... copy them and pasting will result in failure and that isnt useful either.

docs/tutorial.md
```diff
@@ -204,38 +206,38 @@ open http://my-web-server.test
...
+</pre>
 
 The `--tty` and `--interactive` flag allow you to interact with the shell from your host terminal. The `--tty` flag tells the shell in the container that its input is a terminal device, and the `--interactive` flag connects what you input in your host terminal to the input of the shell in the container.
 
 You will often see these two options abbreviated and specified together as `-ti` or `-it`.
 
 ### Access the web server from another container
 
-Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `curl` command to retrieve the `index.html` content from the first container.
+Your web server is accessible from other containers as well as from your host. Launch a second container using your `web-test` image, and this time, specify a `wget` command to retrieve the `index.html` content from the first container.
```
mavenugo, line 239: I think it is useful to have another bash with just `container run -it --rm web-test curl http://192.168.64.3` for easy copy-paste for the user. (we have it just below for `test` domain below).

docs/tutorial.md
jglogan: copy/paste multiple lines should work as long as there's no interactive input on any command

jglogan: It seems odd to me to use console format for something that's not a terminal session - prompts, input, and output.

Users may want to copy/paste and then edit the command line, as opposed to typing everything in. zsh and just about any modern shell allows editing multi-line commands.

docs/tutorial.md
jglogan: done

PR #43, opened 2025-06-09
docs: improve instructions for building with local containerization

egernst:
If they're just pulling containerization, suggest following build env setup directions from containerization repo so they get a cross-platform SDK setup.


egernst:
Depends on https://github.com/apple/containerization/pull/41 - marking as draft for now

PR #118, opened 2025-06-10
Add new `system status` command

### LINKED ISSUE #117: Add command to check system status
Container services can be started and stopped using the `system start` and `system stop` commands. But there is no straightforward way for users to check if the services were already started.

It would be nice to have a `system status` command that shows if the services are running or not.

gonzolino:
Container services can be started and stopped using the `system start` and `system stop` commands. But there is no straightforward way for users to check if the services were already started.

This PR adds a new `system status` command that shows wether the container services are running or not.

The command only checks the status of the apiserver for now, but it could be expanded in the future to show more details.

Fixes #117.

### DIFF
--- a/Sources/CLI/System/SystemStatus.swift
+++ b/Sources/CLI/System/SystemStatus.swift
@@ -1,0 +1,52 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the container project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerClient
+import ContainerPlugin
+import ContainerizationError
+import Foundation
+import Logging
+
+extension Application {
+    struct SystemStatus: AsyncParsableCommand {
+        static let configuration = CommandConfiguration(
+            commandName: "status",
+            abstract: "Show the status of `container` services"
+        )
+
+        @Option(name: .shortAndLong, help: "Launchd prefix for `container` services")
+        var prefix: String = "com.apple.container."
+
+        func run() async throws {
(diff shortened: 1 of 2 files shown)


Sources/CLI/System/SystemStatus.swift
```diff
@@ -0,0 +1,51 @@
...
+            abstract: "Show the status of `container` services"
+        )
+
+        @Option(name: .shortAndLong, help: "Launchd prefix for `container` services")
+        var prefix: String = "com.apple.container."
+
+        func run() async throws {
+            let isRegistered = try ServiceManager.isRegistered(fullServiceLabel: "\(prefix)apiserver")
+            if !isRegistered {
+                print("apiserver is not running")
```
katiewasnothere: Consider changing this wording to
```suggestion
                print("apiserver is not registered with launchd")
```

katiewasnothere:
Thanks for the change! :)

Sources/CLI/System/SystemStatus.swift
gonzolino: Thanks for the suggestion! It is definitely more precise than my wording.

Updated it to "apiserver is not running and not registered with launchd", as I think end users might not know what "registered with launchd" means.

Sources/CLI/System/SystemStatus.swift
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+        @Option(name: .shortAndLong, help: "Launchd prefix for `container` services")
+        var prefix: String = "com.apple.container."
+
+        func run() async throws {
+            let isRegistered = try ServiceManager.isRegistered(fullServiceLabel: "\(prefix)apiserver")
+            if !isRegistered {
+                print("apiserver is not running and not registered with launchd")
+                return
```
adityaramani: I think we should call `Application.Exit` with an error code here. So that the command exits with a non zero status?

Sources/CLI/System/SystemStatus.swift
```diff
@@ -0,0 +1,51 @@
...
+            }
+
+            // Now ping our friendly daemon. Fail after 10 seconds with no response.
+            do {
+                print("Verifying apiserver is running...")
+                try await ClientHealthCheck.ping(timeout: .seconds(10))
+                print("apiserver is running")
+            } catch {
+                print("apiserver is not running")
```
adityaramani, line 47: Same as above - we should exit out with a zero error code  to signify something went wrong

adityaramani (changes requested):
Thanks for the PR! Just had a couple minor comments

Sources/CLI/System/SystemStatus.swift
gonzolino: Good idea! I changed both occurrences to exit with status code 1.

adityaramani (approved):

PR #143, opened 2025-06-11
Fix missing HOME environment variable for root user

### LINKED ISSUE #108: Why isn't `HOME` set?
I noticed something unusual for containers. When running a container, `$HOME` is not set. Is this expected?

```console
% container run --rm -it fedora 
[root@14ea7d31-b047-42a7-8940-217909861924 /]# cd
bash: cd: HOME not set
[root@14ea7d31-b047-42a7-8940-217909861924 /]# 
exit
% container run --rm -it ubuntu
root@940bfeb9-ebec-4492-b5ac-5739273f662a:/# cd
bash: cd: HOME not set
root@940bfeb9-ebec-4492-b5ac-5739273f662a:/# 
exit
```

nandsha:
## Summary

  Fix containers failing with "bash: cd: HOME not set" error when running commands that depend on the HOME environment variable.

  ## Problem

  Containers running as root user without an explicitly set HOME environment variable would fail when executing commands like `cd` with the error:
  bash: cd: HOME not set

  This occurred because the container runtime did not provide a default HOME environment variable for the root user when none was specified.

  ## Solution

  Add logic to `modifyingEnvironment()` in `SandboxService.swift` to set `HOME=/root` for root users when HOME is not already defined.

  User-provided HOME values are always preserved.

  ## Testing

  Add integration tests that verify:
  - `TestDefaultHomeEnvironment`: HOME is set to /root for root user by default
  - `TestCustomHomeEnvironment`: User-provided HOME values are preserved

  ## Test Results

  All tests pass:
  CURRENT_SDK=1 make test         # ✅ All unit tests pass
  CURRENT_SDK=1 make integration  # ✅ Integration tests pass including new tests

  Fixes #108


dcantah:
Thanks for the change! I'm wondering if we should do this in one of the components over in containerization instead that runs in the VM. The rationale is we can actually parse /etc/passwd and other files in the containers rootfs to find what the users home is set to.

dcantah:
Oh, interesting. We actually should be setting it based on this logic already https://github.com/apple/containerization/blob/main/vminitd/Sources/vminitd/Server%2BGRPC.swift#L843-L857. I'll try and see what might be going awry there, but I don't believe here is the right place to solve it

nandsha:
Thanks for the feedback @dcantah! Closing this PR since you've identified the fix at containerization vminitd (apple/containerization#104). Appreciate the guidance!

PR #144, opened 2025-06-11
Improve accuracy of progress updates

dkovba:
This PR resolves the problem of dropped progress updates and ensures the accuracy of the information provided in the progress bar. Additionally, it adds the displaying of the finished state to the progress bar.

### DIFF
--- a/Sources/TerminalProgress/ProgressBar+Add.swift
+++ b/Sources/TerminalProgress/ProgressBar+Add.swift
@@ -92,9 +92,14 @@ extension ProgressBar {
 
     /// Performs an addition to the current tasks.
     /// - Parameter tasks: The tasks to add to the current tasks.
-    public func add(tasks toAdd: Int, render: Bool = true) {
-        let newTasks = state.tasks + toAdd
-        set(tasks: newTasks, render: render)
+    public func add(tasks delta: Int, render: Bool = true) {
+        _state.withLock {
+            let newTasks = $0.tasks + delta
+            $0.tasks = newTasks
+        }
+        if render {
+            self.render()
+        }
     }
 
     /// Sets the total tasks.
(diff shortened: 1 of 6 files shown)


Sources/TerminalProgress/ProgressBar+Add.swift
```diff
@@ -61,23 +61,24 @@ extension ProgressBar {
...
     public func checkIfFinished() {
-        if let totalTasks = state.totalTasks {
+        var finished = true
+        var defined = false
+        if let totalTasks = state.totalTasks, totalTasks > 0 {
             // For tasks, we're showing the current task rather then the number of completed tasks.
-            guard state.tasks > totalTasks else {
-                return
-            }
+            finished = finished && state.tasks > totalTasks
```
adityaramani: When will `state.tasks` be greater than `totalTasks` ?

Sources/TerminalProgress/ProgressBar.swift
```diff
@@ -20,8 +20,9 @@ import SendableProperty
 /// A progress bar that updates itself as tasks are completed.
 public final class ProgressBar: Sendable {
     let config: ProgressConfig
+    // `@SendableProperty` adds `_state: Synchronized<State>`, which can be updated inside a lock using `_state.withLock()`.
```
adityaramani, line 23: Would this property should up in IDEs? Would the user be able to navigate into the def for the property?

adityaramani (commented):
Thanks! Had a couple small questions. Will also wait for more views on this

Sources/TerminalProgress/ProgressBar+Add.swift
dkovba: When we're running the last task, `state.tasks == totalTasks`. Therefore, `state.tasks > totalTasks` indicates that all tasks have been completed.

However, we can count the number of completed tasks instead, and add one when displaying the currently running task (for example, `[1/4]` for the first task). Would this be less confusing?

Sources/TerminalProgress/ProgressBar.swift
dkovba: It it possible to see the definition of the `_state` property by expanding the macro using the corresponding command in the context menu:

Screenshot 2025-06-12 at 4 29 54 PM

Sources/TerminalProgress/ProgressBar+Add.swift
dkovba: We should consider a progress bar finished when `state.tasks == totalTasks` and all other criteria are met. Fixed!

dcantah (approved):

dcantah (approved):

adityaramani (approved):

PR #152, opened 2025-06-11
Add issue templates for bugs and features

katiewasnothere:
I would love feedback on how people feel about these issue templates, if there should be more or less templates, or if there are any fields that people think we should add. 

You can see an example of how to use these by testing opening an issue on my fork [here](https://github.com/katiewasnothere/container/issues).

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: container community support
+    url: https://github.com/apple/container/discussions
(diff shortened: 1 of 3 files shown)


adityaramani (approved):

egernst (approved):
Thanks!

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
+contact_links:
+  - name: container community support
```
dkovba, line 3: It would look better if we capitalize all options, including the one that starts on “container”.

PR #171, opened 2025-06-12
Fix incorrect version number in documentation

pschlump:
Version of OS should be 16 not 26


pschlump:
We are a few years from version 26 of MacOS.   Looks like a typo.

visuphi:
@pschlump they jumped from macOS 15 to macOS Tahoe 26, so 26 is correct.

Source: https://www.apple.com/os/macos/

egernst:
Thanks for the PR, but yes @visuphi is correct here - it's correct as written. Thanks for sending PR though!

PR #177, opened 2025-06-12
FIx CI/CD formatting check fail

makhov:
Fix typos in Builder.grpc.swift that makes CI/CD pipeline fail


katiewasnothere:
Hi @makhov thanks for the fix! Closing this since the issue has already been resolved with https://github.com/apple/container/pull/176

PR #178, opened 2025-06-12
fix(common.yml): globalize CURRENT_SDK, improve shell safety and  imp…

Thedarkmatter10:
## 🔧 Improvements Summary

This PR introduces three improvements focused on safety, maintainability, and readability of the GitHub Actions workflow.

---

### 1. Define Global Environment Variable

**Before:**

Environment variable `CURRENT_SDK` was defined repeatedly in multiple steps.

**After:**

 Declared in gloabally one time. 

#### Why This Matters:

-  Eliminates duplication across steps.
-  Makes it easier to update or remove the variable in the future.
-  Still allows per-step override when necessary.

### 2. Fix Unsafe Shell Conditional on inputs.release
- Using **[[ ... ]]** **instea**d of **[ ... ]** for conditionals.
- Adding **double quotes** around inputs and refs to **avoid** evaluation issues.
 **PREVIOUSLY FIXED** Containerization project. [https://github.com/apple/containerization/pull/68](url)
  
 ### 3.  Removed EXCLUSION AND TODO comment.
  #### Affected Steps:

 `check Formatting`

`make proto`

### Improvements:

Now that the repositories are public, we no longer need to exclude files like Package.swift and Package.resolved from formatting and proto checks.

- Removed EXCLUDES logic
- Removed related TODO comments
- Updated git diff checks to include all files

@wlan0 @katiewasnothere

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -33,26 +33,18 @@ jobs:
           ./scripts/install-hawkeye.sh
           make fmt
 
-          EXCLUDES=(':(exclude)Package.swift' ':(exclude)Package.resolved')
-
-          if ! git diff --quiet -- . "${EXCLUDES[@]}"; then
+          if ! git diff --quiet -- . ; then
             echo "❌ The following files require formatting or license header updates:"
-            git diff --name-only -- . "${EXCLUDES[@]}"
+            git diff --name-only -- .
             false
           fi
 
       - name: Check protobuf
         run: |
           make protos
-
-          # TODO [launch]: TEMPORARILY we need to exclude these files since we had to modify them to add
-          # the github token for pulling the private repos.
-
-          EXCLUDES=(':(exclude)Package.swift' ':(exclude)Package.resolved' ':(exclude)Protobuf.Makefile')
-
-          if ! git diff --quiet -- . "${EXCLUDES[@]}"; then
+          if ! git diff --quiet -- . ; then
             echo "❌ The following files require formatting or license header updates:"
-            git diff --name-only -- . "${EXCLUDES[@]}"
+            git diff --name-only -- .
             false
           fi
   
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -16,64 +16,87 @@ jobs:
     permissions:
       contents: read
       packages: read
+    env:
+      CURRENT_SDK: y
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
       - name: Update containerization
         run: |
           /usr/bin/swift package update containerization
+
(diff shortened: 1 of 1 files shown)


.github/workflows/common.yml
```diff
@@ -16,46 +16,66 @@ jobs:
...
         run: |
           make protos
 
           # TODO [launch]: TEMPORARILY we need to exclude these files since we had to modify them to add
           # the github token for pulling the private repos.
-          if ! git diff --quiet -- . ':(exclude)Package.swift' ':(exclude)Package.resolved' ':(exclude)Protobuf.Makefile'; then echo "The following files require formatting or license header updates:\n$(git diff --name-only)" ; false ; fi
-        env:
-          CURRENT_SDK: y
+
+          EXCLUDES=(':(exclude)Package.swift' ':(exclude)Package.resolved' ':(exclude)Protobuf.Makefile')
```
katiewasnothere: Hi thanks for the fix! We can actually just remove these exclusions in this step and the formatting step. This was needed while we were testing because the repos were private. Now that they're public, we should not have these exclusions.

Thedarkmatter10: Hi, I've removed the temporary exclusions from both the formatting and protobuf check steps, as suggested. The workflow now checks all files without skipping any, since the repositories are public. I also cleaned up the related TODO comments .

Let me know if any further adjustments are needed. Thanks!!

katiewasnothere:
Please run `make fmt` then it LGTM :)

Thedarkmatter10:
I’m on Windows and don’t have Swift installed, so I’m unable to run `make fmt` for the Package.swift file. Could you please help format it on your end, or let me know an alternative workaround?
@katiewasnothere

katiewasnothere:
@Thedarkmatter10 Can you rebase with main again? I think that should resolve the formatting error

Thedarkmatter10:
Rebased with main as suggested. Please let me know if anything else is needed!
@katiewasnothere

katiewasnothere (approved):

.github/workflows/common.yml
```diff
@@ -16,64 +16,79 @@ jobs:
...
           mv bin/${{ env.BUILD_CONFIGURATION }}/bundle/container-dSYM.zip outputs
+
       - name: Test the container project
         run: |
           launchctl setenv HTTP_PROXY $HTTP_PROXY
           make test cleancontent install-kernel integration
         env:
           DEVELOPER_DIR: "/Applications/Xcode_16.3.app/Contents/Developer"
-          CURRENT_SDK: y
+          CURRENT_SDK: y # explicitly repeated due to local env block
```
dkovba, line 77: @katiewasnothere Do we still need `CURRENT_SDK`?

Issue #80, opened 2025-06-10
XPC connection error: Connection invalid


sriaradhyula:
Basic commands like `container ls` or `container build -f Dockerfile .` give this error `Error: internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"")`

Here is my system profile. I am running M1 Pro

```
❯ system_profiler SPHardwareDataType
Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro18,1
      Model Number: MK183LL/A
      Chip: Apple M1 Pro
      Total Number of Cores: 10 (8 performance and 2 efficiency)
      Memory: 16 GB
      System Firmware Version: 11881.121.1
      OS Loader Version: 11881.121.1
      Serial Number (system): <redacted>
      Hardware UUID: <redacted>
      Provisioning UDID: <redacted>
      Activation Lock Status: Disabled
```

```
❯ uname -a
Darwin <redacted> 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:49 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6000 arm64
```

dejanceltra:
You need to start container service via `container system start`, before running commands.

babldev:
Equivalent Docker error for reference:

```
docker: Cannot connect to the Docker daemon at unix:///Users/brady/.docker/run/docker.sock. Is the docker daemon running?

Run 'docker run --help' for more information
```

Something similar would be beneficial here.

artemijan:
I thought the installer should do it automatically. Or say few words in installation guide about this?

katiewasnothere:
Hi thanks for the report! @artemijan the CLI tutorial [here](https://github.com/apple/container/blob/main/docs/tutorial.md#try-out-the-container-cli) does mention running system start. However, we can definitely mention this in other places if it would be more clear that way.

artemijan:
Yes, sure. Readme has "installation" section. The majority of the users don't go in tutorial because almost everyone knows what is container from the docker. They just install it and use.

vndee:
Same issue here, it's better to make it clear in the installation instruction.

afbjorklund:
It is not only after installation, it seems to be needed again after reboot?

PR #179, opened 2025-06-12
provide suggestion if xpc 'Connection invalid' error encountered

heywoodlh:
Closes https://github.com/apple/container/issues/80

Adds the following help message if you try to run `container` against a host that hasn't started the container system:

```
❯ /usr/local/bin/container list
Error: internalError: "failed to list containers" (cause: "interrupted: "Connection invalid: ensure container system has been started with `container system start`"")

❯ /usr/local/bin/container run -it --rm docker.io/alpine
Error: interrupted: "Connection invalid: ensure container system has been started with `container system start`"
```

### DIFF
--- a/Sources/CLI/Application.swift
+++ b/Sources/CLI/Application.swift
@@ -21,6 +21,7 @@ import CVersion
 import ContainerClient
 import ContainerLog
 import ContainerPlugin
+import ContainerizationError
 import ContainerizationOS
 import Foundation
 import Logging
@@ -176,7 +177,13 @@ struct Application: AsyncParsableCommand {
                 Self.printModifiedHelpText()
                 return
             }
-            Application.exit(withError: error)
+            let errorAsString: String = String(describing: error)
+            if errorAsString.contains("XPC connection error") {
+                let modifiedError = ContainerizationError(.interrupted, message: "\(error)\nEnsure container system service has been started with `container system start`.")
+                Application.exit(withError: modifiedError)
+            } else {
+                Application.exit(withError: error)
+            }
         }
     }
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,12 @@ Download the latest signed installer package for `container` from the [GitHub re
 
 To install the tool, double click the package file and follow the instructions. Enter your administrator password when prompted, to give the installer permission to place the installed files under `/usr/local`.
 
+Start the system service with:
+
+```
+container system start
+```
+
 ### Uninstall
 
 Use the `uninstall-container.sh` script to remove `container` from your system. To remove your user data along with the tool, run:


katiewasnothere (approved):

adityaramani (changes requested):
Thanks for the PR!  Im trying to think about this a little bit. 

I think I'd like it better if this check happens somewhere in the CLI - im not sure where the right place to put it in is (in a way that we dont need to update every command). Maybe `ArgumentParser` - the library we use to write the CLI has an exit hook? 

The reason for this is that `XPCClient` in theory is a (mostly) generic XPC client type and could be used for other services that may uses a different bootstrap pattern.

What are your thoughts on this?

heywoodlh:
My thought was to emulate a similar UX to the Docker client since it's the dominant container platform as highlighted here: https://github.com/apple/container/issues/80#issuecomment-2963762739

Honestly, I'm not super concerned about where it's at. I think it makes sense to have on every command where the XPC client fails to connect -- as that's a fatal error.

Regardless, I'm happy to rework it however is most desired -- let me know what you'd prefer and I'll do it. 😄

adityaramani:
Yes totally agree about having the check there - just that I think its better if its somewhere in the CLI. Let me look a bit more to see if there is an easy way to add it. 

This will also allow others to share their thoughts

heywoodlh:
I'll spend some time reworking it into the CLI as you suggested and see what I can come up with. Likely will get around to it tomorrow! 😄

adityaramani:
Went through the code a bit - I think we can perform that check here? https://github.com/apple/container/blob/main/Sources/CLI/Application.swift#L179 

And it should account for all the commands we have?

heywoodlh:
Latest iteration of my PR, the error-handling looks like this:

```
❯ container system stop

❯ ./bin/container list
Error: internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"")
Ensure container system has been started with `container system start`.

❯ ./bin/container run -it --rm docker.io/alpine:latest
Error: interrupted: "XPC connection error: Connection invalid"
Ensure container system has been started with `container system start`.
```

Please let me know if you have any further feedback on wording, etc. :smile: Sorry for taking a while to get around to this!

Sources/CLI/Application.swift
```diff
@@ -176,7 +176,15 @@ struct Application: AsyncParsableCommand {
                 Self.printModifiedHelpText()
                 return
             }
-            Application.exit(withError: error)
+            // Error: internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"") ensure container system has been started with `container system start`
+            let errorAsString: String = String(describing: error)
+            if errorAsString.contains("XPC connection error") {
+                let modifiedErrorDescription = "\(error)\nEnsure container system has been started with `container system start`."
+                let modifiedError = NSError(domain: "", code: 1, userInfo: [NSLocalizedDescriptionKey: modifiedErrorDescription])
+                Application.exit(withError: modifiedError)
```
adityaramani: Can we switch this to be ContainerizationError over NSError? Or Did you see that using NSError was causing the `exit()` to print the message better?

Sources/CLI/Application.swift
```diff
@@ -176,7 +176,15 @@ struct Application: AsyncParsableCommand {
                 Self.printModifiedHelpText()
                 return
             }
-            Application.exit(withError: error)
+            // Error: internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"") ensure container system has been started with `container system start`
```
adityaramani: We could probably just change this to say we are trying to detect this condition / remove this comment entirely

heywoodlh: Oops! This was a note I left so I could remember what the error contents said exactly. Removed now :smile:

Sources/CLI/Application.swift
heywoodlh: Done in latest rev!

heywoodlh: However, is there a way with a `ContainerizationError` to remove the quotes at the beginning and end of the error output?

```
❯ ./bin/container list
Error: internalError: "internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"")
Ensure container system service has been started with `container system start`."
```

heywoodlh: Also, noticed in my output that `internalError` is repeated twice. Is there a more generic `ContainerizationError` type that is not `.internalError` that I could use?

adityaramani: We do have `.interrupted` which may make sense? or `.invalidState` ?
I need to check if we can do something about the quotes

heywoodlh: Interrupted looks good to me!

```
❯ ./bin/container list
Error: interrupted: "internalError: "failed to list containers" (cause: "interrupted: "XPC connection error: Connection invalid"")
Ensure container system service has been started with `container system start`."
```

adityaramani (approved):
Thanks! LGTM! Will wait for the others to look if we want to change the wordings. 

You just need to run `make fmt` and re-push

katiewasnothere (approved):

heywoodlh:
@katiewasnothere I forgot to run `make fmt` before pushing! That has been fixed on latest rev. :smile:

PR #188, opened 2025-06-13
Throw errors in ServiceManager

eliseomartelli:
Add Error throwing to Service Manager

### DIFF
--- a/Sources/ContainerPlugin/ServiceManager.swift
+++ b/Sources/ContainerPlugin/ServiceManager.swift
@@ -62,19 +62,23 @@ public struct ServiceManager {
 
         let null = FileHandle.nullDevice
         let stdoutPipe = Pipe()
+        let stderrPipe = Pipe()
         launchctl.standardOutput = stdoutPipe
-        launchctl.standardError = null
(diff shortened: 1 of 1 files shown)


Sources/ContainerPlugin/ServiceManager.swift
```diff
@@ -70,13 +70,11 @@ public struct ServiceManager {
         launchctl.waitUntilExit()
         let status = launchctl.terminationStatus
         guard status == 0 else {
-            // TODO: review error handling
-            return []
+            throw ContainerizationError(.internalError, message: "Command `launchctl list` failed with status \(status)")
```
makhov: I think it makes sense to [read an error from stderr](https://github.com/apple/container/blob/167d926b4dca1ba065d2566c46392dfb812aad2c/Sources/ContainerPlugin/ServiceManager.swift#L66) and add it to the message

eliseomartelli: That actually makes sense 👍

adityaramani (approved):

adityaramani:
Thanks for the fix!

PR #198, opened 2025-06-13
Wait for IO streams to complete before a process exits

adityaramani:
Implements a way for the CLI to wait until the IO streams from the SandboxService have been drained before closing them.

Follows the same pattern as https://github.com/apple/containerization/pull/110

This change also performs some cleanup in the `SandboxService.startProcess` method - splitting the code paths to handle the init process and an exec'd process into two different private methods to make easier reading

### DIFF
--- a/Sources/CLI/RunCommand.swift
+++ b/Sources/CLI/RunCommand.swift
@@ -18,6 +18,7 @@ import ArgumentParser
 import ContainerClient
 import Containerization
 import ContainerizationError
+import ContainerizationExtras
 import ContainerizationOS
 import Foundation
 import NIOCore
@@ -169,6 +170,13 @@ struct ProcessIO {
     let stdin: Pipe?
     let stdout: Pipe?
     let stderr: Pipe?
+    var ioTracker: IoTracker?
+
+    struct IoTracker {
+        let stream: AsyncStream<Void>
+        let cont: AsyncStream<Void>.Continuation
+        let configuredStreams: Int
+    }
 
     let stdio: [FileHandle?]
 
@@ -224,7 +232,11 @@ struct ProcessIO {
             }
             return Pipe()
         }()
+
+        var configuredStreams = 0
+        let (stream, cc) = AsyncStream<Void>.makeStream()
         if let stdout {
+            configuredStreams += 1
             let pout: FileHandle = {
                 if let current {
                     return current.handle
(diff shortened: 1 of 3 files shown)


Sources/CLI/RunCommand.swift
```diff
@@ -251,25 +264,54 @@ struct ProcessIO {
...
+                    counter -= 1
+                    if counter == 0 {
+                        ioTracker.cont.finish()
+                        break
+                    }
+                }
+            }
+        } catch {
+            log.error("Timeout waiting for IO to complete : \(error)")
+            throw error
```
adityaramani, line 314: I think it makes sense to throw here since `self` is a struct and we cannot update  `self.ioTracker` to be nil (without going throw a few hoops) - At least this way the client will know that something is off and they wont repeatedly call this `wait` method

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -154,88 +154,111 @@ public actor SandboxService {
...
+        self.setState(.starting)
+        do {
+            try await container.start()
+            let waitFunc: ExitMonitor.WaitHandler = {
+                let code = try await container.wait()
+                if let out = stdio[1] {
+                    try self.closeHandle(out.fileDescriptor)
+                }
+                if let err = stdio[2] {
+                    try self.closeHandle(err.fileDescriptor)
```
adityaramani, line 217: This is where we call close on the handles that is passed in from the CLI - wondering if its better to move this inside of the LinuxProcess, where we close the relays? Its effectively the same, but that the library handles it for us

dcantah: I'd have to look, but at the library level the client passes us protocols, we don't get FileHandle's or the type directly. None of the protocols have a `close` method required for them right now, so we can really only close the vsock fds that we dial internally on behalf of the client and not what we're going to relay the vsock output to right now.

dcantah: I'm not a huge fan of closing the stdio they passed internally in the library though, as they may still want valid descriptors after the container exits (imagine if they passed an fd for a regular file, and the user wants to read the file after wait() returns. They'd need to re-open a new fd). In this case it'd be fine as they're dup'd by xpc when you send them iirc, but it wouldn't be good for all use cases I think.

PR #218, opened 2025-06-16
Fix Race Condition in Container Removal (#130)

ramsyana:
This PR resolves a race condition when removing a container immediately after stopping it, caused by the `stop` command returning before the container fully transitions to the stopped state (#130).

**Changes:**
- Enhanced `TestCLIRmRace.swift` with robust test logic and helper methods (`containerExists`, `safeRemove`).
- Improved error handling to distinguish race conditions from successful removals.
- Added exponential backoff retry logic for cleanup operations.
- Updated `CLITest.swift` with missing `doRemove` method.
- Fixed `BuilderStart.swift` to handle `.stopping` case.
- Improved error messages with container ID for better debugging.

**Testing:**
- ✅ All tests pass (`make test`, `make integration`).
- ✅ Verified on macOS 26.
- ✅ Race condition test validates success and failure scenarios.
- ✅ Code formatted (`make fmt`).

Hopefully, this will pass the integration tests on GitHub.


dcantah (approved):

dcantah:
Thanks for this

PR #221, opened 2025-06-17
makefile: Change build_bin_dir to be lazily evaluated

dcantah:
The bin directory for placing built binaries to should be lazily evaluated, as we change BUILD_CONFIGURATION during `make release`.


dkovba (approved):
👍

PR #223, opened 2025-06-17
Match names from both containerd and containerization annotations

### LINKED ISSUE #211: [Bug]: FROM local image tries to pull from docker.io first
### I have done the following

### Steps to reproduce

```sh
> cat Dockerfile1
FROM alpine:latest

CMD ["echo", "Hello from the trivial Docker image!"]

> container build -t trivial-image -f Dockerfile1 .
# ...
Successfully built trivial-image:latest

> cat Dockerfile2
FROM trivial-image

CMD ["echo", "Hello from the other trivial Docker image!"]

> container build -t trivial-image2 -f Dockerfile2 .
[+] Building 0.0s (0/1)
 => [resolver] fetching image...docker.io/library/trivial-image                                                                                                                        0.0s
Error: internalError: "HTTP request to https://registry-1.docker.io/v2/library/trivial-image/manifests/latest failed with response: 401 Unauthorized. No credentials found for host registry-1.docker.io"
```

### Current behavior

`container` first looks for the image on docker.io instead of locally.

### Expected Behavior

First check for local images.

### Environment

```markdown
- OS: macOS 15.5 (24F74)
- Xcode: 26.0 beta (17A5241e)
- Container: 555dbfc (main as of writing)
```

### Code of Conduct

makhov:
Fixes #211

Currently, it matches only with `.containerizationImageName` annotation, but it contains name without `docker.io/library` prefix. `.containerdImageName` annotation has the prefix.


Tests/CLITests/Subcommands/Build/CLIBuilderTest.swift
```diff
@@ -67,7 +67,7 @@ extension TestCLIBuildBase {
             let newTempDir: URL = try createTempDir()
             let newDockerfile: String =
                 """
-                 FROM local-only:\(imageName)
+                 FROM \(imageName)
```
makhov, line 70: I think this is a typo in the test, because `imageName` has `local-only` prefix already:
https://github.com/apple/container/blob/main/Tests/CLITests/Subcommands/Build/CLIBuilderTest.swift#L63

Sources/ContainerClient/Core/ClientImage.swift
```diff
@@ -443,4 +444,13 @@ extension ImageDescription {
...
         return name
     }
+
+    fileprivate func matchWithNameFromAnnotation(name: String) -> Bool {
+        guard let annotations = self.descriptor.annotations else {
+            return false
+        }
+        
+        return annotations[AnnotationKeys.containerizationImageName] == name
+            || annotations[AnnotationKeys.containerdImageName] == name
```
adityaramani, line 458: This is a little tricky - yeah adding this check would fix the issue we are seeing, but we dont feel its right that an image we build without a domain in its tag, automatically gets assigned one.

I think we should change the behavior so that the reference we get from the build container does not have the domain prepended if its a locally built image

And I think that happens here : 
https://github.com/apple/container-builder-shim/blob/main/pkg/build/frontend.go#L144

Tagging @katiewasnothere for some more input

makhov: What do you mean saying that an image automatically gets domain assigned? It's a fully qualified reference and all normalized names are with domain, so `ParseAnyReference` returns it in this format. 

I don't have a strong opinion here, so it you decide to get rid of the domain here, I can make corresponding changes.

adityaramani: For example if the Dockerfile being built is
```
FROM local-image:test
...
```

The expectation is that somewhere in the build process, the `container-image-service` will be asked about the existence of `local-image:test` by the builder. 

But because the builder is parsing that base image reference using `reference. ParseAnyReference` (which has the assumption that an image should get the default domain of docker.io) - the `container-image-service` ends up getting a request for `docker.io/local-image:test`

And this diff makes it such that we intercept the request b/w the build container and image service to work around that assumption

adityaramani: To add to the above - we still want the nice UX of something like this still working

```
FROM alpine:latest
...
```

And to support that container does some normalization of its own - at a high level it does

- Check if "alpine:latest" exists as an image in the daemon
- If not normalize it using the configured default registry (see `container registry default set`) 
- Check if an image with normalized reference exists in the daemon (will be of the form `<domain>/alpine:latest`)
- If that fails - pull the image from the registry

makhov: Yeah, I got the idea. And I agree that the decision about the default registry should be made on the `container-image-service` side, not on `container-build-shim`. 
I made some changes in the PR, that should both support the flow you described and fix the initial bug. Next I'm going to make a PR to the shim repo to change the reference parsing and get rid of the default docker domain in the request. Does it sound like a good plan for you?

Abdullah4345 (approved):

Sources/ContainerClient/Core/ClientImage.swift
makhov: @adityaramani the fix is a bit more complex, but the idea is the same: we [send](https://github.com/apple/container-builder-shim/pull/19/files) a `resolve` request with non-normalized ref and [respond](https://github.com/apple/container/pull/223/files#diff-4fe5f4ae67660957c7454f40aa68ee6c0d33c96cd0880966b5f8fa0bb130f75bR78) with the normalized one.

katiewasnothere: I think we're still missing some things here. 

I'm not a builder-shim or buildkit expert so I've been digging around as well to get more familiar. From what I can tell, the `resolveStates` call in builder-shim is only used for setting up input information that can be used if there's a named context stage. For example, a dockerfile containing something like this

```
FROM alpine:latest AS base
COPY --from=base /some/file /some/file
```

As a result, when we call into buildkit to convert Dockerfile2LLB, buildkit will still attempt to make calls to resolve the images referenced. Inconveniently for us, buildkit will parse the image names itself before requesting the images, see [here](https://github.com/moby/buildkit/blob/7841a73c183b0166dcede61eb07ece92e5bc28c8/frontend/dockerfile/dockerfile2llb/convert.go#L466), so we'd still get the request to resolve docker.io/library/<local image name>.

katiewasnothere: Okay after lengthy investigation, I think there's not really a way here for us to avoid buildkit adding the docker.io/library prefix to an image reference other than us modifying the dockerfile that we send to builder, which seems like a bad plan. Given that, I think our only option is to match the reference to either the container.name or containerization.name as you've done here. I don't think we need the builder side changes though. 

@adityaramani is out for a few days, but I'd like to wait for his input here.

makhov: I think we call Dockerfile2LLB after we resolve images: https://github.com/apple/container-builder-shim/blob/35406d303bc165a090df812d75971d90d4e24f9f/pkg/build/frontend.go#L75-L80

But I agree that it becomes too complex and fragile. We can skip `container-builder-shim` changes then and keep this PR.

katiewasnothere: The image resolving we do in the builder shim itself is only used by buildkit for named contexts (see [here](https://github.com/moby/buildkit/blob/870aae6921d715664522cd8f4d1976e9b6920bc4/frontend/dockerui/namedcontext.go#L58) and [here](https://github.com/moby/buildkit/blob/870aae6921d715664522cd8f4d1976e9b6920bc4/frontend/dockerfile/dockerfile2llb/convert.go#L506))

adityaramani: Do you think its possible to make a variation of this change in `BuildImageResolver.swift` when we try to fetch the image?

Something like this? 

```
        let img = try await {
            if ref.startsWith("docker.io/library") {
               trimmedRef = ref["docker.io/library".count + 1 : ]
                if i = ClientImage.get(trimmedRef) {
		return i
                 }
    
           guard let img = try? await ClientImage.pull(reference: ref, platform: platform) else {
                return try await ClientImage.fetch(reference: ref, platform: platform)
            }
            return img              
}
        }()
```

The idea behind this is to restrict this change only to affect the build flow. Do ya'll think this would work?

katiewasnothere: I think I'd rather we use the containerization.name as the source of truth. It's possible that someone has the images `alpine` (local) and `docker.io/library/alpine` (pulled). If we always stripped out `docker.io/library` for the image on a build, we'd always return the local `alpine` image.

wlan0:
Hi @makhov 👋🏾 , 

Thanks a lot for putting the effort into this and suggesting improvements here. I really appreciate your contribution!

However, this issue is best addressed by fixing the resolver logic itself, rather than at this layer of the stack. Specifically, the resolver shouldn't automatically prefix a default registry if none was provided. Here's the PR addressing this: https://github.com/apple/container-builder-shim/pull/34

I’m happy to discuss this further if you have questions

makhov:
@wlan0 No problem! I left a comment in the mentioned PR, we can discuss further details there, so feel free to close this one

katiewasnothere:
Closing in favor of https://github.com/apple/container-builder-shim/pull/34. Thanks again @makhov!

PR #224, opened 2025-06-17
Remove async from AttachmentAllocator lookup

eliseomartelli:
The `lookup(hostname:)` method in `AttachmentAllocator` no longer requires
async, as it only performs a synchronous dictionary lookup.

### DIFF
--- a/Sources/Services/ContainerNetworkService/NetworkService.swift
+++ b/Sources/Services/ContainerNetworkService/NetworkService.swift
@@ -101,7 +101,7 @@ public actor NetworkService: Sendable {
         }
 
         let hostname = try message.hostname()
-        let index = try await allocator.lookup(hostname: hostname)
+        let index = try allocator.lookup(hostname: hostname)
         let reply = message.reply()
         guard let index else {
             return reply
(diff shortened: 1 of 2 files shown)


Sources/Services/ContainerNetworkService/NetworkService.swift
```diff
@@ -101,7 +101,7 @@ public actor NetworkService: Sendable {
         }
 
         let hostname = try message.hostname()
-        let index = try await allocator.lookup(hostname: hostname)
+        let index = try allocator.lookup(hostname: hostname)
```
dcantah, line 104: Because AttachmentAllocator is an actor, callers still need to access every method asynchronously. The removal of the `async` from the `func lookup(hostname: String)` definition does not change this, it's solely to mark that that method does not call anything async inside of itself.

dcantah: If you get rid of the diff in this file this change LGTM

crosbymichael:
Thank you for the change.  When you have time, please update the change. I'll close this and when you have the updated patch, feel free to open a new PR with the change.

PR #226, opened 2025-06-17
Improve error handling in answerHost

eliseomartelli:
Refactor `answerHost` to throw descriptive `DNSResolverError` instances
instead of returning nil when IP allocation or parsing fails.


makhov:
Arer you sure we need to throw errors here? As far as I understood, the server collects results from different places and return `.nonExistentDomain` if nothing found

https://github.com/apple/container/blob/main/Sources/DNSServer/DNSServer%2BHandle.swift#L59-L61

Sources/APIServer/ContainerDNSHandler.swift
```diff
@@ -76,16 +76,19 @@ struct ContainerDNSHandler: DNSHandler {
 
     private func answerHost(question: Question) async throws -> ResourceRecord? {
         guard let ipAllocation = try await networkService.lookup(hostname: question.name) else {
-            return nil
+            throw DNSResolverError.serverError("No IP allocation found for hostname: \(question.name)")
```
jglogan: As @makhov indicates, handlers implement a chain of command; `nil` should be returned if there exists the possibility for a downstream handler to service a query and this throw precludes this possibility. Could you revert this change in your PR?

Guarding against errors in the data returned by the network service is very reasonable though.

Sources/APIServer/ContainerDNSHandler.swift
```diff
@@ -76,16 +76,19 @@ struct ContainerDNSHandler: DNSHandler {
...
         }
 
         let components = ipAllocation.address.split(separator: "/")
-        guard components.count > 0 else {
-            return nil
+        guard !components.isEmpty else {
+            throw DNSResolverError.serverError("Invalid IP format: empty address")
```
jglogan, line 84: This seems a good fix to me; it's also an unexpected error (it would indicate a bug in the network service).

Sources/APIServer/ContainerDNSHandler.swift
```diff
@@ -76,16 +76,19 @@ struct ContainerDNSHandler: DNSHandler {
...
         }
-        guard let ip = IPv4(String(components[0])) else {
-            return nil
+
+        let ipString = String(components[0])
+        guard let ip = IPv4(ipString) else {
```
jglogan, line 88: Likewise a good fix.

jglogan (changes requested):

Sources/APIServer/ContainerDNSHandler.swift
eliseomartelli: Thank you for the review. I also appreciate @makhov’s thorough comment. 
I amended the commit, to revert back to returning nil.

Abdullah4345 (approved):

jglogan (approved):

jglogan:
@eliseomartelli Sorry for the delay, I was traveling this week. Approved and merged, thanks for improving those error responses!

PR #236, opened 2025-06-19
Add socket publishing functionality

Reneechang17:


Sources/ContainerClient/Core/ContainerConfiguration.swift
```diff
@@ -44,6 +46,43 @@ public struct ContainerConfiguration: Sendable, Codable {
...
+        case dns
+        case rosetta
+        case hostname
+        case initProcess
+        case platform
+        case resources
+        case runtimeHandler
+    }
+
+    // Customizing the decoder to handle the optional publishedSockets field.
```
jglogan: Use three slashes, and the comment should read as something like:

```swift
/// Create a configuration from the supplied Decoder, initializing missing
/// values where possible to reasonable defaults.
```

Issue #55, opened 2025-06-09
compose support?


magnusviri:
Will you be creating a tool that allows users to use docker_compose.yml files?

See [https://docs.docker.com/compose/](https://docs.docker.com/compose/).

awilfox:
Yes, I would also love to see Compose support for `container`.  Is this something that would accepted as a contribution?

bergrahm:
I am a heavy-weight compose user. Having the option to drop and replace would dramatically increase the chance for me to adopt this tool without hassle. Great work by the way!

codedoga:
+1

psy0rz:
Without compose support i will just keep using docker desktop for now. What would be the advantage of this new system?

kv5h:
I'm heavily depending on Compose, so yearning for the support 🐳

koi-tattoo:
I've rolled up a script to emulate some functionalities of docker-compose [https://gist.github.com/koi-tattoo/1decd7f52f55ddb902e8a9eb78f8c587](https://gist.github.com/koi-tattoo/1decd7f52f55ddb902e8a9eb78f8c587).
I mostly generated it through claude 4 sonnet as I was just trying to have something working fast. It managed to up this compose file [https://gist.github.com/koi-tattoo/5f91de153768865d6d29f3773564a75d](https://gist.github.com/koi-tattoo/5f91de153768865d6d29f3773564a75d) decently, maybe you guys can try to use it to easily test apple containers through your already established compose files.

Edit: see the issue docker/compose/issues/12934, maybe it is on docker-compose's end to implement this feature.

t3chn0m4g3:
yes, compose is the peanut-butter and jelly!

psy0rz:
Dont get me wrong its super cool apple does this!

bgottsch:
+1

some integration pointers:
- naming: `container compose`?
- load balancers like Traefik as a container

llmons:
+1

pszymkowiak:
+1

dmatton:
+1

koi-tattoo:
According to [https://github.com/docker/compose/issues/12934](https://github.com/docker/compose/issues/12934), it could be done if the apple container runtime supports the Moby API ([https://github.com/moby/moby/blob/master/api/README.md](https://github.com/moby/moby/blob/master/api/README.md))

mani-arjunan:
+1

tarikweiss:
+1

[...]

PR #239, opened 2025-06-19
Add Compose

Mcrich23:
There has been a lot of discussion around a docker-compose counterpart for Container. This moves that discussion a bit further by introducing a basic translation layer for compose.

This pr introduces a new command group into the cli: `Compose`. It has two subcommands: `up` and `down`.

The origin behind this pr being made comes from the desire to run [Vapor](https://vapor.codes) on Apple Containers. Hopefully, efforts like this make it possible quite soon.

> **Note:** This is an early implementation of compose and there are features missing in Container that makes it not up to complete feature parity. This is bound to change as features such as port customization are introduced.


Mcrich23:
Referencing the discussions/issues about how best to support compose: #55 #194 #101

stevenanthonyrevo (approved):
I've reviewed as much as I can!

I'm fairly limited due to my current MacOS version as anything below 6.2.0 will just error out.  

Although, thanks for your work on this feature request as I was really hopeful for such a feature!

So on Good Faith, I'm going to Approve! If anyone else wants to review, feel free!

Mcrich23:
Thank you! I am also planning to add support for the newfound networking capabilities in the coming days. 

Also, if anyone knows how to restructure compose to call these commands without invoking the command line, I would be very grateful to hear about it!

Mcrich23:
To address these things before the merge, I am going to make this pull request into a draft. If you want to try the tool compiled, you can get it at https://github.com/mcrich23/Container-Compose.

Mcrich23:
More discussion [here](https://github.com/apple/container/discussions/320)

jglogan:
@Mcrich23 Can you refactor the PR so that the compose subcommand is a plugin? I'm working on developer documentation for extending `container` but that will be evolving alongside your work.

* Create a directory `Plugins/compose` and relocate your subcommand there.
* Remove explicit wiring of the compose subcommand from CLI/Application.swift.
* Update Package.swift to create a new "compose" executableTarget based on Plugins/compose.
* Create a `config/compose-config.json` file.
* Update the Makefile to add installation and signing for the compose plugin.

compose-config.json (use your GitHub profile URL for `author` for now):

```json
{
    "abstract" : "Orchestrate container resources using ",
    "author": ""
}
```

add to the `install` target for `Makefile`:

```
	@echo Installing compose into $(DESTDIR)...
	@$(SUDO) mkdir -p $(join $(DESTDIR), libexec/container-plugins/compose/bin)
	@$(SUDO) install $(BUILD_BIN_DIR)/compose $(join $(DESTDIR), libexec/container-plugins/compose/bin/compose)
	@$(SUDO) install config/compose-config.json $(join $(DESTDIR), libexec/container-plugins/compose/config.json)
	@$(SUDO) codesign $(CODESIGN_OPTS) $(join $(DESTDIR), libexec/container-plugins/compose/bin/compose)
```

Mcrich23:
Will do! I did the Application thing to make my other tool work using the same logic, but it sounds like a plugin might also still allow that to work without duplicating code.

Mcrich23:
Question @jglogan. Compose now relies on internal tools instead of just calling the command line for a better experience. How should I maintain this when moving it to a plugin?

Mcrich23:
Nevermind. It seems that I am able to list it as a dependency

Mcrich23:
I think that is all correct now. @jglogan please let me know if I did the makefile stuff wrong. I am still not too great with it.

Mcrich23:
This also seems to be happening when running it via Xcode: 
```
Error: failed to find plugin named container-compose
Usage: container [--debug] <subcommand>
  See 'container --help' for more information.
```

Mcrich23:
Unfortunately, it is now throwing me many symbolic issues. Likely because of a circular reference?

Mcrich23:
I managed to get it all to work although it does require extracting the main cli logic to a target that is then referenced by the executable.

Makefile
```diff
@@ -68,7 +68,13 @@ install: installer-pkg
 		rm -rf $${temp_dir} ; \
 	else \
 		$(SUDO) installer -pkg $(PKG_PATH) -target / ; \
-	fi 
+	fi
+	
+	@echo Installing compose into $(DESTDIR)...
```
jglogan: Cool! Next step is to take what's here and create three corresponding targets for compose, so that we wind up with separate installers for the core app and for compose:

| target | compose target | description |
|-------|----------------|-------------|
| `install` | `install-compose` | Installs the package normally, or to `DESTDIR` if `SUDO` not defined |
| `$(STAGING_DIR)` | `$(STAGING_DIR_COMPOSE)` | Stages binaries, config files, uninstall script for package creation |
| `installer-pkg` | `installer-pkg-compose` | Creates an installer package |

Mcrich23: Done!

Mcrich23:
@jglogan Just want to checkin about the next steps for getting this merged.

stevenanthonyrevo:
@jglogan I appreciate the work on the feature. 

Can you explain why you included TCP and UDP source files and why this needed to be changed to a plugin format? 

I'd like to see a simplified approach without any external dependencies.

Mcrich23:
@stevenanthonyrevo can you please clarify on what you mean by UDP and TCP files? I assume you are referencing the codable structures I use to make compose possible but I just want to confirm. 

As for the plugin stuff, my understanding is that we are making compose an optional install through the plugin system provided by Swift Argument Parser. 

Also the two dependencies being added make yaml file parsing easy as well as enable print coloring for different containers when logging multiple at once.

niteshbalusu11:
@Mcrich23 first of all great work on the PR. I tried to pull in your changes locally and run it. Build was successful but I am running into issues. Does it not work with services in a compose file?

Screenshot 2025-07-25 at 11 55 51 AM

stevenanthonyrevo:
Sure @Mcrich23, after taking a look at the required files for docker compose. 

I noticed a few tests had been added possibly to setup the SocketForwarder. I'm not opposed to the structure but rather confused on a particular classname. 

I did not come across any coverage needed for the particular class, when searching for a **EchoServer**. 

Maybe these extra tests got added in as extra or they may be required.

Here are the files: 
https://github.com/Mcrich23/container/blob/add-compose/Tests/SocketForwarderTests/TCPEchoServer.swift
https://github.com/Mcrich23/container/blob/add-compose/Tests/SocketForwarderTests/UDPEchoServer.swift 

Are these necessary for the TCPForwarder and UDPForwarder?

jglogan:
@Mcrich23 I just got some time to start looking at the PR again.

This is a good start. What would you say is the minimum viable product for compose? What parts of the compose spec would it implement?

Rather than pile everything into one PR, I think it'll be easier to have a maintainable plugin if we start with the smallest usable set of functionality that we can deliver, ensure that it's solid, and then build features from there. It's ok if on the initial commit we have a compose that can do "toy" examples.

With that base in place you might even be able to get other community members to incrementally add similarly high quality features.

Regarding Compose types, have you looked into generating the Swift code from the [spec](https://github.com/compose-spec/compose-spec) itself using [quicktype](https://quicktype.io). If that worked then we could automate that (running quicktype under container since it could be a manual step that we perform only when updates are required).

I also rebased your code and gave it a try with a quick example I threw together. A few notes from that exercise:

- I had to rename my `docker-compose.yaml` (the name I typically use) to `docker-compose.yml` (the name it looks like the plugin requires, and by the way a perfect example of limiting the feature set – accept only that name now, and add `compose -f file {up|down}` later on).
- Building a Dockerfile from the `server` subdirectory failed, IIRC it was trying to locate the file in the top level project directory. Again, the initial product could just support `image`.
- With `image` it appears that it is always pulling the image instead of checking whether a local image is present. With my image named `server` that means that it tries to pull `docker.io/library/server`, which understandably fails.

Once I pushed my image I was able to get to the point where the database and API server were both starting up. The API server couldn't reach the DB server, but that's because the system I ran on is still on macOS 15 so the containers are network-isolated.

Note that I'm not using any IP addresses for the config. The plugin could support a DNS domain option on the command line and/or read the DNS domain default and apply it to the resolv.conf files in the containers using the `--dns-domain` option of the run command.

I've attached the example here:

[20250724-container-compose.tgz](https://github.com/user-attachments/files/21434211/20250724-container-compose.tgz)

Mcrich23:
> Sure @Mcrich23, after taking a look at the required files for docker compose.
> 
> I noticed a few tests had been added possibly to setup the SocketForwarder. I'm not opposed to the structure but rather confused on a particular classname.
> 
> I did not come across any coverage needed for the particular class, when searching for a **EchoServer**.
> 
> Maybe these extra tests got added in as extra or they may be required.
> 
> Here are the files: https://github.com/Mcrich23/container/blob/add-compose/Tests/SocketForwarderTests/TCPEchoServer.swift https://github.com/Mcrich23/container/blob/add-compose/Tests/SocketForwarderTests/UDPEchoServer.swift
> 
> Are these necessary for the TCPForwarder and UDPForwarder?

@stevenanthonyrevo I don't recall touching these files for any reason. I can look at them and see but my best guess is that something is showing up from me merging main into this branch to avoid future merge conflicts.

Mcrich23:
> @Mcrich23 I just got some time to start looking at the PR again.
> 
> This is a good start. What would you say is the minimum viable product for compose? What parts of the compose spec would it implement?
> 
> Rather than pile everything into one PR, I think it'll be easier to have a maintainable plugin if we start with the smallest usable set of functionality that we can deliver, ensure that it's solid, and then build features from there. It's ok if on the initial commit we have a compose that can do "toy" examples.
> 
> With that base in place you might even be able to get other community members to incrementally add similarly high quality features.
> 
> Regarding Compose types, have you looked into generating the Swift code from the [spec](https://github.com/compose-spec/compose-spec) itself using [quicktype](https://quicktype.io). If that worked then we could automate that (running quicktype under container since it could be a manual step that we perform only when updates are required).
> 
> I also rebased your code and gave it a try with a quick example I threw together. A few notes from that exercise:
> 
> * I had to rename my `docker-compose.yaml` (the name I typically use) to `docker-compose.yml` (the name it looks like the plugin requires, and by the way a perfect example of limiting the feature set – accept only that name now, and add `compose -f file {up|down}` later on).
> * Building a Dockerfile from the `server` subdirectory failed, IIRC it was trying to locate the file in the top level project directory. Again, the initial product could just support `image`.
> * With `image` it appears that it is always pulling the image instead of checking whether a local image is present. With my image named `server` that means that it tries to pull `docker.io/library/server`, which understandably fails.
> 
> Once I pushed my image I was able to get to the point where the database and API server were both starting up. The API server couldn't reach the DB server, but that's because the system I ran on is still on macOS 15 so the containers are network-isolated.
> 
> Note that I'm not using any IP addresses for the config. The plugin could support a DNS domain option on the command line and/or read the DNS domain default and apply it to the resolv.conf files in the containers using the `--dns-domain` option of the run command.
> 
> I've attached the example here:
> 
> [20250724-container-compose.tgz](https://github.com/user-attachments/files/21434211/20250724-container-compose.tgz)

@jglogan Thanks for all of the feedback and support. I like the idea of merging in a MVP and then letting the community continue to build on it.

 For the docker compose file differences, I can totally add a file parameter. It should be a very quick fix.

As for the local image building, I don't know what your docker-compose file looks like, but I implemented local image building based on how Vapor does it with the context parameter. Like this: 
```
  app:
    image: assignment-manager-api:latest
    build:
      context: .
```

Finally, as of right now, inter-container communication is supported via passing in ips through the environment. If you name a container in the environment, it can replace the value with the ip. This was done so people didn't have to mess with dns if they don't want to, but I can see how it could become challenging. I don't know enough about how container DNS works, but I could try to implement it. I just worry that it may not go super well and could cause other issues.

Mcrich23:
@niteshbalusu11 fixed

Mcrich23:
While I did add a direct way to change the file path of the docker compose file @jglogan, I also just added inherent support for either .yaml or .yml

Mcrich23:
Also, we might be able to add automatic spec updating, but it could break with edge cases like @niteshbalusu11 encountered where Yams decodes something as a string instead of an array of strings because of a small difference in how the yaml file is written. I would suggest that kind of change to be its own pull request.

niteshbalusu11:
@Mcrich23 I do not know what currently is supported or not. I did pull in your latest changes and it still breaks for me but with a different error.

```
❯ compose up

Error: typeMismatch(Yams.Node.Sequence, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "services", intValue: nil), _DictionaryCodingKey(stringValue: "bark", intValue: nil), CodingKeys(stringValue: "depends_on", intValue: nil)], debugDescription: "Expected to decode Sequence but found Node instead.", underlyingError: nil))
```


If it helps, here's the compose file I am working with:

```
services:
  bitcoind:
    image: bitcoin/bitcoin:29
    command:
      [
        "bitcoind",
        "-regtest",
        "-printtoconsole",
        "-rpcbind=0.0.0.0",
        "-rpcallowip=0.0.0.0/0",
        "-fallbackfee=0.00001",
        "-server",
        "-txindex",
        "-blockfilterindex",
        "-coinstatsindex",
        "-rpcauth=second:23f161a911aa7a75e5ac78a8d41d5313$$cae8671bb16c4e6ac6fcf3e98195bf0a5de8ee9ec536471df3f5caf97bc16437",
      ]
    volumes:
      - bitcoind:/home/bitcoin/.bitcoin
    ports:
      - "18443:18443"

  aspd:
    image: docker.io/secondark/aspd:latest
    volumes:
      - aspd:/data/aspd
      - postgres:/var/lib/postgresql/data
    ports:
      - "3535:3535"

  bark:
    image: docker.io/secondark/bark:latest
    volumes:
      - bark:/root
    depends_on:
      aspd:
        condition: service_started

volumes:
  aspd:
  bark:
  bitcoind:
  postgres:

```

Mcrich23:
Ah. I see. The issue is that you are defining a condition for aspd. Nothing is supported for depends_on beyond a straight array of other services

niteshbalusu11:
Thank you. Doesn't have to be part of this PR but better errors will help :)

Mcrich23:
I can update the parsing so it works and ignores those specifications if we want, but that could cause confusion in the future

Mcrich23:
> Thank you. Doesn't have to be part of this PR but better errors will help :)

Yes, this error is coming straight from the Yams library. Maybe we can implement more descriptive errors in the future, but that is a large undertaking that would be best in a new PR.

Mcrich23:
@jglogan to your question earlier about a compose mvp: 

This is just my opinion, and it isn't what we have to do, but I think that overall, what we have now is a mvp for compose. Outside of dns support, the limiting factor feels to be mostly limitations in Container. While there are some configuration things (like the depends_on stuff) that likely could be implemented without additional work on Container itself, those are very advanced features that most people aren't using and I feel like support for those can come later. Right now, I believe that we should be focusing on tightening the bolts for what we have now (maybe add dns support), and merge this in to allow more people to use Container comfortably.

jglogan:
@niteshbalusu11 

> I do not know what currently is supported or not.

This is why I suggest we start with a minimal PR first, with types generated from the JSON schema, and a README.md that describes exactly what is supported, getting that merged, and then adding features one at a time after that.

My take on MVP for Compose:

- `compose up` and `compose down` read `docker-compose.yml` or `docker-compose.yaml` only
- `compose up` works in interactive mode only, `compose down` to shut the configuration down (this didn't work for me; I had to kill my compose up terminal)
- binds compose file content to data types generated from JSON spec
- Supports only these fields:
  - `name`
  - `services` (zero or more)
    - `image` (pulls from remote repository only)
    - `environment`

Emitting log messages for anything that's populated but not supported is a nice to have, but unless there's a clever way to do that (I can suggest some possibilities) it might be better to live without it in the interest of keeping the PR compact.

How difficult would it be to create a new PR containing that feature set based on what you have now?

Mcrich23:
@jglogan Why not support the other stuff i've already built that work consistently like networking and drive linking? Also, I couldn't figure out how to run code on shutdown which is why compose up actually launches everything and then attaches to containers logs. To shutdown the containers, you have to stop the compose up command and then run compose down.

Mcrich23:
For binding to the json spec, while it might be possible (I would have to look), my guess is that yams will run into enough edge cases with weird but supported yaml files, that it could cause a lot of friction as opposed to implementing stuff by hand. We can start with what quicktype generates, but additional work might be required after it spits stuff out.

jglogan:
> We can start with what quicktype generates, but additional work might be required after it spits stuff out.

This may well be the case. The best way to manage that might be to create a separate `swift-docker-compose-types` library that starts with a `0.1.0-{compose-spec-commit-id}` version for the original generated files of the current compose spec.

If patches are necessary, they are applied to produce `0.1.1`, `0.1.2` and so on.

Should it become necessary to pick up a new version of the compose-spec, that's `0.2.0-{commit-id}`, and whatever patches from 0.1 can be cherry picked to make a `0.2.1`.

jglogan:
> @jglogan Why not support the other stuff i've already built that work consistently like networking and drive linking?

It's simply what I explained above. Let's say I'm submit a contribution to a big open source project, say Linux, or an Apache Foundation or CNCF project, and you're a maintainer making decisions about all the contributions that come in, including mine. You need to make sure that anything that is contributed is maintainable and reliable over time, and not only by the original contributor who may or may not be engaged with the project in the future.

The way to do that is to avoid large PRs like this one, and instead break epic features into small deliverable units just as one would in most any agile software development process. It's far easier to ensure that cross-cutting concerns (logging, error handling, security, and so on) are addressed properly with fewer files, and once those are addressed it establishes a positive pattern for the future commits. It makes everyone's job easier, especially in the case that others see your promising work and want to contribute as well.

From there we add container, network, and storage features one item at a time, with README documentation so it's crystal clear what's currently supported and what isn't.

I can't be sure that anything "works consistently" yet because there are 35 files in the PR and zero tests. Again, with a smaller commit, it's simpler to include unit tests for what can be unit tested (granted, much of what container does is about side effects, but there may be some code here that could benefit from unit coverage).

I hope this explains why, but please follow up if I'm not communicating the need clearly!

jglogan:
> I couldn't figure out how to run code on shutdown which is why compose up actually launches everything and then attaches to containers logs. To shutdown the containers, you have to stop the compose up command and then run compose down.

With a small initial PR, we can focusing on solving that rather large usability problem so that while it's a very limited feature set to start with, it's one that works very well.

Mcrich23:
Thanks for that explanation @jglogan. I am pretty busy for the next month or two, but I can send a sort of "State of Compose" surrounding how key components work and what is supported if you or others want to start pulling my work apart into smaller PRs. Just let me know 😄.

I tried to keep everything documented, which should help this effort.

Also, I totally agree that tests are needed, I just have not had times to write them. Hopefully that will happen as this is broken down into smaller PRs.

jglogan:
@Mcrich23 All good! You've got a great proof of concept in place. There's definitely interest in this feature so I hope we'll all be able to find a little time in coming weeks to take what's here and build out a really cool compose plugin. Better to have it move slowly and thoughtfully than create something too quickly that's difficult for users and a maintenance burden.

Mcrich23:
I have noticed that some repos also do decision markdown files to talk about stuff. Is that something we want to do for compose?

Also, would it be better to do this as a git submodule and maintain a separate issue and pr list for compose?

Mcrich23:
# The State of Compose
This is just a brief overview of current support and how certain things work!

cc: @jglogan @stevenanthonyrevo @niteshbalusu11

> **Note:** For all of these, shortcut definitions may not parse correctly.

## Full/Semi-Supported Features
### Networking
| Feature                          | Supported | Notes |
| -------------------------------- | --------- | -------------------- |
| Attaching Containers to Networks | ✅         |  |
| All Other Features/Configs       | ❌         |                      |

### Environment

| Feature                          | Supported | Notes |
| -------------------------------- | --------- | -------------------- |
| Reading from .env file properties passed through compose file | ✅         |  |
| Shared environment values      | ✅         |                      |
| Resolving Named Services to IP Addresses     | ✅         |   Cross-container communication is easily supported by passing in the ip address through the environment. Just name a container and compose will automatically resolve the IP address of the created container. Make sure to list it as a dependency so the container is already running.  |

### Depended By
| Feature                          | Supported | Notes        |
| -------------------------------- | --------- | -------------------- |
| Basic Service Listing  | ✅         |  |
| All Other Features/Configs for listed services      | ❌         |                      |

### Volume Linking
| Feature                          | Supported | Notes        |
| -------------------------------- | --------- | -------------------- |
| Anonymous Linking  | ✅         | When linking a volume not backed by a location on the host, compose fills that in by linking it to `~/.containers/Volumes/projectName/actualVolumeName`.  | 
| Named Volumes      | ✅         |                      |

### Images
| Feature                          | Supported | Notes        |
| -------------------------------- | --------- | -------------------- |
| Pulling Image | ✅         |  Does easily support private images | 
| Auto Detecting Local Source to Build  | ❌         |   | 
| Normal Dedicated Build Parameters     | ✅         |   Use like <br> `build: ` <br> `  context: .` <br>  `  dockerfile: ./Dockerfile`  |
| Shorthand and Abnormal Build Parameter setups | ❌         |   | 

## Documented Unsupported Features

- Docker Compose Spec Version Switching
- Deploy Config

mazdak:
> To address these things before the merge, I am going to make this pull request into a draft. If you want to try the tool compiled, you can get it at https://github.com/mcrich23/Container-Compose.

@Mcrich23 I have a similar branch with a complete implementation of "compose" with tests and everything. I will try to make this into a standalone plugin as suggested by @jglogan 

https://github.com/mazdak/container/tree/mazdak/container-compose

p.s. how are you working around the permission issues with Postgres (not being allowed to chown/chmod)?

Mcrich23:
@mazdak Sounds good! What do you mean about the issues with Postgres and chown?

mazdak:
> @mazdak Sounds good! What do you mean about the issues with Postgres and chown?

I am unable to get the standard Postgres or redis images to boot up because their entrypoint script tries to modify the mounted volume permissions. This is a limitation of VirtioFS if I understand correctly. You won't notice this issue if you are not using a persistent mounted volume.

Mcrich23:
It could be cause i'm expressly setting the container to be owned by the user? But haven't encountered that error in my testing. Maybe take a look at my implementation

jglogan:
> p.s. how are you working around the permission issues with Postgres (not being allowed to chown/chmod)?

https://github.com/apple/container/issues/333#issuecomment-3076436493

jglogan:
> Finally, as of right now, inter-container communication is supported via passing in ips through the environment. If you name a container in the environment, it can replace the value with the ip. This was done so people didn't have to mess with dns if they don't want to, but I can see how it could become challenging.

Adding the feature for requesting a specific IP (#282) isn't terribly hard, it's just that DNS names work today (but requires configuration as you note) so it's not our top priority.

Mcrich23:
> > p.s. how are you working around the permission issues with Postgres (not being allowed to chown/chmod)?
> 
> 
> 
> https://github.com/apple/container/issues/333#issuecomment-3076436493

Yeah. I guess the way that Vapor does it by default avoids this issue so I never encountered that specifically.

jglogan:
> Regarding Compose types, have you looked into generating the Swift code from the [spec](https://github.com/compose-spec/compose-spec) itself using [quicktype](https://quicktype.io). If that worked then we could automate that (running quicktype under container since it could be a manual step that we perform only when updates are required).

This is a rough sketch of how one could create a standalone swift package that provides a reader using Compose types generated from the [JSON schema specification](https://github.com/compose-spec/compose-go/blob/main/schema/compose-spec.json) (the one under compose-spec/compose-go seems to be the canonical copy).

[swift-docker-compose-types.tgz](https://github.com/user-attachments/files/21519910/swift-docker-compose-types.tgz)

Mcrich23:
@jglogan I totally agree that the ideal way to do this is by autogenerated codable structures via quicktype. My wonder is if totally valid yaml that slightly differs from the mainstream definition will throw errors. One example is: 

```yaml
depends_on: foo
```
versus
```yaml
depends_on: 
  - foo
```

jglogan:
> totally valid yaml

This particular case doesn't seem totally valid.

```console
% docker compose -f docker-compose-syntax.yaml up -d
validating docker-compose-syntax.yaml: services.server.depends_on must be a list
% echo $?
1
```

docker-compose-syntax.yaml:

```
% cat docker-compose-syntax.yaml 
name: compose-test
services:
  db:
    image: postgres:latest
    restart: always
    environment:
      POSTGRES_USER: todo
      POSTGRES_PASSWORD: secret
      POSTGRES_DB: todos
    volumes:
      - pgdata:/var/lib/postgresql/data
    ports:
      - "5432:5432"

  server:
    image: server
    depends_on: db
    ports:
      - "3000:3000"
    environment:
      DATABASE_URL: postgres://todo:secret@compose-test-db.test:5432/todos

volumes:
  pgdata:
```

Mcrich23:
I apologize, that is a bad example, but it is true for other fields.

jglogan:
> I apologize, that is a bad example, but it is true for other fields.

No apologies needed! I'm just curious to see cases where the parser for Docker Compose would admit cases that don't comply to the JSON schema.

This doc describes everything the Compose parser does to transform YAML text into bound types: https://github.com/compose-spec/compose-go/blob/main/parsing.md.

Mcrich23:
@jglogan So sorry for the delay. A lot going on personally including preparing for college (whoohoo!). Anyway, [this](https://github.com/apple/container/pull/239#issuecomment-3119642511) is one example that comes to mind.

I feel like we maybe start with QuickType and find a way to fix these one off issues as we go. The good news is that people won't have an issue as long as people follow typical syntax instead of shortcuts.

Mcrich23:
I wonder if we can make a general decoding pattern that allows all string arrays to be a single string outside of an array, etc.

Maybe member attribute or accessory macros could auto fill it so that just gets prepended to structs and we can still use QuickType.

kmk142789 (approved):

TomasLudvik:
@Mcrich23 any progress here? I can see it has been approved last week. Can we move it forward? I am excited to see this feature in next release :)

ManuW:
Looks like the compose tool is also available as external tool: [container-compose](https://github.com/Mcrich23/Container-Compose).

An internal tool would be good, but it's not a must. “Docker Compose” also started out as an external tool.
I would close this PR for now and refer to the external version.

Mcrich23:
This PR has remained open to help provide information to the community around building a tool. The other compose pr currently open is way more likely to be merged. But at the moment one of the biggest challenges overall is the desire to do small and incremental PRs for the plugin.

atovstonog:
@Mcrich23 any updates here?

Mcrich23:
@atovstonog If you look above, this PR will not be merged due to the size of the PR. It is staying open to push for others to see the conversation as to how it could be built. 

In the meantime, I have made a 3rd party tool called [Container-Compose](https://github.com/mcrich23/container-compose). You can even install it via homebrew!

adamcee:
Hi all. Appreciate the thoughtful discussion around this. I see there is a third party tool [Container-Compose](https://github.com/mcrich23/container-compose) which looks great and which I will be using to get the compose feature with Apple Container.

If possible I'd prefer to have the tool not be third party simply from a dependency management perspective - I installed `container` as recommended, but currently need to use `brew` to install `container-compose`. Let me know if I'm over-thinking this.

But, in that vein, I am trying to read thru the various discussion threads, and it sounds like the issue is the PR size and maybe some elements of the general approach? That the feature set was not minimal enough?

Asking because I'd like to understand this well enough to see if I could take a crack at moving the ball forward on having `compose` not be a 3rd party lib, if that is still desirable. Either by contributing to this PR or using it as a reference. 

If possible could someone summarize:

- Is there still desire for `compose` to be internal?
- At a high level what needs to be done (blockers, etc) to accomplish that as desired?

Or, point me to the thread where that is discussed. There are several related Issues and I was struggling to find the key info.

jglogan:
Closing for now, as we're not considering compose for upstream at this time. Please develop this as an external plugin for now. If you have any questions regarding plugin development (we're still getting docs together) feel free to open a discussion topic!

arch-fan:
What about making Compose a core feature? Or is that not in the plans? I’d be willing to work on an implementation.

Mcrich23:
> Closing for now, as we're not considering compose for upstream at this time. Please develop this as an external plugin for now. If you have any questions regarding plugin development (we're still getting docs together) feel free to open a discussion topic!

Sounds good @jglogan and I am still happy to write those docs. I just feel that it is a waste of time without a way to pass down option groups into subsequent actions.

luisneto:
@jglogan : Would it be possible to help us/me understand the reason(s) you are not considering compose for upstream at this time?
Thanks in advance!

luisneto:
In addition, I'm confused about what's going on here, actually. Your comment says "Closing for now", but this pull request is still open. 
What you did close was pull request https://github.com/apple/container/pull/1394. 

Did you really mean to write that comment here, @jglogan?

jglogan:
@luisneto It's discussed to some extent above, and in discussion #194.

[Docker Compose](https://github.com/docker/compose) is a free-standing project. Its source isn't mingled with Docker Engine, CLI, buildkit, or the SDKs because its concerns are separate from those of the other Docker/Moby ecosystem projects. There's no need to colocate them in one project.

The maintainers of container have limited bandwidth and are focused on the container core functionality at present, trying to refine it and the Containerization library to the point where we can better support plugins for ecosystem integration. Our focus, for now, is on this, and we welcome contributions in this area. We understand the community desire for a Compose-like feature and get that it's frustrating not to have such a key piece of the puzzle (we'd love it too), but upstream in apple/container is not the place for it. @Mcrich23, @mazdak, the socktainer folks, and others are all working hard at this and I'm sure could benefit from contributions as well.

andrew-waters:
@Mcrich23 thanks for keeping this open. The thread is genuinely useful reading for anyone who tries this next, and this is meant as a contribution to that rather than a plug.

I took @jglogan's answer literally and built it as an external plugin: **https://compose.andon.dev**

For anyone arriving here wondering whether the plugin route is actually viable, it is, and it is worth saying plainly: a binary and a `config.toml` in the plugin directory and the stock CLI picks it up, so `container compose up` behaves like any other subcommand without a line of this landing upstream. It builds against released `apple/container` with no patched runtime. The option-group limitation you raised above is real and is the main thing that still makes a plugin more awkward to write than it should be.

The other thing I would offer back is about structure. The hard part here is not the command surface, it is the parsing and the plan, and that is why each of these efforts keeps rewriting the same two things. In mine that part is three libraries whose only dependency is a YAML parser: they cannot start a container and do not link anything that can, and `ContainerAPIClient` appears solely in the executable that carries a plan out. So it can be linked on its own. [Orchard](https://github.com/andrew-waters/orchard) depends on `ComposePlanner` directly and never invokes the plugin binary, and anyone building a UI or another CLI could do the same.

Open source, and I would rather the planning layer became something shared than everyone kept writing their own.

Mcrich23:
No problem! If you don't mind, I would appreciate it if you could speak more to the OptionGroup problem. I think helping the team at Apple understand the use case can make my PR around it more likely to get merged.

PR #243, opened 2025-06-21
Adds `container network` for macOS 26.

jglogan:


Sources/CLI/Application.swift
```diff
@@ -112,42 +109,6 @@ struct Application: AsyncParsableCommand {
         return PluginLoader(pluginDirectories: pluginDirectories, pluginFactories: pluginFactories, defaultResourcePath: statePath, log: log)
     }()
 
-    func validate() throws {
```
jglogan, line 115: NOTE: no changes...just reordered private functions to the end of the compilation unit

Abdullah4345 (approved):

jglogan:
Example usage (mistakenly posted to related container PR, moved here):

```console
% container system start
% container network create foo
foo
% container network ls
NETWORK  STATE    SUBNET
default  running  192.168.64.0/24
foo      running  192.168.65.0/24
% container run -d --name web-default --network default --rm python:slim python3 -m http.server --bind 0.0.0.0 8000
web-default                      
% container run -d --name web-foo --network foo --rm python:slim python3 -m http.server --bind 0.0.0.0 8000        
web-foo                          
% curl http://web-default.test:8000
<!DOCTYPE HTML>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Directory listing for /</title>
</head>
<body>
<h1>Directory listing for /</h1>
...
</body>
</html>
% curl http://web-foo.test:8000
<!DOCTYPE HTML>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Directory listing for /</title>
</head>
<body>
<h1>Directory listing for /</h1>
...
</body>
</html>
% container network rm foo
Error: internalError: "delete failed for one or more networks: ["foo"]"
% container stop web-foo
web-foo
% container network rm foo
foo
%
```

jglogan:
Here's the interface info for a container attached to two networks:

```console
# ip addr 
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 1280 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,UP,LOWER_UP> mtu 1280 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 16:9d:c1:ff:2c:8c brd ff:ff:ff:ff:ff:ff
    inet 192.168.64.2/24 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fd47:1444:7a25:889c:149d:c1ff:feff:2c8c/64 scope global dynamic mngtmpaddr 
       valid_lft 2591912sec preferred_lft 604712sec
    inet6 fe80::149d:c1ff:feff:2c8c/64 scope link 
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,UP,LOWER_UP> mtu 1280 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 3a:ca:72:23:77:80 brd ff:ff:ff:ff:ff:ff
    inet 192.168.65.2/24 scope global eth1
       valid_lft forever preferred_lft forever
    inet6 fd01:b7b1:c35b:1f82:38ca:72ff:fe23:7780/64 scope global dynamic mngtmpaddr 
       valid_lft 2591912sec preferred_lft 604712sec
    inet6 fe80::38ca:72ff:fe23:7780/64 scope link 
       valid_lft forever preferred_lft forever

# ip route
default via 192.168.64.1 dev eth0 
192.168.64.0/24 dev eth0 proto kernel scope link src 192.168.64.2 
192.168.65.0/24 dev eth1 proto kernel scope link src 192.168.65.2 
```

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
```diff
@@ -0,0 +1,100 @@
...
+
+class TestCLINetwork: CLITest {
+    private static let retries = 10
+    private static let retryDelaySeconds = Int64(3)
+
+    @available(macOS 26, *)
+    @Test func testNetworkCreateAndUse() async throws {
+        defer {
+
+        }
```
katiewasnothere: Was this meant to be deleted?

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
```diff
@@ -0,0 +1,100 @@
...
+            let container = try inspectContainer(name)
+            #expect(container.networks.count > 0)
+            let cidrAddress = try CIDRAddress(container.networks[0].address)
+            let url = "http://\(cidrAddress.address):\(port)"
+            var request = HTTPClientRequest(url: url)
+            request.method = .GET
+            let client = getClient()
+            defer { _ = client.shutdown() }
+            var retriesRemaining = Self.retries
+            retryLoop: while retriesRemaining > 0 {
```
katiewasnothere: Could we instead have a boolean indicating success? So then the while loop changes to 

```
while retriesRemaining > 0 && !success {
...
}
```

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
jglogan: It should be deleted, but it prompted me to look over the test once more and I think what might have happened here was that I had some network cleanup code in a defer, but also I needed to delete the network at the start of the test to ensure a known initial condition.

I think it'd be good to do a belt and suspenders for the network cleanup, so I'm testing a change where I defer delete the network at the end of the test, and also delete the network at the start.

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
jglogan: Sure, I can see how that coalesces all the decision making in a single place. I'm modifying the test already so let me make that change.

My motivation for the original pattern is that I try to avoid `var` unless I absolutely need it.

jglogan: The other thing here is that we should never see anything but a 200 if we get an HTTP response, so I think we can `#require(.ok)` here and get rid of the labeled break as well.

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
```diff
@@ -0,0 +1,99 @@
...
+                    let response = try await client.execute(request, timeout: .seconds(Self.retryDelaySeconds))
+                    try #require(response.status == .ok)
+                    success = true
+                } catch {
+                    print("request to \(url) failed, error \(error)")
+                    try await Task.sleep(for: .seconds(Self.retryDelaySeconds))
+                }
+                retriesRemaining -= 1
+            }
+            #expect(success, "Request to \(url) failed after \(Self.retries) retries")
```
katiewasnothere: It's possible we broke out before the full self.retries value. This should be `self.retries - retriesRemaining`.

Issue #203, opened 2025-06-14
[Bug]: Package receipt version is 0


kevinmcox:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

`pkgutil --pkg-info com.apple.container-installer`
package-id: com.apple.container-installer
version: 0

### Current behavior

The version of the release packages (package receipt) is currently `0` and does not match the release version of `0.1.0`.

### Expected Behavior

The package receipt should always match the version of the release.

Enterprise software installation tools rely on the version number to keep corporate fleets patched.

### Environment

```markdown
- OS: macOS 26.0 Beta (25A5279m)
- Xcode: N/A
- Container: Container CLI version 0.1.0
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #249, opened 2025-06-23
Add version info to package installer

katiewasnothere:
Fixes https://github.com/apple/container/issues/203

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ installer-pkg: $(STAGING_DIR)
 	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. --entitlements=signing/container-network-vmnet.entitlements $(join $(STAGING_DIR), libexec/container/plugins/container-network-vmnet/bin/container-network-vmnet)
 
 	@echo Creating application installer
-	@pkgbuild --root $(STAGING_DIR) --identifier com.apple.container-installer --install-location /usr/local $(PKG_PATH)
+	@pkgbuild --root $(STAGING_DIR) --identifier com.apple.container-installer --install-location /usr/local --version ${RELEASE_VERSION} $(PKG_PATH)
 	@rm -rf $(STAGING_DIR)
 
 .PHONY: dsym


Makefile
```diff
@@ -107,7 +107,7 @@ installer-pkg: $(STAGING_DIR)
 	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. --entitlements=signing/container-network-vmnet.entitlements $(join $(STAGING_DIR), libexec/container/plugins/container-network-vmnet/bin/container-network-vmnet)
 
 	@echo Creating application installer
-	@pkgbuild --root $(STAGING_DIR) --identifier com.apple.container-installer --install-location /usr/local $(PKG_PATH)
+	@pkgbuild --root $(STAGING_DIR) --identifier com.apple.container-installer --install-location /usr/local --version ${RELEASE_VERSION} $(PKG_PATH)
```
jglogan, line 110: This will break for any build that is not on a tag, due to the trailing text from `get describe --tags`.

This change will, if there is trailing text, grab the number of commits past the most recent tag and apply it as a fourth component so that you'll get something like:

```console
% pkgutil --pkg-info com.apple.container-installer
package-id: com.apple.container-installer
version: 0.1.0.52
volume: /
location: usr/local
install-time: 1750473741
```

If there's no trailing text, you'll just get the three-component tag.

jglogan (changes requested):

Makefile
katiewasnothere: This is what I see when I run it locally:

```
kathrynbaldauf@Kathryns-MacBook-Pro container % pkgutil --pkg-info com.apple.container-installer
package-id: com.apple.container-installer
version: 0.1.0-53-g1cbb941
volume: /
location: usr/local
install-time: 1751054654
```

I think it's reasonable to have this at the end for packages that are NOT official release packages, such as for a dev testing locally.

jglogan: Yep! My mistake, if pkgutil seems fine with these, I do like that we see everything from `git describe`.

PR #259, opened 2025-06-26
Add container attach/detach support

### LINKED ISSUE #378: [Request]: Attach the terminal to a running container.

This functionality corresponds to `docker container attach` in Docker Desktop.

### Code of Conduct

jacques-n:
Closes #378.

Add container attach/detach support

This change introduces a new `container attach` command that allows users to attach
to running containers with terminal support.

## Key changes:
- Move stdio management for interactive run from client to server so that sessions can exist beyond CLI lifetime.
- Update stdio to use a ring buffer so attach can get recent history.
- Enable legacy (client driven stdio) and server stdio coexistence.

## CLI Changes
- Added `container attach <id>` subcommand for attaching to an existing session.
- Added `--legacy-stdio` run flag to let newer clients talk to older servers.
- Added `--detach-keys` option to run and attach to set detachment keys (defaulted to ctrl-p,ctrl-q)

## Testing
- Added integration tests for attach functionality
- Added session management tests
- Added fallback/legacy mode tests

## Build changes
- Added a couple of additional makefile flags to make it easier to target tests: `INTEGRATION_TEST_FILTER`, `INTEGRATION_TEST_SKIP`, `TEST_FILTER`
- Removed the hand written TestCLI filter list and used a single filter with `--no-parallel`

### DIFF
--- a/Sources/CLI/Container/ContainerAttach.swift
+++ b/Sources/CLI/Container/ContainerAttach.swift
@@ -1,0 +1,82 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the container project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerClient
+import ContainerizationError
+import ContainerizationOS
+import Foundation
+
+extension Application {
+    struct ContainerAttach: AsyncParsableCommand {
+        static let configuration = CommandConfiguration(
+            commandName: "attach",
+            abstract: "Attach to a running container with PTY session support")
+
+        @OptionGroup
+        var global: Flags.Global
+        
+        @Flag(name: .long, help: "Do not send buffered history when attaching")
+        var noHistory = false
+        
+        @Option(name: .customLong("detach-keys"), help: "Override the key sequence for detaching a container")
+        var detachKeys: String?
+
+        @Argument(help: "Container ID or name to attach to")
+        var container: String
+
+        func run() async throws {
+            // Get container info
+            let container = try await ClientContainer.get(id: container)
+            try ensureRunning(container: container)
+            
+            // Check if container has terminal enabled
+            guard container.configuration.initProcess.terminal else {
+                throw ContainerizationError(
+                    .invalidArgument,
+                    message: "Cannot attach to container without terminal enabled"
+                )
+            }
+            
+            log.info("Attaching to container", metadata: ["container": "\(container.id)"])
+            
+            // Create attach client process
+            let process = try await container.attachToSession(sendHistory: !noHistory)
+            
+            // Run the attached session
+            let exitCode = try await runAttachedSession(process: process)
+            
+            if exitCode != 0 {
+                throw ArgumentParser.ExitCode(exitCode)
+            }
+        }
+        
+        private func runAttachedSession(process: ClientProcess) async throws -> Int32 {
+            // Get handles from the attach response
+            let handles = try await process.getAttachHandles()
+            
+            // Create ProcessIO for the attached session with detach key support
+            let io = try ProcessIO.createServerOwned(tty: true, handles: handles, detachKeys: detachKeys)
+            
+            // Run with standard process handling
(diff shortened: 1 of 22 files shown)


jacques-n:
Seeing some failures here. Moving back to draft until resolved. Will also rebase.

jacques-n:
Simplified some things and rebased. Added more integration tests and confirmed passing integration and unit tests.

dcantah:
We haven't forgotten you 😄. Couple questions before I get started. What was the rationale behind needing to support sessions that exist past the lifetime of some attached client? Is that the entire reason behind swapping the direction that IO is established? The other thing I'm not sure of if we'd want to support as well is the ringbuffer of output before attaching. I haven't dug too far into it, but while the asking for the history is optional, it looks like if we ask for a pty we always fill the buffer? To me this feature is somewhat awkward as (and this is likely just me being used to other tools) I'd expect that if I was attaching I would solely be getting output from that point in time. We have `logs` today that can give you a breakdown/historical/last N lines view of IO already, so I'm hesitant to add in this extra logic.

jacques-n:
> We haven't forgotten you 😄

😄

>  What was the rationale behind needing to support sessions that exist past the lifetime of some attached client?

It's pretty common to start a container with a disconnected interactive terminal. I think that is why it is supported in docker. It allows one to separate container creation from exposing that terminal. We use it for starting interactive applications from remote clients without having to guarantee continuous remote connectivity. Think old school client/server. Could one reproduce similar by creating this on top of container by creating a secondary long-lived daemon to hold the sessions? Yes, but it creates a lot of additional complexity and you're basically replicating a bunch of container functionality.

> ring buffer.

Yeah, I considered both options. But the primary use case I think user expectation would be playback. Imagine starting an interactive detached container and then attaching to it a few seconds later. Do you expect to see what just happened? I suspect people would. The best analog would be screen. (I admit that It's nowhere near a perfect analog.) If you ran a command in screen and then detach then reattach, you don't lose context. 

One alt impl would be to internally invoke the log infra. The concern was the complexity of trying to split the stream correctly--e.g. ensuring exactly once semantics. That's what the ring buffer provides that I think would be invasive to add to the logs.

dcantah:
>It's pretty common to start a container with a disconnected interactive terminal.

I think I need to remember how IO was done here (I usually stay over in library land 😬) a tad more, but my remembrance was: We allocate 2 pipes (and 3 if -i was supplied) and send them to the daemon to use to send output to the client. However, the other bit I vaguely remember (you'd think I'd remember more as I added it..) output is always asked for from the process because we always write to a log file in addition to the pipes sent over by the client. This is how `container log` functions, we just tail/read(2) etc. this log file by having the daemon just send us the fd of it over xpc, and if any pipes were sent by the client we just multiplex the writes to the pipes as well. Every container is always ran in a separate process that the daemon spins up, so there should always be something holding open the IO for the lifetime of the container, that's the part that I wasn't super clear on if you found anything that falls over today. I'd imagine with the way IO is setup attach could basically be: send 2/3 new pipe fds to Sandbox process (the process running the container/VM), lock on some data structure that holds the current set of fds to multiplex to, and then continue writing.

dcantah:
After a reworking of some types locally, I'm somewhat convinced server initiated stdio may be a blessing. I'll be looking at this tomorrow morning

Sources/CLI/Container/ContainerAttach.swift
```diff
@@ -0,0 +1,82 @@
...
+        @Argument(help: "Container ID or name to attach to")
+        var container: String
+
+        func run() async throws {
+            // Get container info
+            let container = try await ClientContainer.get(id: container)
+            try ensureRunning(container: container)
+            
+            // Check if container has terminal enabled
+            guard container.configuration.initProcess.terminal else {
```
dcantah, line 47: Why? If you just wanted to attach to see stdout/err really quick this shouldn't be required I'd imagine. Likewise even if you wanted to pipe some stdin I'd think you wouldn't need a pty

Sources/CLI/Container/ContainerAttach.swift
```diff
@@ -0,0 +1,82 @@
...
+            abstract: "Attach to a running container with PTY session support")
+
+        @OptionGroup
+        var global: Flags.Global
+        
+        @Flag(name: .long, help: "Do not send buffered history when attaching")
+        var noHistory = false
+        
+        @Option(name: .customLong("detach-keys"), help: "Override the key sequence for detaching a container")
+        var detachKeys: String?
```
dcantah, line 36: Could not be happier about this 😆 I was honestly hoping someone would implement it.

dcantah:
One thing I can confidently say before reviewing the rest, I think we'd either want to move wholesale to server provided stdio or client and not a mix. I know you left in the old way so it's not an even bigger change/controversial, but we'd definitely just want one or the other. I'm honestly somewhat leaning towards the server end like I mentioned after trying to do a refactor the other day that was miserable, but I'll need to keep reading here and thinking about pitfalls

Mcrich23:

jacques-n:
Closing as not a priority. Others are free to pick up this code and start again if it becomes a priority.

Issue #103, opened 2025-06-10
Is Rosetta required for arm images defined in a Dockerfile?


mac-quantum:
```sh
λ container build -t container-name .
Error: unknown: "Error Domain=VZErrorDomain Code=2 "Rosetta is not installed" UserInfo={NSLocalizedFailure=Invalid virtual machine configuration., NSLocalizedFailureReason=Rosetta is not installed}"
```

I receive this error even if I explicitly mention the architecture with `--arch arm64` is Rosetta a requirement even for arm images?

_Originally posted by @maclong9 in https://github.com/apple/container/discussions/85_

dcantah:
Shouldn't be, I can take a look at what might be going wrong

mac-quantum:
```
# ================================
# Build image
# ================================
FROM swift:6.1-jammy as build

# Install OS updates
RUN export DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true \
    && apt-get -q update \
    && apt-get -q dist-upgrade -y \
    && apt-get install -y libjemalloc-dev \
    && rm -rf /var/lib/apt/lists/*

# Set up a build area
WORKDIR /build

# First just resolve dependencies.
# This creates a cached layer that can be reused
# as long as your Package.swift/Package.resolved
# files do not change.
COPY ./Package.* ./
RUN swift package resolve

# Copy entire repo into container
COPY . .

# Build everything, with optimizations, with static linking, and using jemalloc
RUN swift build -c release \
    --static-swift-stdlib \
    -Xlinker -ljemalloc

# Switch to the staging area
WORKDIR /staging

# Copy main executable to staging area
RUN cp "$(swift build --package-path /build -c release --show-bin-path)/App" ./

# Copy static swift backtracer binary to staging area
RUN cp "/usr/libexec/swift/linux/swift-backtrace-static" ./

# Copy resources bundled by SPM to staging area
RUN find -L "$(swift build --package-path /build -c release --show-bin-path)/" -regex '.*\.resources$' -exec cp -Ra {} ./ \;

# Copy any resouces from the public directory if the directories exist
# Ensure that by default, neither the directory nor any of its contents are writable.
RUN [ -d /build/public ] && { mv /build/public ./public && chmod -R a-w ./public; } || true

# ================================
# Run image
# ================================
FROM ubuntu:jammy

# Make sure all system packages are up to date, and install only essential packages.
[...]

PR #273, opened 2025-06-30
[Build] Disable rosetta during builds with a UserDefault

wlan0:
Fixes https://github.com/apple/container/issues/103

### DIFF
--- a/Sources/CLI/Builder/BuilderStart.swift
+++ b/Sources/CLI/Builder/BuilderStart.swift
@@ -194,7 +194,8 @@ extension Application {
                     options: []
                 ),
             ]
-            config.rosetta = true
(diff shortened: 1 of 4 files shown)


katiewasnothere:
Should we also add a `builder` subcommand to set/unset this?

crosbymichael:
How does a user set this?  We should document either using `defaults` to set this or a subcommand to set this like @katiewasnothere suggested.

I think `defaults` works right now and we could add an issue in case a community member wants to contribute a subcommand.

wlan0:
Updated the docs to explain how to set user defaults in general. That seemed to be missing

katiewasnothere (approved):

PR #274, opened 2025-06-30
Add unpack strategy to SnapshotStore

adityaramani:
Define a `UnpackStrategy` function type in the `SnapshotStore` to give more control over how an image is unpacked.

Previously, we were creating a 512 GB sparse block file for the initial file system of a container, which is overkill.

With this change, the vminit image is unpacked to a smaller block file, while container images are unpacked to the 512 GB block

Follows the same pattern as https://github.com/apple/container/blob/main/Sources/Helpers/RuntimeLinux/RuntimeLinuxHelper.swift#L71

### DIFF
--- a/Sources/Services/ContainerImagesService/Server/SnapshotStore.swift
+++ b/Sources/Services/ContainerImagesService/Server/SnapshotStore.swift
@@ -21,6 +21,7 @@ import ContainerizationExtras
 import ContainerizationOCI
 import ContainerizationOS
 import Foundation
+import Logging
 import TerminalProgress
 
 public actor SnapshotStore {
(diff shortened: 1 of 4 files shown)


Sources/Services/ContainerImagesService/Server/SnapshotStore.swift
```diff
@@ -21,21 +21,41 @@ import ContainerizationExtras
...
 
+    /// Return the Unpacker to use for a given image.
+    /// If the given platform for the image cannot be unpacked return `nil`.
+    public typealias UnpackStrategy = @Sendable (Containerization.Image, Platform) async throws -> Unpacker?
+
+    public static let defaultUnpackStrategy: UnpackStrategy = { image, platform in
+        guard platform.os == "linux" else {
+            return nil
+        }
+        var minBlockSize = 512.gib()
```
jglogan, line 40: Is this a disk/filesystem size, or a block size? Is the `blockSizeInBytes` parameter to EXT4Unpacker a bit misleading?

jglogan (approved):

Sources/Services/ContainerImagesService/Server/SnapshotStore.swift
```diff
@@ -21,21 +21,41 @@ import ContainerizationExtras
...
+        if image.reference == ClientDefaults.get(key: .defaultInitImage) {
+            minBlockSize = 512.mib()
```
crosbymichael, line 42: Let's see if we can move this to a UserDefault in a followup so that end users can customize this size.

crosbymichael (approved):

PR #299, opened 2025-07-05
added ps alias to "list" subcommand

heywoodlh:
As a `docker` main, `docker ps` is one of my go-to commands. This PR adds `ps` as another alias to the `list` subcommand. :smile:

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,12 +1,14 @@
 version: 1
 builder:
   configs:
-    - documentation_targets:
-      - ContainerSandboxService
-      - ContainerNetworkService
-      - ContainerImagesService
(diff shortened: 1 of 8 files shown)


dcantah (approved):
LGTM at least, but I'm sure others might have differing opinions so will wait for one other review

dcantah:
From chatting with the others it doesn't seem like this is a likely accept. We'd like to keep a unified UX for the list APIs, so `ls, list` for any of the resource types (containers, images etc.), and don't want to deviate from this. So while I do think this would be nice for folks with muscle memory from other tools, I don't believe it's worth compromising on that. We really appreciate the contribution!

PR #309, opened 2025-07-07
Mark SandboxSnapshot init as public for sandbox plugins

ArnavReddy:
Changed `SandboxSnapshot` initializer from `package init()` to `public init()` to enable external package consumers to create and use `SandboxSnapshot` instances in their custom Sandbox plugins. 

The `SandboxSnapshot` is used for `container list` functionality


dcantah:
CI's failing for an unrelated reason, but:
```
Test testRunCommandTmpfs() passed after 16.504 seconds.
```

All of the cli tests are taking 15+ seconds which seems insane.. Wonder if `stop()` regressed somehow

adityaramani (approved):

PR #312, opened 2025-07-08
ProcessIO: Don't error if stdin isn't a pty and !-i

dcantah:
We don't need to error unless we're supplying io.

### DIFF
--- a/Sources/CLI/RunCommand.swift
+++ b/Sources/CLI/RunCommand.swift
@@ -194,7 +194,7 @@ struct ProcessIO {
 
     static func create(tty: Bool, interactive: Bool, detach: Bool) throws -> ProcessIO {
         let current: Terminal? = try {
-            if !tty {
+            if !tty || !interactive {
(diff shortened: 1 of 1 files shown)


dcantah:
@jglogan Will undraft after you run your experiments

Sources/CLI/RunCommand.swift
```diff
@@ -194,7 +194,7 @@ struct ProcessIO {
 
     static func create(tty: Bool, interactive: Bool, detach: Bool) throws -> ProcessIO {
         let current: Terminal? = try {
-            if !tty {
+            if !tty || tty && !interactive {
```
jglogan: `if !tty || !interactive`?

dcantah: Good catch, no need for the first

jglogan:
Yep, this worked, thanks!

jglogan (approved):

PR #313, opened 2025-07-08
CLI: Add --platform flag

### LINKED ISSUE #231: [Request]: `container run`/`container create` should support a `--platform=<OS>/<ARCH>` flag

The `docker run` and `docker create` commands take a `--platform=<OS>/<ARCH>` flag to select the OS and architecture of a multiarch image.   `container run` and `container create` do not understand this flag;  they expect separate `--os <OS>` and `--arch <ARCH>` flags.   This means that it is not trivial for a script or tool to support both the `docker` and `container` CLIs at the same time.

`podman` supports both forms:  `--os <OS>` / `--arch <ARCH>` and `--platform <OS>/<ARCH>` (the two forms are exclusive and only one or the other can be used in the same command).  This makes it easy to substitute `podman` for `docker` in a script without rewriting it.

As an example, Swift SDK Generator currently uses the `--platform` flag [here](https://github.com/swiftlang/swift-sdk-generator/blob/af1bea6204fbd43ebd14ad62a6e05233b8ba2d30/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator.swift#L83).

```
 OVERVIEW: Run a container

 USAGE: container run [<options>] <image> [<arguments> ...]

 ARGUMENTS:
   <image>                 Image name
   <arguments>             Container init process arguments

 OPTIONS:
   -w, --cwd, --workdir <cwd>
                           Current working directory for the container
   -e, --env <env>         Set environment variables
   --env-file <env-file>   Read in a file of environment variables
   --uid <uid>             Set the uid for the process
   --gid <gid>             Set the gid for the process
   -i, --interactive       Keep Stdin open even if not attached
   -t, --tty               Open a tty with the process
   -u, --user <user>       Set the user for the process
   -c, --cpus <cpus>       Number of CPUs to allocate to the container
   -m, --memory <memory>   Amount of memory in bytes, kilobytes (K), megabytes
                           (M), or gigabytes (G) for the container, with MB
                           granularity (for example, 1024K will result in 1MB
                           being allocated for the container)
   -d, --detach            Run the container and detach from the process
   --entrypoint <entrypoint>
                           Override the entrypoint of the image
   --mount <mount>         Add a mount to the container
                           (type=<>,source=<>,target=<>,readonly)
   --tmpfs <tmpfs>         Add a tmpfs mount to the container at the given path
   --name <name>           Assign a name to the container. If excluded will be a
                           generated UUID
   --remove, --rm          Remove the container after it stops
>  --os <os>               Set OS if image can target multiple operating systems
>                          (default: linux)
>  -a, --arch <arch>       Set arch if image can target multiple architectures
>                          (default: arm64)
   -v, --volume <volume>   Bind mount a volume into the container
   -k, --kernel <kernel>   Set a custom kernel path
   --cidfile <cidfile>     Write the container ID to the path provided
   --no-dns                Do not configure DNS in the container
   --dns <dns>             DNS nameserver IP address
   --dns-domain <dns-domain>
                           Default DNS domain
   --dns-search <dns-search>
                           DNS search domains
   --dns-option <dns-option>
                           DNS options
   -l, --label <label>     Add a key=value label to the container
   --disable-progress-updates
                           Disable progress bar updates
   --scheme <scheme>       Scheme to use when conntecting to the container
                           registry. One of (http, https, auto) (default: auto)
   -h, --help              Show help information.
```

### Code of Conduct

dcantah:
Fixes #231

--platform is useful as it's what most folks are used to from other container tools. We support two separate flags today that when combined (--os and --arch) do essentially the same thing, but this is a combined single string form.

### DIFF
--- a/Sources/CLI/BuildCommand.swift
+++ b/Sources/CLI/BuildCommand.swift
@@ -87,6 +87,9 @@ extension Application {
             ["linux"]
         }()
 
+        @Option(name: .long, help: "set the builds platform. This takes precedence over --os and --arch")
+        var platform: [String] = []
+
         @Option(name: .long, help: ArgumentHelp("Progress type - one of [auto|plain|tty]", valueName: "type"))
         var progress: String = "auto"
 
@@ -219,6 +222,16 @@ extension Application {
                     }
                     let platforms: [Platform] = try {
                         var results: [Platform] = []
+                        if !self.platform.isEmpty {
+                            for platform in self.platform {
+                                guard let p = try? Platform(from: platform) else {
+                                    throw ValidationError("invalid platform specified \(platform)")
+                                }
+                                results.append(p)
+                            }
+                            return results
+                        }
+
                         for o in self.os {
                             for a in self.arch {
                                 guard let platform = try? Platform(from: "\(o)/\(a)") else {
--- a/Sources/CLI/Image/ImagePull.swift
+++ b/Sources/CLI/Image/ImagePull.swift
@@ -37,7 +37,15 @@ extension Application {
         @OptionGroup
         var progressFlags: Flags.Progress
 
-        @Option(help: "Platform string in the form 'os/arch/variant'. Example 'linux/arm64/v8', 'linux/amd64'") var platform: String?
+        @Option(help: "Platform string in the form 'os/arch/variant'. Example 'linux/arm64/v8', 'linux/amd64'. This takes precedence over --os and --arch")
+        var platform: String?
+
+        @Option(name: .customLong("os"), help: "Set OS if image can target multiple operating systems")
+        public var os: String?
+
+        @Option(
+            name: [.customLong("arch"), .customShort("a")], help: "Set arch if image can target multiple architectures")
+        public var arch: String?
 
         @Argument var reference: String
 
@@ -55,6 +63,10 @@ extension Application {
             var p: Platform?
             if let platform {
                 p = try Platform(from: platform)
+            } else if os != nil || arch != nil {
+                let o = os ?? ""
+                let a = arch ?? ""
+                p = try Platform(from: "\(o)/\(a)")
             }
 
             let scheme = try RequestScheme(registry.scheme)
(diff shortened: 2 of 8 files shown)


tamird:
Will this also work with `container build`?

dcantah:
@tamird In it's current state here no, but it's trivial to add support if we want. My team is currently debating whether we want this in general though, so we'll see 😬. If they do want this flag I'll add to build as well.

tamird:
Sounds good. More broadly, do you folks care about compatibility with docker's porcelain?

dcantah:
@tamird It depends. We'd like to not be beholden to the UX of other tools, but I'd say we'd take any suggestions on a case by case basis if we think it's a good idea. Do you have any things you'd like to see besides --platform as an example?

tamird:
@dcantah nah, currently this is the main one I need. It's a bummer that `--os` and `--arch` are *not* accepted by docker. This makes it difficult to use `container` as a drop-in replacement the way I'd have liked.

jglogan:
Let's not take a scattershot approach to this. With the change this is where we are:

- `container create` and `container run` support `--os`, `--arch`, and `--platform`
- `container build` supports `--os` and `--arch`
- ``container image pull`, `container image push`, and `container image save` support `--platform`

This PR should bring these commands (and any others that deal in platform specification) into harmony, and we should have sane input validation to ensure that the CLI rejects invalid or ambiguous combinations with clear error messages.

I think that `--platform` makes sense because it not only helps with tooling compatibility (the reason for the ask in the first place), but it also covers the relatively rare case where the user needs to specify a variant.

dcantah:
Ack, will amend this Thursday

tamird:
Another thing to mention about the `--platform` flag is that in other tools it is used to [build multiarch images](https://docs.docker.com/build/building/multi-platform/).

It would be great to have support for that here, too.

dcantah:
@tamird Yep. I think it's worthwhile to have it on build/pull/create/run.

tamird:
@dcantah just to clarify further: in order to build multiarch images you pass a comma-separate list of platforms as the value of `--platform` e.g. `--platform linux/amd64,linux/arm64`.

dcantah:
@tamird Yep. It's tricky as I don't believe the argument parser we use will give us that by default. I'll have to write some extra glue.

dcantah:
I keep trying to find time to update this but to no avail. @jglogan may take a stab at it

egernst:
I'll go ahead and close this one since we have this addressed in https://github.com/apple/container/pull/545. Thanks Danny.

PR #318, opened 2025-07-09
Use `swift package edit` instead of `CONTAINERIZATION_PATH`.

### LINKED ISSUE #294: [Request]: Use `swift package edit` instead of `CONTAINER_PATH` for local development.

`CONTAINER_PATH` was introduced in Package.swift as a workaround for the seeming problem that `swift package edit` wasn't well supported in Xcode.

However, as this is an SPM project, it's much cleaner to rely on that than to carry around cruft in Package.swift.

There are two parts to the work:

- [ ] Remove the conditional dependency setup from Package.swift.
- [ ] Update [scripts/install-init.sh](https://github.com/apple/container/blob/main/scripts/install-init.sh) so that it supports branch dependencies as well as path dependencies.

### Code of Conduct

jglogan:
- Closes #294.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -6,6 +6,9 @@ To build the `container` project, you need:
 - macOS 15 minimum, macOS 26 beta recommended
 - Xcode 26 beta
 
+> [!IMPORTANT]
+> There is a bug in the `vmnet` framework on macOS 26 beta that causes network creation to fail if the `container` helper applications are located under your `Documents` or `Desktop` directories. If you use `make install`, you can simply run the `container` binary in `/usr/local`. If you prefer to use the binaries that `make all` creates in your project `bin` and `libexec` directories, locate your project elsewhere, such as `~/projects/container`, until this issue is resolved.
+
 ## Compile and test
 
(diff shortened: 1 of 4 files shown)


katiewasnothere (approved):

katiewasnothere (approved):

BUILDING.md
```diff
@@ -38,58 +41,60 @@ to prepare your build environment.
...
+    /usr/bin/swift package edit --path ../containerization containerization
+    /usr/bin/swift package update containerization
     ```
 
-6. Build `container`.
+    > [!IMPORTANT]
+    > If you are using Xcode, you will need to temporarily modify `Package.swift` instead of using `swift package edit`, using a path dependency in place of the versioned `container` dependency:
+    >
+    >    ```swift
+    >    .package(.path: "../containerization"),
```
dkovba, line 65: @jglogan It should be `path` - not `.path`.

katiewasnothere: This has been resolved with https://github.com/apple/container/pull/327 already

PR #322, opened 2025-07-10
CLI: Handle stdin stream if it's a regular file

dcantah:
readabilityHandler is a bit crummy if it's a regular file. This just writes our own loop and exits and shuts down the write end of the pipe if we get to the end of the file.

### DIFF
--- a/Sources/CLI/RunCommand.swift
+++ b/Sources/CLI/RunCommand.swift
@@ -214,13 +214,21 @@ struct ProcessIO {
         if let stdin {
             if interactive {
                 let pin = FileHandle.standardInput
-                pin.readabilityHandler = { handle in
-                    let data = handle.availableData
-                    if data.isEmpty {
+                let stdinOSFile = OSFile(fd: pin.fileDescriptor)
+                let pipeOSFile = OSFile(fd: stdin.fileHandleForWriting.fileDescriptor)
+                try stdinOSFile.makeNonBlocking()
+                nonisolated(unsafe) let buf = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: Int(getpagesize()))
+
+                pin.readabilityHandler = { _ in
+                    Self.streamStdin(
+                        from: stdinOSFile,
+                        to: pipeOSFile,
+                        buffer: buf,
+                    ) {
                         pin.readabilityHandler = nil
-                        return
+                        buf.deallocate()
+                        try? stdin.fileHandleForWriting.close()
                     }
-                    try! stdin.fileHandleForWriting.write(contentsOf: data)
                 }
             }
             stdio[0] = stdin.fileHandleForReading
@@ -294,6 +302,39 @@ struct ProcessIO {
         )
     }
 
+    static func streamStdin(
+        from: OSFile,
+        to: OSFile,
+        buffer: UnsafeMutableBufferPointer<UInt8>,
+        onErrorOrEOF: () -> Void,
+    ) {
+        while true {
+            let (bytesRead, action) = from.read(buffer)
+            if bytesRead > 0 {
+                let view = UnsafeMutableBufferPointer(
+                    start: buffer.baseAddress,
+                    count: bytesRead
+                )
+
+                let (bytesWritten, _) = to.write(view)
+                if bytesWritten != bytesRead {
+                    onErrorOrEOF()
+                    return
+                }
+            }
+
+            switch action {
+            case .error(_), .eof, .brokenPipe:
+                onErrorOrEOF()
+                return
+            case .again:
+                return
+            case .success:
+                break
+            }
+        }
+    }
+
     public func wait() async throws {
         guard let ioTracker = self.ioTracker else {
             return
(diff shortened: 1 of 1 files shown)


dcantah:
@jglogan This hopefully should get piping a file in working, will test tomorrow. We'll need both this and bump of containerization that has the closing stdin work however.

jglogan:
@dcantah 

Can you check to see if terminal I/O works for you in this branch with:

```bash
container run -it --rm alpine:latest sh
```

dcantah:
@jglogan It doesn't, it's why it's in draft right now. `FileHandle.read(upToCount` reads until EOF by default, so I'll need to write my own read(2) loop (and make stdin nonblocking).

dcantah:
@jglogan tty fixed, and I can confirm we close the relay on eof for regular files. What's the avenue for testing against containerization `main` these days? It seem `make init-block` returns the version as unspecified

Sources/CLI/RunCommand.swift
```diff
@@ -315,3 +356,96 @@ struct ProcessIO {
         }
     }
 }
+
+struct OSFile: Sendable {
```
dcantah, line 360: Note: I've just copied a smaller portion of this same type we have for vminitd here. I'll see if it makes sense to make that public over in CZ, and we can just use directly.

jglogan:
> What's the avenue for testing against containerization `main` these days?

What's missing from [the build documentation](https://github.com/apple/container/blob/main/BUILDING.md#develop-using-a-local-copy-of-containerization) is that you need to set up the default init image as `vminit:latest`, and then restart everything so that both the CLI and the API server pick up the correct default init image.

```bash
make all install
defaults write com.apple.container.defaults image.init vminit:latest
container system stop
container system start
container run -d --rm --name test python:slim python3 -m http.server --bind 0.0.0.0 8000
echo hello | container exec -t exec -i test cp /dev/stdin /tmp/hello.txt
```

FWIW I'm seeing that this seems to suffice for the `exec` test:
```swift
diff --git a/Sources/CLI/RunCommand.swift b/Sources/CLI/RunCommand.swift
index 9a7f0d7..1d3f0cc 100644
--- a/Sources/CLI/RunCommand.swift
+++ b/Sources/CLI/RunCommand.swift
@@ -218,6 +218,7 @@ struct ProcessIO {
                     let data = handle.availableData
                     if data.isEmpty {
                         pin.readabilityHandler = nil
+                        try? stdin.fileHandleForWriting.close()
                         return
                     }
                     try! stdin.fileHandleForWriting.write(contentsOf: data)
```

That we fetch the init filesystem in the client at https://github.com/apple/container/blob/main/Sources/ContainerClient/Utility.swift#L113, and then again when creating the container in the API server https://github.com/apple/container/blob/main/Sources/APIServer/Containers/ContainersService.swift#L165 is curious.

Sources/CLI/RunCommand.swift
```diff
@@ -214,13 +214,21 @@ struct ProcessIO {
         if let stdin {
             if interactive {
                 let pin = FileHandle.standardInput
-                pin.readabilityHandler = { handle in
-                    let data = handle.availableData
-                    if data.isEmpty {
+                let stdinOSFile = OSFile(fd: pin.fileDescriptor)
+                let pipeOSFile = OSFile(fd: stdin.fileHandleForWriting.fileDescriptor)
+                nonisolated(unsafe) let buf = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: Int(getpagesize()))
+                try stdinOSFile.makeNonBlocking()
```
jglogan, line 219: does an exception here leak `buf`?

Sources/CLI/RunCommand.swift
```diff
@@ -294,6 +302,39 @@ struct ProcessIO {
...
     }
 
+    static func streamStdin(
+        from: OSFile,
+        to: OSFile,
+        buffer: UnsafeMutableBufferPointer<UInt8>,
+        onErrorOrEOF: () -> Void,
+    ) {
+        while true {
+            let r = from.read(buffer)
```
jglogan: destructure into `(bytesRead, action)` for readability

or at least rename `r.read` to `r.bytesRead`, `r.readCount`, etc

Sources/CLI/RunCommand.swift
```diff
@@ -294,6 +302,39 @@ struct ProcessIO {
...
+            if r.read > 0 {
+                let view = UnsafeMutableBufferPointer(
+                    start: buffer.baseAddress,
+                    count: r.read
+                )
+
+                let w = to.write(view)
+                if w.wrote != r.read {
```
jglogan: destructure into `(bytesWritten, _)`

make it obvious that action is not used (and prompts the reviewer to check that an `w.action is not ok ⇔ bytesRead != bytesWritten`)

jglogan (changes requested):
mainly readability nits

Sources/CLI/RunCommand.swift
dcantah: Good catch yep, although if we did throw we'd just exit the process so not too much of a worry. I can move this one line down so we mark as non blocking first

Issue #326, opened 2025-07-11
[Bug]: Access the container from another container


zdk:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. https://github.com/apple/container && cd container 
2. Ensure in main branch, run `make all test integration && make install`
3. Run, container system start
4. Follow tutorial doc https://github.com/apple/container/blob/main/docs/tutorial.md from the start until this section https://github.com/apple/container/blob/main/docs/tutorial.md#access-the-web-server-from-another-container

### Current behavior

Got the error: `curl: (7) Failed to connect to 192.168.64.4 port 80 after 3073 ms: Could not connect to server` as shown in the screenshot:

<img width="1724" height="458" alt="Image" src="https://github.com/user-attachments/assets/242c6065-3089-417f-b0dc-9ed9d2c27815" />

### Expected behavior

```
% container run -it --rm web-test curl http://192.168.64.4
<!DOCTYPE html><html><head><title>Hello</title></head><body><h1>Hello, world!</h1></body></html>
```

### Environment

```markdown
- OS: macOS 15.5 (24F74)
- Xcode: Version 26.0 beta 3 (17A5276g)
- Container: container CLI version 0.2.0-15-g7272328 (build: debug, commit: 7272328)
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

afbjorklund:
https://github.com/apple/container/blob/main/docs/technical-overview.md#macos-15-limitations

> Container-to-container communication over the virtual network is not possible.

zdk:
@afbjorklund Oh, i see. Thanks!

afbjorklund:
Does make it slightly less useful...

zdk:
@afbjorklund Just followed along the tuturial and definitely, I overlook the `technical-overview.md` doc, ha.

afbjorklund:
There should probably be a warning in the tutorial, next to the instructions to do something that does not work.

zdk:
@afbjorklund Let me submit the PR for it 😂

afbjorklund:
I think you want to run containers instead of containers, if you want something that works also on macOS 15 and earlier.

_(some find the terminology used here confusing)_

i.e. Linux containers versus Apple containers

![Image](https://github.com/user-attachments/assets/27175be5-2179-457c-be96-1b60aea04885)

[...]

PR #328, opened 2025-07-11
Docs: Add macOS 15 container-to-container communication limitation note to the tutorial

zdk:
Add note about macOS 15 network isolation preventing container-to-container communication

Ref: https://github.com/apple/container/issues/326


afbjorklund:
Here is a workaround:

* #329

zdk:
> Here is a workaround:
> 
> * [Container-to-container networking with NoRouter #329](https://github.com/apple/container/discussions/329)

Aweeesomeeee, thanks!

The workaround works brilliantly (as it should work 😂 )

As shown in the following screenshot:

CleanShot 2568-07-12 at 16 07 22@2x

zdk:
Gist: https://gist.github.com/zdk/575fbb36a913e20f21f6c10e8ec26734
For those who need the actual config.

Issue #254, opened 2025-06-24
[Bug]: `container image pull` does not work for Azure Container Registry even if logged in (401 Unauthorized)


vnea:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. Login to the Azure Regisry (I followed [this documentation](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-authentication?tabs=azure-cli#az-acr-login-with---expose-token)) :

```sh
TOKEN=$(az acr login --name myregistry --expose-token --output tsv --query accessToken)
container registry login myregistry.azurecr.io --username 00000000-0000-0000-0000-000000000000 --password-stdin <<< $TOKEN
```

2. I got the message:
```
Login succeeded
```

3. Try to pull an image:
```
container image pull myregistry.azurecr.io/api/dummy:latest
```

### Current behavior

I got the follow error:

```
Error: internalError: "HTTP request to https://myregistry.azurecr.io/oauth2/token?client_id=containerization-registry-client&service=myregistry.azurecr.io&scope=repository:api/dummy:pull failed with response: 401 Unauthorized. No credentials found for host myregistry.azurecr.io"
```

### Expected Behavior

The image should be pulled.

### Environment

```markdown
- OS: 15.5 (24F74)
- Xcode: N/A
- Container: container CLI version 0.2.0 (build: release, commit: 3b5c253)
```

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

katiewasnothere:
Hi @vnea thanks for the report! Can you confirm that the pull command prompted you to sign into your keychain?

vnea:
> Hi [@vnea](https://github.com/vnea) thanks for the report! Can you confirm that the pull command prompted you to sign into your keychain?

No it didn't ask me to do that, it directly showed me the error.

vnea:
Also I don't know if it can help to resolve the issue but when I run this command:

```sh
 TOKEN=$(az acr login --name myregistry --expose-token --output tsv --query accessToken)
```

It gives me this warning message:

```
[...]

PR #331, opened 2025-07-11
Handle when keychain query returns an unhandled error

katiewasnothere:
Depends on https://github.com/apple/containerization/pull/210
Related to https://github.com/apple/container/issues/254

### DIFF
--- a/Sources/Services/ContainerImagesService/Server/ImageService.swift
+++ b/Sources/Services/ContainerImagesService/Server/ImageService.swift
@@ -166,7 +166,13 @@ extension ImagesService {
             return try await body(authentication)
         }
         let keychain = KeychainHelper(id: Self.keychainID)
-        authentication = try? keychain.lookup(domain: host)
+        do {
+            authentication = try keychain.lookup(domain: host)
+        } catch let err as KeychainHelper.Error {
+            guard case .keyNotFound = err else {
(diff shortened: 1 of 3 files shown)


Sources/Services/ContainerImagesService/Server/ImageService.swift
```diff
@@ -166,7 +166,13 @@ extension ImagesService {
             return try await body(authentication)
         }
         let keychain = KeychainHelper(id: Self.keychainID)
-        authentication = try? keychain.lookup(domain: host)
+        do {
+            authentication = try keychain.lookup(domain: host)
+        } catch let err as KeychainHelper.Error {
+            guard case .keyNotFound = err else {
+                throw ContainerizationError(.internalError, message: "\(String(describing: err)). Error querying keychain for \(host)")
```
dcantah: `ContainerizationError` has a `cause` field that you can include the original error on:
```swift
throw ContainerizationError(.internalError, message: "Error querying keychain for \(host)", cause: err)
```

dcantah: We also don't handle any of the other ways `.get()` and `.exists()` can fail internally in `.lookup()`, this only catches `.keyNotFound`.

katiewasnothere: We intentionally only want to catch the keyNotFound error, since we may still be able to pull/push the image in that scenario.

PR #335, opened 2025-07-15
Make image details public

Typ0genius:
I want to access the different variants of an image in my app, which information is only available in the ImageDetail. Although the function ClientImage.details() -> ImageDetail and its return type are already public, its properties aren’t, which prevents access to this information.

### DIFF
--- a/Sources/ContainerClient/Core/ImageDetail.swift
+++ b/Sources/ContainerClient/Core/ImageDetail.swift
@@ -18,14 +18,14 @@ import Containerization
 import ContainerizationOCI
 
 public struct ImageDetail: Codable {
-    let name: String
-    let index: Descriptor
-    let variants: [Variants]
+    public let name: String
+    public let index: Descriptor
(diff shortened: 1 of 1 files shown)


katiewasnothere:
Hi @Typ0genius thanks for the change! Could you add information to the PR description on the motivation behind this change?

Typ0genius:
@katiewasnothere Sure, I have updated the description.

katiewasnothere (approved):

LiamKearn:
Forgive if I'm wrong but I'm very much dropping into the project looking at recent commits to see if stuff I want has happened.

Isn't this the CLI/tool and public programmatic APIs should be done via the swift framework?

dcantah:
@LiamKearn While this repo IS where you can build the daemon/CLI, the project is still importable/usable as a dependency if you wanted to interact with the daemon programmatically as it contains the client we use in the cli.

Issue #302, opened 2025-07-06
[Request]: add started date in containers metadata


benoitf:
### Feature or enhancement request details

if I have like 10 containers running, I find interesting to be able to see which one was started recently, (or could sort it)

but `container ls --format json | jq .` is not showing any date

it's the `CreatedAt` or `Created` field in docker/podman engine

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

zk2k2:
Hi, can you please assign me this? I would like to contribute, thanks!

zk2k2:
Hello @adeebashraf @benoitf , I have submitted a PR for the issue, let me know if I need to change anything, thanks!

PR #336, opened 2025-07-15
Add StartedAt field to container list output

zk2k2:
## Summary

This pull request adds support for tracking and exposing the container `startedAt` timestamp throughout the container lifecycle.

Fixes [apple/container#302](https://github.com/apple/container/issues/302).

## Changes

- Added `startedAt: Date?` to:
  - `ContainerSnapshot`
  - `ClientContainer`
  - `PrintableContainer`
- Timestamp is assigned in `ContainersService.containerStartHandler` using `Date()`
- Propagated the value through `Item.asSnapshot()`
- Included `startedAt` in:
  - `container list --format json` output
  - Optionally included in `asRow` for table display (ISO 8601 format)

### DIFF
--- a/Sources/APIServer/Containers/ContainersService.swift
+++ b/Sources/APIServer/Containers/ContainersService.swift
@@ -41,6 +41,7 @@ actor ContainersService {
     struct Item: Sendable {
         let bundle: ContainerClient.Bundle
         var state: State
+        var startedAt: Date?
 
         enum State: Sendable {
             case dead
@@ -81,7 +82,7 @@ actor ContainersService {
             do {
                 let bundle = ContainerClient.Bundle(path: dir)
                 let config = try bundle.configuration
-                results[config.id] = .init(bundle: bundle, state: .dead)
+                results[config.id] = .init(bundle: bundle, state: .dead, startedAt: nil)
                 let plugin = runtimePlugins.first { $0.name == config.runtimeHandler }
                 guard let plugin else {
                     throw ContainerizationError(.internalError, message: "Failed to find runtime plugin \(config.runtimeHandler)")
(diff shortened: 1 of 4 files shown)


jglogan:
@zk2k2 Can you run `make fmt` on your end and push the changes? Right now your PRB breaks because format checks are failing.

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -41,6 +41,7 @@ actor ContainersService {
     struct Item: Sendable {
         let bundle: ContainerClient.Bundle
         var state: State
+        var startedAt: Date?
```
jglogan, line 44: `startedAt` is really only relevant when the container is alive, isn't it? We might want to consider [associating this value](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/#Associated-Values) with `State.alive`.

jglogan: The thing seems a bit awkward on the SandboxService side of things (no fault of yours, you're just having to work with it) is that we have a model for runtime state that uses different types on the client (RuntimeStatus) and server (State) side.

The approach in the NetworkService is to have a NetworkState enum type that has an associated value for configuration data for the non-running states, and both config and runtime state values for the running states, and NetworkState is used on both client and server.

Let me look into whether we can take this approach here and I'll offer some guidance.

jglogan: @zk2k2 If you're still interested in moving this forward, have a look at the ContainersService now. The `ContainersService.Item` type has now been eliminated which should make adding this field a lot simpler.

zk2k2: Hi @jglogan, sure I'll check it out

Mcrich23:

kmk142789 (approved):

saehejkang:
@zk2k2 Are you still interested in wrapping this PR up?

zk2k2:
Hi @saehejkang , sorry but I've actually been busy with other stuff, it's all yours if you want it :)

PR #338, opened 2025-07-15
Adds TCP and UDP port forwarders.

jglogan:


Sources/SocketForwarder/GlueHandler.swift
```diff
@@ -0,0 +1,121 @@
...
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+
+final class GlueHandler {
+
+    private var partner: GlueHandler?
+
+    private var context: ChannelHandlerContext?
+
```
dkovba, line 24: Might be better to remove empty lines.

Sources/SocketForwarder/UDPForwarder.swift
```diff
@@ -0,0 +1,211 @@
...
+    typealias OutboundOut = AddressedEnvelope<ByteBuffer>
+
+    private let clientAddress: SocketAddress
+    private let serverAddress: SocketAddress
+    private let frontendChannel: any Channel
+    private let log: Logger?
+
+    private let lock: NSLock = NSLock()
+    private nonisolated(unsafe) var queuedPayloads: Deque<ByteBuffer>
+    private nonisolated(unsafe) var channel: (any Channel)?
```
dkovba: Consider replacing with a state struct inside a `Mutex`:
```
    private let state: Mutex<State>
    
    private struct State {
        var queuedPayloads: Deque<ByteBuffer>
        var channelTest: (any Channel)?
    }
```
In the initializer:
```
        let state = State(queuedPayloads: Deque(), channelTest: nil)
        self.state = Mutex(state)
```
An example of a use:
```
        state.withLock {
            if !$0.queuedPayloads.isEmpty { ... }
        }
```

Sources/SocketForwarder/UDPForwarder.swift
```diff
@@ -0,0 +1,211 @@
...
+    typealias OutboundOut = AddressedEnvelope<ByteBuffer>
+    private let maxProxies = 256
+
+    private let proxyAddress: SocketAddress
+    private let serverAddress: SocketAddress
+    private let eventLoopGroup: any EventLoopGroup
+    private let log: Logger?
+
+    private nonisolated(unsafe) var proxies: LRUCache<String, ProxyContext>
+    private let lock = NSLock()
```
dkovba: Consider replacing with a `Mutex`:
```
    private let proxies: Mutex<LRUCache<String, ProxyContext>>
```
In the initializer:
```
        self.proxies = Mutex(LRUCache())
```
An example of a use:
```
            try proxies.withLock {
                if let context = $0.get(key) { ... }
            }
```

jglogan: also done

Sources/SocketForwarder/GlueHandler.swift
jglogan: done, thx!

Sources/SocketForwarder/UDPForwarder.swift
jglogan: done, thx!

Sources/SocketForwarder/UDPForwarder.swift
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    private let clientAddress: SocketAddress
+    private let serverAddress: SocketAddress
+    private let frontendChannel: any Channel
+    private let log: Logger?
+
+    private let state: Mutex<State>
+
+    private let lock: NSLock = NSLock()
```
dcantah: Is lock used anywhere anymore?

Sources/SocketForwarder/LRUCache.swift
```diff
@@ -0,0 +1,107 @@
...
+    public func get(_ key: K) -> V? {
+        guard let node = members[key] else {
+            return nil
+        }
+        listRemove(node: node)
+        listInsert(node: node, after: tail)
+        return node.value
+    }
+
+    public func put(key: K, value: V) throws {
```
dcantah: Is this supposed to be a general purpose lru cache? I'd expect if someone put a new value in for a key we already had we'd just update the entry (and fix up the positioning).

Sources/SocketForwarder/LRUCache.swift
```diff
@@ -0,0 +1,107 @@
...
+        if members[key] != nil {
+            throw KeyExistsError()
+        }
+
+        let node = Node(key: key, value: value)
+        listInsert(node: node, after: tail)
+        members[key] = node
+    }
+
+    public func evict() -> (K, V)? {
```
dcantah: I'm surprised to see this, I'd think the constructor would take in some bounds (which for this is the maximum number of proxies we allow) and then the evictions would happen internally in the object. I see we use this so that we can get a handle to what is being evicted so we can call close on it, but couldn't the type we're adding just have a destructor that calls close internally? It didn't seem like `ProxyContext.context.close()` throws so this should function

dcantah: The reason this is somewhat odd to me is the eviction is handled externally by the user of this structure which doesn't seem common, although this is a tailored type for our needs so it's not the biggest concern.

Sources/SocketForwarder/LRUCache.swift
```diff
@@ -0,0 +1,107 @@
...
+            self.key = key
+            self.value = value
+        }
+    }
+
+    private var head: Node?
+    private var tail: Node?
+    private var members: [K: Node]
+
+    public init() {
```
dcantah: nit: The constructor and methods likely don't need to be public as the type itself isn't

Sources/SocketForwarder/UDPForwarder.swift
jglogan: that should be gone, forgot to delete, thanks!

Sources/SocketForwarder/LRUCache.swift
jglogan: see below

Sources/SocketForwarder/LRUCache.swift
jglogan: When I was writing this I thought I'd put the minimal amount of stuff in the cache and leave the sizing entirely to the client. I expressly didn't want to couple the cache with anything special about the type; that is to say I didn't want to make it a cache of `V: Evictable` where that protocol defined some `evictYourself()`.

Perhaps it's better called a LRUTracker since it's up to the client to bound the size, whereas LRUCache evokes a cache with a predefined maximum number of entries?

I'm fine giving it fixed size behavior. I think if we had `LRUCache(size: UInt)` and changed put to:

```swift
func put(key: K, value: V) -> V?
```

maybe that makes the semantics more clear?

dcantah: Yea, that could work!

jglogan: I'll push the other changes now and then knock out that change.

jglogan: done

Sources/SocketForwarder/LRUCache.swift
jglogan: replaced throw with `-> (K, V)?` as discussed

Issue #97, opened 2025-06-10
Mapping Ports?


scottfwalter:
Is there a way map ports?  For example port 8088 on my local machine, but port 80 inside the container?  I looked at the how to documentation page and I did not find any info.  Thanks.

BenHall:
You should be able to solve this using `socat` to bind 8088 to the IP:80 of the container, similar to https://github.com/apple/container/blob/main/docs/technical-overview.md#container-to-host-networking

scottwalter-nice:
Hopefully, a command line option will be added for a better developer experience.

lperdereau:
Hey everyone!

I wanted to share a handy shell alias I created for forwarding ports to a container using socat. This alias simplifies the process and makes it easier to manage port forwarding with just a single command. 😅 

Note: I'm note proud of it

```sh
alias container-port-forward='function _container_port_forward() {
    local container_id="$1"
    local src_port="$2"
    local dst_port="$3"
    sudo socat TCP-LISTEN:"$src_port",fork TCP:$(container inspect "$container_id" | jq -r ".[0].networks[0].address" | cut -d'/' -f1):"$dst_port"
}; _container_port_forward'
```

```sh
container-port-forward <containerd-id> <local-port> <container-port>
```

scottfwalter:
@lperdereau  nice job!

tidbeck:
Thanks for the workaround @lperdereau! Is there any similar approach to reach a service inside the container that binds to `127.0.0.1` (as it will normally not be accessible from another IP) without modifying the image?

jglogan:
@lperdereau Nice! Thank you for sharing that alias.

@tidbeck Sorry for the long delay, I just came back to this issue and noticed your comment.

Container-to-host networking is a slightly different problem than publishing container ports to the host. Right now that's not possible in the nice easy way that Docker, Colima, et al do it (just use `host.docker.internal` in the container and it magically gets your traffic to your host-based service).

We documented this limitation and the socat-based workaround is here: https://github.com/apple/container/blob/main/docs/technical-overview.md#container-to-host-networking

jglogan:
Port forwarding (`--publish/-p`) will be done in two parts, to keep the PRs manageable.

#338 adds TCP and UDP forwarders.

The next PR will plumb `--publish` from CLI options to ContainerClient configuration to running the forwarders in SandboxService.

jglogan:
Closed by #338, #352, and #365.

```console
% container run --help
[...]

PR #352, opened 2025-07-18
Adds `--publish` flag for forwarding traffic to container ports.

jglogan:


jglogan:
Closes #97

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -54,6 +59,7 @@ public actor SandboxService {
         self.interfaceStrategy = interfaceStrategy
         self.log = log
         self.monitor = ExitMonitor(log: log)
+        self.eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
```
katiewasnothere, line 62: Should we be calling `shutdownGracefully` anywhere for this?

katiewasnothere (approved):

Sources/Services/ContainerSandboxService/SandboxService.swift
jglogan: probably should let the caller allocate and manage the lifecycle

PR #362, opened 2025-07-22
Named Volumes: stubs for CLI and option processing, and volume data structure

realrajaryan:
Added stubs for CLI and option processing for the named volumes feature, and a volume data structure.


Sources/CLI/Volume/VolumeList.swift
```diff
@@ -0,0 +1,48 @@
...
+extension Application.VolumeCommand {
+    struct VolumeList: AsyncParsableCommand {
+        static let configuration = CommandConfiguration(
+            commandName: "list",
+            abstract: "List volumes",
+            aliases: ["ls"]
+        )
+
+        @Flag(name: .shortAndLong, help: "Don't truncate output")
+        var noTrunc: Bool = false
```
dcantah: Is this just if some of the headers go off the screen, don't replace them with `...`?

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+    // The mount point of the volume on the host.
+    public var mountpoint: String
+    public var createdAt: Date
+    // User-defined key/value metadata.
+    public var labels: [String: String]
+    // Driver-specific options.
+    public var options: [String: String]
+    public var scope: VolumeScope
+    public var size: UInt64?
+    public var usageData: VolumeUsageData?
```
dcantah: Document the rest

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+
+/// Scope of volume availability.
+public enum VolumeScope: String, Sendable, Codable, CaseIterable {
+    case local = "local"
+    case global = "global"
+}
+
+/// Usage information for a volume.
+public struct VolumeUsageData: Sendable, Codable, Equatable {
+    public var size: Int64
```
dcantah: I'd make this have the unit in the name, e.g. `sizeInBytes` or whatever the actual unit is.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+        self.refCount = refCount
+    }
+}
+
+/// Request to create a new volume.
+public struct VolumeCreateRequest: Sendable, Codable {
+    public var name: String
+    public var driver: String
+    public var driverOpts: [String: String]
+    public var labels: [String: String]
```
dcantah: I know it's tedious, but documentation here and for anywhere that doesn't 🐸

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+    public static func createVolumeImage(for name: String, sizeInBytes: UInt64 = 1024 * 1024 * 1024) throws {
+        let blockPath = blockPath(for: name)
+        let fm = FileManager.default
+
+        // Create sparse file
+        fm.createFile(atPath: blockPath, contents: nil, attributes: nil)
+
+        // Truncate to desired size to create sparse file
+        let fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: blockPath))
+        defer { fileHandle.closeFile() }
```
dcantah: Prefer .close()

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+            return "Storage error: \(message)"
+        }
+    }
+}
+
+/// Volume storage management utilities.
+public struct VolumeStorage {
+    public static let volumesDirectory = URL(fileURLWithPath: NSHomeDirectory())
+        .appendingPathComponent("Library")
+        .appendingPathComponent("Application Support")
```
dcantah:
```swift
       let root = FileManager.default.urls(
            for: .applicationSupportDirectory,
            in: .userDomainMask
        ).first
```

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+        case .driverNotSupported(let driver):
+            return "Volume driver '\(driver)' is not supported"
+        case .storageError(let message):
+            return "Storage error: \(message)"
+        }
+    }
+}
+
+/// Volume storage management utilities.
+public struct VolumeStorage {
```
adityaramani, line 89: More of a higher level question - should this logic be in the daemon? Creating a new `VolumeService`? 

This is allow us to not have hardcoded paths for the App Support dir, and also follow the pattern we use for the other types?

Sources/CLI/Volume/VolumeList.swift
jglogan: @dcantah I know with `container image ls` the image digest is always elided with `...` at 24 characters. This actually seems a good idea but it might be good to indicate what would be truncated here (`Don't truncate {} output`), and also apply it uniformly wherever else we're dropping information (if it's only image, I wouldn't be against revising this PR with that change).

Sources/CLI/Volume/VolumeCreate.swift
```diff
@@ -0,0 +1,63 @@
...
+    struct VolumeCreate: AsyncParsableCommand {
+        static let configuration = CommandConfiguration(
+            commandName: "create",
+            abstract: "Create a volume"
+        )
+
+        @Argument(help: "Volume name")
+        var name: String
+
+        @Option(name: .long, help: "Volume driver name")
```
jglogan: Would it be easier to add it at the whatever point when we implement another driver instead of baking this in today? I'm not totally against building out the whole thing if we're certain this is the model we want, just asking the question.

Sources/CLI/Volume/VolumeCreate.swift
```diff
@@ -0,0 +1,63 @@
...
+
+            // Parse driver options
+            let parsedDriverOpts = parseKeyValuePairs(driverOpts)
+            let parsedLabels = parseKeyValuePairs(labels)
+
+            print("Driver options: \(parsedDriverOpts)")
+            print("Labels: \(parsedLabels)")
+        }
+
+        private func parseKeyValuePairs(_ pairs: [String]) -> [String: String] {
```
jglogan: Extract into ContainerClient utility and write a unit test? This is very generic functionality (`--network` could add key-value opts), and might also exist elsewhere already.

Sources/CLI/Volume/VolumeInspect.swift
```diff
@@ -0,0 +1,40 @@
...
+        static let configuration = CommandConfiguration(
+            commandName: "inspect",
+            abstract: "Display detailed information on one or more volumes"
+        )
+
+        @Argument(help: "Volume name(s)")
+        var names: [String]
+
+        @Option(name: .shortAndLong, help: "Format the output using the given Go template")
+        var format: String?
```
jglogan: We should drop this option for a couple reasons:

- Even if there's a suitable Swift version of `pkg/template` we should discuss whether that's the way we want to go for custom formatting.
- In any case it'd be better to implement this later, in a cross-cutting PR so that we don't have one command with it and a bunch of others that should have it but don't.

Sources/ContainerClient/Core/Volume.swift
jglogan: > ```swift
> let root = FileManager.default.urls(
>             for: .applicationSupportDirectory,
>             in: .userDomainMask
>         ).first
> ```

Better: `Application.appRoot` where you need the app data root.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+        self.labels = labels
+        self.options = options
+        self.scope = scope
+        self.size = size
+        self.usageData = usageData
+    }
+}
+
+/// Scope of volume availability.
+public enum VolumeScope: String, Sendable, Codable, CaseIterable {
```
jglogan: As with the options, I'm wondering if we keep the model simple to begin with – dealing with just our initial named volume implementation and the limited set of mount opts I'd expect us to support – and just make sure that what we do is open for extension.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,223 @@
...
+    public init(name: String, force: Bool = false) {
+        self.name = name
+        self.force = force
+    }
+}
+
+/// Response containing a list of volumes.
+public struct VolumeListResponse: Sendable, Codable {
+    public var volumes: [Volume]
+    public var warnings: [String]?
```
jglogan: Why `[Type]?` and not just `[Type]`? Optional-of-list is often redundant and needlessly complicates handling.

Sources/ContainerClient/Core/Volume.swift
jglogan: > More of a higher level question - should this logic be in the daemon? Creating a new `VolumeService`?

Since this is just client stubs to keep this small, perhaps we leave it here knowing that it will wind up in the service.

jglogan (changes requested):

Sources/CLI/Volume/VolumeList.swift
dcantah: Right, but all the info for the volume is user provided, so is it just if the user supplied a name like "my-really-cool-volume-name-that-is-probably-a-bit-too-unique"

Sources/CLI/Volume/VolumeDelete.swift
```diff
@@ -0,0 +1,40 @@
...
+        static let configuration = CommandConfiguration(
+            commandName: "delete",
+            abstract: "Remove one or more volumes",
+            aliases: ["rm"]
+        )
+
+        @Argument(help: "Volume name(s)")
+        var names: [String]
+
+        @Flag(name: .shortAndLong, help: "Force the removal of one or more volumes")
```
jglogan: Can we do without `--force` for now? I'm not sure it serves a common use case.

Sources/CLI/Volume/VolumeList.swift
jglogan: Right now it's the responsibility of `list` subcommand to trim any fields it wants to (`image list` with the manifest digest is the only place I'm aware of today). Perhaps we just drop the option and output the name without trimming.

jglogan (changes requested):

Sources/ContainerClient/Core/Volume.swift
jglogan: yes!

jglogan (approved):

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,237 @@
...
+/// A named volume that can be mounted in containers.
+public struct Volume: Sendable, Codable, Equatable, Identifiable {
+    // id of the volume.
+    public var id: String { name }
+    // Name of the volume.
+    public var name: String
+    // Driver used to create the volume.
+    public var driver: String
+    // The mount point of the volume on the host.
+    public var mountpoint: String
```
dcantah: If I'm following correctly, it's not really a mountpoint, it's just where the block device file is on the host. I'd just call it source.

Sources/ContainerClient/Parser.swift
```diff
@@ -317,10 +317,30 @@ public struct Parser {
                 let s = "mode=\(val)"
                 fs.options.append(s)
             case "source":
-                let absPath = URL(filePath: val).absoluteURL.path
                 switch type {
                 case "virtiofs", "bind":
-                    fs.source = absPath
+                    // Check if it's an absolute directory path first
```
dcantah, line 343: I'd just check if src is a file/dir first, and if that fails assume it's a volume.

dcantah: However, what I'd do here is if you do fall into the "is this a volume" block, I'd query the daemon to see if it actually is a valid volume, and use the returned data to create a .block FSType directly

Sources/ContainerClient/Core/Filesystem.swift
```diff
@@ -115,6 +116,16 @@ public struct Filesystem: Sendable, Codable {
         )
     }
 
+    /// A virtioblk backed filesystem providing a named volume block device.
```
dcantah: Does the existing .block not suffice?

Sources/CLI/Volume/VolumeList.swift
```diff
@@ -0,0 +1,106 @@
...
+extension Application.VolumeCommand {
+    struct VolumeList: AsyncParsableCommand {
+        static let configuration = CommandConfiguration(
+            commandName: "list",
+            abstract: "List volumes",
+            aliases: ["ls"]
+        )
+
+        @Option(name: .long, help: "Provide filter values (e.g., 'dangling=true')")
+        var filter: [String] = []
```
dcantah: What filters do we plan on supporting? I'd honestly skimp on this for now

Sources/ContainerClient/Parser.swift
```diff
@@ -354,17 +373,36 @@ public struct Parser {
...
             let src = String(parts[0])
             let dst = String(parts[1])
 
-            let abs = URL(filePath: src).absoluteURL.path
-            if !FileManager.default.fileExists(atPath: abs) {
-                throw ContainerizationError(.invalidArgument, message: "named volumes are not supported")
+            var sourcePath = src
+
+            // Check if it's an absolute directory path first
+            if src.hasPrefix("/") {
```
dcantah: Same comments as above

Sources/ContainerClient/Parser.swift
```diff
@@ -317,10 +317,29 @@ public struct Parser {
...
+                            throw ContainerizationError(.invalidArgument, message: "path '\(val)' is not a directory")
+                        }
+                        fs.source = absolutePath
+                    } else {
+                        guard VolumeStorage.isValidVolumeName(val) else {
+                            throw ContainerizationError(.invalidArgument, message: "Invalid volume name '\(val)': must match \(VolumeStorage.volumeNamePattern)")
+                        }
+
+                        // Use special marker to indicate this needs volume resolution during semantic validation
+                        fs.source = "named-volume:\(val)"
```
dcantah: This shouldn't be needed anymore with the above idea

Sources/ContainerClient/Utility.swift
```diff
@@ -148,6 +148,20 @@ public struct Utility {
         var mounts = try Parser.mounts(management.mounts)
         mounts.append(contentsOf: tmpfs)
         mounts.append(contentsOf: volumes)
+
+        // Semantic validation: resolve named volume markers to actual paths
```
dcantah: Same comment as above. If we resolve the volumes the second we know it is one, we don't need some sentinel value to tell us later

Sources/ContainerClient/Parser.swift
dcantah: This does make the parser have quite a bit of logic that it doesn't need to care about today, and would also make it async. I don't know which one I hate more though..

dcantah: What we COULD do is:
```swift
struct ParsedVolume {
    let name: String
    let destination: String
}

enum FileSystemOrVolume {
   case fs(Filesystem)
   case volume(ParsedVolume)
}
```

And have this method return something like that instead. Then later on it's very simple to tell whether you don't need to do anything, or you need to query for if it's a volume.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,237 @@
...
+    public var refCount: Int64
+
+    public init(sizeInBytes: Int64, refCount: Int64) {
+        self.sizeInBytes = sizeInBytes
+        self.refCount = refCount
+    }
+}
+
+/// Request to create a new volume.
+public struct VolumeCreateRequest: Sendable, Codable {
```
adityaramani: Having types for the xpc requests / responses is nice - but it is something we avoided in the project until now.
I think we should we consistent in how we do XPC.

Want to know the thoughts of others

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -0,0 +1,115 @@
...
+        self.options = options
+        self.sizeInBytes = sizeInBytes
+        self.usageData = usageData
+    }
+}
+
+/// Usage information for a volume.
+public struct VolumeUsageData: Sendable, Codable, Equatable {
+    // Size of the volume in bytes.
+    public var sizeInBytes: Int64
```
dcantah: Don't we already have this on the top level Volume struct itself?

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -228,12 +228,61 @@ actor ContainersService {
     private func _cleanup(id: String, item: Item) throws {
         self.log.debug("\(#function)")
         let config = try item.bundle.configuration
+
+        // Release volumes before deleting the container
+        Task {
```
adityaramani: Maybe remove the task and mark this method async? The calling method `delete` is async itself right?

Sources/CLI/Volume/VolumeCreate.swift
```diff
@@ -0,0 +1,50 @@
...
+        static let configuration = CommandConfiguration(
+            commandName: "create",
+            abstract: "Create a volume"
+        )
+
+        @Argument(help: "Volume name")
+        var name: String
+
+        @Option(name: .customLong("opt"), parsing: .upToNextOption, help: "Set driver specific options")
+        var driverOpts: [String] = []
```
adityaramani, line 35: Not sure if this is the right place to do it - but we should mark the volume to be mounted as atleast `sync-mode=full`. Since we dont want to deal with data corruption. Will need make sure that these are options are plumbed through to containerization

Tests/ContainerClientTests/VolumeValidationTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        #expect(!VolumeStorage.isValidVolumeName("_private"), "Name starting with underscore should be invalid")
+        #expect(!VolumeStorage.isValidVolumeName("-dash"), "Name starting with hyphen should be invalid")
+    }
+
+    @Test("Unicode and special character handling")
+    func testUnicodeCharacters() {
+        let unicodeNames = [
+            "volume-ñ",  // Non-ASCII letter
+            "volume-中文",  // Chinese characters
+            "volume-🐳",  // Emoji
```
adityaramani: nit: replace emoji? :)

Tests/ContainerClientTests/VolumeValidationTests.swift
```diff
@@ -0,0 +1,134 @@
...
+            "volume-café",  // Accented characters
+            "αβγ",  // Greek letters
+        ]
+
+        for name in unicodeNames {
+            #expect(!VolumeStorage.isValidVolumeName(name), "Unicode name '\(name)' should be invalid")
+        }
+    }
+
+    @Test("Common Docker volume name patterns")
```
adityaramani: Replace docker -> container?

adityaramani (changes requested):

Tests/ContainerClientTests/VolumeValidationTests.swift
jglogan: poo :)

Tests/ContainerClientTests/VolumeValidationTests.swift
jglogan: nit: or just volume

Sources/ContainerClient/Utility.swift
```diff
@@ -144,11 +144,35 @@ public struct Utility {
...
+        for item in (volumesOrFs + mountsOrFs) {
+            switch item {
+            case .filesystem(let fs):
+                resolvedMounts.append(fs)
+            case .volume(let parsed):
+                do {
+                    let readonly = parsed.options.contains("ro")
+                    let volume = try await ClientVolume.reserve(name: parsed.name, readonly: readonly, containerId: id)
+                    let blockMount = Filesystem.block(
+                        format: "ext4",
```
adityaramani: I think the `Volume` object we get from the `reserve` method should hold the information. I dont think we can assume the format is `ext4` if we were to ever add support for other FS formats

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,329 @@
...
+        let unit = String(trimmed[Range(unitRange, in: trimmed)!])
+
+        guard let number = Double(numberString) else {
+            throw VolumeError.storageError("Invalid number in size: '\(numberString)'")
+        }
+
+        let multiplier: UInt64
+        switch unit {
+        case "", "b":
+            multiplier = 1
```
adityaramani: nit: I think we already have some of these defined here. https://github.com/apple/containerization/blob/main/Sources/ContainerizationOS/BinaryInteger%2BExtensions.swift#L30

Might be better to have these in a single spot

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -215,33 +215,80 @@ actor ContainersService {
...
+            // Don't fail the cleanup process if volume release fails
+        }
+
         let label = Self.fullLaunchdServiceLabel(runtimeName: config.runtimeHandler, instanceId: id)
         try ServiceManager.deregister(fullServiceLabel: label)
         try item.bundle.delete()
         self.containers.removeValue(forKey: id)
     }
 
+    private func releaseVolumes(for config: ContainerConfiguration, containerId: String) async throws {
```
adityaramani: Just realized - This will get called only when `_cleanup` is invoked.

Looking at the call chain, this is done only when the container is explicitly deleted with `container rm <name>` or if the container was run with `container run --rm foo bar` 

Is that what we want? The right thing to do is to free up the volume to be used by some another container when it exits right? And not when deleted?

adityaramani: Also, im not sure how I feel about having this clean up logic here (in ContainerService)

I kinda like the way we handle free'ing up the IP / network allocated to the container in the `ContainerSandboxService` https://github.com/apple/container/blob/main/Sources/Services/ContainerSandboxService/SandboxService.swift#L841

I understand that this has its problems - specifically if the sandbox service somehow crashes. Bringing this up to have a discussion

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,329 @@
...
+import ContainerizationEXT4
+import ContainerizationError
+import Foundation
+import Logging
+import SystemPackage
+
+actor VolumesService {
+    private let resourceRoot: URL
+    private let store: ContainerPersistence.FilesystemEntityStore<Volume>
+    private let log: Logger
```
adityaramani, line 32: Should make use of the `AsyncLock` type we have and make sure all the operations in the VolumeService go through it?

This is mainly to deal with re-entrant actor methods, cause I think for example two parallel calls to the `create` method with the same args may lead to a weird situation

adityaramani (commented):
Sorry missed a few things in the prev review - adding a few more comments

Sources/CLI/Volume/VolumeCreate.swift
adityaramani: @dcantah - thoughts? Do we need changes in Containerization to handle this

Sources/APIServer/Containers/ContainersService.swift
dcantah: Yea, the only reason we'd even need a refcount for "is another container using this" in the first place is because virtualization.framework forbids (for good reason) the same block device being used across VMs if they're not read only. After the container exits we're free to decrement the refcount, we don't need to wait for rm so +1 to Aditya's comment

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -215,33 +215,80 @@ actor ContainersService {
...
+        let volumePrefix = "volumes/"
+        let volumeSuffix = "/volume.img"
+
+        for mount in config.mounts {
+            // Check if this is a named volume by looking at the source path
+            if mount.isBlock && mount.source.contains(volumePrefix) && mount.source.hasSuffix(volumeSuffix) {
+                // Extract volume name from path like ".../volumes/myvolume/volume.img"
+                let sourcePath = mount.source
+                if let volumeStart = sourcePath.range(of: volumePrefix)?.upperBound,
+                    let volumeEnd = sourcePath.range(of: volumeSuffix)?.lowerBound
```
dcantah: This is... crummy (and fragile) that we need to do this. There has to be a better way we can just include at least the volume name in what we send over. The `.block` Filesystem type has everything we need except the volume name which we need to dec the refcount right? I honestly wouldn't be opposed to (I know, we were the ones who told you to use .block) a new FS type that is literally just .block, but with a volume name included. OR just extend .block to include an optional volumeName parameter potentially

dcantah: @jglogan @adityaramani thoughts?

adityaramani: Agree - lets avoid checking the path and making a guess based off of that

Sources/CLI/Volume/VolumeCreate.swift
realrajaryan: waiting on this discussion

Sources/APIServer/Volumes/VolumesService.swift
realrajaryan: waiting on the sync mode stuff if I have to touch containerization

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -215,19 +215,20 @@ actor ContainersService {
...
+            try await self._cleanup(id: id, item: item)
         }
     }
 
     private static func fullLaunchdServiceLabel(runtimeName: String, instanceId: String) -> String {
         "\(Self.launchdDomainString)/\(Self.machServicePrefix).\(runtimeName).\(instanceId)"
     }
 
-    private func _cleanup(id: String, item: Item) throws {
+    private func _cleanup(id: String, item: Item) async throws {
```
jglogan: why is this async now?

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,368 @@
...
+    private static let blockFile = "volume.img"
+
+    public init(resourceRoot: URL, log: Logger) throws {
+        try FileManager.default.createDirectory(at: resourceRoot, withIntermediateDirectories: true)
+        self.resourceRoot = resourceRoot
+        self.store = try FilesystemEntityStore<Volume>(path: resourceRoot, type: "volumes", log: log)
+        self.log = log
+    }
+
+    private func parseSize(_ sizeString: String) throws -> UInt64 {
```
jglogan: nit: move private funcs down to bottom

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,368 @@
...
+    private let resourceRoot: URL
+    private let store: ContainerPersistence.FilesystemEntityStore<Volume>
+    private let log: Logger
+    private let lock = AsyncLock()
+
+    // Track volume references separately from persistent volume data
+    private var volumeReferences: [String: VolumeReferences] = [:]
+
+    private struct VolumeReferences {
+        var readwriteContainers: Set<String> = []
```
jglogan: Are we doing this to support shared RO mounts in this PR?

If not in this PR, my inclination is YAGNI for this PR, add it later.

But if we really want this in this PR, then define enum VolumeReference with .readOnly(String) and .readWrite(string).

Can you add unit test coverage for this type as well?

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,368 @@
...
+            multiplier = 1
+        case "k", "kb":
+            multiplier = 1024
+        case "m", "mb":
+            multiplier = 1024 * 1024
+        case "g", "gb":
+            multiplier = 1024 * 1024 * 1024
+        case "t", "tb":
+            multiplier = 1024 * 1024 * 1024 * 1024
+        case "kib":
```
jglogan: Are you getting the decision logic for these multipliers from somewhere else?

Generally for storage, `kb`, `mb`, `gb`, and so on refer to 10^3, 10^6, 10^9, and `kib`, `mib, `gib` are 2^10, 2^20, 2^30.

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,368 @@
...
+        case "tib":
+            multiplier = 1024 * 1024 * 1024 * 1024
+        default:
+            throw VolumeError.storageError("Unsupported size unit: '\(unit)'")
+        }
+
+        let sizeInBytes = UInt64(number * Double(multiplier))
+
+        // Validate reasonable size limits
+        let minSize: UInt64 = 1024 * 1024  // 1MB minimum
```
jglogan: Do we want to do this or let people shoot themselves in the foot? MBPs can have 8TB storage, and you can build a studio with 16TB.

jglogan: min size makes sense, might want to consult with Sid on this

Sources/APIServer/Containers/ContainersService.swift
jglogan: So the rules for volume attach/mount are this?
- You can create any number of RO and RW references to volumes at container create time.
- When attaching a volume at container run:
  - Allow if the mount is RO and all other mounts are RO
  - Allow if the mount is RW and no other mounts
  - Fail otherwise

Seems we would need a separate refcount for containers at rest to guard volume deletion.

dcantah:
We just chatted on this, but definitely add some CLI tests for this. Some ones that make sense to me:
1. Run a container with a volume -> Write to the volume and stop the container -> Run another container with the same volume and verify the write is there.
2. Run two containers with the same volume and verify the second fails (it'll be with the crummy vz error, but honestly just see if exit code is 1).
3. Run a container with a volume and try and delete the volume while it's still running and confirm this is not possible (the deletion fails).

And any others you think would be useful.

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+        let bytes = measurement.converted(to: .bytes).value
+
+        // Validate reasonable size limits
+        let minSize: UInt64 = 1.mib()  // 1mib minimum
+        let maxSize: UInt64 = 1.tib()  // 1tib maximum
+
+        let sizeInBytes = UInt64(bytes)
+
+        guard sizeInBytes >= minSize else {
+            throw VolumeError.storageError("Volume size too small: minimum 1MB")
```
dcantah: Going to be pedantic, but:
```suggestion
            throw VolumeError.storageError("Volume size too small: minimum 1MiB")
```

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        guard sizeInBytes <= maxSize else {
+            throw VolumeError.storageError("Volume size too large: maximum 1TB")
```
dcantah:
```suggestion
            throw VolumeError.storageError("Volume size too large: maximum 1TiB")
```

dcantah (approved):

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+        guard VolumeStorage.isValidVolumeName(name) else {
+            throw VolumeError.invalidVolumeName("Invalid volume name '\(name)': must match \(VolumeStorage.volumeNamePattern)")
+        }
+
+        let volumes = try await store.list()
+        guard let volume = volumes.first(where: { $0.name == name }) else {
+            throw VolumeError.volumeNotFound(name)
+        }
+
+        // Return volume without usage data since we removed refcounting
```
dcantah: Oh crud, one final thing. I'd remove/change this comment. No one reading this after should have to know the lore of this. Possibly change to something like the below (or remove it, I don't mind either):
```
// NOTE: It may be useful to return some sort of refcount showing
// what containers are using this volume eventually.
```

jglogan (changes requested):
putting a request changes on this while I test something

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -334,6 +335,29 @@ extension ContainersService {
             )
         }
     }
+
+    /// Check if a volume is currently in use by any container
+    public func isVolumeInUse(_ volumeName: String) -> Bool {
```
jglogan: With this approach, what happens with this? I would expect that `volume rm` should fail, and the subsequent `start` succeeds.

What I see it that `volume rm` succeeds and the container start fails with:

```console
% container start pgtest
Error: unknown: "Error Domain=NSPOSIXErrorDomain Code=45 "Operation not supported""
```

```bash
container system start
container run --name pgtest -e POSTGRES_PASSWORD=guest -v pgdata:/var/lib/postgresql -d postgres:latest
container exec -it pgtest df
container stop pgtest
container volume rm pgdata
container start pgtest
```

jglogan (changes requested):
`rm` needs to ensure no containers reference the volume.

Sources/APIServer/APIServer.swift
```diff
@@ -242,6 +245,18 @@ struct APIServer: AsyncParsableCommand {
         return service
     }
 
+    private func initializeVolumeService(root: URL, containersService: ContainersService, log: Logger, routes: inout [XPCRoute: XPCServer.RouteHandler]) async throws {
+        let resourceRoot = root.appendingPathComponent("volumes")
+        let service = try VolumesService(resourceRoot: resourceRoot, log: log)
```
jglogan: What requires us to have a weak reference between the services? Is this just in case that we have circular refs at some point?

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        // Check if volume exists by trying to list and finding it
+        let existingVolumes = try await store.list()
+        guard existingVolumes.contains(where: { $0.name == name }) else {
+            throw VolumeError.volumeNotFound(name)
+        }
+
+        // Check if volume is in use by any container
+        if let containersService = self.containersService, await containersService.isVolumeInUse(name) {
```
jglogan: Can this happen?

- We call isVolume in use here, and get back `false` and proceed to delete the volume.
- In the meantime we have another container starting up and is at https://github.com/apple/container/blob/f2f6f92fd6381c640e751afc61d1ce79e8584171/Sources/APIServer/Containers/ContainersService.swift#L121
- As a result we delete the volume and a container has a broken reference

The operation "check whether no containers reference a volume foo and remove the volume if nor refs" needs to be atomic with respect to "container creates a reference to volume foo"

jglogan (changes requested):

Sources/APIServer/Containers/ContainersService.swift
adityaramani: Also - can we have this method moved to the VolumeService?

The ContainerService object that we hold a ref to has a `list` method that we can use?

Sources/APIServer/Volumes/VolumesService.swift
dcantah: I talked with Raj on this, we're going to make a new closure based `list()` that passes you the current container list in the closure you pass to the method. Like this:

```swift
try await containersService.withContainerList() { containers in 
       for container in containers {
            if container.isRunning() && container.isUsingVolume(foo) {
                throw Woops()
            }
       }
       
       try await volumeStore.delete(foo)
}
```

withContainerList holds a lock (just as list() does) the entire operation.

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    public func delete(name: String) async throws {
+        try await lock.withLock { _ in
+            try await self._delete(name: name)
+        }
+    }
+
+    public func list() async throws -> [Volume] {
+        try await store.list()
```
adityaramani, line 66: Does this need to be in the lock?

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+        labels: [String: String] = [:]
+    ) async throws -> Volume {
+        try await lock.withLock { _ in
+            try await self._create(name: name, driver: driver, driverOpts: driverOpts, labels: labels)
+        }
+    }
+
+    public func delete(name: String) async throws {
+        try await lock.withLock { _ in
+            try await self._delete(name: name)
```
adityaramani, line 61: This is very much a nit - but for code clarity

should we make it such that the `_foo` methods takes in an object of type `AsyncLock.Context`

So these invocations become

```
try await lock.withLock { ctx in
            try await self._delete(name: name, lock: ctx)
}
```

Dont need to make this update in this PR, it can be done as a quick follow up

adityaramani (approved):
Just had one question

Sources/APIServer/Volumes/VolumesService.swift
jglogan: I suppose it could be but the store is an actor with no awaits in it

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        // Check if volume exists by trying to list and finding it
+        let existingVolumes = try await store.list()
+        guard existingVolumes.contains(where: { $0.name == name }) else {
+            throw VolumeError.volumeNotFound(name)
+        }
+
+        // Check if volume is in use by any container atomically
+        let volumeInUse = try await containersService.withContainerList { containers in
```
jglogan: Shouldn't the volume delete itself be executed in the block so that there's no chance of a reference being created to the volume after this block exits? Right now it seems that if another thread is creating a container that attaches the volume at the same time we're at line 197, there's a race still.

Sources/APIServer/Volumes/VolumesService.swift
```diff
@@ -0,0 +1,220 @@
...
+            }
+            return false
+        }
+
+        if volumeInUse {
+            throw VolumeError.volumeInUse(name)
+        }
+
+        try await store.delete(name)
+        try removeVolumeDirectory(for: name)
```
dcantah: These lines should be inside the closure itself (for whatever reason John's comment isn't popping up in this view, but I'm mirroring his comment)

dcantah: (197 to 202)

jglogan (approved):

jglogan (changes requested):
Need to fix build errors after pulling volume delete into closure.

jglogan (approved):

k13w:
How do I create a volume with names? Can you give me an example of how it would look in my LocalStack setup?

```
  localstack:
    image: localstack/localstack:3.1
    ports:
      - "127.0.0.1:4510-4559:4510-4559" # external service port range
      - "127.0.0.1:4566:4566" # LocalStack Edge Proxy
    environment:
      - DEBUG=${DEBUG:-0}
      - SERVICES=sns,sqs,s3
    volumes:
      - "${TMPDIR:-/tmp}/localstack:/var/lib/localstack"
      - "/var/run/docker.sock:/var/run/docker.sock"
      - ./localstack/start-localstack.sh:/etc/localstack/init/ready.d/start-localstack.sh
    networks:
      - dev
```

PR #363, opened 2025-07-22
Bump Containerization to 0.5.0

dcantah:
0.5.0 introduces a new way to configure the containers and execs. This is now done all upfront at constructor time in a callback style. I'm very happy with the config improvements, but because IO can only be setup at constructor time this makes it so that we need to supply IO at creation time of the VM or exec, which isn't the end of the world. All that really changes is `boostrap()` and `createProcess()` now take in IO instead of slightly later in `process.start()`


adityaramani:
Skimmed through the changes they look fine to me

crosbymichael (approved):

PR #364, opened 2025-07-23
Transition to `Mutex`

dkovba:
Due to the reduced use of the macro, we can now fully transition to `Mutex`.

### DIFF
--- a/Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift
+++ b/Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift
@@ -21,22 +21,18 @@ import ContainerizationExtras
 import Dispatch
 import Foundation
 import Logging
-import SendableProperty
 import SystemConfiguration
 import XPC
 import vmnet
+import Synchronization
 
 /// Creates a vmnet network with reservation APIs.
 @available(macOS 26, *)
 public final class ReservedVmnetNetwork: Network {
-    @SendablePropertyUnchecked
-    private var _state: NetworkState
+    private let _state: Mutex<NetworkState>
     private let log: Logger
 
-    @SendableProperty
-    private var network: vmnet_network_ref?
-    @SendableProperty
-    private var interface: interface_ref?
+    private let network = Mutex<vmnet_network_ref?>(nil)
     private let networkLock = NSLock()
 
     /// Configure a bridge network that allows external system access using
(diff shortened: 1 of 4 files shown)


Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift
```diff
@@ -21,22 +21,18 @@ import ContainerizationExtras
...
-    @SendablePropertyUnchecked
-    private var _state: NetworkState
+    private let _state: Mutex<NetworkState>
     private let log: Logger
 
-    @SendableProperty
-    private var network: vmnet_network_ref?
-    @SendableProperty
-    private var interface: interface_ref?
+    private let network = Mutex<vmnet_network_ref?>(nil)
```
dkovba: @crosbymichael Would it be appropriate to move `_state` and `network` inside a single "state" mutex in this type? The caveat is that we already have a public property called `state`.

dkovba: @jglogan Would it make sense to rename `public var state: NetworkState ` to `public var networkState: NetworkState`, so we could have `state` for the new `private let state: Mutex<State>`?

jglogan:
```swift

    private struct ThreadSafeState {
        var state: NetworkState
        var network: vmnet_network_ref?
    }

    private let threadSafeState: Mutex<ThreadSafeState>
```
```

Comment (dkovba) [reply Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift]: We've agreed on the `State` struct and `stateMutex` property.

File: Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift [hunk 2]
```diff
@@ -51,23 +49,25 @@ public final class ReservedVmnetNetwork: Network {
...
+            try handler(state.network.map { try Self.serialize_network_ref(ref: $0) })
         }
     }
 
     public func start() async throws {
-        guard case .created(let configuration) = _state else {
-            throw ContainerizationError(.invalidArgument, message: "cannot start network that is in \(_state.state) state")
+        let networkState = stateMutex.withLock { $0.networkState }
+        guard case .created(let configuration) = networkState else {
+            throw ContainerizationError(.invalidArgument, message: "cannot start network that is in \(networkState.state) state")
```
dcantah: You can double start this today. If we care about that you'll need to shove the whole statement in the lock closure.

dkovba: Done

Sources/Services/ContainerNetworkService/ReservedVmnetNetwork.swift
```diff
@@ -51,23 +49,27 @@ public final class ReservedVmnetNetwork: Network {
...
     }
 
     public func start() async throws {
-        guard case .created(let configuration) = _state else {
-            throw ContainerizationError(.invalidArgument, message: "cannot start network that is in \(_state.state) state")
+        let configuration = try stateMutex.withLock {
+            guard case .created(let configuration) = $0.networkState else {
+                throw ContainerizationError(.invalidArgument, message: "cannot start network that is in \($0.networkState.state) state")
+            }
+            return configuration
```
dcantah: I more meant we'll need to do `startNetwork` in the lock as well to prevent double starts, or add a new .starting state and set it in this closure, but just throwing the `startNetwork` call in here should be fine.

dcantah: You can just remove the state setting in startNetwork and move it here directly after the startNetwork() call in the closure.

dcantah: Something like:
```swift
    public func start() async throws {
        try stateMutex.withLock {
            guard case .created(let configuration) = $0.networkState else {
                throw ContainerizationError(.invalidArgument, message: "cannot start network that is in \($0.networkState.state) state")
            }
            let networkInfo = try startNetwork(configuration: configuration, log: log)
            $0.network = networkInfo.vmnetRef
            $0.networkState = NetworkState.running(
                configuration,
                NetworkStatus(
                    address: networkInfo.subnet.description,
                    gateway: networkInfo.gateway.description
                )
            )
        }
    }
```

We can have startNetwork return all the info we need (just make some struct with all of the fields we need to change the state to running, and return this).

dkovba: My fault. Will fix it shortly.

dcantah (approved):
LGTM!

Issue #376, opened 2025-07-26
[Request]: Virtualization support in containers


dcantah:
### Feature or enhancement request details

Containerization already exposes the nested virtualization capabilities on any SoCs that support it. We should expose a flag on run/create that expose it as well. The kata kernel we use by default doesn't have kvm on iirc, so a user would need to provide their own kernel however.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #377, opened 2025-07-26
Add virtualization support for containers

dcantah:
Closes https://github.com/apple/container/issues/376

On supported SoCs virtualization.framework has nested virtualization
support, and it's already exposed in Containerization. This exposes
a --virt flag on create and run to expose this to users here. If
the host doesn't support it we'll fail when going to run the container.
The guest kernel also must have virtualization enabled, which our default
kernel does not so one must be supplied.

```
./bin/container run -it --rm --virtualization ubuntu:latest bash
Error: unsupported: "nested virtualization is not supported on the platform"
```

And a successful run with a supplied kernel:

```
./bin/container run --rm --virtualization -it --kernel ~/Downloads/kernel.arm64 ubuntu:latest sh -c "dmesg | grep kvm"
[    0.017245] kvm [1]: IPA Size Limit: 40 bits
[    0.017499] kvm [1]: GICv3: no GICV resource entry
[    0.017501] kvm [1]: disabling GICv2 emulation
[    0.017506] kvm [1]: GIC system register CPU interface enabled
[    0.017685] kvm [1]: vgic interrupt IRQ9
[    0.017893] kvm [1]: Hyp mode initialized successfully
```

### DIFF
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -148,6 +148,13 @@ public struct Flags {
 
         @Option(name: [.customLong("label"), .short], help: "Add a key=value label to the container")
         public var labels: [String] = []
+
+        @Flag(
+            name: [.customLong("virtualization")],
+            help:
+                "Expose virtualization capabilities to the container. (Host must have nested virtualization support, and guest kernel must have virtualization capabilities enabled)"
+        )
+        public var virtualization: Bool = false
     }
 
     public struct Progress: ParsableArguments {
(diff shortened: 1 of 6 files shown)


Sources/ContainerClient/Flags.swift
```diff
@@ -148,6 +148,13 @@ public struct Flags {
 
         @Option(name: [.customLong("label"), .short], help: "Add a key=value label to the container")
         public var labels: [String] = []
+
+        @Flag(
+            name: [.customLong("virt")],
+            help:
+                "Expose virtualization capabilities to the container. (Host must have nested virtualization support, and guest kernel must have virtualization capabilities enabled"
```
jglogan: Missing closing parenthesis.

jglogan (changes requested):
nit: typo

jglogan (changes requested):
Typo in help

jglogan (approved):

PR #379, opened 2025-07-27
Explain how to install a release build in BUILDING.md

justinyaodu:
When users run commands with a debug build, they see a warning that debug builds may have worse performance. Add instructions to BUILDING.md so that users who wish to install from source know how to get a release build.

I tested the command locally and ran `container --version` to verify that I got a release build.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -26,6 +26,7 @@ make install
 Or to install a release build, with better performance than the debug build:
 
 ```bash
+BUILD_CONFIGURATION=release make all test integration
 BUILD_CONFIGURATION=release make install
 ```
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -23,6 +23,12 @@ Copy the binaries to `/usr/local/bin` and `/usr/local/libexec` (requires enterin
 make install
 ```
 
+Or to install a release build, with better performance than the debug build:
+
+```bash
+BUILD_CONFIGURATION=release make install
+```
+
 ## Compile protobufs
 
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -23,6 +23,12 @@ Copy the binaries to `/usr/local/bin` and `/usr/local/libexec` (requires enterin
 make install
 ```
 
+Or to install a release build, with better performance than the debug build:
+
+```bash
+BUILD_CONFIGURATION=release make install
```
jglogan, line 30: @justinyaodu Thanks for the contribution! This makes sense.

Since it's necessary for `BUILD_CONFIGURATION` to be defined consistently across targets, could you amend this so it shows both commands?

```bash
BUILD_CONFIGURATION=release make all test integration
BUILD_CONFIGURATION=release make install
```

Of course you can set `BUILD_CONFIGURATION` in the environment; you can also set these as make variables (putting `BUILD_CONFIGURATION` after make) instead of using the environment. We can assume a certain amount of developer expertise with make and variables, just show the release version of these two commands, and let developers figure out what suits them best.

jglogan (changes requested):

BUILDING.md
justinyaodu: Now that you mention it, I actually did get confused by this. At some point I ran `make all` with `BUILD_CONFIGURATION=release` and `make install` without it, and still got the debug build. Thanks for the great suggestion! I have updated the PR.

jglogan: lgtm, thanks!

jglogan (approved):

PR #388, opened 2025-07-28
Clarified documentation

sam-lord:
The demonstration is that patch versions can be updated without breaking changes, and that minor versions include breaking changes.

Please excuse the small documentation commit, just caused me a moment of confusion.


dcantah (approved):
Good catch 🐸

PR #393, opened 2025-07-29
Fix user arg passthrough for container create

### LINKED ISSUE #395: [Bug]: `container create` doesn't accept options process arguments.
### I have done the following

### Steps to reproduce

If I use any CLI option (such as `--bind` in the below example) in the process arguments of `container create`, the CLI tries to handle the options as subcommand options, instead of just leaving them as process args.

### Current behavior

```console
% container create -it --rm --network foo --name web python:slim python3 -m http.server --bind 0.0.0.0 8000
Error: Unknown option '--bind'. Did you mean '--uid'?
Usage: container create [<options>] <image> [<arguments> ...]
  See 'container create --help' for more information.
```

### Expected behavior

`container create` should succeed and `container start` should successfully start the container.

### Environment

```markdown
- OS: macOS 26 preview
- Xcode: 26 beta
- Container: main
```

### Relevant log output

```shell

```

### Code of Conduct

katiewasnothere:
This matches other container commands that rely on user arguments at the end, such as `container run`. Closes #395.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,7 @@ integration: init-block
 	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLINetwork
 	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunLifecycle
 	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIExecCommand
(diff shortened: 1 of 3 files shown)


dcantah (approved):

dcantah:
I approved a little hastily, can we make a simple test to create some container where the args uses something we would interpret as a flag and verify we don't fail (or whatever you think would be a good test).

katiewasnothere:
@dcantah added a simple test!

PR #397, opened 2025-07-30
Optionally resolve wrapper index to single-platform manifest based on `com.apple.containerization.index.indirect` annotation

### LINKED ISSUE #212: [Bug]: `index.digest` not consistent after pulls
### I have done the following

### Steps to reproduce

Pull the following image and dump its digest with `jq`. You'll get different results each time.

```
$ image="gcr.io/k8s-staging-cri-tools/test-image-digest@sha256:9700f9a2f5bf2c45f2f605a0bd3bce7cf37420ec9d3ed50ac2758413308766bf"
$ container image pull $image
$ container image inspect $image | jq '.[].index.digest'
``` 

### Current behavior

The digest is changed by every pull action:

```
$ image="gcr.io/k8s-staging-cri-tools/test-image-digest@sha256:9700f9a2f5bf2c45f2f605a0bd3bce7cf37420ec9d3ed50ac2758413308766bf"
$ container image pull $image
$ container image inspect $image | jq '.[].index.digest'
"sha256:0570a026b6348a1b02bae16156ccfc945aff10e0c322507e83e7ea962988e56c"
$ container image inspect $image | jq '.[].index.digest'
"sha256:0570a026b6348a1b02bae16156ccfc945aff10e0c322507e83e7ea962988e56c"
$ container image pull $image
$ container image inspect $image | jq '.[].index.digest'
"sha256:7fb9571b54af27a9f25e08ebe305aeba2c5e87d0c11f8fbe4411945ef8fec186"
```

### Expected Behavior

The digest should always be `sha256:9700f9a2f5bf2c45f2f605a0bd3bce7cf37420ec9d3ed50ac2758413308766bf`.

### Environment

```markdown
- OS: macOS 15.5 (24F74)
- Xcode: Version 26.0 beta (17A5241e)
- Container: container CLI version 0.1.0 (build: release, commit: 0fd8692)
```

### Code of Conduct

stevapple:
This PR:
- adds a `package func resolved()` on `ClientImage` that makes use of the new `com.apple.containerization.index.indirect` annotation to identify and resolve wrapper indices created by Containerization;
- replace the digest displayed in `container image list` with the one of the resolved manifest;
- use the resolved manifest for `container image inspect` if the index is a wrapper.

Fixes #212, as shown with the following local result:
image

### DIFF
--- a/Sources/ContainerClient/Core/ClientImage.swift
+++ b/Sources/ContainerClient/Core/ClientImage.swift
@@ -74,11 +74,16 @@ public struct ClientImage: Sendable {
     package func resolved() async throws -> Descriptor {
         let index = try await self.index()
         let indirect = index.annotations?[AnnotationKeys.containerizationIndexIndirect]
+        // If this is not an indirect index, return its own descriptor
         guard let indirect, ["1", "true"].contains(indirect.lowercased()) else {
             return self.descriptor
         }
+        // For indirect indices, return the first (and only) manifest
         guard let manifest = index.manifests.first else {
-            throw ContainerizationError(.internalError, message: "Failed to resolve indirect index: \(self.digest)")
+            throw ContainerizationError(
+                .internalError,
+                message: "Failed to resolve indirect index \(self.digest): no manifests found"
+            )
         }
         return manifest
     }


stevapple:
@adityaramani may you review the pull if you have time?

Sources/ContainerClient/Core/ClientImage.swift
```diff
@@ -69,6 +69,16 @@ public struct ClientImage: Sendable {
         }
         return try content.decode()
     }
+
+    /// Returns the resolved OCI descriptor for the image.
+    package func resolved() async throws -> Descriptor {
+        let index = try await self.index()
+        if index.annotations?[AnnotationKeys.containerizationIndexIndirect] == "true" {
+            return index.manifests.first!
```
adityaramani: This should work all the time - but just in case somehow something went wrong I think we should check for a nil value here

Maybe we can do something like this. What do you think?
```
guard index.annotations?[AnnotationKeys.containerizationIndexIndirect] == nil else {
   return self.descriptor
}
guard let manifest = index.manifests.first else {
   throw ContainerizationError(.internalError, message: "Malformed image .... ")
}
return manifest
```

PR #419, opened 2025-08-01
Select alternate data path with `container system start --app-root path`.

### LINKED ISSUE #418: [Request]: `container` can use a custom application data directory.

As a container user, I can run the tool using an alternate data directory, so that I can switch contexts for temporary testing, and so that I can store the data on a larger storage device than my system volume.

The system works with a single data directory at a time, selectable by an `--app-data` option on `container system start`.

Relocation of data directories is not in the scope of this issue (some files may container full paths at present).

### Code of Conduct

jglogan:
Closes #418.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,17 +140,21 @@ integration: init-block
 	@echo Running the integration tests...
 	bin/container system start $(SYSTEM_START_OPTS) && \
 	echo "Starting CLI integration tests" && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLINetwork && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunLifecycle && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIExecCommand && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLICreateCommand && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunCommand && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIImagesCommand && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunBase && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIBuildBase && \
-	$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIVolumes && \
-	echo Ensuring apiserver stopped after the CLI integration tests && \
-	scripts/ensure-container-stopped.sh
+	{ \
+		exit_code=0; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLINetwork || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunLifecycle || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIExecCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLICreateCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIImagesCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunBase || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIBuildBase || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIVolumes || exit_code=1 ; \
+		echo Ensuring apiserver stopped after the CLI integration tests ; \
+		scripts/ensure-container-stopped.sh ; \
+		exit $${exit_code} ; \
+	}
 
 .PHONY: fmt
 fmt: swift-fmt update-licenses


Makefile
```diff
@@ -124,28 +131,26 @@ test:
...
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIExecCommand
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLICreateCommand
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunCommand
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIImagesCommand
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIRunBase
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIBuildBase
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) --filter TestCLIVolumes
-	@echo Ensuring apiserver stopped after the CLI integration tests...
-	@scripts/ensure-container-stopped.sh
+	bin/container system start $(SYSTEM_START_OPTS) ; \
```
realrajaryan: Should we use `&&` instead of the trailing `;` after `bin/container system start $(SYSTEM_START_OPTS)`, to make sure the tests don't run in case this fails?

Sources/ContainerClient/Core/ClientHealthCheck.swift
```diff
@@ -26,9 +27,19 @@ extension ClientHealthCheck {
         XPCClient(service: serviceIdentifier)
     }
 
-    public static func ping(timeout: Duration? = .seconds(5)) async throws {
+    public static func ping(timeout: Duration? = .seconds(5)) async throws -> SystemHealth {
         let client = Self.newClient()
         let request = XPCMessage(route: .ping)
-        try await client.send(request, responseTimeout: timeout)
+        let reply = try await client.send(request, responseTimeout: timeout)
+        guard let appRootValue = reply.string(key: .appRoot), let appRoot = URL(string: appRootValue) else {
```
realrajaryan, line 34: The server uses `appRoot.formatted()` (which produces a human readable string) to encode the URL, but the client uses `URL(string:)` (which expects a proper URL format) to decode it. Would these always be compatible?

Sources/CLI/Application.swift
```diff
@@ -174,6 +131,53 @@ struct Application: AsyncParsableCommand {
...
+            appBundlePluginsURL,
+            installRootPluginsURL,
+        ].compactMap { $0 }
+
+        let pluginFactories = [
+            DefaultPluginFactory()
+        ]
+
+        guard let systemHealth = try? await ClientHealthCheck.ping(timeout: .seconds(10)) else {
+            throw ContainerizationError(.timeout, message: "unable to retrieve application data root from API server")
```
realrajaryan, line 166: The CLI now gets app root from health check `systemHealth.appRoot` while the API server independently resolves it from environment/default `ApplicationRoot.url`, is it possible if the API server restarts without the same `--app-root` flag, it will use a different app root than what CLI operations were using, causing CLI to look for previously created resources (like .build directories) in the wrong location?

- User starts API server with `--app-root /tmp/custom-data`
- CLI operations use health check -> get `/tmp/custom-data` from server
- CLI creates resources at `/tmp/custom-data/.build`
- API server crashes/restarts without `--app-root` flag
- API server now uses ApplicationRoot.url -> falls back to default `~/Library/Application Support/com.apple.container`
- Subsequent CLI operations health check -> get default path from server
-  CLI looks for resources at `~/Library/Application Support/com.apple.container/.build` ->
  files not found

.github/workflows/common.yml
```diff
@@ -68,7 +68,10 @@ jobs:
       - name: Test the container project
         run: |
           launchctl setenv HTTP_PROXY $HTTP_PROXY
-          make test cleancontent install-kernel integration
+          APP_ROOT=$(mktemp -d -p "${RUNNER_TEMP}")
+          trap 'rm -rf "${APP_ROOT}"; echo Removing data directory ${TMP_DIR}' EXIT
```
realrajaryan: Do you mean –
```suggestion
          trap 'rm -rf "${APP_ROOT}"; echo Removing data directory ${APP_ROOT}}' EXIT
```

Sources/CLI/Application.swift
```diff
@@ -304,8 +308,14 @@ struct Application: AsyncParsableCommand {
...
-    static func printModifiedHelpText() {
-        let altered = Self.pluginLoader.alterCLIHelpText(
+    static func printModifiedHelpText(pluginLoader: PluginLoader?) async {
+        let original = Application.helpMessage(for: Application.self)
+        guard let pluginLoader else {
+            print(original)
+            print("PLUGINS: not available, run `container system start`")
+            return
+        }
+        let altered = pluginLoader.alterCLIHelpText(
```
realrajaryan: nit: could just re-use `original` from line 312.
```suggestion
        let altered = pluginLoader.alterCLIHelpText(original: original)
```

Sources/CLI/Application.swift
jglogan: Let me write out what I think is happening and then I'll look at the scenario.

- `container system start` gets the app root from `--app-root`, defaulting to `~/Library/AS/com.apple.container`, and launches the API server, setting `CONTAINER_APP_ROOT` in the launchd plist.
- The API server launches and gets app root from `CONTAINER_APP_ROOT` (or default). All of the services, including the health check, use this value.
- The API server, via the plugin loader, starts all the other helpers, setting the env variable `CONTAINER_APP_ROOT` in the launchd plists using the same value as what gets returned by the health check.
- The CLI (other than `container system start`) and any other plugins use the health check to get the same application root as the API server and all of the helpers.

jglogan: > API server crashes/restarts without --app-root flag

If the API server or any helper were actually to restart (we don't automatically restart XPC services today), the `CONTAINER_APP_ROOT` setting would still be in the plist.

The only time the plists get regenerated is via `container system start`, so that if a user were to restart w/o the `--app-root` flag, that's their choice and they get what they asked for which is to use the default app data location.

.github/workflows/common.yml
jglogan: yes! good catch

Sources/CLI/Application.swift
jglogan: fixed!

Makefile
jglogan: Originally I had this in a `bash -ec` but the quoting got hellish.

I'll make the change...

Sources/ContainerClient/Core/ClientHealthCheck.swift
jglogan: yup: should be `.absoluteString`

realrajaryan (approved):
lgtm

Issue #375, opened 2025-07-26
[Request]: Structured documentation for the container commands


Daniel-Nashed:
### Feature or enhancement request details

Hi team!

I was looking for information about adding more memory and CPU information and it turned out it be in the howto section.
IMHO a project like this needs a bit more structured documentation.

This information can still be listed in a howto. But all parameters should be documented in reference material.

We had to learn structuring repositories and add GitHub pages to better structure the information in our projects.
A good structured documentation will pay back from my experience in other projects.

### Idea

- **reference material** --> a full reference of all commands
- **howto** --> quick information to perform a specific task
- **tutorials** --> more comprehensive to learn how functionality belongs together

The documentation page referenced in the project is more for an API documentation.
It would make sense to have a reference for the container command, too.
This could have the same style.

### Reference

Here is a good reference for a documentation structure and what it is used for.
Maybe you find it useful. It helped me a lot in other projects.

https://docs.divio.com/documentation-system/

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

adeebashraf:
Need to figure out how to use generated docs

Daniel-Nashed:
You should look into what other parts of Apple use on GitHub to document.
There are good projects around. You could use GitHub pages. Or mkdocs for example.
It would make sense to pick something other teams already use.

I am happy to provide feedback and ideas. I am maintaining multiple projects and documentation is always a challenge.

Let me know if there is anything I can do to help by giving feedback etc.

-- Daniel

jglogan:
@Daniel-Nashed Thanks for the request! We actually used the Divio doc as a guide when we put together the original documentation, but we had limited time and resources for the tech documentation, so that's why you see the tutorial and how-to docs but no reference.

Could you have a look at the PR that @wlan0 filed (see above) and provide feedback? My primary concern about all of this is that as our documentation surface area grows, it gets harder and harder to keep docs current with each new feature. Not saying that the answer is "don't document"; rather, I'm curious about ways we can keep the documents current, accurate, and readable without completely bogging down contributors.

Perhaps this is an area where it'd be easy to get GenAI into the loop.

PR #445, opened 2025-08-04
Create command-reference.md

wlan0:
Fixes https://github.com/apple/container/issues/375

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -1,0 +1,632 @@
+# Container CLI Command Reference
+
+## Core Commands
+
+### `container run`
+
+Runs a container from an image. If a command is provided, it will execute inside the container; otherwise the image's default command runs. By default the container runs in the foreground and STDIN remains closed unless `-i`/`--interactive` is specified.
+
+**Usage**
+
+```bash
+container run [OPTIONS] IMAGE [COMMAND] [ARG...]
+```
+
+**Options**
+
+*   **Resource management**
+    *   `-c, --cpus <number>`: number of CPUs to allocate
+    *   `-m, --memory <size>`: memory limit (K/M/G suffixes)
+*   **Process control**
+    *   `-e, --env <key=value>`: set environment variables
+    *   `--env-file <file>`: read environment variables from a file
+    *   `-i, --interactive`: keep STDIN open for interactive processes
+    *   `-t, --tty`: allocate a pseudo-TTY
+    *   `-u, --user <user>`: run as the specified user name or UID
+    *   `--uid <uid>`: run as the specified numeric user ID
+    *   `--gid <gid>`: run as the specified numeric group ID
+    *   `--cwd <path>`: set the working directory inside the container
+*   **Container management**
+    *   `--name <name>`: assign a name to the container
+    *   `-d, --detach`: run the container in the background (daemon mode)
+    *   `--rm`: automatically remove the container when it exits
+    *   `--cidfile <file>`: write the container ID to a file
+*   **Registry**
+    *   `--scheme <scheme>`: registry scheme (`auto`, `http`, or `https-`)
+*   **Progress**
+    *   `--disable-progress-updates`: disable progress display for image pulls and pushes
+*   **Global**
+    *   `--debug`: enable debug logging
+    *   `-h, --help`: show help
+
+**Examples**
+
+```bash
+# run a container and attach an interactive shell
+container run -it ubuntu:latest /bin/bash
+
+# run a background web server
+container run -d --name web -p 8080:80 nginx:latest
+
+# set environment variables and limit resources
+container run -e NODE_ENV=production --cpus 2 --memory 1G node:18
+```
+
+### `container build`
+
+Builds an OCI image from a local build context. It reads a Dockerfile (default `Dockerfile`) and produces an image tagged with `-t` option. The build runs in isolation using BuildKit, and resource limits may be set for the build process itself.
+
+**Usage**
+
+```bash
+container build [OPTIONS] PATH
+```
+
+**Options**
+
+*   **Resource management**
+    *   `-c, --cpus <number>`: CPUs to allocate to the build process (default 2)
+    *   `-m, --memory <size>`: memory for the build process (default 2048MB)
+*   **Build configuration**
+    *   `--build-arg <key=value>`: build-time variables passed to the Dockerfile
+    *   `-f, --file <path>`: path to the Dockerfile (default `Dockerfile`)
+    *   `-l, --label <key=value>`: add metadata labels to the image
+    *   `--no-cache`: disable cache usage
+    *   `-o, --output <config>`: specify build output (default `type=oci`)
+    *   `--arch <arch>`: target architecture (default `arm64`)
+    *   `--os <os>`: target operating system (default `linux`)
+    *   `--progress <type>`: progress output mode: `auto`, `plain`, or `tty`
+    *   `--vsock-port <port>`: vsock port used by BuildKit (default 8088)
+    *   `-t, --tag <name>`: set image name and tag
+    *   `--target <stage>`: set the target stage for multi-stage builds
+    *   `-q, --quiet`: suppress build output
+*   **Hidden options**
+    *   `--cache-in <config>`: configure build cache imports
+    *   `--cache-out <config>`: configure build cache exports
+*   **Global**
+    *   `--debug`: enable debug logging
+    *   `-h, --help`: show help
+
+**Examples**
+
+```bash
+# build an image and tag it as my-app:latest
+container build -t my-app:latest .
+
+# use a custom Dockerfile
+container build -f docker/Dockerfile.prod -t my-app:prod .
+
+# pass build args
+container build --build-arg NODE_VERSION=18 -t my-app .
+
+# build the production stage only and disable cache
+container build --target production --no-cache -t my-app:prod .
+```
+
+## Container Management
+
+### `container create`
+
+Creates a container from an image without starting it. This command accepts most of the same process/resource/management flags as `container run`, but leaves the container stopped after creation.
+
+**Usage**
+
+```bash
+container create [OPTIONS] IMAGE [COMMAND] [ARG...]
+```
+
+**Typical use**: create a container to inspect or modify its configuration before running it.
+
+### `container start`
+
+Starts one or more stopped containers. You can attach to the container's output streams and optionally keep STDIN open.
+
+**Usage**
+
+```bash
+container start [OPTIONS] CONTAINER...
+```
+
+**Options**
+
+*   `-a, --attach`: attach to STDOUT/STDERR of the container(s)
+*   `-i, --interactive`: attach STDIN for interactive sessions
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container stop`
+
+Stops running containers gracefully by sending a signal. A timeout can be specified before a SIGKILL is issued. If no containers are specified, nothing is stopped unless `--all` is used.
+
+**Usage**
+
+```bash
+container stop [OPTIONS] [CONTAINER...]
+```
+
+**Options**
+
+*   `-a, --all`: stop all running containers
+*   `-s, --signal <signal>`: signal to send (default SIGTERM)
+*   `-t, --time <seconds>`: timeout in seconds before killing the container (default 5)
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container kill`
+
+Immediately kills running containers by sending a signal (defaults to `SIGKILL`). Use with caution: it does not allow for graceful shutdown.
+
+**Usage**
+
+```bash
+container kill [OPTIONS] [CONTAINER...]
+```
+
+**Options**
+
+*   `-s, --signal <signal>`: signal to send (default `KILL`)
+*   `-a, --all`: kill all running containers
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container delete (rm)`
+
+Removes one or more containers. If the container is running, you may force deletion with `--force`. Without a container ID, nothing happens unless `--all` is supplied.
+
+**Usage**
+
+```bash
+container delete [OPTIONS] [CONTAINER...]
+```
+
+**Options**
+
+*   `-f, --force`: remove running containers by sending SIGKILL
+*   `-a, --all`: remove all containers
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container list (ls)`
+
+Lists containers. By default only running containers are shown. Output can be formatted as a table or JSON.
+
+**Usage**
+
+```bash
+container list [OPTIONS]
+```
+
+**Options**
+
+*   `-a, --all`: include stopped containers
+*   `-q, --quiet`: display only container IDs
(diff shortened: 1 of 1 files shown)


katiewasnothere:
@wlan0 could we use something like https://github.com/apple/swift-argument-parser/blob/main/Tools/generate-manual/GenerateManual.swift to generate man pages instead? I'd like something that we don't need to manually update

wlan0:
> @wlan0 could we use something like https://github.com/apple/swift-argument-parser/blob/main/Tools/generate-manual/GenerateManual.swift to generate man pages instead? I'd like something that we don't need to manually update

Let’s skip man pages entirely. Every detail a user would look for in man container - subcommands, flags, examples already comes straight from our source of truth via container --help.

Our commands are largely static and any future changes automatically flow into --help. 

Let’s keep Markdown as the only human-readable doc format and rely on container --help for terminal users

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+    *   `-m, --memory <size>`: memory limit (K/M/G suffixes)
+*   **Process control**
+    *   `-e, --env <key=value>`: set environment variables
+    *   `--env-file <file>`: read environment variables from a file
+    *   `-i, --interactive`: keep STDIN open for interactive processes
+    *   `-t, --tty`: allocate a pseudo-TTY
+    *   `-u, --user <user>`: run as the specified user name or UID
+    *   `--uid <uid>`: run as the specified numeric user ID
+    *   `--gid <gid>`: run as the specified numeric group ID
+    *   `--cwd <path>`: set the working directory inside the container
```
katiewasnothere: There's a few commands here that have additional option names. Should we include those as well?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+*   **Container management**
+    *   `--name <name>`: assign a name to the container
+    *   `-d, --detach`: run the container in the background (daemon mode)
+    *   `--rm`: automatically remove the container when it exits
+    *   `--cidfile <file>`: write the container ID to a file
+*   **Registry**
+    *   `--scheme <scheme>`: registry scheme (`auto`, `http`, or `https-`)
+*   **Progress**
+    *   `--disable-progress-updates`: disable progress display for image pulls and pushes
```
katiewasnothere: I think this isn't the right description. This disables the progress animation for updates. That could include other things than image pulls and pushes.

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+```bash
+container build [OPTIONS] PATH
+```
+
+**Options**
+
+*   **Resource management**
+    *   `-c, --cpus <number>`: CPUs to allocate to the build process (default 2)
+    *   `-m, --memory <size>`: memory for the build process (default 2048MB)
```
katiewasnothere: Should we include the K/M/G suffixes here as well?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+*   **Build configuration**
+    *   `--build-arg <key=value>`: build-time variables passed to the Dockerfile
+    *   `-f, --file <path>`: path to the Dockerfile (default `Dockerfile`)
+    *   `-l, --label <key=value>`: add metadata labels to the image
+    *   `--no-cache`: disable cache usage
+    *   `-o, --output <config>`: specify build output (default `type=oci`)
+    *   `--arch <arch>`: target architecture (default `arm64`)
+    *   `--os <os>`: target operating system (default `linux`)
+    *   `--progress <type>`: progress output mode: `auto`, `plain`, or `tty`
+    *   `--vsock-port <port>`: vsock port used by BuildKit (default 8088)
```
katiewasnothere: This is used by the builder shim right? Not buildkit itself

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+    *   `-t, --tag <name>`: set image name and tag
+    *   `--target <stage>`: set the target stage for multi-stage builds
+    *   `-q, --quiet`: suppress build output
+*   **Hidden options**
+    *   `--cache-in <config>`: configure build cache imports
+    *   `--cache-out <config>`: configure build cache exports
```
katiewasnothere: I don't think we should include these if they're hidden and aren't supported yet

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+```bash
+container create [OPTIONS] IMAGE [COMMAND] [ARG...]
+```
+
+**Typical use**: create a container to inspect or modify its configuration before running it.
+
+### `container start`
+
+Starts one or more stopped containers. You can attach to the container's output streams and optionally keep STDIN open.
```
katiewasnothere: I don't think container start supports specifying multiple containers to start in the same command

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+**Key flags**
+
+*   `-e, --env <key=value>`: set environment variables inside the exec session
+*   `--env-file <file>`: read environment variables from a file
+*   `-i, --interactive`: keep STDIN open
+*   `-t, --tty`: allocate a TTY
+*   `-u, --user <user>`: run as a specific user
+*   `--uid <uid>`, `--gid <gid>`: specify numeric UID/GID
+*   `--cwd <path>`: set working directory
```
katiewasnothere: This is also missing the other options

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container logs`
+
+Fetches logs from a container. You can tail the logs (`-f`/`--follow`), restrict the number of lines shown, or view boot logs.
```
katiewasnothere: I think the phrase "tail the logs" referencing following the logs is a bit confusing since this command also supports -n, which acts like `tail` in linux. 

```suggestion
Fetches logs from a container. You can follow the logs(`-f`/`--follow`), restrict the number of lines shown, or view boot logs.
```

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+**Options**
+
+*   `-q, --quiet`: show only image names
+*   `-v, --verbose`: produce verbose listing
+*   `--format <format>`: `table` (default) or `json`
+*   **Global**: `--debug`, `-h`/`--help`
+
+### `container image pull`
+
+Pulls an image from a registry. Supports specifying a platform and controlling progress display.
```
katiewasnothere, line 289: I'm not sure where we landed with whether or not we want to pull all platforms by default if no platform is specified. But should we include that default info here?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+    *   `-d, --detach`: run the container in the background (daemon mode)
+    *   `--rm`: automatically remove the container when it exits
+    *   `--cidfile <file>`: write the container ID to a file
+*   **Registry**
+    *   `--scheme <scheme>`: registry scheme (`auto`, `http`, or `https-`)
+*   **Progress**
+    *   `--disable-progress-updates`: disable progress display for image pulls and pushes
+*   **Global**
+    *   `--debug`: enable debug logging
+    *   `-h, --help`: show help
```
katiewasnothere: Should we include "--version" here too?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+**Usage**
+
+```bash
+container image push [OPTIONS] REFERENCE
+```
+
+**Options**
+
+*   `--platform <platform>`: platform to push (defaults to all available)
+*   `--scheme <scheme>`: registry scheme
```
katiewasnothere: Can we include the default and options for this field as well?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+```bash
+container image inspect [OPTIONS] IMAGE...
+```
+
+Only global flags (`--debug`, `-h`/`--help`) are available.
+
+## Builder Management
+
+The builder commands manage the BuildKit instance used for image builds. These commands are available when the system supports BuildKit.
```
katiewasnothere: What do we mean by "these commands are available when the system supports BuildKit"?

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+The `registry default` group allows setting, unsetting, and inspecting the default registry used when no registry is specified on image references.
+
+*   `container registry default set [OPTIONS] HOST`: Set the default registry.
+    *   `--scheme <scheme>`: registry scheme (`auto`, `http`, `https`)
+*   `container registry default unset (clear)`: Clears the default registry configuration.
+*   `container registry default inspect`: Displays the current default registry, if any.
+
+## System Management
+
+System commands manage the container apiserver, logs, DNS settings and kernel. These are primarily for macOS hosts.
```
katiewasnothere: These are only for macOS hosts

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+**Usage**
+
+```bash
+container system start [OPTIONS]
+```
+
+**Options**
+
+*   `-p, --path <path>`: path to the `container-apiserver` binary
```
katiewasnothere: I think this option is out of date, as well as the "--kernel-install" option below. We're missing `--app-root` and `--install-root`

docs/command-reference.md
```diff
@@ -0,0 +1,632 @@
...
+
+**Options**
+
+*   `--last <duration>`: show logs for the given duration (e.g., `1h`, `30m`)
+*   `-f, --follow`: stream logs as they are generated
+*   **Global**: `--debug`, `-h`/`--help`
+
+### DNS management (`container system dns`)
+
+Commands for managing local DNS configurations used by containers. Available on macOS hosts.
```
katiewasnothere: This is also only available on macOS hosts 😁

katiewasnothere:
I think overall we should keep the language/descriptions of the subcommands and options the same as in the help output. Otherwise the difference in language could become confusing (for example, the comment above about referring to "tailing the logs" as using the `--follow` option while also having an option on the same command that operates like `tail` in linux). 

IMO this doc really should just be the help output with a bit more description as needed (for example, the description of the run command in this PR is very helpful and augments what's in the help output) and some example use cases as you already have.

docs/command-reference.md
wlan0: This is used by builder

wlan0:
@katiewasnothere addressed all comments

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
...
+*   `container system kernel set [OPTIONS]`: Install or update the kernel.
+    *   `--binary <path>`: path to a kernel binary (can be used with `--tar` inside a tar archive)
+    *   `--tar <path | URL>`: path or URL to a tarball containing kernel images
+    *   `--arch <arch>`: target architecture (`arm64` or `x86_64`)
+    *   `--recommended`: download and install the recommended default kernel for your host
+    *   **Global**: `--debug`, `--version`, `-h`/`--help`
+
+***
+
+Command availability may vary depending on host operating system and macOS version.
```
jglogan: This should go up top.

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
+# Container CLI Command Reference
```
jglogan, line 1: Do we need a PR template that includes a reminder (or checklist if we also want a reminder for unit/integration tests and anything else) for updating docs?

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
...
+
+```bash
+container registry login [OPTIONS] SERVER
+```
+
+**Options**
+
+*   `-u, --username <username>`: username for the registry
+*   `--password-stdin`: read the password from STDIN (non-interactive)
+*   `--scheme <scheme>`: registry scheme (`auto`, `http`, `https`)
```
katiewasnothere: Can we add the default here as well?

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
...
+    *   `--cidfile <cidfile>`: Write the container ID to the path provided
+    *   `--no-dns`: Do not configure DNS in the container
+    *   `--dns <dns>`: DNS nameserver IP address
+    *   `--dns-domain <dns-domain>`: Default DNS domain
+    *   `--dns-search <dns-search>`: DNS search domains
+    *   `--dns-option <dns-option>`: DNS options
+    *   `-l, --label <label>`: Add a key=value label to the container
+    *   `--virtualization`: Expose virtualization capabilities to the container. (Host must have nested virtualization support, and guest kernel must have virtualization capabilities enabled)
+*   **Registry/progress/global**
+    *   `--scheme <scheme>`: Scheme to use when connecting to the container registry. One of (http, https, auto) (default: auto)
```
katiewasnothere: Could we do this for all the places that mention scheme?  

```suggestion
    *   `--scheme <scheme>`: Scheme to use when connecting to the container registry. One of (`http`, `https`, `auto`) (default: `auto`)
```

docs/command-reference.md
jglogan: if we wanted a more closed-form approach to this (one that we could validate via build), swift-argument-parser offers an `--experimental-dump-help` that writes all help output to a JSON file. If we wanted to merge informative content with the normative material from the dump, we could maintain the informative material as a set of files that could be JSON patched in, and then rendered to markdown or whatever we want

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
...
+container system logs [OPTIONS]
+```
+
+**Options**
+
+*   `--last <duration>`: Fetch logs starting from the specified time period (minus the current time); supported formats: m, h, d (default: 5m)
+*   `-f, --follow`: Follow log output
+*   **Global**: `--debug`, `--version`, `-h`/`--help`
+
+### DNS management (`container system dns`) (macOS only)
```
katiewasnothere: Should we have another subsection here that lists out the different subcommands like we did for the builder management commands? Same for the `container system kernel` section below? Just for consistency

docs/command-reference.md
```diff
@@ -0,0 +1,709 @@
...
+
+Commands for managing local DNS configurations used by containers. Available on macOS hosts.
```
katiewasnothere: Can we just remove the `Available on macOS hosts` part?

docs/command-reference.md
katiewasnothere: Another option is argument parser's `generate-docc-reference`. Doesn't look like this plugin is feature rich enough for us yet, but maybe in the future it will be.

docs/command-reference.md
```diff
@@ -0,0 +1,751 @@
...
+
+**Options**
+
+*   `--binary <path>`: Path to a kernel binary (can be used with `--tar` inside a tar archive)
+*   `--tar <path | URL>`: Path or URL to a tarball containing kernel images
+*   `--arch <arch>`: Target architecture (`arm64` or `x86_64`)
+*   `--recommended`: Download and install the recommended default kernel for your host
+*   **Global**: `--debug`, `--version`, `-h`/`--help`
+
+***
```
jglogan, line 751: Is the horizontal rule necessary? We don't use it in any of the other docs.

jglogan (commented):
One nit, see comment.

jglogan (approved):

Issue #83, opened 2025-06-10
Suggestion: pull platform specific containers by default to lessen the load on docker.io (eg 350MB vs 8GB)


mbhutton:
The default behaviour results in pulling images sometimes >20x greater than what's needed.

Eg. attempting to pull the `python` image started an 8GB download:
```
 container images pull python
⠦ [1/2] Fetching image 0% (15 of 112 blobs, 2.0 MB/7.92 GB, 810 KB/s) [9s]
```

Specifying my local platform reduces this substantially:
```
container images pull --platform linux/arm64/v8  python
⠇ [1/2] Fetching image 4% (2 of 10 blobs, 14.4/354.7 MB, 554 KB/s) [29s]
```

Pulling images for the local platform by default would:
* Save users time (eg pulling the multi-platform python above would take ~3 hours)
* Save users confusion (I wondered whether it was pulling the wrong image, knowing that it's usually far smaller)
* Save bandwidth for docker.io who could otherwise get hammered by people pulling for the wrong platforms
* Save the planet 🌏 

This project looks great btw, very exciting, congratulations for the launch! 🚀

crosbymichael:
Thank you for the issue. We will prioritize resolving this issue.

Sunsvea:
Created https://github.com/apple/container/pull/455 to address this. It seems the `bug` label was removed from this issue so we may have to wait and see if this behavior was intentional or not.

PR #455, opened 2025-08-06
fix: Pull platform-specific container images by default

Sunsvea:
This change addresses issue https://github.com/apple/container/issues/83 by modifying the default behavior of `container images pull` to automatically select the current platform when no `--platform` flag is specified. 

This reduces bandwidth consumption by up to ~95% (from ~8GB to ~350MB for multi-platform images like `python`).

## What changed
- Modified `ImagePull.swift` to default to `Platform.current` instead of pulling all platforms
- Added test coverage with `testPullDefaultPlatform()`
- Maintains backward compatibility with `--platform` flag

## Why this change
Users reported excessive bandwidth usage when pulling container images, as the default behavior downloads all available platforms. This particularly impacts users on limited connections and increases registry load unnecessarily.

## Testing
- New test confirms single-platform pulling behavior
- Manual tests show bandwidth reduction from ~8GB to ~350MB for multi-arch images
- Backward compatibility verified with `--platform` flag

## Impact
- Faster pull times for end users
- Reduced load on container registries
- Zero breaking changes to existing workflows

### DIFF
--- a/Tests/CLITests/Subcommands/Images/TestCLIImages.swift
+++ b/Tests/CLITests/Subcommands/Images/TestCLIImages.swift
@@ -176,10 +176,13 @@ extension TestCLIImagesCommand {
 
     @Test func testPullDefaultPlatform() throws {
         do {
+            // Use busybox for this test to avoid cache conflicts with other tests
+            let testImage = busybox
+            
             // Pull without specifying platform - should default to current platform
-            try doPull(imageName: alpine)
+            try doPull(imageName: testImage)
 
-            let output = try doInspectImages(image: alpine)
+            let output = try doInspectImages(image: testImage)
             #expect(output.count == 1, "expected a single image inspect output, got \(output)")
 
             // Verify only current platform was pulled


dcantah:
cc @adityaramani as this behavior was chosen for a reason

adityaramani:
The reason for pulling (and pushing) all platforms by default is that - if a user decided to do something like the following, and foo is a multi platform image

```
container image pull foo
container image tag foo bar
container image push bar
```
We should be keeping the sha of the index the same for the image we are pushing out, which would mean that we have to push out all the layers / sub-layers referenced in the index, which would include all the platforms. And to do that, we need to have all the layers present on disk


That being said - we have quite a few issues around the same area, and are open to changing the default behavior to work with only the current platform.

The changes required for this would be
- Change the default for pull / push commands to be the current platform (you have done this for pull)
- Add in a flag / a way for the user to specify that they want to pull / push all platforms. Maybe a `--all-platforms` flag?

Im not sure if we should have a different default for pull (current platform only) vs push (all platforms), but if we do we'd need some error handling and think about the UX

Sunsvea:
> The reason for pulling (and pushing) all platforms by default is that - if a user decided to do something like the following, and foo is a multi platform image
> 
> ```
> container image pull foo
> container image tag foo bar
> container image push bar
> ```
> 
> We should be keeping the sha of the index the same for the image we are pushing out, which would mean that we have to push out all the layers / sub-layers referenced in the index, which would include all the platforms. And to do that, we need to have all the layers present on disk
> 
> That being said - we have quite a few issues around the same area, and are open to changing the default behavior to work with only the current platform.
> 
> The changes required for this would be
> 
>     * Change the default for pull / push commands to be the current platform (you have done this for pull)
> 
>     * Add in a flag / a way for the user to specify that they want to pull / push all platforms. Maybe a `--all-platforms` flag?
> 
> 
> Im not sure if we should have a different default for pull (current platform only) vs push (all platforms), but if we do we'd need some error handling and think about the UX

@adityaramani I admit I like the sound of a 'smart' push workflow that cares about the local image state (i.e. push all platforms if local image has multiple platforms) but I feel it would introduce too much ambiguity.

I would lean toward a coherent workflow for both push/pull such as:
1. pull current platform by default
2. push current platform by default
3. use `--all-platforms` for both when doing multi-arch workflows
4. likewise use `--platform=arm64` when pulling or pushing to specific platforms.

Lastly, this would mean that we're updating the default behavior for push/pull. We'd need to consider how we'll communicate the change to anyone who's currently using this tool...

I'd be keen to hear some feedback from others.

bennettp123:
Another option might be to allow people to opt in (or out) of the new behaviour using something like

```bash
defaults write com.apple.container.defaults push.platforms <platforms>
defaults write com.apple.container.defaults pull.platforms <platforms>
```

Mcrich23:

PR #467, opened 2025-08-07
Fixes for install root and plugin detection.

jglogan:
- Sets up API server as source of truth for installation root, similarly to what was done for the data root. `system start` establishes the install root, setting the environment variable `CONTAINER_INSTALL_ROOT` when launching the API server.
- The API server propagates the environment variable when launching helpers, and returns the install root to the CLI via the health check XPC.
- Includes several fixes for detecting plugins that use app bundle layout.

### DIFF
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -15,7 +15,6 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
-import ContainerPlugin
 import ContainerizationError
 import Foundation
 
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import ContainerPlugin
 import ContainerizationError
 import Foundation
 
--- a/Sources/ContainerPlugin/PluginLoader.swift
+++ b/Sources/ContainerPlugin/PluginLoader.swift
@@ -98,9 +98,9 @@ extension PluginLoader {
 
             // Get all entries under the parent directory
             guard
-                var dirs = try? fm.contentsOfDirectory(
+                let urls = try? fm.contentsOfDirectory(
                     at: pluginDir,
-                    includingPropertiesForKeys: [.isDirectoryKey],
+                    includingPropertiesForKeys: [.isDirectoryKey, .isSymbolicLinkKey],
                     options: .skipsHiddenFiles
                 )
             else {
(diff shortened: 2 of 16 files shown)


Sources/ContainerClient/Flags.swift
```diff
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import ContainerPlugin
```
realrajaryan: nit: is this import needed?

Sources/ContainerPlugin/PluginLoader.swift
```diff
@@ -95,12 +106,15 @@ extension PluginLoader {
             else {
                 continue
             }
+
+            // Filter out all but plugin installation directories
             dirs = dirs.filter {
-                $0.isDirectory
+                $0.resolvingSymlinksInPath().isDirectory
```
realrajaryan: Calling `resolvingSymlinksInPath().isDirectory` for every item creates unnecessary URL objects and performs filesystem operations even for regular directories which is expensive. We could instead request both `.isDirectoryKey` and `.isSymbolicLinkKey` up front in `contentsOfDirectory`, then:
  - If `.isDirectoryKey` is true, keep it
  - If it's a symlink, use `FileManager.fileExists(atPath:isDirectory:)` to check the target

Sources/ContainerPlugin/PluginLoader.swift
```diff
@@ -143,12 +159,16 @@ extension PluginLoader {
...
 
+    /// Locate a plugin with a specific name.
     public func findPlugin(name: String, log: Logger? = nil) -> Plugin? {
         do {
+            // Try to load the plugin from each parent directory, using each plugin factory
             return
                 try pluginDirectories
                 .compactMap { installURL in
-                    try pluginFactories.compactMap { try $0.create(installURL: installURL.appending(path: name)) }.first
+                    try pluginFactories
```
realrajaryan: This double `compactMap { … }.first` creates temporary arrays at both levels. Using `.lazy` or nested loops would avoid the extra allocations and stop searching as soon as a plugin is found

Sources/ContainerClient/Flags.swift
jglogan: no, I had started to put `--install-root` in the global flags and backed that out but not the import!

Sources/ContainerPlugin/PluginLoader.swift
jglogan: I just didn't stop to think, would nested loops be clearer here. Fixed

realrajaryan (approved):

PR #472, opened 2025-08-08
Native Builder: tokenize each input line individually

katiewasnothere:
This PR removes tokenizing by logical line in favor of tokenizing by actual line. This will make it easier to return line and column number on an error in the future.

### DIFF
--- a/Sources/NativeBuilder/ContainerBuildParser/Docker/DockerfileParser.swift
+++ b/Sources/NativeBuilder/ContainerBuildParser/Docker/DockerfileParser.swift
@@ -23,29 +23,33 @@ public struct DockerfileParser: BuildParser {
         var instructions = [any DockerInstruction]()
         let lines = input.components(separatedBy: .newlines)
         var lineIndex = 0
+
         while lineIndex < lines.count {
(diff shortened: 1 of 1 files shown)


wlan0:
First off, thank you for the work here! 

Improving diagnostics in our Dockerfile parser is a worthy goal. However, splitting tokenization by physical lines changes the semantics of the parser and inevitably diverges from the Dockerfile spec, which is defined over logical instructions.

Instead of retokenizing per physical line, we should preserve the current, spec‑accurate logical parsing and introduce a SourceMap that records how the canonicalized input (the one the tokenizer sees) relates back to the original file. 

This is the same pattern used in compilers and bundlers (TypeScript, Go compiler, etc.). They don’t distort parsing to improve diagnostics, instead they record a mapping from transformed text back to source text and use it to render precise error locations.

Here is a pointer to buildkit's types for source mapping as reference: https://github.com/moby/buildkit/blob/ec7a152a9ae1911f69b53b00a4039a445f9b1fcd/client/llb/sourcemap.go#L78

wlan0 (changes requested):
More info in PR comments

katiewasnothere:
I'm going to close this for now while I investigate @wlan0's suggestion. Thanks!

PR #473, opened 2025-08-08
Complete implementation of Snapshot, Diff and Cache in NativeBuilder

wlan0:


Sources/NativeBuilder/ContainerBuildCache/ContentAddressableCache.swift
```diff
@@ -128,12 +125,8 @@ public actor ContentAddressableCache: BuildCache {
...
                 _ = try await contentStore.completeIngestSession(sessionId)
 
-                // Calculate total size
-                let _ =
-                    manifestSize
-                    + [snapshotLayer, environmentLayer, metadataLayer]
-                    .compactMap { $0 }
-                    .reduce(0) { $0 + $1.descriptor.size }
+                // Calculate total size (manifest + snapshot content only)
+                // Note: Snapshot size is already accounted for in the content store
```
katiewasnothere, line 129: Is this comment still relevant?

crosbymichael:
If all of these separate commits can be built and tested independently, could we have them as separate PRs?  This is only because I'm lazy and Github review tooling is built around reviewing the entire submitted change and not commits.

dcantah:
>If all of these separate commits can be built and tested independently, could we have them as separate PRs? This is only because I'm lazy and Github review tooling is built around reviewing the entire submitted change and not commits.

Ditto. If this commands a thorough review (which every change does) it would take quite a while to get through this as is.. I'd prefer 5 scoped PRs.

PR #483, opened 2025-08-11
CLI (run): Only ask for stdin for -i

dcantah:
Today if you asked for a pty and also wanted to detach we'd open stdin and then immediately close that pipe which would close the guest relay as well. I don't believe we need stdin open unless it's asked for with -i. Truthfully there's one extra bit here that is needed which is if -i and -d are supplied we need to tell the daemon to open stdin, but not send an fd (as they're client supplied and the client/cli is going to exit immediately). That will need to be a followup, as it's mainly useful for attach which we don't have today.

### DIFF
--- a/Sources/CLI/RunCommand.swift
+++ b/Sources/CLI/RunCommand.swift
@@ -205,30 +205,28 @@ struct ProcessIO {
         var stdio = [FileHandle?](repeating: nil, count: 3)
 
         let stdin: Pipe? = {
-            if !interactive && !tty {
+            if !interactive {
                 return nil
             }
             return Pipe()
         }()
 
         if let stdin {
-            if interactive {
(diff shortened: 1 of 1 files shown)


Sources/CLI/RunCommand.swift
```diff
@@ -205,30 +205,28 @@ struct ProcessIO {
         var stdio = [FileHandle?](repeating: nil, count: 3)
 
         let stdin: Pipe? = {
-            if !interactive && !tty {
+            if !interactive {
```
jglogan, line 208: Should ProcessIO live in the CLI target, or ContainerClient, at this point?

jglogan (approved):
lgtm, but see the general comment regarding location of the ProcessIO type

Issue #272, opened 2025-06-28
[Bug]: Cannot connect to internet on Fedora


osalbahr:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

```console
% container run --rm -it fedora
# curl apple.com
```

### Current behavior

Cannot connect to internet.

### Expected behavior

Can connect to internet.

### Environment

```markdown
- OS: macOS 15.5 (24F74)
- Xcode: Version 26.0 beta (17A5241e)
- Container: CLI version 0.2.0
```

### Relevant log output

```shell
% container run --rm -it fedora
[root@4388c973-1612-4c2f-b7e8-1fbec90ac50d /]# curl apple.com
curl: (6) Could not resolve host: apple.com
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

osalbahr:
I'd like to add that the `fedora` container can connect to the internet on `podman` and `docker`. And that the same issue happens on the `fedora:40`, `fedora:41, and `fedora:43` containers.

jglogan:
Hi @osalbahr, did you have a look at the macOS 15 networking limitations?

osalbahr:
Hi @jglogan, I wasn't able to find this issue mentioned in https://github.com/apple/container/blob/main/docs/technical-overview.md#macos-15-limitations. And I doubt it's a macOS 15 limitation because I remember it used to work fine with container 0.1.0 and this issue does not happen with the Ubuntu container.

Is this issue not reproducible on macOS 26? Unfortunately I don't have access to a macOS 26 system to test.

jglogan:
@osalbahr I can't reproduce it on may macOS 15 system. Are you sure that the IP configured on your network interface for your fedora container corresponds to the subnet for the vmnet bridge interface?

```console
% container --version
container CLI version 0.2.0 (build: release, commit: 3b5c253)
% container run -it --rm fedora bash               
[root@befb5890-6e68-4e94-82d5-b59cbbe4d88f /]# yum -y install iputils iproute
...
[root@befb5890-6e68-4e94-82d5-b59cbbe4d88f /]# ping -c 1 example.com
PING example.com (96.7.128.175) 56(84) bytes of data.
64 bytes from a96-7-128-175.deploy.static.akamaitechnologies.com (96.7.128.175): icmp_seq=1 ttl=54 time=40.4 ms

--- example.com ping statistics ---
[...]

PR #485, opened 2025-08-11
Clarify macOS version support and minor style tweaks

timsneath:
Clarify that container is only supported on macOS 26, and remove language that implies that it's also designed for macOS 15.

This change also includes a couple of minor Apple Style Guide changes, and links to the OCI container image format for clarity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 `container` is a tool that you can use to create and run Linux containers as lightweight virtual machines on your Mac. It's written in Swift, and optimized for Apple silicon.
 
-The tool consumes and produces OCI-compliant container images, so you can pull and run images from any standard container registry. You can push images that you build to those registries as well, and run the images in any other OCI-compliant application.
+The tool consumes and produces [OCI-compliant container images](https://github.com/opencontainers/image-spec), so you can pull and run images from any standard container registry. You can push images that you build to those registries as well, and run the images in any other OCI-compliant application.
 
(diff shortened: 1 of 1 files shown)


petr-ujezdsky:
So.. Can we use `container` on macOS 15 or not? It used to work, with some limitations described in https://github.com/apple/container/blob/main/docs/technical-overview.md

I was also suggesting to improve tutorials with respect to these limitations (common pitfalls) for users using macOS 15, see https://github.com/apple/container/issues/272#issuecomment-3025664397

I see it still available via `brew` for sequoia and later
https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/c/container.rb#L16

timsneath:
Hi Petr, we haven't made changes to explicitly block installation on macOS 15, but we built Containerization for macOS 26. We don't test it on older versions or intend to backport fixes to that release, so we want to clarify that Containerization is unsupported on that version.

petr-ujezdsky:
Ok, thanks for that :) This PR is about "Clarifying supported macOS version" but the change made that information actually more opaque IMO. All I am saying is, that the current wording

> We do not support older versions of macOS

could be a bit misleading. At first glance it would simply mean I can not install it on any other macOS than 26. Which is not true.

> "Feature not supported" means a piece of hardware, software, or a service does not understand or cannot use a specific function, file type, or feature.

As a newcomer I was very happy to see macOS 15 mentioned explicitly in top-level README with some additional info about caveats. I suggest to be more explicit on which OS can run this (15 and 26) and which OS you are accepting bug reports and feature requests (26 only).

PR #493, opened 2025-08-13
Native Builder: Concrete Snapshotter and Differ implementations

wlan0:


Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
```diff
@@ -0,0 +1,487 @@
...
+
+        // Metadata fields that contribute to metadata-only classification
+        let modeChanged = (old.mode ?? 0) != (new.mode ?? 0)
+        let uidChanged = (old.uid ?? 0) != (new.uid ?? 0)
+        let gidChanged = (old.gid ?? 0) != (new.gid ?? 0)
+
+        // Timestamps (mtime/ctime) are intentionally ignored for diff classification.
+        // They are not part of DiffKey semantics to ensure determinism across filesystems.
+        let mtimeChanged = false
+        let ctimeChanged = false
```
katiewasnothere, line 62: Can we remove these two but keep the comment around?

Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
```diff
@@ -0,0 +1,487 @@
...
+        let xattrsDigestChanged: Bool = {
+            let od = digestOfXAttrs(old)
+            let nd = digestOfXAttrs(new)
+            return od != nd
+        }()
+
+        // Aggregate metadata-only fields
+        let anyMetadataChange =
+            modeChanged || uidChanged || gidChanged || mtimeChanged || ctimeChanged || xattrsDigestChanged
+            || (!sizeChanged && old.type != .regular && new.type != .regular && (old.size ?? -1) != (new.size ?? -1))
```
katiewasnothere: Can we move this line out to it's own variable? Just for ease of reading

Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
```diff
@@ -0,0 +1,487 @@
...
+        let sec = Int64(st.st_mtimespec.tv_sec)
+        let nsec = Int64(st.st_mtimespec.tv_nsec)
+        return sec &* 1_000_000_000 &+ nsec
+    }
+
+    static func epochNanos(fromCTime st: stat) -> Int64 {
+        let sec = Int64(st.st_ctimespec.tv_sec)
+        let nsec = Int64(st.st_ctimespec.tv_nsec)
+        return sec &* 1_000_000_000 &+ nsec
+    }
```
katiewasnothere, line 461: can we just have one version of this function that takes in a timespec instead of the whole stat?

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
```diff
@@ -0,0 +1,401 @@
...
+        inspector: any FileAttributeInspector = DefaultFileAttributeInspector(),
+        fileDiffer: FileDiffer = FileDiffer(),
+        options: InspectorOptions = InspectorOptions()
+    ) {
+        self.inspector = inspector
+        self.fileDiffer = fileDiffer
+        self.options = options
+    }
+
+    // Safety: ensure no multiplicity-sensitive duplicates leak to DiffKey materialization.
```
katiewasnothere: Can you elaborate on what this comment means?

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
```diff
@@ -0,0 +1,401 @@
...
+                        // Device nodes (char/block) are intentionally excluded from diffs; they do not affect rebuilds.
+                        if attrs.type == .characterDevice || attrs.type == .blockDevice {
+                            return nil
+                        }
+                        return makeAdded(from: attrs, path: rel)
+                    }
+                    inFlight += 1
+                    if inFlight >= concurrency {
+                        if let d = try await group.next(), let dd = d { additions.append(dd) }
+                        inFlight -= 1
```
katiewasnothere, line 113: Can we use a semaphore instead of tracking the number of in-queue tasks there are? I'm not super familiar with the task groups in swift yet, but since this block of code isn't running concurrently, is it possible that there are actually much fewer than the limit in the queue by the time we have waited for one to finish and potentially added another? Then we wouldn't decrement the count correctly and end up limiting the concurrency more than we intended?

katiewasnothere: I've been informed that DispatchSemaphore is considered a big no-no  😁Is there another option or is this considered the standard?

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
```diff
@@ -0,0 +1,401 @@
...
+        // Fallback: try with original paths
+        let urlPath = BinaryPath(url: url)
+        let basePath = BinaryPath(url: base)
+
+        if let relative = urlPath.relativePath(from: basePath) {
+            return relative
+        }
+
+        // Last resort - shouldn't happen in normal usage
+        return urlPath.lastPathComponent
```
katiewasnothere, line 305: When would these happen?

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveDiffer.swift
```diff
@@ -0,0 +1,649 @@
...
+                                Archiver.ArchiveEntryInfo(
+                                    pathOnHost: sidecarHost,
+                                    pathInArchive: URL(fileURLWithPath: sidecarRel)
+                                )
+                            )
+                        }
+                    }
+                }
+
+            case .modified(let m):
```
katiewasnothere, line 315: Is this doing the same thing as the add case?

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveDiffer.swift
```diff
@@ -0,0 +1,649 @@
...
+
+            default:
+                continue
+            }
+        }
+
+        // Best-effort apply xattrs if present (no-op if unsupported)
+        // NOTE: Setting xattrs is platform-specific; intentionally skipped here to avoid
+        // adding Darwin imports. XAttrs were captured for determinism but not required by tests.
+        _ = pendingXAttrs  // Reserved for future use
```
katiewasnothere: These aren't actually applied, can we update the comment here and above?

wlan0:
There is no need to get bogged down in detailed scrutiny of this PR before the product ships. Blocking productivity for something that hasn't yet launched is counterproductive.

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveSnapshotter.swift
```diff
@@ -0,0 +1,360 @@
...
+        guard let content = try await store.get(digest: currentLayerDigest) else {
+            throw TarArchiveSnapshotterError.invalidSnapshotState("missing content for layer \(currentLayerDigest)")
+        }
+        let bytes = try content.data()
+
+        // Heuristic: use declared media type if present, else default to gzip (our differ default)
+        let mediaType = node.state.layerMediaType ?? "application/vnd.oci.image.layer.v1.tar+gzip"
+
+        // Persist to a temp file for ArchiveReader
+        let suffix = mediaType.contains("+gzip") ? ".tar.gz" : ".tar"
```
katiewasnothere: Could we reference DiffFormat's media type instead?

katiewasnothere:
> There is no need to get bogged down in detailed scrutiny of this PR before the product ships. Blocking productivity for something that hasn't yet launched is counterproductive.

@wlan0 I disagree that we shouldn't scrutinize the details of this PR. This is a lot of code and we shouldn't be adding tech debt from the get go, especially for a project this young and of this expected size. The tech debt will just compound over time and be harder to clean up. Even if we're trying to work fast, we should be producing readable and maintainable code. Doubly so if we want other members of the community to be able to contribute to the code easily.

wlan0:
> I disagree that we shouldn't scrutinize the details of this PR. This is a lot of code and we shouldn't be adding tech debt from the get go, especially for a project this young and of this expected size. The tech debt will just compound over time and be harder to clean up. Even if we're trying to work fast, we should be producing readable and maintainable code. Doubly so if we want other members of the community to be able to contribute to the code easily.

Thanks for your thoughtful input. While I agree that maintainability and readability are critical, it's equally important that feedback remains actionable, objective, and focused on tangible improvements rather than subjective preferences or questions.

Given our stage and pace, we must prioritize progress to ensure the project's momentum doesn't stall. If you have clear, specific technical concerns, please highlight them explicitly and if you have a better approach, suggest them. Otherwise, subjective style or preference-based suggestions should be considered separately and shouldn't prevent this PR from merging promptly.

Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
wlan0: Thanks! I’m intentionally keeping `mtimeChanged`/`ctimeChanged` set to `false` and listed in the aggregation. 

Tar formats don’t reliably carry ctime (and mtime is often normalized for reproducibility), so we ignore time changes by design. 

Keeping the named flags in anyMetadataChange documents that decision and prevents future regressions if we revisit timestamp semantics. If you prefer, I can add a brief inline comment clarifying this, but I’d like to keep the flags visible for readability and intent.

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
wlan0: Thanks for digging into the parallelism!

I’m deliberately using withTaskGroup here. The inFlight/next() throttling pattern caps concurrency reliably, with at most a transient underutilization of one slot. I’d prefer to stick with the task‑group approach here.

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveDiffer.swift
wlan0: Great question. Even though both branches eventually emit an `entry`, `.add` handles paths absent in the baseline, whereas `.modified` applies changes to an existing path with different preconditions and bookkeeping. Keeping them separate keeps those invariants explicit.

Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
wlan0: Good thought. I kept two tiny helpers on purpose: the stat time fields differ across platforms (e.g., `st_mtimespec` vs `st_mtim`), and I want to keep that platform detail out of call sites. The duplication is minimal but preserves clarity (`mtime` vs `ctime`) and avoids leaking `timespec` bridging/imports

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
wlan0: hardlinks, bind mounts and symlinks can lead to multiple entires producing the same diffKey. The order of these keys showing up between different traversals is non-deterministic. This comment states that duplicates should not be materialized to ensure that downstream consumers can always expect DiffKeys in the same, stable order.

I'll add a clarifying note in the comment as well

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveDiffer.swift
wlan0: >   // Best-effort apply xattrs if present (no-op if unsupported)

Do you mean this comment? 

>    // Accumulate sidecar xattrs by path; applied best-effort after entry creation.

and this one?

wlan0: I'll update these two for now, pls let me know if you had a different one in mind

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
wlan0: This is due to swift URL's non-deterministic behavior

- The first call to relativePath computes relative path based on standardizedFileURL 
 - The second call to relativePath computes relative path based on direct URL (without calling standardizedFileURL)
- If both of these do not yield a relative path, then this acts as a safe fallback. This can happen if `standardizedFIleURL` does not work as expected (which we've run into before)

Worst case, it leads to a key being counted as a .modified or .added for a particular path.

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveDiffer.swift
katiewasnothere: The updated comments look good thanks!

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
katiewasnothere: The updated comment looks good! Thanks for the explanation :)

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
```diff
@@ -0,0 +1,637 @@
...
+        #else
+        let dev: dev_t = dev_t(devRaw)
+        #endif
+        let mode: mode_t = S_IFCHR | mode_t(perms)
+        let r: Int32 = url.path.withCString { mknod($0, mode, dev) }
+        #endif
+        if r != 0 { throw NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: [NSLocalizedDescriptionKey: "mknod failed"]) }
+    }
+
+    private func computeKey(base: URL?, target: URL, options: InspectorOptions? = nil) async throws -> DiffKey {
```
katiewasnothere, line 139: We have `computeKey` and `computeDiffKey` that do very similar things. Is there a way we can take out some of the utility functions and put them in a common place between these tests?

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
```diff
@@ -0,0 +1,637 @@
...
+#else
+import Darwin
+#endif
+
+@Suite
+struct DiffKeySwiftTestingSuite {
+
+    // MARK: - Utilities
+
+    struct TempDir {
```
katiewasnothere, line 28: Should we be using the TestUtils functions instead where possible?

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
```diff
@@ -0,0 +1,637 @@
...
+            }
+        }
+        if bindRes != 0 {
+            throw NSError(domain: NSPOSIXErrorDomain, code: Int(errno), userInfo: [NSLocalizedDescriptionKey: "bind(AF_UNIX) failed"])
+        }
+    }
+
+    private func createCharDevice(_ url: URL, major: UInt32, minor: UInt32, perms: Int = 0o644) throws {
+        try FileManager.default.createDirectory(at: url.deletingLastPathComponent(), withIntermediateDirectories: true)
+        #if os(Linux)
```
katiewasnothere: These tests only ever run on Mac right? Can we remove the if/elses for linux?

Sources/NativeBuilder/ContainerBuildSnapshotter/DirectoryDiffer.swift
dkovba: There is `AsyncSemaphore` in `ExecutionDispatcher.swift`, which seems reasonable to use here.

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
wlan0: yeah, this a good idea!

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
wlan0: good catch

Sources/NativeBuilder/ContainerBuildSnapshotter/FileMetadataDiffer.swift
katiewasnothere: This is non-blocking, but in general I feel like this separation is not really providing additional clarity. I felt more confused as to why we kept them separate and was concerned there was some nuance I just didn't have the context for.

Sources/NativeBuilder/ContainerBuildSnapshotter/TarArchiveSnapshotter.swift
wlan0: This is the path of a temporary file being staged for archive reader

Tests/NativeBuilderTests/ContainerBuildSnapshotterTests/DiffKeyEdgeCasesTests.swift
katiewasnothere: Hey @wlan0 were you planning to make this change in this PR?

wlan0: Yeah I missed this one in my last push

wlan0: I just re-reviewed this. There are different functions. There is no need to find a common abstraction

wlan0:
Closing this since we are moving to feature branch for native builder work

PR #238, opened 2025-06-19
Docs: clarify Xcode 26 beta must be the active developer directory

unitof:
And add instructions for setting it so

Related: https://github.com/apple/containerization/issues/66

PR #499, opened 2025-08-14
Xcode 26 active developer directory instructions

unitof:
Re-do of https://github.com/apple/container/pull/238, now with verified commit

Related: apple/containerization#66


unitof:
Was able to reopen original #238 after it was auto-closed, so closing this as duplicate

PR #502, opened 2025-08-15
Add SSH auth socket forwarding

### LINKED ISSUE #498: [Request]: Add `--ssh` flag for ssh auth forwarding to container create/run.

At present, I can mount my host `SSH_AUTH_SOCK` into a container using a technique like this, so that I can do `git clone` using ssh:

```bash
container run --volume "${SSH_AUTH_SOCK}:/run/ssh-auth.sock" --env SSH_AUTH_SOCK=/run/ssh-auth.sock ...
```

This works fine as long as my login session lasts. If I logout, login, and `container start` the container, the auth forwarding doesn't work anymore because `SSH_AUTH_SOCK` points to a new location. Since the mount is defined in the container configuration, the mount becomes stale.

Adding a boolean `--ssh` flag (defaulting to false so that the user opts into the special behavior) would allow the system to specify the mount anew every time the container starts. It improves UX because the user no longer needs to remember the `-v`/`-e` dance, and it may help with delivering #421.

### Code of Conduct

Reneechang17:
Add `--ssh` flag to forward the host's SSH agent socket into the container, so we can use SSH authentication for things like cloning private repos, and also updates the socket path every time the container starts to handle socket path changes like reboot/re-login.

Closes #498

### DIFF
--- a/Sources/Services/ContainerSandboxService/SandboxService.swift
+++ b/Sources/Services/ContainerSandboxService/SandboxService.swift
@@ -765,11 +765,14 @@ public actor SandboxService {
         czConfig.process.arguments = [process.executable] + process.arguments
         czConfig.process.environmentVariables = process.environment
 
-        if czConfig.sockets.contains(where: {
-            $0.destination == URL(fileURLWithPath: Self.sshAuthSocketGuestPath)
-        }) {
-            if config.ssh && !czConfig.process.environmentVariables.contains(where: { $0.starts(with: "\(Self.sshAuthSocketEnvVar)=") }) {
-                czConfig.process.environmentVariables.append("\(Self.sshAuthSocketEnvVar)=\(Self.sshAuthSocketGuestPath)")
+        // Add SSH_AUTH_SOCK if ssh forwarding is enabled
+        if config.ssh {
+            if czConfig.sockets.contains(where: {
+                $0.destination == URL(fileURLWithPath: Self.sshAuthSocketGuestPath)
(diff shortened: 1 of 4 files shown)


Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -748,6 +762,15 @@ public actor SandboxService {
     ) {
         czConfig.process.arguments = [process.executable] + process.arguments
         czConfig.process.environmentVariables = process.environment
+
+        if czConfig.sockets.contains(where: {
```
jglogan: This will automatically set an environment variable even if the user specify `-ssh`, in the case where they've mounted a socket at `/run/host-services/ssh-auth.sock`, and they've not set the variable explicitly?

That's probably harmless but it might be unexpected. Using `if config.ssh && config.sockets.contains...` would ensure that we only do this when `--ssh` is specified. If `--ssh` isn't present, then nothing ever happens.

jglogan (changes requested):

Sources/Services/ContainerSandboxService/SandboxService.swift
Reneechang17: Make sense - it's good point to ensure we only set the env var when the flag is specified.

jglogan: can the `config.ssh` test move out of the closure to L768?

PR #516, opened 2025-08-16
Add the support for ARG in the native builder parser

dkovba:
This PR adds the support for the ARG instruction in the native builder parser, implements the support for different kinds of ARGs, and performs the substitution of ARG variables in the supported instructions. Resolves https://github.com/apple/container/issues/437.

Some features are currently blocked and not included into this PR:
- [Native builder: add the current target platform we're building use it to set automatic platform ARGs #522](https://github.com/apple/container/issues/522)
- [Native builder: add the support for the BuildKit built-in ARGs #523](https://github.com/apple/container/issues/523)
- [Native builder: ensure pre-defined ARGs are excluded from the output of the history #524](https://github.com/apple/container/issues/524)
- [Native builder: add the support for stage references and ARG inheritance #525](https://github.com/apple/container/issues/525)


Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
```diff
@@ -0,0 +1,99 @@
...
+
+            let varName = String(input[varRange])
+            let replacement = resolveArg(key: varName, inFromContext: inFromContext) ?? ""
+
+            let matchRange = match.range
+            let adjustedMatchRange = NSRange(location: matchRange.location + offset, length: matchRange.length)
+
+            if let replacementRange = Range(adjustedMatchRange, in: result) {
+                result.replaceSubrange(replacementRange, with: replacement)
+                offset += replacement.count - matchRange.length
```
wlan0: `String.count` and `Range.length` work on different principles

`String.count`  counts as grpaheme clusters (i.e. Ç would be 1 count)
`Range.length` counts as utf-16 (i.e Ç would be two counts)

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// Add a FROM-only build argument (ARG before the first FROM).
+    /// These arguments are only available in FROM instructions, not within stages.
+    @discardableResult
+    public func fromOnlyArg(_ name: String, defaultValue: String? = nil) -> Self {
+        buildArgs[name] = defaultValue
+        return self
+    }
+
+    /// Check if there's an active stage.
+    public var hasActiveStage: Bool {
```
wlan0, line 39: this could be `internal`

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// These arguments are available within the stage.
+    @discardableResult
+    public func arg(_ name: String, defaultValue: String? = nil) throws -> Self {
+        let operation = MetadataOperation(
+            action: .declareArg(name: name, defaultValue: defaultValue)
+        )
+        return try add(operation)
+    }
+
+    /// Add a FROM-only build argument (ARG before the first FROM).
```
wlan0, line 30: Buildkit calls them `meta args`. For parity, it would be good to call them the same here

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// Substitute ARG variables in a string.
+    /// - Parameters:
+    ///   - input: The string that may contain `${ARG}` references.
+    ///   - inFromContext: Whether this substitution is happening in a FROM instruction context.
+    /// - Returns: The string with ARG variables substituted.
+    public func substituteArgs(_ input: String, inFromContext: Bool) -> String {
+        let pattern = #"\$\{([A-Za-z_][A-Za-z0-9_]*)\}"#  // `${ARG}`
+        let regex = try! NSRegularExpression(pattern: pattern)
```
wlan0: make this a singleton, otherwise it compiles on each invocation

```
static let regex: NSRegularExpression = {
     try! NSRegularExpression(pattern: #"\$\{([A-Za-z_][A-Za-z0-9_]*)\}"# ) // `${ARG}`
}()
```

Sources/NativeBuilder/ContainerBuildParser/Docker/Instructions/ExposeInstruction.swift
```diff
@@ -17,14 +17,14 @@
 import ContainerBuildIR
 
 struct ExposeInstruction: DockerInstruction {
-    let ports: [PortSpec]
+    let ports: [String]
 
-    init(_ rawPorts: [String]) throws {
-        self.ports = try rawPorts.map(parsePort)
+    init(_ ports: [String]) throws {
```
wlan0: remove `throws`

wlan0 (changes requested):
Only fix needed is to fix NSRange vs. String counting. 

LGTM otherwise!

Also, the current impl doesn't support for the instructions right?

 - `ARG var=` i.e. no value
 - `RUN $var` i.e. without enclosing `${}`
 - `ARG var2=${var1:-default}`

Could we add an issue for these (if true)

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
```diff
@@ -244,6 +244,143 @@ import Testing
...
+
+        // Verify ARG substitution
+        let stage = graph.stages[0]
+        if case .registry(let imageRef) = stage.base.source {
+            #expect(imageRef.stringValue == "alpine:latest")
+        } else {
+            #expect(Bool(false), "Expected registry image source")
+        }
+
+        if let run = stage.nodes[1].operation as? ExecOperation {
```
katiewasnothere, line 338: Can we add guards for accessing the nodes in the stage?

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
```diff
@@ -827,4 +964,49 @@ extension ParserTest {
             try parsePort(testCase)
         }
     }
+
+    @Test func testArgSubstitution() throws {
```
katiewasnothere: Should we move this test to the graph builder test file?

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerInstructionVisitor.swift
```diff
@@ -23,6 +23,9 @@ protocol InstructionVisitor {
     func visit(_ cmd: CMDInstruction) throws
     func visit(_ label: LabelInstruction) throws
     func visit(_ expose: ExposeInstruction) throws
+    func visit(_ arg: ArgInstruction) throws
+    // (1) Added instructions have to perform the ARG substitution
+    // (2) Update `testSimpleDockerfileArgInInstructions()`
```
katiewasnothere: Are these out of date comments?

dkovba: No, these are instructions for engineers who will be adding new instructions.

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
```diff
@@ -827,4 +964,49 @@ extension ParserTest {
...
+            let tokens: [Token] = [.stringLiteral("ARG"), .stringLiteral(testCase)]
+            let _ = try DockerfileParser().tokensToArgInstruction(tokens: tokens)
+        }
+    }
+
+    @Test func testArgWithSpacesAndQuotes() throws {
+        let dockerfile =
+            #"""
+            FROM alpine:latest AS build
+            ARG MESSAGE=Hello, World!
```
katiewasnothere: Is this a valid arg value?

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerfileParser.swift
```diff
@@ -363,4 +365,38 @@ public struct DockerfileParser: BuildParser {
 
         return try ExposeInstruction(rawPorts)
     }
+
+    func tokensToArgInstruction(tokens: [Token]) throws -> ArgInstruction {
```
katiewasnothere, line 369: the ARG instruction supports specifying multiple arg name+values in a single line (see [here](https://docs.docker.com/reference/dockerfile/#arg)). This function does not support that.

```
ARG <name>[=<default value>] [<name>[=<default value>]...]
```

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
dkovba: `Ç` actually worked. But something like `e\u{0301}` failed. Fixed.

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
dkovba: Making it `internal` led to `'hasActiveStage' is inaccessible due to 'internal' protection level` in ContainerBuildParser.

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
dkovba: The documentation for the ARG instruction says "build arguments". Please see a link below posted by @katiewasnothere.

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
dkovba: Done

Sources/NativeBuilder/ContainerBuildParser/Docker/Instructions/ExposeInstruction.swift
dkovba: Fixed

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
dkovba: Crashing immediately with an index out of bounds error in a test provides a clear, specific failure message that pinpoints exactly what assumption was violated, whereas a guard might obscure the root cause. This will also make it easier to add new tests for added instructions in the future.

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
dkovba: Done

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
dkovba: The documentation by the link below doesn't restrict the characters allowed in default values.

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder+Arg.swift
```diff
@@ -0,0 +1,101 @@
...
+                return defaultValue
+            }
+            // A stage-local ARG without a default value - a redeclared global FROM-only ARG
+            return buildArgs[key]
+        }
+
+        return nil
+    }
+
+    private static let argRegex: NSRegularExpression = {
```
wlan0, line 100: 👍

Tests/NativeBuilderTests/ContainerBuildParserTests/ParserTests.swift
katiewasnothere: Yeah the documentation is sparse in a lot of places. I think the behavior here though is that if there's spaces in the value, the value needs to be surrounded by quotes.

dkovba: An unquoted space seems to be reserved for:
```
ARG <name>[=<default value>] [<name>[=<default value>]...]
```

dkovba: Fixed

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerfileParser.swift
dkovba: Fixed

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerInstructionVisitor.swift
```diff
@@ -23,6 +23,16 @@ protocol InstructionVisitor {
...
+    func visit(_ arg: ArgInstruction) throws
+    // These instructions have to perform the ARG substitution checked in `testSimpleDockerfileArgInInstructions()`:
+    // - ADD
+    // - ENV
+    // - STOPSIGNAL
+    // - USER
+    // - VOLUME
+    // - WORKDIR
+    // - ONBUILD
+    // - ENTRYPOINT
```
wlan0, line 35: non-blocking, but CMD, LABEL, FROM also need substitution

edit: looks like cmd already substitutes, its just not mentioned in this comment

dkovba: We substitute ARGs in all existing commands in this PR.

wlan0 (approved):

Issue #432, opened 2025-08-04
Native builder parser support ADD instruction


katiewasnothere:

katiewasnothere:
Native builder work is not in our near term plan. Closing for now, but may reopen in the future when we're ready to work on native builder again.

PR #521, opened 2025-08-19
Add support for ADD in native builder parser

katiewasnothere:
Closes https://github.com/apple/container/issues/432

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "fbda395f591cc8b30792f91f1a1ab7f08c6dec742cc97861ef0a03caf7e6ddb4",
+  "originHash" : "29003ad0ff703439089a8e3a9d41bfe1a1d447679a40bffe701c492de6d099db",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/containerization.git",
       "state" : {
-        "revision" : "53021bec3606d4ca28e45c5f1d57461b652bb23a",
-        "version" : "0.6.0"
+        "revision" : "4099bdccd615bca49a2ea9cdb5b20dc5c5bd0fea",
+        "version" : "0.6.2"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ import PackageDescription
 let releaseVersion = ProcessInfo.processInfo.environment["RELEASE_VERSION"] ?? "0.0.0"
 let gitCommit = ProcessInfo.processInfo.environment["GIT_COMMIT"] ?? "unspecified"
 let builderShimVersion = "0.6.0"
-let scVersion = "0.6.0"
+let scVersion = "0.6.2"
 
 let package = Package(
     name: "container",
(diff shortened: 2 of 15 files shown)


Sources/NativeBuilder/ContainerBuildParser/Docker/Instructions/AddInstruction.swift
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        guard let destination = destination else {
+            throw ParseError.missingRequiredField("destination")
+        }
+
+        if let cs = checksum, !cs.hasPrefix("sha256:") {
+            throw ParseError.unexpectedValue
+        }
+
```
wlan0, line 58: I believe checksum only applies for HTTP(S) sources. https://docs.docker.com/reference/dockerfile/#add---checksum

Please add a guard for that

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerInstructionVisitor.swift
```diff
@@ -140,6 +141,12 @@ extension DockerInstructionVisitor {
         try graphBuilder.copyFromContext(paths: copy.sources, to: copy.destination, chown: copy.chown, chmod: copy.chmod)
     }
 
+    func visit(_ add: AddInstruction) throws {
+        // TODO katiewasnothere: plumb through "--link" option
```
wlan0, line 176: should we add an issue for this?

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder.swift
```diff
@@ -205,6 +205,22 @@ public final class GraphBuilder {
         return try add(operation)
     }
 
+    @discardableResult
+    public func addFiles(
```
wlan0, line 209: Since the number of these methods is increasing, it would be good to group these by operation type and into multiple files i.e. 

GraphBuilder+SourceOp
GraphBuilder+MetaOp 
GraphBuilder+RunOp

These could be done in a follow-up PR

Sources/NativeBuilder/ContainerBuildParser/Docker/DockerInstructionVisitor.swift
katiewasnothere: Add an issue here https://github.com/apple/container/issues/534. There is also a matching one for the copy option.

Sources/NativeBuilder/ContainerBuildIR/Graph/GraphBuilder.swift
katiewasnothere: Good idea, I'll add this in a follow up PR

Sources/NativeBuilder/ContainerBuildParser/Docker/Instructions/AddInstruction.swift
katiewasnothere: Added a guard for that and for `keepGitDir`

katiewasnothere:
Updated the PR to include arg substitution and updated related tests

wlan0 (approved):

katiewasnothere:
Closing for now. We'll be creating a new branch that we make native builder changes to going forward to enable faster iteration in the development process.

PR #530, opened 2025-08-20
CLI: Timeout if container.stop doesn't return in time

dcantah:
There's been bugs during dev where stop is hosed for one reason or another. To be safe as most folks will likely run `system stop` as a destructive "turn it off and on again" toggle, let's ensure we always make forward progress if something in stop is awry.

### DIFF
--- a/Sources/CLI/Container/ContainerStop.swift
+++ b/Sources/CLI/Container/ContainerStop.swift
@@ -17,6 +17,7 @@
 import ArgumentParser
 import ContainerClient
 import ContainerizationError
+import ContainerizationExtras
 import ContainerizationOS
 import Foundation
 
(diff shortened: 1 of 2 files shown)


Sources/CLI/Container/ContainerStop.swift
```diff
@@ -78,9 +79,18 @@ extension Application {
                 for container in containers {
                     group.addTask {
                         do {
-                            try await container.stop(opts: stopOptions)
-                            print(container.id)
-                            return nil
+                            return try await Timeout.run(seconds: UInt32(stopOptions.timeoutInSeconds + 2)) {
```
realrajaryan, line 82: Seems like we don't do any input validation on `timeoutInSeconds` and it can be negative. So, a `--time -10` input with the wraparound `UInt32(-10 + 2) = UInt32(-8)` would convert to `4294967288` seconds putting us in 136 years timeout here 😰?

dcantah: The timeout for stop should ideally be unsigned.

dcantah: Although docker does use negative values to signify there IS no timeout, which is interesting.

dcantah:
I'm honestly inclined to drop this change, I think it needs more thought

Issue #466, opened 2025-08-07
[Bug]: Does not use proxy provided by environment


SuzukiHonoka:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Set proxy envirment by use following vars:
- http_proxy
- https_proxy

Start the container system.

### Current behavior

It does not use the proxy settings in the environment.

### Expected behavior

Use proxy settings in environment.

### Environment

```markdown
- OS: 15.6
- Xcode: 16.4
- Container: container CLI version 0.3.0 (build: release, commit: 3fcf647)
```

### Relevant log output

```shell
#printenv|grep proxy                
no_proxy=127.0.0.1,localhost,0.0.0.0,.docker.internal,127.17.0.0/16,.cn,gitlab.kpy.cloud
http_proxy=http://127.0.0.1:10828
https_proxy=http://127.0.0.1:10828

#container s start
Verifying apiserver is running...
No default kernel configured.
Install the recommended default kernel from [https://github.com/kata-containers/kata-containers/releases/download/3.17.0/kata-static-3.17.0-arm64.tar.xz]? [Y/n]: y
Installing kernel...
Error: unknown: "StreamClosed(streamID: HTTP2StreamID(1), errorCode: HTTP2ErrorCode<0x1 ProtocolError>, location: "NIOHTTP2/HTTP2StreamChannel.swift:892")"
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
@SuzukiHonoka Good catch! This turns out to be a containerization issue, but in container the same limitation exists for  [kernel download](https://github.com/apple/container/blob/main/Sources/APIServer/Kernel/FileDownloader.swift#L60).

I filed apple/containerization#266 for the issue that you reported here; we can use this one for kernel download.

These should be easy fixes. Best would be to implement a pubic utility function down in Containerization to create the appropriate `HttpProxy?` based on URL and the standard environment variables (both upper and lower case), and then use that for both image fetch and kernel download.

Consider filing PRs!

SuzukiHonoka:
Thanks, I tested the 'HTTP_PROXY', but it does not work too, I don't see the traffic for downloding kernel at all.
Environment check:

```shell
# printenv|grep PROXY
HTTP_PROXY=http://127.0.0.1:10828
HTTPS_PROXY=http://127.0.0.1:10828
```

PR #533, opened 2025-08-22
feat(ContainerClient): add proxy utility

mayooot:
fix: https://github.com/apple/container/issues/466

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "e8b443090b53c809167aee6992ace3db1aee9822463df42e3fdc3319d0324a4b",
+  "originHash" : "90f79c1cf088508fab9c425d97b2cae0d922db9b9f1298f80e07bfb75096114d",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -22,7 +22,7 @@
     {
       "identity" : "dns",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/Bouke/DNS.git",
+      "location" : "https://github.com/Bouke/DNS",
       "state" : {
         "revision" : "78bbd1589890a90b202d11d5f9e1297050cf0eb2",
         "version" : "1.2.0"
@@ -31,7 +31,7 @@
     {
       "identity" : "dnsclient",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/orlandos-nl/DNSClient.git",
+      "location" : "https://github.com/orlandos-nl/DNSClient",
       "state" : {
         "revision" : "551fbddbf4fa728d4cd86f6a5208fe4f925f0549",
         "version" : "2.4.4"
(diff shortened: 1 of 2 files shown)


mayooot:
@jglogan Could you please take a look?

egernst:
@mayooot - thanks for the PR. 

From reading the original issue discussion, I think the direction suggested was to add the utility in github.com/apple/containerization since we'll also want to address this there, and then use that utility in container?

mayooot:
> @mayooot - thanks for the PR. 
> 
> 
> 
> From reading the original issue discussion, I think the direction suggested was to add the utility in github.com/apple/containerization since we'll also want to address this there, and then use that utility in container?

Ok, thanks for your reply. Do you have any suggestions on where to place this utility?

jglogan:
> Do you have any suggestions on where to place this utility?

@mayooot The issue in the containerization repo is apple/containerization#255. Having it there could allow us to do the right thing for image downloads as well as kernel.

I'd start by having a look at the issue (see the comment at https://github.com/apple/containerization/issues/255#issuecomment-3256865988) and then adapting what's in your PR to that. IMO ContainerizationExtras is the best place to put it the proxy utility.

Then this PR would just contain your kernel download fix. I'll create another issue for the image pull/push fix in container so that we can keep this PR small and quick to merge.

jglogan:
Added #578 for image push/pull.

mayooot:
> > Do you have any suggestions on where to place this utility?
> 
> @mayooot The issue in the containerization repo is [apple/containerization#255](https://github.com/apple/containerization/issues/255). Having it there could allow us to do the right thing for image downloads as well as kernel.
> 
> I'd start by having a look at the issue (see the comment at [apple/containerization#255 (comment)](https://github.com/apple/containerization/issues/255#issuecomment-3256865988)) and then adapting what's in your PR to that. IMO ContainerizationExtras is the best place to put it the proxy utility.
> 
> Then this PR would just contain your kernel download fix. I'll create another issue for the image pull/push fix in container so that we can keep this PR small and quick to merge.

Thanks a lot, I'll check it out ASAP.

jglogan:
For this PR, this will probably need to be fixed: https://github.com/apple/container/blob/cb2f42442096f950f70b63ef1e4ce263793089ce/.github/workflows/common.yml#L70

Might even be good to add a little fuzzing in the git workflow so that it randomly chooses HTTPS_PROXY or https_proxy.

jglogan:
@mayooot Have a go at an updated version once we merge #606. You should be able to bring in your proxy utility then.

Did you want to have a crack at apple/containerization#255 and then the container updates to use the affected image store APIs?

mayooot:
> @mayooot Have a go at an updated version once we merge #606. You should be able to bring in your proxy utility then.
> 
> Did you want to have a crack at [apple/containerization#255](https://github.com/apple/containerization/issues/255) and then the container updates to use the affected image store APIs?

Sure, I can take a look.

PR #536, opened 2025-08-22
ContainerService: Move force delete logic to daemon

dcantah:
Today force deleting (if a container is running then stop()'ing first) is handled entirely in the cli, which is brittle. The CLI doesn't know if the container was started with --rm so it would have to do a weird timeout + list dance to check if the containers gone after stopping. This change remedies this by just having the daemon take in a `force` boolean to the delete rpc. If this is provided and the container is running then we'll stop first, and then cleanup. We can additionally not cleanup if --rm was provided as the daemon has the data to determine if a container asked for autoRemove.

### DIFF
--- a/Sources/CLI/Container/ContainerDelete.swift
+++ b/Sources/CLI/Container/ContainerDelete.swift
@@ -80,33 +80,23 @@ extension Application {
             var failed = [String]()
             let force = self.force
             let all = self.all
-            try await withThrowingTaskGroup(of: ClientContainer?.self) { group in
+            try await withThrowingTaskGroup(of: String?.self) { group in
                 for container in containers {
                     group.addTask {
                         do {
-                            // First we need to find if the container supports auto-remove
-                            // and if so we need to skip deletion.
-                            if container.status == .running {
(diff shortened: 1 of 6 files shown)


Sources/CLI/Container/ContainerDelete.swift
```diff
@@ -80,33 +80,29 @@ extension Application {
...
-                                    if all {
-                                        return nil
+                                    guard all else {
+                                        throw ContainerizationError(
+                                            .invalidState,
+                                            message: "container is running"
+                                        )
                                     }
-                                    throw ContainerizationError(.invalidState, message: "container is running")
+                                    return nil
```
realrajaryan: Since we no longer error here, we can simplify this nesting and get rid of that comment.

```swift
  if container.status == .running && !force {
      guard all else {
          throw ContainerizationError(.invalidState, message: "container
   is running")
      }
      return nil  // Skip running container when using --all
  }
  ```

dcantah: @realrajaryan done

PR #541, opened 2025-08-25
Integration: Log container bootlog in run test failures

dcantah:
If any of the container run tests fail on the basic setup phases (e.g. actually starting the container) we should log the bootlog to see if we can glean any info off of this.

### DIFF
--- a/Tests/CLITests/Subcommands/Run/TestCLIRunOptions.swift
+++ b/Tests/CLITests/Subcommands/Run/TestCLIRunOptions.swift
@@ -14,8 +14,6 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-//
-
 import AsyncHTTPClient
 import ContainerClient
 import ContainerizationExtras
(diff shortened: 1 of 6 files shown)


Tests/CLITests/Subcommands/Run/TestCLIRunOptions.swift
```diff
@@ -24,61 +22,80 @@ import Foundation
 import Testing
 
 class TestCLIRunCommand: CLITest {
-    @Test func testRunCommand() throws {
+    func runHelper(name: String, testCase: () throws -> Void) {
```
dkovba, line 25: Can we wrap the async version of this method `runHelperAsync()` into `Task {...}` to avoid code duplication?

dkovba (approved):

Tests/CLITests/Subcommands/Run/TestCLIRunOptions.swift
dcantah: Can you describe more what you want? I still can't use the sync helper in a Task body if the callback needs to call an async method (only one of our tests). This seemed the least tedious of the two between just solely having an async helper and needing to annotate all of the runHelper calls with await.

dkovba: The use of `Task {...}` led to:
> error: passing closure as a 'sending' parameter risks causing data races between code in the current task and concurrent 

Adding `async` to test functions and `await` to `runHelperAsync()` might still be better than code duplication.

PR #545, opened 2025-08-27
Uniform support for `--platform`, `--os`, `--arch`.

jglogan:
- Extends #313 (@dcantah) so that all of `container create`, `container run`, `container build`, `container image pull`, and `container image save` accept the three options.
- `container build` creates a platform list as the union of the `--platform` list with all combinations of the architecture and OS lists.
- All other commands work accept a single platform, preferring the `--platform` option over `--arch` and `--os` when both are specified. `--os` defaults to `linux`, and `--arch` defaults to the host architecture.


dcantah:
Needs a `make fmt`

dcantah (approved):

dcantah (approved):

PR #550, opened 2025-08-28
Test fixes for 0.4.1 release.

jglogan:
- Disabled tests that failed with release config.
- Changed so merge builds build with release config so this doesn't burn us next release.
- Added `swift --version` to Makefile so we can see what we're running up in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,7 @@ all: init-block
 .PHONY: build
 build:
 	@echo Building container binaries...
+	@$(SWIFT) --version
 	@$(SWIFT) build -c $(BUILD_CONFIGURATION)
 
 .PHONY: container
--- a/Tests/NativeBuilderTests/ContainerBuildIRTests/DependencyAnalysisTests.swift
+++ b/Tests/NativeBuilderTests/ContainerBuildIRTests/DependencyAnalysisTests.swift
@@ -67,8 +67,7 @@ struct DependencyAnalysisTests {
         #expect(analyzedStage.nodes[2].dependencies.count == 1)
     }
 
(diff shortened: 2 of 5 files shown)


Tests/NativeBuilderTests/ContainerBuildIRTests/DependencyAnalysisTests.swift
```diff
@@ -132,7 +133,8 @@ struct DependencyAnalysisTests {
             "Entrypoint should depend on copy operation")
     }
 
-    @Test func stageReferenceResolution() throws {
+    //@Test
```
madrob: I think using Test(.disabled) is preferred.

Makefile
```diff
@@ -50,6 +50,7 @@ all: init-block
 .PHONY: build
 build:
 	@echo Building container binaries...
+	@$(SWIFT) --version
```
madrob, line 53: Do you want this on every build or only if CI=true?

jglogan: It's fine on every build - most developers won't even notice it on their local builds but it'll be there if they want a quick check

Tests/NativeBuilderTests/ContainerBuildIRTests/DependencyAnalysisTests.swift
jglogan: TIL!

jglogan: TIL! Fixing.

```swift
@Test(.disabled())
```

PR #567, opened 2025-09-02
APIServer: Make container deletes a little more robust

dcantah:
Today it's possible we receive the container exit event sent to the daemon after a user might've invoked `container delete`. The aftermath of that unfortunate scenario is that you would be met with an error stating we can't delete the container because it is running. This is not really common unless someone was doing kill/stop+delete in some script, but it still is an issue that needs to be fixed. The true source of truth for state is actually the sandboxservice hosting the container, so if we were guaranteed that that is still alive, we can query and take some action based on the possible mismatch in current snapshot state and what the sb service is telling us.

To make this possible I've introduced a new rpc on the sandbox service named shutdown. Shutdown's purpose is to run some cleanup code (if there is any) and then to wind down the process. Prior to this RPC the sandbox service would basically just call exit(0) on its own accord after the container exited. Now this process is driven by the APIServer instead, which I like a little bit more. The true win here though is that because we know as long as we haven't sent the shutdown RPC that the process is alive (barring a bug), we can query the state of the container the whole time.

### DIFF
--- a/Sources/APIServer/Containers/ContainersService.swift
+++ b/Sources/APIServer/Containers/ContainersService.swift
@@ -122,7 +122,11 @@ actor ContainersService {
             throw ContainerizationError(.exists, message: "hostname(s) already exist: \(conflictingHostnames)")
         }
 
-        self.containers[configuration.id] = ContainerSnapshot(configuration: configuration, status: .stopped, networks: [])
+        self.containers[configuration.id] = ContainerSnapshot(
+            configuration: configuration,
+            status: .stopped,
+            networks: []
+        )
 
         let runtimePlugin = self.runtimePlugins.filter {
             $0.name == configuration.runtimeHandler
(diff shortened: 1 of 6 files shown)


Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -207,36 +211,63 @@ actor ContainersService {
...
+                // state here isn't updated yet. We are perfectly fine to delete the container,
+                // so let's update our state and do just that.
+                if ctrState.status == .stopped {
+                    let snapshot = ContainerSnapshot(
+                        configuration: item.configuration,
+                        status: .stopped,
+                        networks: []
+                    )
+                    await self.setContainer(id, snapshot, context: context)
+                    if !autoRemove {
```
jglogan, line 237: perhaps a one-line explanation of why we don't invoke this for the auto-remove case? who's responsible for cleanup?

Sources/APIServer/Containers/ContainersService.swift
```diff
@@ -207,36 +211,63 @@ actor ContainersService {
...
+                        try await self.cleanup(id: id, item: item, context: context)
+                    }
+                    return
+                }
+
+                if !force {
```
jglogan, line 243: nit: perhaps L243-L258 in an else, and then 237-240 can go away in favor of 259-262?

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -634,28 +633,44 @@ public actor SandboxService {
...
+        reply.set(key: .fd, value: fh)
+        return reply
+    }
+
+    public func shutdown(_ message: XPCMessage) async throws -> XPCMessage {
+        self.setState(.shuttingDown)
+
+        Task {
+            do {
+                try await Task.sleep(for: .seconds(3))
```
jglogan, line 665: If we're in the `.shuttingDown` state, the state going back to the containers helper will be `.stopped`, right? Is there any issue with the state being that way for 3 secs while the launch agent is still running? I'm thinking of a new container with the same name being created/run immediately.

dcantah:
I have a much better idea for this I'll be opening a change for

PR #569, opened 2025-09-03
Make "Plugin not found" message more descriptive

mazdak:
## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Description
Users frequently saw “failed to find plugin …” when the system services weren’t running; the message didn’t explain the root cause or where plugins are looked up. No change to plugin execution flow; only error messaging and path discovery hints are improved for a better UX.

## Motivation and Context
Users frequently saw “failed to find plugin …” when the system services weren’t running; the message didn’t explain the root cause or where plugins are looked up. I was confronted with this several times during plugin development because the container system was not running, but the error message led me to believe the files were missing.

## Testing
- [X] Tested locally
- [X ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/CLI/DefaultCommand.swift
+++ b/Sources/CLI/DefaultCommand.swift
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import Foundation
 import ContainerClient
 import ContainerPlugin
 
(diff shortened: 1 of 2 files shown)


dcantah:
This makes me think we should expose a `plugins` command and subsystem in the daemon, where one of the bits of info you could get from this would be the plugin directories so we don't need to do all of that logic here to craft them again.

dcantah:
Also, the license headers can be generated via `make fmt`, you don't need to add them manually.

Sources/CLI/DefaultCommand.swift
```diff
@@ -45,8 +46,43 @@ struct DefaultCommand: AsyncParsableCommand {
...
+            )
+        }
+
         guard let plugin = pluginLoader?.findPlugin(name: command), plugin.config.isCLI else {
-            throw ValidationError("failed to find plugin named container-\(command)")
+            throw ValidationError(
+                """
+                Plugin 'container-\(command)' not found.
+                - If system services are not running, start them with: container system start
+                - If the plugin isn't installed, ensure it exists under:
```
jglogan, line 87: @mazdak this all looks good. This one sentence is a bit confusing. What do you think about

```
- Check to see that the plugin exists under:
```

jglogan: What I see in my local build looks good.

Only other nit is that a blank line before `Usage:` might improve readability.

```console
% bin/container foo
Error: Plugin 'container-foo' not found.
- If system services are not running, start them with: container system start
- If the plugin isn't installed, ensure it exists under:
  - {my-project-path}/libexec/container-plugins/foo
  - {my-project-path}/libexec/container/plugins/foo
Usage: container [--debug] <subcommand>
  See 'container --help' for more information.
```

mazdak: I have made both those changes.

jglogan:
@mazdak Try rebasing and running `make fmt` - your build is failing the format checks. Otherwise everything looks good.

mazdak:
> @mazdak Try rebasing and running `make fmt` - your build is failing the format checks. Otherwise everything looks good.

Done and Done @jglogan

jglogan:
@mazdak Merged, thank you for the contribution!

Issue #574, opened 2025-09-04
[Request]: Add tests for installing kernels


katiewasnothere:
### Feature or enhancement request details

We currently support installing kernels from local files on disk and from within tar archives. We should add tests that both of these scenarios work as expected, including when the kernel is a symlink to another file. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

katiewasnothere:
Related to https://github.com/apple/container/issues/475

katiewasnothere:
Closed by https://github.com/apple/container/pull/579

Issue #475, opened 2025-08-09
[Bug]: `kernel set` doesn't handle symlinks within the tarball


bennettp123:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. `curl -O 'https://github.com/kata-containers/kata-containers/releases/download/3.19.1/kata-static-3.19.1-arm64.tar.xz`
2. ``container system kernel set --tar `pwd`/kata-static-3.19.1-arm64.tar.xz --binary opt/kata/share/kata-containers/vmlinux.container``
    * _note: inside kata-static-3.19.1-arm64.tar.xz, vmlinux.container is a symlink to vmlinux-6.12.36-160._
3. `container run <...>`
4. `defaults delete com.apple.container.defaults`
5. `container system kernel set --recommended`
6. `container run <...>`

### Current behavior

* Step 2: the requested kernel image is extracted and activated (as expected)
* Step 3: `container run <...>` fails with the following error:
    ```
    Error: unknown: "Error Domain=VZErrorDomain Code=1 "The virtual machine failed to start." UserInfo={NSLocalizedFailure=Internal Virtualization error., NSLocalizedFailureReason=The virtual machine failed to start.}"
    ```
* Step 5: `container system kernel set --recommended` fails with the following error:
    ```
    Installing the recommended kernel from https://github.com/kata-containers/kata-containers/releases/download/3.17.0/kata-static-3.17.0-arm64.tar.xz...
    Error: unknown: "Error Domain=NSCocoaErrorDomain Code=516 "“vmlinux-6.12.28-153” couldn’t be copied to “kernels” because an item with the same name already exists." UserInfo={NSSourceFilePathErrorKey=/var/folders/gt/ptf_r7hs7y3gw1mzqczbx78w0000gn/T/80190809-A9EC-4F5E-BC71-A639505C33BD/vmlinux-6.12.28-153, NSUserStringVariant=(
        Copy
[...]

PR #575, opened 2025-09-04
Support following symlinks in tar when setting default kernel

katiewasnothere:
## Type of Change
- [x] Bug fix

## Description
This PR checks if the found entry in the archive for the requested kernel path is a symlink, and if so, follows the symlink. I've opened a separate issue to track creating tests for kernel downloading here https://github.com/apple/container/issues/574 and will create a follow-up PR for those. 

## Motivation and Context
This PR fixes an issue reported in https://github.com/apple/container/issues/475 where installing a kernel from a tar archive does not follow symlinks in the archive. 

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/APIServer/Kernel/KernelService.swift
+++ b/Sources/APIServer/Kernel/KernelService.swift
@@ -82,8 +82,7 @@ actor KernelService {
         await progressUpdate?([
             .setDescription("Unpacking kernel")
         ])
-        let archiveReader = try ArchiveReader(file: tarFile)
-        let kernelFile = try archiveReader.extractFile(from: kernelFilePath, to: tempDir)
+        let kernelFile = try self.extractFile(tarFile: tarFile, at: kernelFilePath, to: tempDir)
         try self.installKernel(kernelFile: kernelFile, platform: platform)
 
         if !FileManager.default.fileExists(atPath: tar.absoluteString) {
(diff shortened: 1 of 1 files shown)


Sources/APIServer/Kernel/KernelService.swift
```diff
@@ -112,13 +111,27 @@ actor KernelService {
...
+        var target = at
+        var archiveReader = try ArchiveReader(file: tarFile)
+        var (entry, data) = try archiveReader.extractFile(path: target)
+
+        // if the target file is a symlink, get the data for the actual file
+        if entry.fileType == .symbolicLink, let symlinkRelative = entry.symlinkTarget {
+            // the previous extractFile changes the underlying file pointer, so we need to reopen the file
+            // to ensure we traverse all the files in the archive
+            archiveReader = try ArchiveReader(file: tarFile)
+            let symlinkTarget = URL(filePath: target).appending(path: symlinkRelative)
```
wlan0: The path should be relative to the containing dir, not to the file itself. 

i.e.  if the symlink is 


`vmlinuz -> vmlinuz-<ver>
`
The current logic will resolve symlinkTarget to `vmlinuz/vmlinuz-<ver>` instead of `vmlinuz-<ver>`

katiewasnothere: Fixed! Thanks for pointing that out :)

PR #579, opened 2025-09-05
Add force option to kernel set and add tests for kernel setting

katiewasnothere:
## Type of Change
- [x] New feature  

## Description
Add option to force kernel setting and tests for CLI `kernel set`. Related to https://github.com/apple/container/pull/575. 

## Motivation and Context
This PR adds additional tests to ensure that we can set kernels from local files, remote tar files, and local tar files. A new option `force` is added to the `kernel set` subcommand which will overwrite an existing kernel with the same name if one exists to make testing easier. The tests ensure that a container can be started with the newly set kernel and resets to the default recommended kernel when complete. 

## Testing
- [x] Tested locally
- [x] Added/updated tests

### DIFF
--- a/Sources/APIServer/Kernel/KernelService.swift
+++ b/Sources/APIServer/Kernel/KernelService.swift
@@ -37,10 +37,19 @@ actor KernelService {
 
     /// Copies a kernel binary from a local path on disk into the managed kernels directory
     /// as the default kernel for the provided platform.
-    public func installKernel(kernelFile url: URL, platform: SystemPlatform = .linuxArm) throws {
+    public func installKernel(kernelFile url: URL, platform: SystemPlatform = .linuxArm, force: Bool) throws {
         self.log.info("KernelService: \(#function) - kernelFile: \(url), platform: \(String(describing: platform))")
         let kFile = url.resolvingSymlinksInPath()
         let destPath = self.kernelDirectory.appendingPathComponent(kFile.lastPathComponent)
+        if force {
+            do {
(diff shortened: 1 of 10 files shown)


Sources/APIServer/Kernel/KernelService.swift
```diff
@@ -37,10 +37,13 @@ actor KernelService {
 
     /// Copies a kernel binary from a local path on disk into the managed kernels directory
     /// as the default kernel for the provided platform.
-    public func installKernel(kernelFile url: URL, platform: SystemPlatform = .linuxArm) throws {
+    public func installKernel(kernelFile url: URL, platform: SystemPlatform = .linuxArm, force: Bool) throws {
         self.log.info("KernelService: \(#function) - kernelFile: \(url), platform: \(String(describing: platform))")
         let kFile = url.resolvingSymlinksInPath()
         let destPath = self.kernelDirectory.appendingPathComponent(kFile.lastPathComponent)
+        if force {
+            try? FileManager.default.removeItem(at: destPath)
```
dcantah: I wouldn't ignore the error, because if the remove fails we're just going to hit the existing error we already were before (which will likely be more confusing to debug if we did set --force)

dcantah (commented):
LGTM, but one comment. Will re-review after!

Sources/APIServer/Kernel/KernelService.swift
katiewasnothere: Updated! PTAL

dcantah (approved):

PR #603, opened 2025-09-11
Expose Command Structs for Plugins

Mcrich23:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Plugins technically exist, but to add shortcuts or to do existing things with functions in `container` requires calling a compiled binary. This pull request aims to remove that hurdle and instability by exposing commands as a new `ContainerCLI` target.

Simply import `ContainerCLI` and you can access almost any command as if it were a native part of the binary. This makes plugin development significantly easier.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

> No new tests or documentation in code needed?


Mcrich23:
@jglogan This also accomplishes some of the work that I found necessary when developing my initial compose effort, because it increases safety and reduces the developer workload to reproduce, funnel, or output commands being run in the background. It also will encourage devs to use the commands internal to container instead of modifying containers and other things directly.

jglogan:
How are you running the commands? Are you using ParseableCommand.parse()?

Mcrich23:
You can do that, or just init it and then manually set every property after it. 

Because of how Swift parser operates, we can't just init it with all of the properties, so we have to instead init it empty and then set all of the properties. 

An example of this is in my Compose proposal.

jglogan:
Having the ParseableCommand types public so we can all import the library and use parse() totally makes sense.

I'd prefer that that be just the API, instead of making it so that any change to the member properties is potentially a breaking change. Could we start with this PR simply allowing clients to use parse(), without changing member visibility?

Mcrich23:
@jglogan We could put that in the documentation, but `AsyncParsableCommand` requires an empty public init to be specified in any public structure that conforms to it, so it would have to be available technically.

Mcrich23:
However, we could make all properties and functions private to force use of the parse method.

jglogan:
> However, we could make all properties and functions private to force use of the parse method.

Exactly! Could you try that approach - making the CLI library such that you can import it and use parse, but the property visibility isn't public? I'm curious how that'd work for your plugin.

jglogan:
Also, do you have an enhancement issue open for making the CLI library importable? If not, please create one and I'll assign it to you. It'd be best to have an issue backing each PR (especially for larger changes or those affecting the API). Thanks!

Mcrich23:
I didn't have one, so I just made #609. But as I try to update my Compose plugin to use entirely parse, it is becoming increasingly difficult to pass down global flags that I want to respect inherently such as `Flags.Global.verbose` or `Flags.Progress.disableProgressUpdates`.

I strongly think that a great DX is the only way that people will make plugins for `compose`, so I think we kinda only have two ways to move forward. Either, we expose every property on commands, or we can expose flag groups in a setable manner that allows developers to pass down flag groups.

If we do the latter, this would not be required because `.parse()` would set them by default, but it would also allow them to be easily overriden and treated in a sort of "environment" manner.

Mcrich23:
@jglogan We can undo it, but I decided to go ahead and make all of the OptionGroup properties public to make this possible. Again these are not required to be set when using `.parse` it just makes passing down use flags far easier.

jglogan:
The reason I don't want to see any of the properties public is that it those are implementation, not interface. The interface for a ParseableCommand are the parse/validate/run methods.

As soon as we make the properties the interface, it reduces our ability to change anything without breaking clients and creates the possibility to create all kinds of commands that break invariants.

Let's make all the properties non-public and just use parse() to populate the fields.

Separately, is there a reason you're preferring to invoke the CLI from a plugin as opposed to using ContainerClient? Wondering if there's something about it that needs improvement for your use case. It is still pretty rough as we've been focusing more on fixes and features than refining the "SDK".

Mcrich23:
I see. I just wonder if more boilerplate code and code breaking may occur if developers want to pass down certain flag structures to commands because the only way to do this if those properties are not public is via adding a line for every property, which would also require updating any time that structure got a new parameter.

Separately, I really like invoking the commands for some things instead of ContainerClient because I know that for basic things like ImagePulling, I can get it very easy and the abstraction to the command means that I also am safe if the internal logic and internal APIs change, because that command will be updated and I just call to that.

Mcrich23:
By the way, one example of a benefit from using the command instead of recreating the action is that I automatically get progress bars and related loading terminal stuff without having to build it in for each command. It is already properly used and compartmentalized for readable code. And I can just get it for free!

jglogan:
> I just wonder if more boilerplate code and code breaking may occur if developers want to pass down certain flag structures to commands because the only way to do this if those properties are not public is via adding a line for every property, which would also require updating any time that structure got a new parameter.

Perhaps? Perhaps not. Still, this library is a command line interface library, and as such the interface is the command line (via the protocol conformance to ParseableCommand) and not the properties.

Mcrich23:
Yes, but I think an exception should be made for OptionGroups.

dcantah:
I am not a huge fan of this. If a user/plugin author wants to do {thing that CLI command does} we should try and expose helpers that do the {thing} as much as we can within reason, and not the CLI commands and properties themselves. I think our maintenance/backwards compat should begin and end at the CLI interface; having to worry about changing something in the properties/commands themselves and possibly breaking someone seems like much more than I'd like to sign up for this early in the project.

Obviously that comes with the work of exposing helpers for {things}, but in the long run I think we can build up a decent arsenal of them. Which command would you say you would get the most value out of being exposed, maybe we can start there

Mcrich23:
I agree for most things. I am just saying that passing down parameters for flags such as `Flags.Global` is harder to do when doing it all with `.parse` and developers don't inherit updates for it when new parameters are added. They have to go and add support for those parameters.

The reason why I find passing stuff down in the first place is to respect requests by the user. If the user wants verbose output, it happens with every command and minimal work from me. If they want to set the current working directory, I don't want to have to pass that down in string form if it can happen with type safe swift.

I completely agree that we shouldn't necessarily be exposing properties, but I think we have to find a way to pass down OptionGroups in a manner that does not require developers to do more than re-compile plugins when a new version of container is released if they want to support new flags in that release.

Mcrich23:
My big concern is that if this is not implemented correctly, developers will turn towards using the shell environment and it will make plugins less portable than if we do. I'm not saying we have to expose everything, but I think finding a middle ground to encourage developers to use existing APIs and avoid the shell environment is likely important to all of us.

jglogan:
> I agree for most things. I am just saying that passing down parameters for flags such as `Flags.Global` is harder to do when doing it all with `.parse` and developers don't inherit updates for it when new parameters are added. They have to go and add support for those parameters.

Can you provide a concrete example of what you're saying?

Mcrich23:
Happy to!

Here is what you are proposing would be required to pass down any specified flags from `Flags.Management`:
```swift
@OptionGroup
var managementFlags: Flags.Management

mutating func run2() async throws {
    var arguments: [String] = ["hello-world"]

    // Expand managementFlags into CLI arguments
    if managementFlags.detach {
        arguments.append("--detach")
    }

    if let entryPoint = managementFlags.entryPoint {
        arguments.append(contentsOf: ["--entrypoint", entryPoint])
    }

    for mount in managementFlags.mounts {
        arguments.append(contentsOf: ["--mount", mount])
    }

    for port in managementFlags.publishPorts {
        arguments.append(contentsOf: ["--publish", port])
    }

    for socket in managementFlags.publishSockets {
        arguments.append(contentsOf: ["--publish-socket", socket])
    }

    for tmp in managementFlags.tmpFs {
        arguments.append(contentsOf: ["--tmpfs", tmp])
    }

    if let name = managementFlags.name {
        arguments.append(contentsOf: ["--name", name])
    }

    if managementFlags.remove {
        arguments.append("--rm")
    }

    if let platform = managementFlags.platform {
        arguments.append(contentsOf: ["--platform", platform])
    }

    arguments.append(contentsOf: ["--os", managementFlags.os])
    arguments.append(contentsOf: ["--arch", managementFlags.arch])

    for vol in managementFlags.volumes {
        arguments.append(contentsOf: ["--volume", vol])
    }

    if let kernel = managementFlags.kernel {
        arguments.append(contentsOf: ["--kernel", kernel])
    }

    for network in managementFlags.networks {
        arguments.append(contentsOf: ["--network", network])
    }

    if !managementFlags.cidfile.isEmpty {
        arguments.append(contentsOf: ["--cidfile", managementFlags.cidfile])
    }

    if managementFlags.dnsDisabled {
        arguments.append("--no-dns")
    }

    for dns in managementFlags.dnsNameservers {
        arguments.append(contentsOf: ["--dns", dns])
    }

    if let domain = managementFlags.dnsDomain {
        arguments.append(contentsOf: ["--dns-domain", domain])
    }

    for search in managementFlags.dnsSearchDomains {
        arguments.append(contentsOf: ["--dns-search", search])
    }

    for option in managementFlags.dnsOptions {
        arguments.append(contentsOf: ["--dns-option", option])
    }

    for label in managementFlags.labels {
        arguments.append(contentsOf: ["--label", label])
    }

    if managementFlags.virtualization {
        arguments.append("--virtualization")
    }

    if managementFlags.ssh {
        arguments.append("--ssh")
    }

    // Now parse with all expanded args
    var createCommand = try Application.ContainerCreate.parse(arguments)
    try await createCommand.run()
}
```

Here is what I would like to be possible:
```swift
@OptionGroup
var managementFlags: Flags.Management

mutating func run() async throws {
    var createCommand = try Application.ContainerCreate.parse(["hello-world"])
    createCommand.managementFlags = managementFlags
    try await createCommand.run()
}
```

I just want to be able to pass in user configuration stuff with a minimal amount of work. Plus in the world that we do the former, adding a new argument to `Flags.Management` would mean that developers would have to manually add support for it. Meanwhile the latter would just require recompilation.

jglogan:
> I am not a huge fan of this. If a user/plugin author wants to do {thing that CLI command does} we should try and expose helpers that do the {thing} as much as we can within reason, and not the CLI commands and properties themselves. I think our maintenance/backwards compat should begin and end at the CLI interface

I'm okay with being able to use swift-argument-parser to run commands without having to exec them. The library provides an API for that. The above case notwithstanding, we shouldn't be making the properties on the command public though.

Mcrich23:
I agree with @jglogan and like the simplicity of calling into the existing commands instead of creating a common layer for both to run commands through, because doing so would cause more issues when updating any code at all.

Sources/CLI/Application.swift
```diff
@@ -35,12 +35,13 @@ nonisolated(unsafe) var log = {
     return log
 }()
 
-@main
-struct Application: AsyncParsableCommand {
+public struct Application: AsyncParsableCommand {
     @OptionGroup
-    var global: Flags.Global
+    public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerCreate.swift
```diff
@@ -33,21 +35,21 @@ extension Application {
         var arguments: [String] = []
 
         @OptionGroup
-        var processFlags: Flags.Process
+        public var processFlags: Flags.Process
```
jglogan: Remove public visibility for all of these.

Sources/CLI/Container/ContainerDelete.swift
```diff
@@ -33,12 +35,12 @@ extension Application {
         var all = false
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerLogs.swift
```diff
@@ -22,14 +22,16 @@ import Dispatch
...
+        public init() {}
+        
+        public static let configuration = CommandConfiguration(
             commandName: "logs",
             abstract: "Fetch container stdio or boot logs"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerStart.swift
```diff
@@ -33,12 +35,12 @@ extension Application {
         var interactive = false
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/System/SystemLogs.swift
```diff
@@ -22,16 +22,18 @@ import Foundation
...
 
-        static let configuration = CommandConfiguration(
+        public static let configuration = CommandConfiguration(
             commandName: "logs",
             abstract: "Fetch system logs for `container` services"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/RunCommand.swift
```diff
@@ -26,36 +26,37 @@ import NIOPosix
...
-        static let configuration = CommandConfiguration(
+    public struct ContainerRunCommand: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "run",
             abstract: "Run a container")
 
         @OptionGroup
-        var processFlags: Flags.Process
+        public var processFlags: Flags.Process
```
jglogan: Remove public visibility for all of these.

Sources/CLI/Registry/RegistryDefault.swift
```diff
@@ -34,22 +35,23 @@ extension Application {
...
+    public struct DefaultSetCommand: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "set",
             abstract: "Set the default registry"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility for these.

Sources/CLI/Registry/Logout.swift
```diff
@@ -20,17 +20,18 @@ import Containerization
...
+        public init() {}
+        public static let configuration = CommandConfiguration(
             abstract: "Log out from a registry")
 
         @Argument(help: "Registry server name")
         var registry: String
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Registry/Login.swift
```diff
@@ -37,9 +38,9 @@ extension Application {
         var server: String
 
         @OptionGroup
-        var registry: Flags.Registry
+        public var registry: Flags.Registry
```
jglogan: Remove public visibility.

Sources/CLI/Network/NetworkList.swift
```diff
@@ -35,18 +36,18 @@ extension Application {
         var format: ListFormat = .table
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Network/NetworkInspect.swift
```diff
@@ -21,18 +21,19 @@ import Foundation
...
-        static let configuration = CommandConfiguration(
+    public struct NetworkInspect: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "inspect",
             abstract: "Display information about one or more networks")
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Network/NetworkDelete.swift
```diff
@@ -31,12 +32,12 @@ extension Application {
         var all = false
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Network/NetworkCreate.swift
```diff
@@ -22,18 +22,19 @@ import Foundation
...
+        public static let configuration = CommandConfiguration(
             commandName: "create",
             abstract: "Create a new network")
 
         @Argument(help: "Network name")
         var name: String
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageTag.swift
```diff
@@ -30,9 +31,9 @@ extension Application {
         var target: String
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageSave.swift
```diff
@@ -23,14 +23,15 @@ import Foundation
...
+    public struct ImageSave: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "save",
             abstract: "Save an image as an OCI compatible tar archive"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageRemove.swift
```diff
@@ -79,20 +81,22 @@ extension Application {
         }
     }
 
-    struct ImageRemove: AsyncParsableCommand {
+    public struct ImageRemove: AsyncParsableCommand {
+        public init() {}
+        
         @OptionGroup
-        var options: RemoveImageOptions
+        public var options: RemoveImageOptions
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageRemove.swift
```diff
@@ -21,18 +21,20 @@ import ContainerizationError
...
+        
         @Flag(name: .shortAndLong, help: "Remove all images")
         var all: Bool = false
 
         @Argument
         var images: [String] = []
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImagePush.swift
```diff
@@ -21,26 +21,27 @@ import ContainerizationOCI
...
+    public struct ImagePush: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "push",
             abstract: "Push an image"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility for these.

Sources/CLI/Image/ImagePull.swift
```diff
@@ -22,20 +22,20 @@ import ContainerizationOCI
...
-        static let configuration = CommandConfiguration(
+    public struct ImagePull: AsyncParsableCommand {
+        public static let configuration = CommandConfiguration(
             commandName: "pull",
             abstract: "Pull an image"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility for these.

Sources/CLI/Image/ImagePrune.swift
```diff
@@ -19,15 +19,16 @@ import ContainerClient
...
-        static let configuration = CommandConfiguration(
+    public struct ImagePrune: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "prune",
             abstract: "Remove unreferenced and dangling images")
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageLoad.swift
```diff
@@ -22,14 +22,15 @@ import Foundation
...
+    public struct ImageLoad: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "load",
             abstract: "Load images from an OCI compatible tar archive"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageList.swift
```diff
@@ -160,16 +162,17 @@ extension Application {
...
+    public struct ImageList: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "list",
             abstract: "List images",
             aliases: ["ls"])
 
         @OptionGroup
-        var options: ListImageOptions
+        public var options: ListImageOptions
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageList.swift
```diff
@@ -34,19 +35,20 @@ extension Application {
         var format: ListFormat = .table
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Image/ImageInspect.swift
```diff
@@ -21,18 +21,19 @@ import Foundation
...
-        static let configuration = CommandConfiguration(
+    public struct ImageInspect: AsyncParsableCommand {
+        public init() {}
+        public static let configuration = CommandConfiguration(
             commandName: "inspect",
             abstract: "Display information about one or more images")
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/DefaultCommand.swift
```diff
@@ -21,13 +21,13 @@ import Darwin
...
 struct DefaultCommand: AsyncParsableCommand {
-    static let configuration = CommandConfiguration(
+    public static let configuration = CommandConfiguration(
         commandName: nil,
         shouldDisplay: false
     )
 
     @OptionGroup(visibility: .hidden)
-    var global: Flags.Global
+    public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerStop.swift
```diff
@@ -39,9 +41,9 @@ extension Application {
         var containerIDs: [String] = []
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerExec.swift
```diff
@@ -21,24 +21,26 @@ import ContainerizationOS
...
+    public struct ContainerExec: AsyncParsableCommand {
+        public init() {}
+        
+        public static let configuration = CommandConfiguration(
             commandName: "exec",
             abstract: "Run a new command in a running container")
 
         @OptionGroup
-        var processFlags: Flags.Process
+        public var processFlags: Flags.Process
```
jglogan: Remove public visibility for these.

Sources/CLI/Container/ContainerInspect.swift
```diff
@@ -20,18 +20,20 @@ import Foundation
...
+    public struct ContainerInspect: AsyncParsableCommand {
+        public init() {}
+        
+        public static let configuration = CommandConfiguration(
             commandName: "inspect",
             abstract: "Display information about one or more containers")
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerKill.swift
```diff
@@ -36,9 +38,9 @@ extension Application {
         var containerIDs: [String] = []
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

Sources/CLI/Container/ContainerList.swift
```diff
@@ -38,9 +40,9 @@ extension Application {
         var format: ListFormat = .table
 
         @OptionGroup
-        var global: Flags.Global
+        public var global: Flags.Global
```
jglogan: Remove public visibility.

jglogan (changes requested):

Mcrich23:
@jglogan am I correct in understanding that the idea of passing down this information will not be made easy for developers?

On a related note, I did just come up with the idea that we could make a macro to generate a function at runtime that returns an array of the relevant arguments to pass in, making both sides happy. 

Is this something I should move forward on in addition to making these properties internal?

jglogan:
> @jglogan am I correct in understanding that the idea of passing down this information will not be made easy for developers?

Look, if that's what you want to extract from what I wrote above, fine. But we're not going to break the encapsulation on those CLI properties, not right now anyway.

I've explained to you before that you don't need to shove the whole world in one PR. Go back and review what I wrote re the compose plugin and you'll see that the message has been consistent on this. It's a general best practice here, and for most OSS projects.

I really appreciate the work you've done here, and I'm happy that you're making good progress with the compose plugin. I've asked over on the swift-argument-parser forum to see what options there might be for your CLI use case that don't involve exposing those properties.

Mcrich23:
@jglogan Thank you so much. I don't want to cause any tension at all. I'm not sure if you can tell, but I am new to contributing to big (but heavily new) OSS projects. I really appreciate all of your thoughts, patience, time, and discourse. 

I have reverted the commit making these properties public and look forward to this PR being merged. If you want to start or have me create an issue or discussion about the best way to handle passing groups down, I would be happy to chat more about the ideal implementation for this.

Thank you again and I look forward to working on this with you, @dcantah, and others on this.

jglogan:
It's all good. Both things can be true – exposing those fields breaks encapsulation, which makes APIs more brittle, while at the same time it makes your quite reasonable use case painful.

I'm wondering about directory and target names? What if the "ExecutableCLI" directory now just becomes "CLI", and we use "ContainerCommands" as the library, target, and project directory name for what is presently "CLI".

Now would be the time to rename things since nobody yet can import the command library.

Mcrich23:
That sounds good. I didn't want to put this forward with git possibly thinking I was deleting and adding everything. Especially because of merge conflicts. 

But I am happy to do this when I am at my computer next.

Mcrich23:
Another option is that we rename the library and target in the `Package.swift` file now and then rename the folder in a second pull request.

jglogan:
That's a good idea. We can rename in a follow-up. Let me take one more pass through the PR.

Mcrich23:
So do you want me to rename it in Package.swift for the sake of maintaining the same import name starting with this or should I wait for the other pr?

jglogan:
> So do you want me to rename it in Package.swift for the sake of maintaining the same import name starting with this or should I wait for the other pr?

The first seems good to me. Let me read that back to make sure I understand you:

This PR, rename the library and target name for the now-importable commands, but leave the source directory names as is.

Follow-on PR, rename the source directory names for consistency with the target names (should be renames only, I think).

Also, you'll need to `make fmt` (we really need to add an opt-in pre-commit hook). And likely a rebase since we got rid of the `defaults` subcommands in favor of `container system property`.

Mcrich23:
Alright. This looks like it should be ready!

jglogan:
Kicked off the workflow, let's see how the build looks.

jglogan:
Hmm, that was quick:

```
❌ The following files require formatting or license header updates:
Sources/ExecutableCLI/Executable.swift
```

Mcrich23:
ah yes. forgot about that. one sec

Mcrich23:
I believe the issue was a rogue comment near the top

Mcrich23:
Actually, before we merge this, should I update the readme with information about developing plugins?

jglogan:
Nah, for that, could you create a git issue for plugin development information? You and I can put together an outline there first, and then we'll do that in a dedicated PR.

It's long overdue. I have some thoughts scratched down in a local branch but have been getting sidetracked a lot.

Sources/CLI/Image/ImageList.swift
```diff
@@ -37,16 +38,17 @@ extension Application {
         var global: Flags.Global
     }
 
-    struct ListImageImplementation {
-        static private func createHeader() -> [[String]] {
+    public struct ListImageImplementation {
```
jglogan: Sorry, just noticed this one. Do you need to make this public today for your plugin?

That we're hardwiring ListImageImplementation into ImageList is curious; I hadn't really looked at what was being done here before. Perhaps we just keep this stuff as is and come up with a better way later to achieve reusability for what's here.

Sources/CLI/Network/NetworkList.swift
```diff
@@ -22,31 +22,32 @@ import Foundation
...
+        @OptionGroup
+        var global: Flags.Global
+
+        public func run() async throws {
             let networks = try await ClientNetwork.list()
             try printNetworks(networks: networks, format: format)
         }
 
-        private func createHeader() -> [[String]] {
+        func createHeader() -> [[String]] {
```
jglogan: Same here...unless you really need these today for your use case, I'm inclined to keep these private and focus on coming up with a more reusable solution for tabular output.

Sources/CLI/System/Kernel/KernelSet.swift
```diff
@@ -89,7 +90,7 @@ extension Application {
             try await Self.downloadAndInstallWithProgressBar(tarRemoteURL: remoteURL.absoluteString, kernelFilePath: binaryPath, platform: platform, force: force)
         }
 
-        private func getSystemPlatform() throws -> SystemPlatform {
+        func getSystemPlatform() throws -> SystemPlatform {
```
jglogan: Same here, if you don't need them now, let's expose them when and if we come to need them.

Sources/CLI/Volume/VolumeList.swift
```diff
@@ -33,16 +34,16 @@ extension Application.VolumeCommand {
...
         var format: Application.ListFormat = .table
 
-        func run() async throws {
+        public func run() async throws {
             let volumes = try await ClientVolume.list()
             try printVolumes(volumes: volumes, format: format)
         }
 
-        private func createHeader() -> [[String]] {
+        func createHeader() -> [[String]] {
```
jglogan: Same as above.

jglogan:
Builds now! I went through everything and left a few comments regarding some of the private methods that went to default visibility. Was that needed to compile the code without errors? For those (especially the table output stuff which we should just solve in a better way), I'm inclined to leave these private until someone actually needs to use these functions.

Mcrich23:
Sounds good. I will fix these soon

katiewasnothere:
@Mcrich23 Do you have an example of a plugin you're hoping to use these changes with? Draft code is fine, I'm just curious to see how it looks

Mcrich23:
@katiewasnothere yeah. This hasn't been updated entirely for the finalized interface, but it and the conversation here should give you a decent idea of the current API with ContainerCommands. https://github.com/Mcrich23/container/blob/add-compose/Plugins/Compose/ComposeCLI/Commands/ComposeUp.swift

Sources/ExecutableCLI/Executable.swift
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerClient
+import ContainerCommands
+
+@main
+public struct Executable: AsyncParsableCommand {
```
dcantah, line 22: I'm sufficiently dumb, but what is this target for?

Mcrich23: This is the new compose executable entry point since I could not otherwise get an executable and normal target for the same sources. So this is just a proxy making it executable.

jglogan:
Looks like we need `make fmt` once more...

jglogan (approved):

PR #604, opened 2025-09-12
Replace scattered defaults subcommands with `system property`.

### LINKED ISSUE #384: [Request]: Minimal feature set for common defaults management.

As a container user, I can use a common set of CLI commands to get names, descriptions and values for application defaults, as well as set and clear default values, so that I can adapt default behavior to my needs.

Currently defaults management is scattered about, using varying verbs.

Bulk load/merge/extraction of defaults is out of scope and can be handled in follow-up issues.

### Code of Conduct

jglogan:
Common subcommands for all defaults.

- Closes #384.
- Replaces `registry default` and `system dns default` subcommands with `system property`.
- Users can use `system property ls` to see details about each supported default value.
- `system property set` implements reasonable validation for all properties.
- NOTE: Probing of the registry for `registry default set` was removed, which means users will find out about a botched setting when pulling or pushing.
- Updates docs.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [x] Breaking change
- [x] Documentation update

## Motivation and Context
See #384.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs


Sources/CLI/System/Property/PropertySet.swift
```diff
@@ -0,0 +1,77 @@
...
+                    throw ContainerizationError(.invalidArgument, message: "invalid boolean value: \(value)")
+                }
+                DefaultsStore.setBool(value: boolValue, key: key)
+            case .defaultDNSDomain, .defaultRegistryDomain:
+                guard Parser.isValidDomainName(value) else {
+                    throw ContainerizationError(.invalidArgument, message: "invalid domain name: \(value)")
+                }
+                DefaultsStore.set(value: value, key: key)
+            case .defaultBuilderImage, .defaultInitImage:
+                guard (try? Reference(path: value)) != nil else {
```
adityaramani: I think this should be `(try? Reference.parse(value)) != nil`

jglogan: fixed

PR #619, opened 2025-09-17
Use a lock consistently in the `ExecutionContext` class

dkovba:
The `headSnapshot` property was accessed outside a lock in the `prepareSnapshot()` method, while we did more than needed inside a lock in the `childContext()` method.


Sources/NativeBuilder/ContainerBuildExecutor/ExecutionContext.swift
```diff
@@ -201,7 +203,7 @@ public final class ExecutionContext: @unchecked Sendable {
     /// - Returns: A prepared snapshot ready for modification
     /// - Throws: Any errors from the snapshotter
     public func prepareSnapshot(for operationId: UUID) async throws -> Snapshot {
-        let parentCommitted = headSnapshot
+        let parentCommitted = lock.withLock { _headSnapshot }
```
wlan0: headSnapshot is already protected by a lock, is this necessary?

wlan0 (changes requested):
one change seems redundant, but LGTM otherwise

Sources/NativeBuilder/ContainerBuildExecutor/ExecutionContext.swift
dkovba: The mixed use of `headSnapshot` and `_headSnapshot` might be confusing. But this is not incorrect.

wlan0 (approved):

PR #628, opened 2025-09-17
Swap to APIServer for all communications

dcantah:
Today, we have a sort of odd model all things considered. We talk to the APIServer to do the initial container creation, which mostly all of the work is just registering the runtime helper with launchd. After this registration, almost all communication from a client is talking directly to that runtime helper. This has a couple rather annoying issues in that we now need a sort of event/notification channel that the helper will establish with the APIServer for errors/start/exited cases.

If instead of talking to the runtime helper directly, we instead took a detour through the APIServer, the APIServer is clued into exactly what order of operations is occurring. This makes "did starting the container fail? Okay we should clean up" scenarios much simpler, and it also simplifies the clients quite a bit as they don't need this split brained client model, everyone just talks to the APIServer. This change is in pursuit of that. I have reworked our clients, the ContainerService and some of our XPC types to accomplish it.

The biggest "contract" change is in the SandboxService. Today we have on the flag when we register any runtime helper that makes any xpc messages wake up the registered process. This isn't great in scenarios where the process have may crashed, or it exited normally and we're just trying to invoke an RPC on it. Today the helper would spawn again and try and answer our request. It'd be much nicer if we have a connection object that will become invalid if the process that vended it to us is gone. To accomplish this, now the runtime helpers will listen on an anonymous xpc connection and vend endpoints from this via only one handler exposed by the SandboxService (createEndpoint). From that point onwards all communication will be through the endpoint the service vended a client.


Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -145,16 +137,67 @@ extension ClientContainer {
...
+                message: "failed to bootstrap container",
+                cause: error
+            )
+        }
+    }
+
+    public func kill(_ signal: Int32) async throws {
+        do {
+            let request = XPCMessage(route: .containerKill)
+            request.set(key: .id, value: id)
```
wlan0, line 175:
```suggestion
            request.set(key: .id, value: self.id)
```

Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -145,16 +137,67 @@ extension ClientContainer {
...
+            request.set(key: .processIdentifier, value: id)
```
wlan0, line 176: what is processIdentifier used for? In other places, I see it being set to `containerId`

Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -145,16 +137,67 @@ extension ClientContainer {
...
+        let client = Self.newClient()
+
+        for (i, h) in stdio.enumerated() {
+            let key: XPCKeys = {
+                switch i {
+                case 0: .stdin
+                case 1: .stdout
+                case 2: .stderr
+                default:
+                    fatalError("invalid fd \(i)")
```
wlan0, line 150: If a user accidentally sends a longer array, this would crash the whole program. Consider a typed error instead

Sources/ContainerClient/ContainerEvents.swift
```diff
@@ -15,6 +15,5 @@
 //===----------------------------------------------------------------------===//
 
 public enum ContainerEvent: Sendable, Codable {
-    case containerStart(id: String)
```
jglogan, line 18: just delete this file?

Sources/ContainerClient/Core/ClientContainer.swift
dcantah: Some rpcs are only applicable to the entire container (stop, delete etc.) and some can be targeted at specific execs as well

Sources/ContainerClient/Core/ClientContainer.swift
dcantah: Yea this was a copy paste from the type we already had, I'd love to fix this up now though. I've been trying to get rid of some of our fatalError's in CZ recently too

Sources/ContainerClient/ContainerEvents.swift
dcantah: Oh 100%, I forgot to axe this after removing the use of it

Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -145,16 +137,67 @@ extension ClientContainer {
...
+                case 2: .stderr
+                default:
+                    throw ContainerizationError(.invalidArgument, message: "invalid fd \(i)")
+                }
+            }()
+
+            if let h {
+                request.set(key: key, value: h)
+            }
+        }
```
dkovba, line 157: This code repeats 4 times. Consider extracting it into helper methods in `XPCMessage`:
```
 static func stdioKey(for index: Int) throws -> XPCKeys {
      switch index {
      case 0: return .stdin
      case 1: return .stdout
      case 2: return .stderr
      default:
          throw ContainerizationError(.invalidArgument, message: "invalid fd \(index)")
      }
  }

 static func setStdioHandles(on request: XPCMessage, stdio: 
  [FileHandle?]) throws {
      for (index, handle) in stdio.enumerated() {
          if let handle {
              request.set(key: stdioKey(for: index), value: handle)
          }
      }
  }
```

Sources/ContainerClient/SandboxClient.swift
```diff
@@ -34,29 +34,56 @@ public struct SandboxClient: Sendable, Codable {
...
 
     /// Create a container.
-    public init(id: String, runtime: String) {
+    init(id: String, runtime: String, client: XPCClient) {
         self.id = id
         self.runtime = runtime
+        self.client = client
+    }
+
+    public static func create(id: String, runtime: String) async throws -> SandboxClient {
```
jglogan, line 47: docc for this?

Sources/Services/ContainerNetworkService/NetworkClient.swift
```diff
@@ -40,7 +40,6 @@ extension NetworkClient {
     public func state() async throws -> NetworkState {
         let request = XPCMessage(route: NetworkRoutes.state.rawValue)
         let client = createClient()
-        defer { client.close() }
```
jglogan, line 43: Why not close on leaving scope?

dcantah: The XPCClient type now closes in the destructor for you

Sources/ContainerClient/Core/ClientProcess.swift
```diff
@@ -45,79 +45,66 @@ public protocol ClientProcess: Sendable {
 
 struct ClientProcessImpl: ClientProcess, Sendable {
     static let serviceIdentifier = "com.apple.container.apiserver"
+
+    /// ID of the process.
+    public var id: String {
```
jglogan, line 50: Do we want to claim Identifiable protocol conformance?

Sources/ContainerClient/Core/ClientProcess.swift
```diff
@@ -45,79 +45,66 @@ public protocol ClientProcess: Sendable {
...
     /// init process of the container.
     public let processId: String?
 
-    public var id: String {
-        processId ?? containerId
-    }
+    private let client: XPCClient
 
-    init(containerId: String, processId: String? = nil, client: SandboxClient) {
+    init(containerId: String, processId: String? = nil, client: XPCClient) {
```
jglogan: It'd be nice to have a more descriptive name than `client` throughout this type, and for `ClientContainer.newClient()`. This is specifically an XPCClient (a low level client compared to our ClientXXX "SDK" types) that's used to communicate with the API server and gotten from ClientContainer, right?

`clientContainerClient` is like fingernails^v^v^chalkboard though. And `ClientContainer.newClientContainerClient()` is like a fever dream.

What if this was either `apiClient` or `xpcClient` everywhere?

Sources/ContainerClient/Core/ClientProcess.swift
dcantah: I'm too ignorant to know what we'd gain out of that at the moment. It seems mostly used in UI programming to find which item has changed in a view.

jglogan: Yeah it was just a thought. YAGNI

Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -145,16 +137,67 @@ extension ClientContainer {
...
         do {
-            let client = self.sandboxClient
-            try await client.stop(options: opts)
+            let client = Self.newClient()
+            let request = XPCMessage(route: .containerStop)
+            let data = try JSONEncoder().encode(opts)
+            request.set(key: .id, value: self.id)
+            request.set(key: .stopOptions, value: data)
+
+            let responseTimeout = Duration(.seconds(Int64(opts.timeoutInSeconds + 3)))
```
jglogan, line 204: comment explaining why we wait 3 seconds longer than the specified timeout?

and if it's because of [this](https://github.com/apple/container/pull/628/files#diff-7c52675591a7382280e1a7f7be56af20d9bf04627fad5ad8ea93fe0f3517b41cR272) we should use an extracted constant I think

Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -180,46 +223,53 @@ extension ClientContainer {
...
-    /// Send or "kill" a signal to the initial process of the container.
-    /// Kill does not wait for the process to exit, it only delivers the signal.
-    public func kill(_ signal: Int32) async throws {
-        do {
-            let client = self.sandboxClient
-            try await client.kill(self.id, signal: Int64(signal))
+            let data = try JSONEncoder().encode(configuration)
+            request.set(key: .processConfig, value: data)
+
+            for (i, h) in stdio.enumerated() {
```
jglogan, line 247: extract private method?

Sources/ContainerClient/Core/ClientProcess.swift
dcantah: xpcClient is a good idea

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -163,111 +207,289 @@ public actor ContainersService {
...
-        let initImage = try await ClientImage.fetch(reference: ClientImage.initImageRef, platform: platform)
-        var fs = try await initImage.getCreateSnapshot(platform: platform)
-        fs.options = ["ro"]
-        return fs
+    /// Bootstrap the init process of the container.
+    public func bootstrap(id: String, stdio: [FileHandle?]) async throws {
+        self.log.debug("\(#function)")
+        do {
+            try await self.lock.withLock { context in
+                var item = try await self.get(id: id, context: context)
```
jglogan: `s/item/state/g` will make this code a bit more understandable

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -277,104 +499,87 @@ public actor ContainersService {
...
-            let snapshot = ContainerSnapshot(configuration: currentSnapshot.configuration, status: .running, networks: sandboxSnapshot.networks)
-            await self.setContainer(id, snapshot, context: context)
-        } catch {
-            self.log.error(
-                "Failed to handle container start",
-                metadata: [
-                    "id": .string(id),
-                    "error": .string(String(describing: error)),
-                ])
+    private func get(id: String, context: AsyncLock.Context) throws -> ContainerState {
```
jglogan: Suggest `getContainerState`, `_getContainerState`,  and `setContainerState` instead of `get` and `setContainer`. Also move `setContainer` down here with the other private funcs.

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -27,28 +28,42 @@ import Foundation
 import Logging
 
 public actor ContainersService {
+    struct ContainerState {
+        var snapshot: ContainerSnapshot
+        var client: SandboxClient?
+
+        func getClient() throws -> SandboxClient {
+            guard let client else {
+                throw ContainerizationError(.invalidState, message: "no sandbox client exists")
```
jglogan: perhaps include state from snapshot to understand a little more about why things are broken?

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -163,111 +207,289 @@ public actor ContainersService {
...
+        config: ProcessConfiguration,
+        stdio: [FileHandle?]
+    ) async throws {
+        self.log.debug("\(#function)")
+
+        let item = try self._get(id: id)
+        do {
+            let client = try item.getClient()
+            try await client.createProcess(
+                processID,
```
jglogan, line 254: Are two processes allowed to use the same ID? If not, who ensures this?

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -163,111 +207,289 @@ public actor ContainersService {
...
+    public func startProcess(id: String, processID: String) async throws {
+        self.log.debug("\(#function)")
+
+        do {
+            try await self.lock.withLock { context in
+                var item = try await self.get(id: id, context: context)
+                let client = try item.getClient()
+                try await client.startProcess(processID)
+
+                if id == processID {
```
jglogan: We only use it in two places it appears, but the code might read better if we had an `isInitProcess(containerId, processId)`

jglogan (approved):

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -1114,6 +1161,7 @@ extension SandboxService {
         case running
         case stopping
         case stopped(Int32)
+        case shuttingDown
```
jglogan, line 1188: docc this enum and values, mainly so that folks can see the difference between `stopping` and `shuttingDown`.

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -365,7 +411,7 @@ public actor SandboxService {
         log.info("`createProcess` xpc handler")
         return try await self.lock.withLock { [self] _ in
             switch await self.state {
-            case .created, .stopped(_), .starting, .stopping:
+            case .created, .stopped(_), .starting, .stopping, .shuttingDown:
```
jglogan: use `default:` for this?

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -457,33 +503,36 @@ public actor SandboxService {
...
         let reply = try await self.lock.withLock { [self] _ in
             switch await self.state {
-            case .stopped(_), .created, .stopping:
+            case .stopped(_), .created, .stopping, .shuttingDown:
                 return message.reply()
             case .starting:
                 throw ContainerizationError(
                     .invalidState,
-                    message: "cannot stop: container is not running"
+                    message: "cannot stop: container is starting"
```
jglogan: the only way one could hit this would be to do something like?

```
container run --name {id} ... & sleep {some-lucky-value} ; container stop {id}
```

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -866,25 +913,36 @@ public actor SandboxService {
         return container
     }
 
-    private func gracefulStopContainer(_ lc: LinuxContainer, stopOpts: ContainerStopOptions) async throws {
```
jglogan, line 869: Note about code not modified in this PR:

```swift
private nonisolated func configureProcessConfig()
```

This should be able to become:

```swift
private static func configureProcessConfig()
```

Same for `closeHandle()`.

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -225,14 +248,38 @@ public actor SandboxService {
...
+            } catch {
+                self.log.error("failed to sleep before shutting down SandboxService: \(error)")
+            }
+            self.log.info("Shutting down SandboxService")
+            exit(0)
+        }
+
+        return message.reply()
+    }
+
```
jglogan: This is a nit and might be a matter of preference but perhaps we should move the private methods that follow down below `public dial()`.

Sources/Services/ContainerSandboxService/SandboxService.swift
jglogan: Another lowkey obsessive nit in the non-modified code above - move `getDefaultNameserver` out from between the `configure...()` funcs. All the private methods could probably stand to be reordered sensibly. In a swift file that's pushing 1200 lines it will help new developers.

Mcrich23:

Sources/Services/ContainerAPIService/Containers/ContainersService.swift
dcantah: It'll fail at the cz level iirc

dcantah: I can add a little check in SandboxService.

Sources/ContainerClient/Core/ClientContainer.swift
dcantah: I'll leave for a followup. Anything that was mostly already here that can be cleaned up I think would be easier to review standalone

Sources/Services/ContainerSandboxService/SandboxService.swift
dcantah: Yea, basically another term doing stop at the same time as run and getting lucky

Sources/Services/ContainerSandboxService/SandboxService.swift
dcantah: Agree

dcantah:
Converting to draft because there's a small cz change that I want in that will allow us to "fix" some behavior in stop()

jglogan (approved):

wlan0 (approved):

Sources/ContainerClient/Core/ClientContainer.swift
dcantah: John had commented the same, I'm going to do that (and some other cleanups) in a followup

adityaramani (commented):
Changes look good!

PR #637, opened 2025-09-18
misc: update default dns command in tutorial

rnjohn:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
The tutorial.md file is very useful in getting started with `container`, but there is a command that might be outdated because it doesn't currently work (at least to me). The command present in the tutorial for setting the default DNS is this `container system property set dns.domain test`, but according to the --help flag of `container`, the current command to set the default DNS is as follows: `container system dns default set test`

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -98,7 +98,7 @@ Use the `--help` flag to see which abbreviations exist.
 
 ```bash
 sudo container system dns create test
-container system property set dns.domain test
+container system dns default set test
 ```
 
 Enter your administrator password when prompted. The first command requires administrator privileges to create a file containing the domain configuration under the `/etc/resolver` directory, and to tell the macOS DNS resolver to reload its configuration files.


jglogan:
That's a good catch. It's actually the other way around. The documentation in main documents the code in main, and the `container system property` command replaced `container system dns default` following the 0.4.1 release.

If you look at https://github.com/apple/container/blob/0.4.1/docs/tutorial.md, you'll see the expected command.

This situation should be improved. I filed #638 to address making the current release documentation more discoverable.

rnjohn:
@jglogan thank you for answering! English is not my first language so to verify, on version 0.4.1 it is supposed to be `container system property`? because that is the version of `container` that I installed and I had to use `container system dns default`

katiewasnothere:
@rnjohn On `container version 0.4.1` the correct command is `container system dns default`. Once we make a new package of `container` (which we do at the end of the month) and you update to that version, the command will then be `container system property set dns.domain`. Closing for now, please feel free to reopen the issue if we can clarify further. Thanks!

Issue #515, opened 2025-08-16
[Request]: Clean up help for complex commands.


jglogan:
### Feature or enhancement request details

Help output for commands is a bit hectic. We should go through each command and make sure that help output is well organized and easy to scan and understand.

We also need a way to document complex option values (e.g. `--publish`) without cluttering things.

Worst-case example:

```console
% container run --help
Warning! Running debug build. Performance may be degraded.
OVERVIEW: Run a container

USAGE: container run [<options>] <image> [<arguments> ...]

ARGUMENTS:
  <image>                 Image name
  <arguments>             Container init process arguments

OPTIONS:
  -w, --cwd, --workdir <cwd>
                          Current working directory for the container
  -e, --env <env>         Set environment variables
  --env-file <env-file>   Read in a file of environment variables
  --uid <uid>             Set the uid for the process
  --gid <gid>             Set the gid for the process
  -i, --interactive       Keep Stdin open even if not attached
  -t, --tty               Open a tty with the process
  -u, --user <user>       Set the user for the process
  -c, --cpus <cpus>       Number of CPUs to allocate to the container
  -m, --memory <memory>   Amount of memory in bytes, kilobytes (K), megabytes (M), or gigabytes (G) for the container, with MB granularity (for example, 1024K will result in 1MB being allocated
                          for the container)
  -d, --detach            Run the container and detach from the process
  --entrypoint <entrypoint>
                          Override the entrypoint of the image
  --mount <mount>         Add a mount to the container (type=<>,source=<>,target=<>,readonly)
  -p, --publish <publish> Publish a port from container to host (format: [host-ip:]host-port:container-port[/protocol])
  --publish-socket <publish-socket>
                          Publish a socket from container to host (format: host_path:container_path)
  --tmpfs <tmpfs>         Add a tmpfs mount to the container at the given path
  --name <name>           Assign a name to the container. If excluded will be a generated UUID
  --remove, --rm          Remove the container after it stops
  --os <os>               Set OS if image can target multiple operating systems (default: linux)
  -a, --arch <arch>       Set arch if image can target multiple architectures (default: arm64)
  -v, --volume <volume>   Bind mount a volume into the container
  -k, --kernel <kernel>   Set a custom kernel path
  --network <network>     Attach the container to a network
  --cidfile <cidfile>     Write the container ID to the path provided
  --no-dns                Do not configure DNS in the container
  --dns <dns>             DNS nameserver IP address
  --dns-domain <dns-domain>
                          Default DNS domain
[...]

PR #662, opened 2025-09-21
Cleans up system subcommand options.

jglogan:
- Part of #515.
- Order options alphabetically.
- Make descriptions more concise and consistent.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
See #515.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/System/DNS/DNSCreate.swift
+++ b/Sources/ContainerCommands/System/DNS/DNSCreate.swift
@@ -22,14 +22,18 @@ import Foundation
 
 extension Application {
     public struct DNSCreate: AsyncParsableCommand {
-        public init() {}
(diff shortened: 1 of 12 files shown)


Sources/ContainerCommands/System/DNS/DNSCreate.swift
```diff
@@ -22,15 +22,19 @@ import Foundation
 
 extension Application {
     public struct DNSCreate: AsyncParsableCommand {
-        public init() {}
```
katiewasnothere, line 25: Why are we removing these?

Issue #642, opened 2025-09-18
[Request]: UX: Use stdout and stderr, and logging consistently in CLI commands.


jglogan:
### Feature or enhancement request details

Right now some output that should probably go to stderr is just being printed with print().

An example is: https://github.com/apple/container/blob/main/Sources/CLI/System/SystemStart.swift#L95.

All commands should log diagnostics to stderr and print data proper to stdout.

For this we could create a plaintext-to-stderr LogHandler in ContainerClient that all commands use.

At present it appears we do no logging in ContainerClient, with the single Exception of Sources/ContainerClient/ExitMonitor.swift, which is actually only used on the server side. Therefore, we can limit the scope of this to just going through each of the command groups.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vyncint:
Picking up one piece of this.

`container machine run` reports its forced exit after three SIGINT/SIGTERMs with `print()`, so the diagnostic goes to stdout, while the identical `SignalThreshold` handler in `container run` and `container exec` uses `log.warning()` (stderr via `StderrLogHandler`). That handler is only installed when stdout is not a TTY, so the message lands in redirected/piped output.

PR incoming for just that one. Leaving this issue open, since it covers the wider audit.

Issue #507, opened 2025-08-16
[Request]: Consistent, useful behavior for `--debug`.


jglogan:
### Feature or enhancement request details

Tasks:

- [ ] `--debug` is available every that it makes sense.
- [ ] The global flags to which `--debug` belongs appear in consistent order relative to other options across all subcommands.
- [ ] Where `--debug` does appear, it provides clear and actionable output. If the debug output does appears elsewhere than the CLI output (e.g. OS logs), the CLI informs the user where the debug output can be found.
- [ ] If debug is not present and the `CONTAINER_DEBUG` environment variable is truthy, debug is enabled.
- [ ] `CONTAINER_DEBUG` included in help text for debug everywhere (or consider removing it everywhere if we switch to using env variables instead of user defaults for settings). At present all commands seem to display it except `system start`.

Right now these leaf commands do not have `--debug`:

- build
- registry login
- default unset
- default inspect
- volume [*]
- builder [*]
- system dns [*]
- system stop
- system status
- system kernel set
- help

Perhaps we can look at #360 as well while attending to this one.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

saehejkang:
Is this a good issue to get started on and learn more about the codebase?

karenheckel:
@jglogan Hello! We are a group from UT Austin looking for issues to work on for our Virtualization class. We would love to work on this issue if it still needs to be closed.

Pranjal-Jain-01:
Hi! I’d like to work on the remaining parts of this issue (debug availability on missing subcommands, CONTAINER_DEBUG behavior, and help text consistency). Happy to take this if it’s still unassigned.

PR #680, opened 2025-09-24
Adds a plain stderr log handler.

jglogan:
- Initial work for #642. CLI commands will configure this log handler so that they write warnings and errors to stderr, and only proper output will go to stdout.
- Facilitates #507. The `--debug` option can set the logging level appropriately, and CLI commands can just use `log.debug()`.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/Sources/ContainerLog/StderrLogHandler.swift
+++ b/Sources/ContainerLog/StderrLogHandler.swift
@@ -1,0 +1,56 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the container project authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+//
+
+import Foundation
+import Logging
+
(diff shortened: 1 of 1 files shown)


Sources/ContainerLog/StderrLogHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+    public init() {}
+
+    public func log(
+        level: Logger.Level,
+        message: Logger.Message,
+        metadata: Logger.Metadata?,
+        source: String,
+        file: String,
+        function: String,
+        line: UInt
```
katiewasnothere, line 46: Why are some of these unused?

Sources/ContainerLog/StderrLogHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+
+//
+
+import Foundation
+import Logging
+
+/// Basic log handler for where simple message output is needed,
+/// such as CLI commands.
+public struct StderrLogHandler: LogHandler {
+    public var logLevel: Logger.Level = .info
```
katiewasnothere, line 25: should the default here be ".error"?

Sources/ContainerLog/StderrLogHandler.swift
jglogan: Because this logger is used to write plain output to the console on stderr.

Later if we want to get fancy we can add more configurability over the log message structure and then plumb that to CLI options but for now I want to be able to do `log.error("foo")` from somewhere in the CLI code and see:

```console
foo
```

in `errs` when I run `container somecommand 2> errs` and it hits that log statement.

Sources/ContainerLog/StderrLogHandler.swift
jglogan: No, we want to be able to write stuff at different levels to stderr, not just `log.error()`.

When we plumb out `--debug`, the logger will be configured with logLevel `.debug` when the flag is enabled.

PR #681, opened 2025-09-24
Relocate ProcessIO to ContainerClient.

jglogan:
- Makes ProcessIO more generally available.
- `Application.handleProcess(processIO:process:)` becomes `processIO.handleProcess(process:log:)`.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This code shouldn't be in the CLI, it should be part of ContainerClient.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerCommands/Application.swift
```diff
@@ -171,78 +171,6 @@ public struct Application: AsyncParsableCommand {
         )
     }
 
-    public static func handleProcess(io: ProcessIO, process: ClientProcess) async throws -> Int32 {
```
jglogan, line 174: this becomes an instance function in ProcessIO.swift()

Sources/ContainerClient/ProcessIO.swift
```diff
@@ -0,0 +1,381 @@
...
+            stdin: stdin,
+            stdout: stdout,
+            stderr: stderr,
+            ioTracker: ioTracker,
+            stdio: stdio,
+            console: current
+        )
+    }
+
+    public func handleProcess(process: ClientProcess, log: Logger) async throws -> Int32 {
```
jglogan, line 154: relocated from Application.swift:
- now an instance function so the `io` parameter just becomes `self`
- was using the `log` member from Application before, the caller now passes `log` in

Sources/ContainerClient/ProcessIO.swift
```diff
@@ -0,0 +1,381 @@
+//===----------------------------------------------------------------------===//
```
jglogan, line 1: extracted from ContainerRun in the ContainerCommands target to this file in the ContainerClient target

Sources/ContainerCommands/Application.swift
```diff
@@ -314,14 +242,6 @@ extension Application {
         case table
     }
 
-    static let signalSet: [Int32] = [
```
jglogan, line 317: These were only used by handleProcess so they've also moved to ProcessIO

Sources/ContainerClient/ProcessIO.swift
```diff
@@ -0,0 +1,381 @@
...
+import Foundation
+import Logging
+
+public struct ProcessIO: Sendable {
+    let stdin: Pipe?
+    let stdout: Pipe?
+    let stderr: Pipe?
+    var ioTracker: IoTracker?
+
+    static let signalSet: [Int32] = [
```
jglogan, line 28: moved to here from Application.swift

Sources/ContainerCommands/Container/ContainerRun.swift
```diff
@@ -165,285 +165,3 @@ extension Application {
         }
     }
 }
-
```
jglogan, line 168: All of this is now down in ProcessIO.swift in ContainerClient.

Sources/ContainerClient/ProcessIO.swift
dcantah, line 154: I like the ProcessIO move, but I'm not as big of a fan about this method on here. To me anything that would call start() or handle process lifecycle makes more sense on the ClientProcess type itself. A `run()` method on `ClientProcess` might be a decent naming, as that's essentially what this is. We start + wait + consume IO until completion.

dcantah: Thinking more about this, not entirely related but, to even be able to get a ClientProcess object we need to pass the io type into bootstrap() or createProcess(), why don't we just have ClientProcess manage the IO type directly? e.g. it can call closeAfterStart() inside the type and not burden the caller with doing it. Additionally, in process.wait() we can do the `try await io.wait()` just as we do in `handleProcess` right now. Seems like it'd clean things up a bit.

PR #685, opened 2025-09-25
make: add coverage target

egernst:
To facilitate ability to gather code coverage metrics without slowing down our integration target, let's add a coverage target.

From this we can run `xcrun llvm-profdata merge` to get profdata and gather overall code coverage metrics

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,21 @@ install-kernel:
 	@bin/container system stop || true
 	@bin/container system start --enable-kernel-install $(SYSTEM_START_OPTS)
 
+.PHONY: coverage
+coverage: init-block
+	@echo Ensuring apiserver stopped before the CLI integration tests...
+	@bin/container system stop && sleep 3 && scripts/ensure-container-stopped.sh
+	@echo Running the integration tests...
+	@bin/container system start $(SYSTEM_START_OPTS) && \
+	echo "Starting CLI integration tests" && \
+	{ \
+		exit_code=0; \
+		$(SWIFT) test --no-parallel --enable-code-coverage -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) || exit_code=1 ; \
+		echo Ensuring apiserver stopped after the CLI integration tests ; \
+		scripts/ensure-container-stopped.sh ; \
+		exit $${exit_code} ; \
+	}
+
 .PHONY: integration
 integration: init-block
 	@echo Ensuring apiserver stopped before the CLI integration tests...


katiewasnothere:
How much slower does adding `--enable-code-coverage` make running the tests?

egernst:
Not that much. The goal is to run make integration (CLI tests) and the other tests together in a single run, serially (which wouldn’t be ideal otherwise for CI testing)

Makefile
```diff
@@ -136,6 +136,21 @@ install-kernel:
 	@bin/container system stop || true
 	@bin/container system start --enable-kernel-install $(SYSTEM_START_OPTS)
 
+.PHONY: coverage
+coverage: init-block
+	@echo Ensuring apiserver stopped before the CLI integration tests...
+	@bin/container system stop && sleep 3 && scripts/ensure-container-stopped.sh
+	@echo Running the integration tests...
+	@bin/container system start $(SYSTEM_START_OPTS) && \
+	echo "Starting CLI integration tests" && \
```
dkovba: Is this print out accurate?

egernst:
@katiewasnothere @dcantah  - updated to add XCRUN so we get the css output for code coverage

Makefile
```diff
@@ -136,6 +138,30 @@ install-kernel:
...
+		scripts/ensure-container-stopped.sh ; \
+		echo Generating code coverage report... ; \
+		xcrun llvm-profdata merge -sparse $(COV_DATA_DIR)/*.profraw -o $(COV_DATA_DIR)/default.profdata ; \
+		xcrun llvm-cov show --compilation-dir=`pwd` \
+			-instr-profile=$(COV_DATA_DIR)/default.profdata \
+			--ignore-filename-regex=".build/" \
+			--ignore-filename-regex=".pb.swift" \
+			--ignore-filename-regex=".proto" \
+			--ignore-filename-regex=".grpc.swift" \
+			$(BUILD_BIN_DIR)/ContainerPackageTests.xctest/Contents/MacOS/ContainerPackageTests > $(COV_REPORT_FILE) ; \
```
katiewasnothere: Why are we referencing xctest here? This doesn't work for me locally

egernst: Mostly b/c this was the path for build output after running tests (make test) on my system. Maybe we can chat through this one in realtime, or others can TAL on their platform for what's unique here?

katiewasnothere: Found the issue, I have my dev env setup on a case sensitive volume. The xctest bundle that shows up for me is at `$(BUILD_BIN_DIR)/containerPackageTests.xctest/Contents/MacOS/containerPackageTests` so the capitalized "ContainerPackageTests" is what's causing this to fail locally for me. Can we update the PR with the lowercase `containerPackageTests`?

egernst: Ah, this makes sense - will fix.

katiewasnothere (approved):

PR #725, opened 2025-10-06
Fix image save to default to current platform when no platform specified

### LINKED ISSUE #712: [Request]: `Image save` command: platform default to current

When saving an image pulled from remote sources, if the platform is nil, the save will failed with the notFound error:

For example, running 
```
container image save alpine:latest --output /Users/itsuki/Desktop/untitled\ folder/image.tar
```
will give 
```
Error: notFound: "Content with digest sha256:eafc1edb577d2e9b458664a15f23ea1c370214193226069eb22921169fc7e43f"
```

The digest here is not the digest for the image we are trying to save specifically, but the same for any (I think) remote images.

However, specifying the platform to be the current platform, in my case, --platform linux/arm64/v8, the save will success.

### Code of Conduct

bismansahni:
## Summary

Fixes issue #712 where `container image save` command fails with a notFound error when no platform options are specified.

The fix adds a default to use the current platform when `--platform`, `--arch`, and `--os` flags are all unspecified, preventing the platform variable from being nil.

## Changes

- Modified `ImageSave.swift` to default to `Platform.current` when no platform options are provided

## Test plan

- Build the project successfully
- Verify that `container image save` now works without platform flags
- Ensure existing behavior with explicit platform flags remains unchanged

Fixes #712

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "26b26b11a8c840723acebdf378a73b77950dc4c3dfeda984e6337a0ab4ef3f2b",
+  "originHash" : "34442675b179c8ff0a1e41f75f518645e1a7e589de554666d976493153067284",
   "pins" : [
     {
       "identity" : "async-http-client",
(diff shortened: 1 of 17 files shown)


bismansahni:
Hi @katiewasnothere, can you have a look at this pr :)

katiewasnothere:
Hi @bismansahni, sorry for the delay and thank you for the contribution!  

I know we have an open issue [here](https://github.com/apple/container/issues/83) to push/pull the current platform by default. I see there's a PR for that [here](https://github.com/apple/container/pull/455#issuecomment-3161755002), which may be out of date and need to be updated. Based on the [comment in that PR](https://github.com/apple/container/pull/455#issuecomment-3161755002), I'm wondering if it would be better for us to do that change and this change at the same time so we can keep the behavior consistent between image `pull`, `push`, `save`, `load`, etc. 

What do you think? I'm open to other thoughts on the approach here

Issue #559, opened 2025-08-30
[Request]: `image save` and `image load` should support stdin.


jglogan:
### Feature or enhancement request details

At present the save and load commands require a tar file pathname to be supplied. This should be optional, and they should write stdout and read stdin, respectively, when the path is nil.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

saehejkang:
Commenting here that I am working on this issue 🙂.

Issue #268, opened 2025-06-27
[Bug]: Consistent Error Propagation


dkovba:
Created from https://github.com/apple/container/issues/167.

### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Not applicable.

### Current behavior

Inconsistent error propagation.

### Expected Behavior

**Consistent Error Propagation**

- Errors should be properly propagated through the XPC hierarchy
- Failure states should be clearly communicated to clients
- Timeouts should be implemented for all operations

**Graceful Recovery**

- The system should recover from communication failures
- Resources should be properly released even when operations fail
- Failed operations should leave the system in a consistent state

### Environment

```markdown
Not applicable.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

dkovba:
An example where this is reproducible:
1. Trigger an error in vminitd, for example, by adding `throw ContainerizationError(.internalError, message: "Important error message")` to `TerminalIO.start()`.
2. Run a container using `bin/container run -it --rm alpine:latest date`.

Results:
```
Error: interrupted: "internalError: "failed to start container" (cause: "interrupted: "XPC connection error: Connection interrupted"")
Ensure container system service has been started with `container system start`."
```

PR #734, opened 2025-10-07
[image-save-load]: support stdin for filepath

saehejkang:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

Example Commands with Outputs

```
container image save
Pathname for the saved image: test-path/test.yaml
⠦ Saving image(s) [3s]
```

```bash
container image load
Pathname to the image tar archive:  test-path/test.yaml
Files does not exist
```


Sources/ContainerCommands/Image/ImageCommand.swift
```diff
@@ -38,3 +38,8 @@ extension Application {
         )
     }
 }
+
+/// `ImageCommand` errors.
```
jglogan: For now, just use `ContainerizationError` instead of creating new error types. We do have a tech debt issue #268 for improving our error handling.

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -34,14 +34,23 @@ extension Application {
             transform: { str in
                 URL(fileURLWithPath: str, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false)
             })
-        var input: String
+        var input: String?
```
jglogan, line 37: This is fine but what you'll want to do is change L35 to something like this so `input = nil` is transformed properly:

str.map { URL(fileURLWithPath: $0, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false) }

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -34,14 +34,23 @@ extension Application {
...
 
         @OptionGroup
         var global: Flags.Global
 
         public func run() async throws {
-            guard FileManager.default.fileExists(atPath: input) else {
-                print("File does not exist \(input)")
+            var filePath = ""
```
jglogan: This isn't quite what #559 is asking for. If `input == nil` we don't want to prompt the user for a pathname. What `image load` needs to do in this case is:

- Create a temporary file, using `defer {}` to delete the file once we're done using it.
- Read binary data from the standard input, writing it to the temp file.
- Use the path to the temporary file as the path to load.

This allows us to do things like moving images between Docker and `container` without having to mess with intermediate files explicitly:

```bash
docker save foo:latest bar:latest | container image load
container save baz:latest qux:latest | docker image load
```

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -46,7 +46,7 @@ extension Application {
             transform: { str in
                 URL(fileURLWithPath: str, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false)
             })
-        var output: String
+        var output: String?
```
jglogan, line 49: Same as above, map `String?` to get the absolute path.

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -59,6 +59,15 @@ extension Application {
         @Argument var references: [String]
 
         public func run() async throws {
+            var filePath = ""
+            if output == nil {
```
jglogan: Here if output is nil we create a temporary file, defer the delete, do ClientImage.save to that file, and then copy the binary data in the file to the standard output

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: Ahhh, I was unsure why the issue was talking about `stdout`. 

How is the binary data coming from the standard input? Are we prompting anything from the user (would it be the path to the temporary file or the path from the file to read the binary data from)?

jglogan: @saehejkang no prompting whatsoever.

The idea behind the save and load commands is that without any filename, you can use them the same way you'd use many other standard unix commands. Take `tar` for example. If I run it with just the `-t` option:

```bash
% tar -t
I type in some stuff and then type CTRL-D!
^D
tar: Error opening archive: Unrecognized archive format
%
```

In this instance the command will just sit there, because it's reading its input from the standard input. I type some stuff in, hit return, then type ^D (EOF), and I get an error message because what I inputted wasn't a tar archive.

To do something more useful, I could run the command to create an archive file containing `foo`, writing the standard output and redirecting it to `foo.tar`:

```bash
% touch foo
% tar -c -v -f foo.tar foo
a foo
/Users/john/Documents/projects/vessel/vessel main % tar -c foo > foo.tar 
% ls foo*
foo	foo.tar
```

And then I can print the table of contents of the archive by running with the `-t` option and hooking up the standard input to the `foo.tar` file:

```bash
/Users/john/Documents/projects/vessel/vessel main % tar -t < foo.tar    
foo
```

On the load side you will stream data from `stdin` into a temporary file, and then give the temporary file to the API.

On the save side, you will use the API to save into a temporary file, and then stream the data to `stdout`

jglogan: @saehejkang Continuing...

Try a LLM prompt like:

> How do I write something like `cat` in Swift, except it doesn't need to deal with filename arguments, and I want to write stdin to a temporary file, and then read the file out to stdout. It should stream rather than slurp data.

LOL you might want to also ask it to use Swift Foundation APIs instead of low level calls 😄

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: Not really sure why we need to do this 🤔 . 

Also getting an error with that change `Cannot convert value of type 'String.Element' (aka 'Character') to expected argument type 'String'`

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: Funny enough, I only remember how to do this using `C`. I have never worked with `Swift` before, but it was 10000x easier lol.

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -90,7 +90,30 @@ extension Application {
                 throw ContainerizationError(.invalidArgument, message: "failed to save image(s)")
 
             }
-            try await ClientImage.save(references: references, out: output, platform: p)
+
+            let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("temp-file.tar")
```
saehejkang: I was thinking we create the `tempFile` no matter the `--output` flag is enabled, so we don't have to do another check after the `ClientImage.save` call. It is going to be deleted anyway with the `defer`, no matter it is read from or not.

Sources/ContainerCommands/Image/ImageLoad.swift
jglogan: > I have never worked with Swift before, but it was 10000x easier lol.

Agree. In C it is very straightforward.

Sources/ContainerCommands/Image/ImageLoad.swift
jglogan: @saehejkang I'm not exactly sure about why we chose to do that transformation. I'd need to go way back into the archives to git blame it but at this point I wouldn't want to change it on a whim.

Regarding the recommendation about this:

```swift
str.map { URL(fileURLWithPath: $0, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false) }
```

If the compiler isn't complaining about what's there, that suggests that `transform` knows how to deal with `String?` being nil, no what's there should be fine.

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -34,14 +34,35 @@ extension Application {
...
+                }
+
+                let bufferSize = 4096
+                while true {
+                    let chunk = FileHandle.standardInput.readData(ofLength: bufferSize)
+                    if chunk.isEmpty { break }
+                    outputHandle.write(chunk)
+                }
+            }
+            defer {
```
jglogan: Since you're using `try?` you could move this right below L43 to ensure that the file gets removed even if L49 throws.

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -34,14 +34,35 @@ extension Application {
...
-        var input: String
+        var input: String?
 
         @OptionGroup
         var global: Flags.Global
 
         public func run() async throws {
-            guard FileManager.default.fileExists(atPath: input) else {
-                print("File does not exist \(input)")
+            let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("temp-file.tar")
```
jglogan: This isn't safe if executed concurrently.

How about:

```swift
let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("\(UUID().uuidString).tar")
```

Sources/ContainerCommands/Image/ImageSave.swift
jglogan: Yep but same thing as above, we should use a random filename here.

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -90,7 +90,30 @@ extension Application {
...
+                    throw ContainerizationError(.internalError, message: "unable to open temporary file for reading")
+                }
+                
+                let bufferSize = 4096
+                while true {
+                    let chunk = outputHandle.readData(ofLength: bufferSize)
+                    if chunk.isEmpty { break }
+                    FileHandle.standardOutput.write(chunk)
+                }
+            }
```
jglogan, line 119: As for the `container build`-from-stdin PR you opened, could you add tests to save/load using stdin?

Could just be a tweaked version of `testImageSaveAndLoad` in TestCLIImages.swift.

jglogan (changes requested):

Sources/ContainerCommands/Image/ImageSave.swift
saehejkang: I will work on the tests after #827 is merged

jglogan:
@saehejkang Just merged the underlying test change, could you resolve conflicts on this? I'll have a look after and then try building it and doing a little local testing. Thanks!

saehejkang:
@jglogan #827, that also relied on the underlying test change, is ready. I squashed and cleaned up the commits for readability in both PRs.

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -57,7 +81,7 @@ extension Application {
             progress.start()
 
             progress.set(description: "Loading tar archive")
-            let loaded = try await ClientImage.load(from: input)
+            let loaded = try await ClientImage.load(from: input ?? tempFile.path())
```
jglogan, line 84: nit: What do you think about restructuring load like to only run the code that's needed for each case? The only bit of code that repeats is the ClientImage.load line, and we would be removing a conditional from that.

```swift
var loaded: ClientImage
if input == nil {
    // create tmpfile, defer delete, copy stdin to file
    loaded = try await ClientImage.load(from: tempFile.path())
} else {
    // ensure input exists
    loaded = try await ClientImage.load(from: input)
}
```

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -90,7 +90,32 @@ extension Application {
...
+                try? FileManager.default.removeItem(at: tempFile)
+            }
+
+            guard FileManager.default.createFile(atPath: tempFile.path(), contents: nil) else {
+                throw ContainerizationError(.internalError, message: "unable to create temporary file")
+            }
+
+            try await ClientImage.save(references: references, out: output ?? tempFile.path(), platform: p)
+
+            // Write to stdout
```
jglogan: Same for this one, we could put an `else` on this that just saves the image to output, and move the above code into the `output == nil` case.

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: Any reason for this update besides readability (coding standards, etc)? I thought using the ternary operator makes sense in that it is clear to use the input path first, if not nil, and the temp file path otherwise. Just thought calling load once makes more sense and is cleaner. Not 100% adamant on one way but will defer to you for the final decision.

jglogan: Sure! I like to look at it from the point of view of someone who's not you (who wrote it) or I (who have spent some time reviewing it) who's doesn't know the code up and down, and is coming in to try to fix a bug or add a feature.

For me, the notion of

```
if some_condition_we_test_exactly_once {
   do exactly what's needed for this case, no more, no less
   one_non_dry_statement
} else {
   do exactly what's needed for this case, no more, no less
   one_non_dry_statement
}
```

feels most maintainable.

It's only a nit, and actually I think I missed the bit with the progress bar in there - we wouldn't want to repeat that and extracting a method for that code doesn't feel like it'd help much with readability or mainability.

So yeah, the ternary operator statement seems like a good call. What about moving the temp file stuff into one leg of the test and the existence check into the other leg so that we "do exactly what's needed for this case, no more, no less" in both? Am I missing a reason why we can't do that?

saehejkang: There is no reason why we can't do that :) Thank you for the explanation about why it is more maintainable.

Sources/ContainerCommands/Image/ImageSave.swift
saehejkang: I thought updating the save command to "do exactly what's needed for this case, no more, no less" made more sense. Made the appropriate changes here.

jglogan: Thanks!

Yeah to make load work like what you did for save we'd need to extract the progress stuff into a private function and then we could move the tempfile creation inside the `input == nil` leg and the only duplicated lines would be the progress call and the load call. I'm happy leaving that for another day (possibly never), let's get this merged.

What I think is maintainable might not be what another experienced developer thinks, there's definitely some subjectivity and "matter of taste" to it!

PR #739, opened 2025-10-08
[container-run-create]: add support for --network none

### LINKED ISSUE #386: [Request]: Add option to create/run a container that has no network attachment.

The current default for `create` and `run` recognizes that the most common use case is to attach a container to the default network.

We could define a special network name `none` but possibly better would be to have a `--no-network` flag. In either case the client should validate the network options for a container and reject inconsistent inputs.

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #386

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -1,5 +1,8 @@
 name: container project - common jobs
 
+permissions:
+  contents: read
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/docs-release.yml
+++ b/.github/workflows/docs-release.yml
@@ -1,6 +1,10 @@
 # Manual workflow for releasing docs ad-hoc. Workflow can only be run for main or release branches.
 # Workflow does NOT publish a release of container.
 name: Deploy application website
+
+permissions:
+  contents: read
+
 on:
   workflow_dispatch:
 
--- a/.github/workflows/merge-build.yml
+++ b/.github/workflows/merge-build.yml
@@ -1,5 +1,8 @@
 name: container project - merge build
 
+permissions:
+  contents: read
+
 on:
   push:
     branches:
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,5 +1,8 @@
 name: container project - PR build
 
+permissions:
+  contents: read
+
 on:
   pull_request:
     types: [opened, reopened, synchronize]
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,5 +1,8 @@
 name: container project - release build
 
+permissions:
+  contents: read
+
 on:
   push:
     tags:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,5 +1,8 @@
 name: container project - release build
 
+permissions:
+  contents: read
+
 on:
   push:
     tags:


Sources/ContainerClient/Utility.swift
```diff
@@ -179,12 +179,16 @@ public struct Utility {
...
         config.virtualization = management.virtualization
-
-        config.networks = try getAttachmentConfigurations(containerId: config.id, networkIds: management.networks)
-        for attachmentConfiguration in config.networks {
-            let network: NetworkState = try await ClientNetwork.get(id: attachmentConfiguration.network)
-            guard case .running(_, _) = network else {
-                throw ContainerizationError(.invalidState, message: "network \(attachmentConfiguration.network) is not running")
+        
+        if management.noNetwork {
+            config.networks = [AttachmentConfiguration(network: ClientNetwork.defaultNoNetworkName, options: AttachmentOptions(hostname: config.id))]
```
saehejkang: Not sure if it would be appropriate to set the `config.networks` to an empty array or add an attachment like so above (with the `none` network name)

```suggestion
            config.networks = []
```

jglogan:
Hi @saehejkang!

I think that the preference on the issue was to define `none` as a reserved "network name" as you have done with your constant definition, and instead of a new `--no-network` flag, `--network none` would signify that the attachment list would be empty instead of the default.

Could you revise the PR to use `--network none` in place of `--no-network`? I'll also add review comments. Thanks again!

Sources/ContainerClient/Core/ClientNetwork.swift
```diff
@@ -24,6 +24,7 @@ public struct ClientNetwork {
     static let serviceIdentifier = "com.apple.container.apiserver"
 
     public static let defaultNetworkName = "default"
+    public static let defaultNoNetworkName = "none"
```
jglogan: This can just be `noNetworkName`.

Sources/ContainerClient/Flags.swift
```diff
@@ -153,6 +153,9 @@ public struct Flags {
 
         @Option(name: [.customLong("network")], help: "Attach the container to a network")
         public var networks: [String] = []
+
```
jglogan: remove, we'll use `--network none` instead

Sources/ContainerClient/Utility.swift
```diff
@@ -179,12 +179,16 @@ public struct Utility {
...
 
         config.virtualization = management.virtualization
-
-        config.networks = try getAttachmentConfigurations(containerId: config.id, networkIds: management.networks)
-        for attachmentConfiguration in config.networks {
-            let network: NetworkState = try await ClientNetwork.get(id: attachmentConfiguration.network)
-            guard case .running(_, _) = network else {
-                throw ContainerizationError(.invalidState, message: "network \(attachmentConfiguration.network) is not running")
+        
+        if management.noNetwork {
```
jglogan: Here, test to see if none is in management.networks. If it is:
- If management.networks.count != 1, throw a ContainerizationError like "no other networks may be defined along with network `none`"
- Otherwise, proceed, setting the config.networks to `[]`

If `none` isn't in management.networks, we just use the existing logic to use the default network if management.networks.isEmpty, or create attachments for all the specified networks otherwise.

One other place you'll need to make a change is here: https://github.com/apple/container/blob/main/Sources/Services/ContainerAPIService/Networks/NetworksService.swift#L118. We need to throw an error if someone tries to run `container network create none`.

Sources/ContainerClient/Utility.swift
saehejkang: `config.networks = []` is the correct here

Sources/ContainerClient/Utility.swift
saehejkang: Completely slipped my mind to add a check for the the `network create` command 👍🏽

Sources/ContainerClient/Utility.swift
```diff
@@ -180,11 +180,18 @@ public struct Utility {
 
         config.virtualization = management.virtualization
 
-        config.networks = try getAttachmentConfigurations(containerId: config.id, networkIds: management.networks)
-        for attachmentConfiguration in config.networks {
-            let network: NetworkState = try await ClientNetwork.get(id: attachmentConfiguration.network)
-            guard case .running(_, _) = network else {
-                throw ContainerizationError(.invalidState, message: "network \(attachmentConfiguration.network) is not running")
+        if management.networks.contains(ClientNetwork.noNetworkName) {
+            if management.networks.count != 1 {
```
jglogan: nit: use `guard...else`

jglogan (changes requested):
@saehejkang this looks great, sorry for the follow delay and sorry to trouble you with one tiny stylistic nit and then it should be good to go

I'll kick off the workflow on this one though, please check it just in case `make fmt` needs to be run with the little fix.

jglogan (approved):

jglogan:
@saehejkang Merged, thank you!

Issue #459, opened 2025-08-06
[Request]: Assign MAC address for container in `--network`


jglogan:
### Feature or enhancement request details

Related to #282, being able to specify the MAC address when attaching the container to a network would be useful for some esoteric cases.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
Closed by #753.

PR #751, opened 2025-10-11
Specify ip in --network

### LINKED ISSUE #282: [Request]:  A direct option to assign or configure the container's IP address

Currently, containers launched via the container CLI do not offer a way to define or control their IP address. This limits flexibility in several use cases, including:

- Local development scenarios that rely on predictable container IPs
- Inter-container communication, where fixed IPs are preferable
- Integration with tools or services expecting specific network behavior
- Testing network-related functionality that depends on a deterministic IP setup

While `--network` allows attaching to a custom network, it doesn't currently support IP address assignment or any advanced networking options.

Proposed Solution

Introduce an additional flag to `container run`, such as:

```
--ip <ip-address>
```

### Code of Conduct

siikamiika:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Adds Flag.NetworkArg, an `ExpressibleByArgument` struct replacing networkId that optionally parses an arbitrary amount of kv arguments in addition to network id: `<network_id>[:k1=v1,k2=v2,...]`. `ExpressibleByArgument` doesn't support throwing so invalid kv arguments are also parsed into an array and thrown before creating attachment configurations. Here only `ip` will be added to supported keys.

AddressAllocator from containerization used by AttachmentAllocator supports reserving addresses and handles skipping over reserved addresses even when dynamic allocation starts behind and gets ahead of a reserved address. Attempting to reserve invalid addresses (already taken or out of range) throws a pretty low level error and should maybe be improved.

Does not currently implement --ip but can be done and merged with --network in getAttachmentConfigurations. However, swift argument parser doesn't seem to retain positional information relative to other flags and that makes supporting multiple --ip for multiple networks challenging. Also other container tools (only tried podman) seem to error when using --ip or --mac-address with multiple --network s.

The implemented format `--network network_id[:ip=<>]` follows Podman syntax: https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net

Docker style would be `--network name=network_id,ip=<>`: https://docs.docker.com/reference/cli/docker/container/run/#network (apparently `:` is reserved for `--network container:containername`)

Both can be supported simultaneously without breaking changes but maybe just one should be chosen to avoid confusing documentation.

Related:

- The NetworkArg option should be easy to extend to support MAC address in #459.
- The new --mac-address option introduced in #753 could be merged with --network in getAttachmentConfigurations

Fixes #282

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -18,7 +18,6 @@ import ArgumentParser
 import ContainerizationError
 import Foundation
 
-
 public struct Flags {
     public struct NetworkArg: ExpressibleByArgument, Decodable {
         var networkId: String
--- a/Sources/ContainerClient/Utility.swift
+++ b/Sources/ContainerClient/Utility.swift
@@ -241,7 +241,8 @@ public struct Utility {
             // attach the first network using the fqdn, and the rest using just the container ID
             return try networks.enumerated().map { item in
                 guard item.element.invalidArgs.isEmpty else {
-                    throw ContainerizationError(.invalidArgument, message: "invalid network arguments \(item.element.networkId): \(item.element.invalidArgs.joined(separator: ", "))")
+                    throw ContainerizationError(
+                        .invalidArgument, message: "invalid network arguments \(item.element.networkId): \(item.element.invalidArgs.joined(separator: ", "))")
                 }
                 guard item.offset == 0 else {
                     return AttachmentConfiguration(network: item.element.networkId, options: AttachmentOptions(hostname: containerId, ip: item.element.ip))


Sources/Services/ContainerNetworkService/AttachmentAllocator.swift
```diff
@@ -29,12 +29,18 @@ actor AttachmentAllocator {
...
     /// Allocate a network address for a host.
-    func allocate(hostname: String) async throws -> UInt32 {
+    func allocate(hostname: String, staticIndex: UInt32? = nil) async throws -> UInt32 {
         // Client is responsible for ensuring two containers don't use same hostname, so provide existing IP if hostname exists
         if let index = hostnames[hostname] {
             return index
         }
 
+        if let staticIndex {
+            try allocator.reserve(staticIndex)
```
siikamiika, line 39: When trying to reserve gateway ip 192.168.64.1:

```
Error: internalError: "failed to bootstrap container" (cause: "internalError: "failed to bootstrap container 66336145-44ba-4e26-8954-39eecd7f73fa (cause: "unknown: "cannot create index using address 3232251905"")"")
```

When trying to reserve already allocated ip 192.168.64.12:

```
Error: internalError: "failed to bootstrap container" (cause: "internalError: "failed to bootstrap container b7bd2b0f-00db-4cd3-8970-fd55864d6089 (cause: "unknown: "cannot choose already-allocated address 3232251916"")"")
```

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
```diff
@@ -88,6 +88,52 @@ class TestCLINetwork: CLITest {
...
+            }
+
+            let networkInspectOutput = try inspectNetwork(name)
+            guard let networkInspectStatus = networkInspectOutput.status else {
+                throw CLIError.invalidOutput("network inspect output invalid, missing status")
+            }
+            let network = try CIDRAddress(networkInspectStatus.address)
+            var addressBytes = network.address.networkBytes
+            addressBytes[3] = UInt8.random(in: 10...100)
+            let staticIp = try IPv4Address(fromNetworkBytes: addressBytes).description
```
siikamiika, line 111: Getting a static ip for testing is a bit difficult without a static subnet (#737) but should work

jglogan:
@siikamiika thanks! Sorry for the delay, will have a look no later than early next week.

jglogan:
@siikamiika Also see my comment at https://github.com/apple/container/pull/753#issuecomment-3412825966.

I'm just blasting through these so I'm not going to scan the whole issue and changes right now but perhaps you can answer here.

If say we went the "mount option" approach and did:

```bash
container run --network default,ip=192.168.64.14 ...
```

We're now in a condition where we can have two new failure modes that can occur if we `run`, `stop`, `system stop`, and then at some arbitrary time `system start`:

1. The subnet is still `192.168.64.0/24` but another container already is using `192.168.64.14`.
2. The subnet is now `192.168.65.0/24` because I started up some other software before container that took the 64 subnet.

Thoughts on handing these? Feel free to point me at prior comments as needed.

PR #753, opened 2025-10-12
Add --mac-address flag to set custom MAC addresses for containers

### LINKED ISSUE #752: [Request]: Add --mac-address flag to set custom MAC addresses for containers

# Problem

Currently, there is no way to specify a custom MAC address for a container's network interface and the MAC address is auto-generated by the system.

# Proposed Solution

Add a `--mac-address` flag to container run and container create commands:

`container run --mac-address AA:BB:CC:DD:EE:FF ubuntu`
`container create --mac-address AA:BB:CC:DD:EE:FF --name mycontainer ubuntu`

The flag would:

- Accept standard MAC address format: `XX:XX:XX:XX:XX:XX` (with : or - separators)
- Validate the MAC address format before creating the container
- Apply to the first network interface
- Be optional (current auto-generation behavior preserved when not specified)

### Code of Conduct

DSS3113:
## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [x] Documentation update
 
## Motivation and Context
[Why is this change needed?]
Currently, there is no way to specify a custom MAC address for a container's network interface and the MAC address is auto-generated by the system.

#### Use Cases
- **Network Testing**: Developers testing network-dependent applications that need predictable MAC addresses
- **License Management**: Running containerized software with MAC-based license keys
- **Network Automation**: Scripts and tools that expect specific MAC addresses for configuration
- **Debugging**: Consistent MAC addresses across container restarts for easier troubleshooting

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [X] Added/updated docs

## Issue
closes #752

### DIFF
--- a/Sources/ContainerClient/Parser.swift
+++ b/Sources/ContainerClient/Parser.swift
@@ -712,6 +712,76 @@ public struct Parser {
         }
     }
 
+    // MARK: Networks
+
+    /// Parsed network attachment with optional properties
+    public struct ParsedNetwork {
+        public let name: String
+        public let macAddress: String?
+
+        public init(name: String, macAddress: String? = nil) {
+            self.name = name
+            self.macAddress = macAddress
+        }
+    }
+
+    /// Parse network attachment with optional properties
+    /// Format: network_name[,mac=XX:XX:XX:XX:XX:XX]
+    /// Example: "backend,mac=02:42:ac:11:00:02"
+    public static func network(_ networkSpec: String) throws -> ParsedNetwork {
+        guard !networkSpec.isEmpty else {
+            throw ContainerizationError(.invalidArgument, message: "network specification cannot be empty")
+        }
+
+        let parts = networkSpec.split(separator: ",", omittingEmptySubsequences: false)
+
+        guard !parts.isEmpty else {
+            throw ContainerizationError(.invalidArgument, message: "network specification cannot be empty")
+        }
+
+        let networkName = String(parts[0])
+        if networkName.isEmpty {
+            throw ContainerizationError(.invalidArgument, message: "network name cannot be empty")
+        }
+
+        var macAddress: String?
+
+        // Parse properties if any
+        for part in parts.dropFirst() {
+            let keyVal = part.split(separator: "=", maxSplits: 2, omittingEmptySubsequences: false)
+
+            let key: String
+            let value: String
+
+            guard keyVal.count == 2 else {
+                throw ContainerizationError(
+                    .invalidArgument,
+                    message: "invalid property format '\(part)' in network specification '\(networkSpec)'"
+                )
+            }
+            key = String(keyVal[0])
+            value = String(keyVal[1])
+
+            switch key {
+            case "mac":
+                if value.isEmpty {
+                    throw ContainerizationError(
+                        .invalidArgument,
+                        message: "mac address value cannot be empty"
+                    )
+                }
+                macAddress = value
+            default:
+                throw ContainerizationError(
+                    .invalidArgument,
+                    message: "unknown network property '\(key)'. Available properties: mac"
+                )
+            }
+        }
+
+        return ParsedNetwork(name: networkName, macAddress: macAddress)
+    }
+
     // MARK: DNS
 
     public static func isValidDomainName(_ name: String) -> Bool {
--- a/Sources/ContainerClient/Utility.swift
+++ b/Sources/ContainerClient/Utility.swift
@@ -62,6 +62,14 @@ public struct Utility {
         }
     }
 
+    public static func validMACAddress(_ macAddress: String) throws {
+        let pattern = #"^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$"#
+        let regex = try Regex(pattern)
+        if try regex.firstMatch(in: macAddress) == nil {
+            throw ContainerizationError(.invalidArgument, message: "invalid MAC address format \(macAddress), expected format: XX:XX:XX:XX:XX:XX")
+        }
+    }
+
     public static func containerConfigFromFlags(
         id: String,
         image: String,
@@ -176,13 +184,15 @@ public struct Utility {
 
         config.virtualization = management.virtualization
 
+        // Parse network specifications with properties
+        let parsedNetworks = try management.networks.map { try Parser.network($0) }
         if management.networks.contains(ClientNetwork.noNetworkName) {
             guard management.networks.count == 1 else {
                 throw ContainerizationError(.unsupported, message: "no other networks may be created along with network \(ClientNetwork.noNetworkName)")
             }
             config.networks = []
         } else {
-            config.networks = try getAttachmentConfigurations(containerId: config.id, networkIds: management.networks)
+            config.networks = try getAttachmentConfigurations(containerId: config.id, networks: parsedNetworks)
             for attachmentConfiguration in config.networks {
                 let network: NetworkState = try await ClientNetwork.get(id: attachmentConfiguration.network)
                 guard case .running(_, _) = network else {
@@ -220,7 +230,14 @@ public struct Utility {
         return (config, kernel)
     }
 
-    static func getAttachmentConfigurations(containerId: String, networkIds: [String]) throws -> [AttachmentConfiguration] {
+    static func getAttachmentConfigurations(containerId: String, networks: [Parser.ParsedNetwork]) throws -> [AttachmentConfiguration] {
+        // Validate MAC addresses if provided
+        for network in networks {
+            if let mac = network.macAddress {
+                try validMACAddress(mac)
+            }
+        }
+
         // make an FQDN for the first interface
         let fqdn: String?
         if !containerId.contains(".") {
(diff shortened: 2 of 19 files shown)


Sources/ContainerClient/Utility.swift
```diff
@@ -241,13 +254,13 @@ public struct Utility {
             // attach the first network using the fqdn, and the rest using just the container ID
             return networkIds.enumerated().map { item in
                 guard item.offset == 0 else {
-                    return AttachmentConfiguration(network: item.element, options: AttachmentOptions(hostname: containerId))
+                    return AttachmentConfiguration(network: item.element, options: AttachmentOptions(hostname: containerId, macAddress: item.offset == 0 ? macAddress : nil))
```
siikamiika: (disclaimer: I'm not a maintainer and I'm not requesting a change, just interested in this feature because I'm implementing an adjacent feature to --network in #751)

Do you know how docker or podman handle this if there are multiple networks? I think if you do something like `--network net-1 --network net-2 --mac-address ff:ff:ff:ff:ff:ff` this will set the MAC on net-1.

DSS3113: Thanks for the question! Yes, multiple `--network` flags are supported with a single `--mac-address` flag. I tested this and here's the behavior: 

`container run --network net-1 --network net-2 --mac-address 02:42:ac:11:00:02 ubuntu`
**net-1 (first network)**: Gets the specified MAC address `02:42:ac:11:00:02`
**net-2 (subsequent networks)**: Gets auto-generated MAC address

I am not sure about podman but docker simply does not support inputting multiple `--network` flags in the `run` command *when a single `--mac-address` is specified*. [Edited]

`docker run -d --name test-mac-container --mac-address 02:42:ac:11:00:99 --network test-net-1 --network test-net-2 --network test-net-3 alpine sleep 300`
```
docker: Error response from daemon: Container cannot be connected to network endpoints: test-net-1, test-net-2, test-net-3.
See 'docker run --help'.
```

siikamiika: That's interesting, according to [Docker docs](1) multiple networks should be supported. Could it be related to the --mac-address flag even though the error message is a bit vague? (or maybe that's what you meant)

I don't have Docker installed at the moment, but I tried how Podman works in a VM:
```bash
# by default podman uses host networking
# podman also has a default bridge network that can be used with either --network=bridge or --network=podman, same as --network=default or no --network in apple/container

# create two custom named bridge networks:
$ podman network create net-1 # gets the network 10.89.0.0/24
$ podman network create net-2 # gets the network 10.89.1.0/24

$ podman run --rm -it --network=bridge --mac-address=02:00:00:00:00:01 alpine # default bridge network gets static mac (command fails with just --mac-address but that's due to the network type)
$ podman run --rm -it --network=net-1 --mac-address=02:00:00:00:00:01 alpine # custom bridge network net-1 gets static mac
$ podman run --rm -it --network=net-1 --network=net-2 alpine  # starts (random mac for both)
$ podman run --rm -it --mac-address=02:00:00:00:00:01 --network=net-1 --network=net-2 alpine # fails
Error: --mac-address can only be set for a single network: invalid argument
$ podman run --rm -it --network=net-1:mac=02:00:00:00:00:01 --network=net-2 alpine # net-1 gets static mac, net-2 gets random mac
$ podman run --rm -it --network=net-1:mac=02:00:00:00:00:01 --network=net-2:mac=02:00:00:00:00:02 alpine # both get static mac
```

[1]: https://docs.docker.com/reference/cli/docker/container/run/#network

DSS3113: You're right, I apologize. I should've clarified that the docker example failing was in the case when the `--mac-address` flag is specified with multiple networks.

jglogan:
@DSS3113 Thank you! Got a bit of a backlog but let's get the workflow going and I'll try to look it over no later than early next week.

jglogan:
@DSS3113 however, an immediate comment relating to the multiple networks aspect.

What do you think about handling attachment properties similarly to mount options for filesystems?

This has the attractive property of being extensible without cluttering the CLI with a bunch of options.

The drawback is discovery/documentation isn't as simple; there's only so much you can include in the help for a single option.

Example:

```bash
container run --network default --network backend,mac=addr1 --network backend2,mac=addr2
```

DSS3113:
> @DSS3113 however, an immediate comment relating to the multiple networks aspect.
> 
> What do you think about handling attachment properties similarly to mount options for filesystems?
> 
> This has the attractive property of being extensible without cluttering the CLI with a bunch of options.
> 
> The drawback is discovery/documentation isn't as simple; there's only so much you can include in the help for a single option.
> 
> Example:
> 
> ```shell
> container run --network default --network backend,mac=addr1 --network backend2,mac=addr2
> ```

[@jglogan](https://github.com/jglogan) Sure, I can start adjusting it to match this format.

jglogan:
@DSS3113 Building now but there are conflicts (probably because of the just merged `--network none` stuff).

I'll also pull down a patch into my local and test it out, looking fwd to trying it!

DSS3113:
@jglogan Resolved merge conflicts!

Sources/ContainerClient/Parser.swift
```diff
@@ -712,6 +712,83 @@ public struct Parser {
...
+        for part in parts.dropFirst() {
+            let keyVal = part.split(separator: "=", maxSplits: 2, omittingEmptySubsequences: false)
+
+            let key: String
+            let value: String
+
+            if keyVal.count == 2 {
+                key = String(keyVal[0])
+                value = String(keyVal[1])
+            } else if keyVal.count == 1 {
```
jglogan: Is this block necessary if it's the same as the `else` block?

jglogan:
@DSS3113 I'm seeing the following with my local build:

```console
% container run -it --name foo --network default,mac=c1:2f:c2:fd:ce:0b alpine:latest sh
Error: internalError: "failed to bootstrap container" (cause: "internalError: "failed to bootstrap container foo (cause: "unknown: "internal error (13): ip-link-set: netlink response indicates error, rc = -99"")"")
```

Are you seeing that it's working without issue on your system?

jglogan:
Ah, that's a multicast address (I just generated a random one). Let me try with a different one...

jglogan:
@DSS3113 Yep, looks good. Lemme take one last look at the code. Could you look at the one comment I posted above?

```console
% container run -it --rm --network default,mac=c2:2f:c2:fd:ce:0b alpine:latest ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,UP,LOWER_UP> mtu 1280 qdisc pfifo_fast state UP qlen 1000
    link/ether c2:2f:c2:fd:ce:0b brd ff:ff:ff:ff:ff:ff
3: teql0: <NOARP> mtu 1500 qdisc noop state DOWN qlen 100
    link/void 
4: tunl0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN qlen 1000
    link/ipip 0.0.0.0 brd 0.0.0.0
5: sit0@NONE: <NOARP> mtu 1480 qdisc noop state DOWN qlen 1000
    link/sit 0.0.0.0 brd 0.0.0.0
6: ip6tnl0@NONE: <NOARP> mtu 1452 qdisc noop state DOWN qlen 1000
    link/tunnel6 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00 brd 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00
```

jglogan:
@DSS3113 Also it looks like some of your commits in this PR might not have verified sigs.

Could you rebase your branch, reset it back to main, do a single squashed, verified commit, and force push that? I think we're good to go after that.

Sources/ContainerClient/Parser.swift
DSS3113: My bad. I've fixed that now.

DSS3113:
@jglogan Thanks for reviewing! I've done the needful.

jglogan:
Any idea why https://github.com/apple/container/pull/753/commits/4652e2f3dc98c84bf1e3a12c559b5a5bb036062a got pulled into this PR? That's already down in main.

DSS3113:
@jglogan I believe I cherry-picked the commit and that created a duplicate. My bad, it should be fixed now.

jglogan:
@DSS3113 It looks like there are `make fmt` issues, and I still see a merge commit (https://github.com/apple/container/pull/753/commits/72530a37d7314d285ac352f17f192f256866b6ef) that shouldn't be there if your change was cleanly rebased onto the latest `main`. Could you rebase just your changes onto the current main and run `make fmt` to resolve the build issues? I think everything should be good to go then. Thank you!

DSS3113:
@jglogan My sincere apologies! I think it should be good now.

Sources/ContainerCommands/System/DNS/DNSDelete.swift
```diff
@@ -41,7 +41,7 @@ extension Application {
                 try resolver.deleteDomain(name: domainName)
                 print(domainName)
             } catch {
-                throw ContainerizationError(.invalidState, message: "cannot delete domain (try sudo?)")
+                throw ContainerizationError(.invalidState, message: "cannot create domain (try sudo?)")
```
jglogan: @DSS3113 Looks like a regression snuck in here? Other than that it's ready to go.

jglogan (changes requested):

Sources/ContainerCommands/System/DNS/DNSDelete.swift
DSS3113: Fixed!

docs/command-reference.md
```diff
@@ -11,75 +11,68 @@ Runs a container from an image. If a command is provided, it will execute inside
 **Usage**
 
 ```bash
-container run [<options>] <image> [<arguments> ...]
```
jglogan, line 14: I just noticed that there are a lot of changes compared to current main that don't need to be made, along with at least one regression. Could you fix this so that only the PR only updates the material relating to the MAC address? Apologies for the hassle!

docs/command-reference.md
```diff
@@ -92,41 +85,44 @@ container run -d --name web -p 8080:80 nginx:latest
...
+# run a container with a specific MAC address
+container run --network default,mac=02:42:ac:11:00:02 ubuntu:latest
 ```
 
 ### `container build`
 
-Builds an OCI image from a local build context. It reads a Dockerfile (default `Dockerfile`) or Containerfile and produces an image tagged with `-t` option. The build runs in isolation using BuildKit, and resource limits may be set for the build process itself.
-
-When no `-f/--file` is specified, the build command will look for `Dockerfile` first, then fall back to `Containerfile` if `Dockerfile` is not found.
+Builds an OCI image from a local build context. It reads a Dockerfile (default `Dockerfile`) and produces an image tagged with `-t` option. The build runs in isolation using BuildKit, and resource limits may be set for the build process itself.
```
jglogan: Regression?

jglogan (changes requested):
Sorry, only now noticed some seemingly unnecessary diffs in command-reference.md, see comments.

Sources/ContainerCommands/System/DNS/DNSDelete.swift
0777888212: Hello

docs/command-reference.md
DSS3113: I apologize for the trouble. I somehow missed these. Should be fixed now.

docs/command-reference.md
DSS3113: SAA.

Sources/Services/ContainerNetworkService/NetworkService.swift
```diff
@@ -59,21 +59,24 @@ public actor NetworkService: Sendable {
         }
 
         let hostname = try message.hostname()
+        let macAddress = message.string(key: NetworkKeys.macAddress.rawValue)
```
jglogan, line 62: Just a note to myself and any interested parties here, no change required for this PR.

@dcantah reworked the client, API server, and Sandbox Service so that the client only deals with the API server for sandbox management (instead of the doing bootstrap through API server and the rest directly with the sandbox service) for the 0.5.0 release. This gives us the opportunity to make the API server responsible for allocating IP addresses at container start, and freeing the allocations when a container stops for any reason. Currently if `container-linux-runtime` gets terminated (SIGKILL) or faults, the IP will be orphaned.

When we make that change we might want simplify this by only passing what's needed for IP allocation and hostname resolution, and see if we can take the returned `Attachment` and integrate it with the MAC address and other caller-specified properties not related to the core network service functions to produce and `Attachment` that we can transfer over to the sandbox service at container start time. We trade a simpler interface between NetworkService and its caller for a little more complexity creating the Attachment that contains the user-specified network attributes, the allocated attributes, and the vmnet_network_ref.

docs/command-reference.md
jglogan: No worries! I appreciate your persistence.

Not an excuse but a self-criticism 😄 I've been paging out of PR and issue reviews a lot and I fear it's cost contributors several more round trips than the PRs really need. Something for me to work on!

jglogan (approved):

docs/command-reference.md
DSS3113: It’s alright, I understand! Thanks again!

PR #768, opened 2025-10-15
Add support for anonymous volumes

### LINKED ISSUE #726: [Request]: Support anonymous volumes

Rather than forcing pre-creation of named volumes it would be great to match the docker cli and create anonymous volumes on demand as this is incredibly common.
```
$ container run -v /xyz alpine
Error: invalidArgument: "anonymous volumes are not supported" 
```

Similar but slightly different from #690 

### Code of Conduct

realrajaryan:
- Closes #726.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Adds support for anonymous volumes. Users can now create volumes without explicit naming using `-v /path` and `--mount type=volume,dst=/path` syntax.

Usage:
```
# anonymous volume  
container run -v /data alpine
container run -v anon-01k7jpghe4kg4ph5a4vkccksbb:/data alpine

# Multiple anonymous volumes
container run --rm -v /logs -v /cache -v /tmp alpine
```

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/Volume.swift
+++ b/Sources/ContainerClient/Core/Volume.swift
@@ -60,7 +60,7 @@ public struct Volume: Sendable, Codable, Equatable, Identifiable {
 
 extension Volume {
     /// Reserved label key for marking anonymous volumes
-    public static let anonymousLabel = "com.apple.container.volume.anonymous"
+    public static let anonymousLabel = "com.apple.container.resource.anonymous"
 
     /// Whether this is an anonymous volume (detected via label)
     public var isAnonymous: Bool {
@@ -98,20 +98,12 @@ public enum VolumeError: Error, LocalizedError {
 /// Volume storage management utilities.
 public struct VolumeStorage {
     public static let volumeNamePattern = "^[A-Za-z0-9][A-Za-z0-9_.-]*$"
-    public static let anonymousVolumePattern = "^anon-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
     public static let defaultVolumeSizeBytes: UInt64 = 512 * 1024 * 1024 * 1024  // 512GB
 
     public static func isValidVolumeName(_ name: String) -> Bool {
         guard name.count <= 255 else { return false }
 
         do {
-            // Check if it's an anonymous volume name (anon-{uuid})
-            let anonRegex = try Regex(anonymousVolumePattern)
-            if (try? anonRegex.wholeMatch(in: name)) != nil {
-                return true
-            }
-
-            // Check if it's a regular named volume
             let regex = try Regex(volumeNamePattern)
             return (try? regex.wholeMatch(in: name)) != nil
         } catch {
@@ -119,8 +111,8 @@ public struct VolumeStorage {
         }
     }
 
-    /// Generates an anonymous volume name with the format: anon-{uuid}
+    /// Generates an anonymous volume name with UUID format
     public static func generateAnonymousVolumeName() -> String {
-        "anon-\(UUID().uuidString.lowercased())"
+        UUID().uuidString.lowercased()
     }
 }
--- a/Sources/ContainerClient/Parser.swift
+++ b/Sources/ContainerClient/Parser.swift
@@ -25,11 +25,13 @@ public struct ParsedVolume {
     public let name: String
     public let destination: String
     public let options: [String]
+    public let isAnonymous: Bool
 
-    public init(name: String, destination: String, options: [String] = []) {
+    public init(name: String, destination: String, options: [String] = [], isAnonymous: Bool = false) {
         self.name = name
         self.destination = destination
         self.options = options
+        self.isAnonymous = isAnonymous
     }
 }
 
@@ -434,7 +436,8 @@ public struct Parser {
         }
 
         // If it's a volume type but no source was provided, create an anonymous volume
-        if volumeName.isEmpty {
+        let isAnonymous = volumeName.isEmpty
+        if isAnonymous {
             volumeName = VolumeStorage.generateAnonymousVolumeName()
         }
 
@@ -442,7 +445,8 @@ public struct Parser {
             ParsedVolume(
                 name: volumeName,
                 destination: fs.destination,
-                options: fs.options
+                options: fs.options,
+                isAnonymous: isAnonymous
             ))
     }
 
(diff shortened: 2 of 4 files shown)


Sources/ContainerClient/Core/Volume.swift
```diff
@@ -88,16 +105,79 @@ public enum VolumeError: Error, LocalizedError {
 /// Volume storage management utilities.
 public struct VolumeStorage {
     public static let volumeNamePattern = "^[A-Za-z0-9][A-Za-z0-9_.-]*$"
+    public static let anonymousVolumePattern = "^anon-[0-9a-hjkmnp-tv-z]{26}$"
```
jglogan: Why not just use plain UUIDv4? Containers do the same if no `--name` is provided, and it keeps the PR simpler.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -46,13 +50,26 @@ public struct Volume: Sendable, Codable, Equatable, Identifiable {
...
         self.format = format
         self.source = source
         self.createdAt = createdAt
-        self.labels = labels
+        self.isAnonymous = isAnonymous
+        self.createdByContainerID = createdByContainerID
+
+        // Add reserved label for anonymous volumes to persist the flag
+        var finalLabels = labels
+        if isAnonymous {
```
jglogan: What purpose do these labels serve?

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -46,13 +50,26 @@ public struct Volume: Sendable, Codable, Equatable, Identifiable {
         labels: [String: String] = [:],
         options: [String: String] = [:],
         sizeInBytes: UInt64? = nil,
+        isAnonymous: Bool = false,
```
jglogan: I think I'd prefer that we don't do what Docker does, at least for now:

> Just like named volumes, anonymous volumes persist even if you remove the container that uses them, except if you use the --rm flag when creating the container, in which case the anonymous volume associated with the container is destroyed.

IMO it introduces a bunch of server-side complexity that we can add later should we need it. Perhaps we can approach the problem of pruning such volumes creatively to make up for the lack of the `--rm` anonymous volume feature.

Sources/ContainerClient/Utility.swift
```diff
@@ -161,6 +161,7 @@ public struct Utility {
                 resolvedMounts.append(fs)
             case .volume(let parsed):
                 do {
+                    // Try to inspect the volume first
                     let volume = try await ClientVolume.inspect(parsed.name)
```
jglogan: Can inspect throw for reasons other than a volume not existing? Is there a way to determine whether a volume exists without this assumption about inspect's behavior?

Sources/ContainerClient/Core/Volume.swift
egernst: I always wondered what the use case is/was with respect to anonymous _not_ being tied to workload.

dcantah: Is the complexity for the server that much though? I think we're (as we've found 😄) somewhat beholden to the behavior of other tools regardless, and Docker's behavior somewhat makes sense to me here. I think the outcome of not following this --rm behavior is a little more devious than others (like not supporting some flag they did, which is minor in most cases) as if folks have gotten used to --rm being the "nuke this thing please" button, if we didn't follow suit they may be surprised after 100 container runs with -v /whatever to find they have a bunch of volumes filling up their disk.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -88,16 +98,29 @@ public enum VolumeError: Error, LocalizedError {
 /// Volume storage management utilities.
 public struct VolumeStorage {
     public static let volumeNamePattern = "^[A-Za-z0-9][A-Za-z0-9_.-]*$"
+    public static let anonymousVolumePattern = "^anon-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
```
jglogan: My preference is for no `anon-`. Then unnamed containers and volumes follow the same naming convention.

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -58,6 +58,16 @@ public struct Volume: Sendable, Codable, Equatable, Identifiable {
     }
 }
 
+extension Volume {
+    /// Reserved label key for marking anonymous volumes
+    public static let anonymousLabel = "com.apple.container.volume.anonymous"
```
jglogan: Let's discuss using either `com.apple.container.resource.role=anonymous`, or `com.apple.container.resource.anonymous` like you're using it now.

Sources/ContainerClient/Parser.swift
```diff
@@ -459,7 +463,18 @@ public struct Parser {
         let parts = vol.split(separator: ":")
         switch parts.count {
         case 1:
-            throw ContainerizationError(.invalidArgument, message: "anonymous volumes are not supported")
+            // Anonymous volume: -v /path
+            // Generate a ULID-based name for the anonymous volume
```
jglogan: It's a UUID now, could just say "Generate a random name"

Sources/ContainerClient/Utility.swift
```diff
@@ -161,6 +161,7 @@ public struct Utility {
                 resolvedMounts.append(fs)
             case .volume(let parsed):
                 do {
+                    // Try to inspect the volume first
```
jglogan: Could we just check whether the volume is anonymous here, and if so create it, instead of all the retroactive checking in the exception handler?

Sources/ContainerCommands/Volume/VolumeList.swift
```diff
@@ -61,8 +61,13 @@ extension Application.VolumeCommand {
                 return
             }
 
+            // Sort volumes by creation time (newest first)
```
jglogan, line 64: For the future, we should probably sort everything this way.

Sources/ContainerCommands/Volume/VolumeList.swift
```diff
@@ -74,9 +79,11 @@ extension Application.VolumeCommand {
 
 extension Volume {
     var asRow: [String] {
+        let volumeType = self.isAnonymous ? "anonymous" : "named"
```
jglogan, line 82: This seems fine. If we have this, `anon-` on the name really does seem superfluous.

Sources/ContainerClient/Core/Volume.swift
jglogan: It's not whether it's too complex for the server. I'd rather do it as two smaller PRs if we're going to do it.

I also don't like us having to drop a bunch of extra fields in our core structs for anonymous and have to broaden the API when we're already using reserved labels for resource scope/role hints.

Let's get the basic stuff in place and do the cleanup in a follow-up PR.

> I always wondered what the use case is/was with respect to anonymous not being tied to workload.

"I don't care about the volume name but I want the data around for later."

I agree that it's a bit of a tenuous use case.

PR #769, opened 2025-10-16
Implicitly create named volumes

### LINKED ISSUE #690: [Request]: Create volumes automatically

When i spin up a new container, and i have a line like this
`
-v tailscale-ai-state:/var/lib/tailscale \
`
and volume is not created, it fails.

On docker, if volume does not exist, it is created automatically, which is very handy.

### Code of Conduct

realrajaryan:
- Closes #690.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Named volumes are now implicitly created when referenced in container commands. So, if `myvolume` doesn't exist and you run `container run -v myvolume:/data alpine`, it is automatically created.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Utility.swift
+++ b/Sources/ContainerClient/Utility.swift
@@ -284,7 +284,6 @@ public struct Utility {
         let labels = parsed.isAnonymous ? [Volume.anonymousLabel: ""] : [:]
 
         let volume: Volume
-        let volumeCreated: Bool
         do {
             volume = try await ClientVolume.create(
                 name: parsed.name,
@@ -292,31 +291,21 @@ public struct Utility {
                 driverOpts: [:],
                 labels: labels
             )
-            volumeCreated = true
         } catch let error as VolumeError {
             guard case .volumeAlreadyExists = error else {
                 throw error
             }
             // Volume already exists, just inspect it
             volume = try await ClientVolume.inspect(parsed.name)
-            volumeCreated = false
         } catch let error as ContainerizationError {
             // Handle XPC-wrapped volumeAlreadyExists error
             guard error.message.contains("already exists") else {
                 throw error
             }
             volume = try await ClientVolume.inspect(parsed.name)
-            volumeCreated = false
         }
 
-        // Warn user if named volume was auto-created
-        if volumeCreated && !parsed.isAnonymous {
-            let warning = "Volume '\(parsed.name)' does not exist. Creating new volume."
-            let formattedWarning = "\u{001B}[33m\(warning)\u{001B}[0m\n"
-            if let warningData = formattedWarning.data(using: .utf8) {
-                FileHandle.standardError.write(warningData)
-            }
-        }
+        // TODO: Warn user if named volume was auto-created
 
         return volume
     }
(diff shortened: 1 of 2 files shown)


Sources/ContainerClient/Utility.swift
```diff
@@ -185,8 +185,38 @@ public struct Utility {
                         volume = try await ClientVolume.inspect(parsed.name)
                     }
                 } else {
-                    // Named volume so it must already exist
-                    volume = try await ClientVolume.inspect(parsed.name)
+                    // Named volume - create if doesn't exist, inspect if it does
+                    var volumeCreated = false
```
jglogan: This is the smallest of nits, but could you change this to minimize the usage of mutable state?

Using `let volumeCreated: Bool` here helps avoid subtle bugs, since you need to explicitly indicate how things are handled for each conditional block.

It's typically harder for both humans and machines to reason about code that uses `var`.

Sources/ContainerClient/Utility.swift
```diff
@@ -185,8 +185,38 @@ public struct Utility {
...
+                    do {
+                        volume = try await ClientVolume.create(
```
jglogan: This (and the similar code in the anonymous volume block) is inherently racy.

I think that's fine for now, but perhaps create a new issue where we can have a design discussion around how whether we should improve this more generally. What you added in the ContainersService serves as a precedent: https://github.com/apple/container/blob/main/Sources/Services/ContainerAPIService/Containers/ContainersService.swift#L122.

The other benefit is then this exception stuff could be changed to an atomic "check existing volumes and do if doesn't exist" block.

Sources/ContainerClient/Utility.swift
```diff
@@ -185,8 +185,38 @@ public struct Utility {
                         volume = try await ClientVolume.inspect(parsed.name)
                     }
                 } else {
-                    // Named volume so it must already exist
-                    volume = try await ClientVolume.inspect(parsed.name)
+                    // Named volume - create if doesn't exist, inspect if it does
+                    var volumeCreated = false
+                    do {
```
jglogan: Ad we discussed in standup, see if there's enough common code here and for anonymous that an extract-method could DRY things a little.

Sources/ContainerClient/Utility.swift
```diff
@@ -305,4 +277,47 @@ public struct Utility {
...
+                throw error
+            }
+            volume = try await ClientVolume.inspect(parsed.name)
+            volumeCreated = false
+        }
+
+        // Warn user if named volume was auto-created
+        if volumeCreated && !parsed.isAnonymous {
+            let warning = "Volume '\(parsed.name)' does not exist. Creating new volume."
+            let formattedWarning = "\u{001B}[33m\(warning)\u{001B}[0m\n"
```
jglogan: We're not logging anything else in utility today, let's drop the output here for now and add a TODO comment mentioning that we might want to warn the user in the future.

We have #642 for improving the way we report things. At some point we should consider plumbing ContainerClient for Logger so we can provide more, and consistent feedback, using the log handler or a facility atop it (similar to  [this](https://github.com/Textualize/rich)) to do things like ANSI vs plain, etc.

PR #741, opened 2025-10-08
added the new workflow that automatically puts the labels in PR

Ronitsabhaya75:
## Type of Change
- [ ] New feature   : This can help dev teams to particularly look on those things are differently and understand what has been changed through labels

@dcantah can you please review this pr and suggest me if any changes require

PR #776, opened 2025-10-17
Revert "added the new workflow that automatically puts the labels in PR"

katiewasnothere:
Reverts apple/container#741

The action does not have permissions to add labels. Reverting this while we investigate the right fix 

@Ronitsabhaya75 heads up


Ronitsabhaya75:
> Reverts #741
> 
> The action does not have permissions to add labels. Reverting this while we investigate the right fix
> 
> @Ronitsabhaya75 heads up

Thanks for letting me know! I understand the permissions issue.  Happy to help update the PR once you've identified the right  permission configuration.

katiewasnothere:
@Ronitsabhaya75 I think we have two options here:

1. We could follow a similar flow to what is described in [this article](https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/) under Attack Vectors -> Remediation to split the workflow into two, where one is unprivileged and reads the PR content and another is privileged and can run the labeler on the PR number output by the first. 

2. OR we can have a cron job workflow like what is described in the labeler action docs [here](https://github.com/actions/labeler?tab=readme-ov-file#example-workflow-specifying-pull-request-numbers) that runs periodically to label new PRs. This one would be tricky to figure out how to determine what a "new PR" is and we don't want to constantly be running the labeler on PRs that are already labeled.

In either case, I think we can't read the labeler configuration file from the PR since that would require using untrusted files. 

If you want to read more about making GitHub workflows safer, [this](https://openssf.org/blog/2024/08/12/mitigating-attack-vectors-in-github-workflows/) is a good post.

Ronitsabhaya75:
hey @katiewasnothere I can look at option 1 and try to see how can i implement this

Issue #755, opened 2025-10-13
[Request]: Change all documented occurrences of Dockerfile to Containerfile


davpea:
### Feature or enhancement request details

The documentation and help text for the CLI both refer to Dockerfile... why are we selling another organisations products ? If we are compliant with OCI, we should change all occurrences of Dockerfile to Containerfile. Happy to do this if we have agreement from the community.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

davpea:
I've run a `grep -rnw 'Dockerfile'` against the top level folder and these are the places that Dockerfile is mentioned:

./Tests/CLITests/Subcommands/Build/CLIBuildBase.swift:66:        try dockerfileBytes.write(to: tempDir.appendingPathComponent("Dockerfile"), options: .atomic)
./Tests/CLITests/Subcommands/Build/CLIBuildBase.swift:109:            tempDockerfileContext.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderLocalOutputTest.swift:128:                .file("dockerfile-context.txt", content: .data("Dockerfile context file\n".data(using: .utf8)!))
./Tests/CLITests/Subcommands/Build/CLIBuilderLocalOutputTest.swift:240:                tempDockerfileContext.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderTarExportTest.swift:45:                "-f", tempDir.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderTarExportTest.swift:73:                "-f", tempDir.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderTarExportTest.swift:103:                "-f", tempDir.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderTarExportTest.swift:115:                "-f", tempDir.appendingPathComponent("Dockerfile").path,
./Tests/CLITests/Subcommands/Build/CLIBuilderTarExportTest.swift:135:                "-f", tempDir.appendingPathComponent("Dockerfile").path,
./docs/how-to.md:15:When you first run `container build`, `container` starts a *builder*, which is a utility container that builds images from your `Dockerfile`s. As with anything you run with `container run`, the builder runs in a lightweight virtual machine, so for resource-intensive builds, you may need to increase the memory and CPU limits for the builder VM.
[...]

PR #777, opened 2025-10-17
Changed all occurances of Dockerfile to Containerfile

davpea:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ X ] Documentation update

## Motivation and Context
[All occurrences of 'Dockerfile' have been changed to 'Containerfile' to ensure we are using OCI compliant language and not upselling an product, as mentioned in issue #755 ]

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ X ] Added/updated docs

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,94 @@
+{
+    "configurations": [
+        {
+            "type": "swift",
+            "request": "launch",
+            "args": [],
+            "cwd": "${workspaceFolder:apple-container}",
+            "name": "Debug container-runtime-linux",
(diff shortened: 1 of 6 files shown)


dcantah:
Thanks for the issue and suggestion. When we created the project we just went with the most common name from our collective experience. At this point in time I don't think we want to change our usage throughout the repo but what we'd like to implement which we enjoyed from nerdctl is to just have a fallback for Containerfile if a Dockerfile isn't detected (e.g. check for both, but leave the default today).

PR #780, opened 2025-10-17
Add test for trimDigest utility function

Ronitsabhaya75:
## Type of Change
- [x] Test addition

## Motivation and Context
The function removes the "sha256:" prefix from digests. This test  ensures the function works correctly and prevents regressions.

## Testing
- [x] Tested locally (verified with standalone test script)
- [x] Added/updated tests (added testTrimDigestRemovesPrefix)

```
Test behavior:
- Input:  `"sha256:abc123"`
- Expected:  `"abc123"` (prefix removed)
- Actual:  `"abc123"` 
```

### DIFF
--- a/Tests/ContainerClientTests/UtilityTests.swift
+++ b/Tests/ContainerClientTests/UtilityTests.swift
@@ -51,4 +51,10 @@ struct UtilityTests {
         #expect(result["standalone"] == "")
         #expect(result["key2"] == "value2")
     }
+
+    @Test("Trim digest removes sha256 prefix")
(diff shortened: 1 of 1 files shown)


realrajaryan:
Thanks for the contribution! This function is a pretty trivial string-formatting helper mainly used for CLI output, so unit testing it in isolation doesn’t add much real value. If `trimDigest()` were to break, it would be immediately visible in the user-facing output.

If we want to increase coverage meaningfully here, an integration test for the `image list` command (which exercises this logic as part of actual output formatting) would be a better place to validate it end-to-end.

PR #785, opened 2025-10-18
add support for multiple --tag flags in build

### LINKED ISSUE #732: [Request]: Multiple -t flags on build command should be recognized

When I use `docker build` I can use `-t 25.10.1 -t latest` at the same time, and the final image will be tagged with both tags `25.10.1` and `latest`. In case of `container build ... -t 25.10.1 -t latest` it will use just one of them.

### Code of Conduct

realrajaryan:
- Closes #732.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Adds support for specifying multiple `-t` / `--tag` flags with the `build` command. All specified tags will point to the same built image.

Example:
`container build -t myapp:latest -t myapp:v1.0.0 -t myapp:stable .`

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/BuildCommand.swift
+++ b/Sources/ContainerCommands/BuildCommand.swift
@@ -106,7 +106,9 @@ extension Application {
         var quiet: Bool = false
 
         @Option(name: [.short, .customLong("tag")], help: ArgumentHelp("Name for the built image", valueName: "name"))
-        var targetImageName: String = UUID().uuidString.lowercased()
+        var targetImageNames: [String] = {
+            [UUID().uuidString.lowercased()]
+        }()
 
         @Option(name: .long, help: ArgumentHelp("Set the target build stage", valueName: "stage"))
         var target: String = ""
@@ -195,11 +197,11 @@ extension Application {
                     try? FileManager.default.removeItem(at: tempURL)
                 }
 
-                let imageName: String = try {
-                    let parsedReference = try Reference.parse(targetImageName)
+                let imageNames: [String] = try targetImageNames.map { name in
+                    let parsedReference = try Reference.parse(name)
                     parsedReference.normalize()
                     return parsedReference.description
-                }()
+                }
 
                 var terminal: Terminal?
                 switch self.progress {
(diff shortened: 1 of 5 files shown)


Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -106,7 +106,9 @@ extension Application {
         var quiet: Bool = false
 
         @Option(name: [.short, .customLong("tag")], help: ArgumentHelp("Name for the built image", valueName: "name"))
-        var targetImageName: String = UUID().uuidString.lowercased()
+        var targetImageNames: [String] = {
+            [UUID().uuidString.lowercased()]
```
Ronitsabhaya75: Array default is [] but old default was UUID - how is this handled?

```var targetImageName: [String] = []```

File: Sources/ContainerCommands/BuildCommand.swift [hunk 2]
```diff
@@ -267,7 +269,7 @@ extension Application {
                             noCache: noCache,
                             platforms: [Platform](platforms),
                             terminal: terminal,
-                            tag: imageName,
+                            tags: imageNames,
```
Ronitsabhaya75: Does the builder support multiple tags?

dcantah (approved):

dcantah:
Just to be sure, the builder shim doesn't need any updates to support multiple tag headers being sent?

Ronitsabhaya75:
> Just to be sure, the builder shim doesn't need any updates to support multiple tag headers being sent?

I was thinking about multiple tags and what happens when multiple tags are same?

katiewasnothere:
> > Just to be sure, the builder shim doesn't need any updates to support multiple tag headers being sent?
> 
> I was thinking about multiple tags and what happens when multiple tags are same?

@Ronitsabhaya75 if there are multiple of the same tag it looks like we end up with only one version of the image with that tag.

Issue #729, opened 2025-10-07
[Request]: Named volume pre-population from container image


skaegi:
### Feature or enhancement request details

When you mount an **empty** named volume to a container directory (like `-v my_volume:/data`), `container run` should automatically copy any existing files from that directory in the image into the volume on first use.

```
$ echo -e "FROM alpine\nRUN mkdir /xyz && touch /xyz/hello" > Dockerfile
$ container build -t myimage .

$ container run myimage ls -al /xyz/hello
-rw-r--r--    1 root     root             0 Oct  7 00:36 /xyz/hello

$ container v create xyz
xyz
$ container run -v xyz:/xyz myimage ls -al /xyz/hello
ls: /xyz/hello: No such file or directory
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
For reference, this is the desired functionality.

https://docs.docker.com/engine/storage/volumes/#mounting-a-volume-over-existing-data

katiewasnothere:
Closing, thanks @skaegi!

jglogan:
@katiewasnothere I don't see a reason for closing this? My comment was there to provide a reference to the behavior that @skaegi called out.

Reopening this as it relates to another volume-related feature.

katiewasnothere:
Sorry about that! Closing this was an error on my part.

mungler:
Commenting to follow this one as my ticket #745 was closed in favour of this one :)

PR #787, opened 2025-10-19
Volume Ownership Fix for Non-Root Containers in Apple Container

radoxtech:
## Type of Change
- [x] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

#### Observed Problem
When mounting a volume to a container running as a non-root user (e.g., UID 1000), attempts to write files inside the container fail with permission denied errors. Application file operations report "Operation not permitted," even though the volume is mounted correctly. Forcing root execution inside the container resolves it, but breaks security guidelines.

#### Consequence
Non-root containers cannot use volumes for writable storage, limiting persistent data workflows. Developers must either run insecurely as root, manually adjust ownership post-mount (e.g., via `chown -R`), or avoid volumes entirely—disrupting OCI image compatibility and production-like testing on macOS.

#### Root Cause
Apple Container initializes volumes with root privileges on the host during creation. When mounted into a non-root container, the host filesystem permissions do not match the container's runtime UID/GID, preventing writes without escalation.

#### Fix Implemented
Apple Container now automatically adjusts volume ownership before mounting as follows:

- Detects if the volume was already mounted in the past (via mount history); if so, skips to avoid overwriting user data. If zero prior mounts (new volume), aligns ownership to the container's runtime UID/GID.
- Skips if running as root (no issue).
- Sources UID/GID from container's default runtime user or CLI flags at creation.

This ensures non-root containers can write to volumes seamlessly without image modifications or manual intervention, aligning with OCI standards.

#### Reproduction Steps (Apple Container CLI 0.5.0)
Create a simple non-root container image (e.g., based on Alpine with user 1000):
**Dockerfile**
```
FROM alpine:latest
RUN addgroup -g 1000 appgroup && adduser -u 1000 -G appgroup -D appuser
USER appuser
CMD ["sh"]
```
Build the image and create a named volume:
```
container build -t test-nonroot .
container volume create myvol
```
Run with the volume:
```
container run -it --name test-vol -v myvol:/data test-nonroot sh
```
From inside the container, attempt to write a file to /data.
Observed: Fails with "Permission denied."
## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerSandboxService/SandboxService.swift
+++ b/Sources/Services/ContainerSandboxService/SandboxService.swift
@@ -837,13 +837,13 @@ public actor SandboxService {
             self.log.info("Resolved username '\(czConfig.process.user.username)' to uid=\(uid) gid=\(gid)")
         }
 
-        // Skip if running as root (no fix needed, root can write anywhere)
-        guard uid != 0 else {
-            self.log.info("Container running as root (uid=0), skipping volume ownership fix")
+        self.log.info("Container user: uid=\(uid) gid=\(gid)")
+
+        // Skip if running as root with root group (no fix needed, root can write anywhere)
+        guard uid != 0 || gid != 0 else {
+            self.log.info("Container running as root:root (uid=0, gid=0), skipping volume ownership fix")
             return
         }
-
-        self.log.info("Container user: uid=\(uid) gid=\(gid)")
 
         // Find all volume mounts from container config
         let volumeMounts = config.mounts.filter { mount in


Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -759,6 +778,221 @@ public actor SandboxService {
...
+                self.log.warning("Failed to resolve username '\(czConfig.process.user.username)', skipping volume ownership fix")
+                return
+            }
+            uid = resolved.uid
+            gid = resolved.gid
+            self.log.info("Resolved username '\(czConfig.process.user.username)' to uid=\(uid) gid=\(gid)")
+        }
+
+        // Skip if running as root (no fix needed, root can write anywhere)
+        guard uid != 0 else {
```
jglogan: Would it still make sense to update if `uid == 0 && gid != 0`?

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -759,6 +778,221 @@ public actor SandboxService {
         }
     }
 
+    /// Checks if an EXT4 volume has never been mounted (is brand new).
+    /// Uses the EXT4 superblock mountCount field:
+    /// - mountCount == 0: Never mounted - SAFE to change ownership (brand new volume)
+    /// - mountCount > 0: Already been mounted/used - DO NOT change ownership (data may exist)
+    ///
+    /// This function is called BEFORE the volume is mounted, so new volumes will have mountCount=0.
+    private func isFirstMount(volumeImagePath: SystemPackage.FilePath) -> Bool {
```
jglogan, line 787: @dcantah Thinking about this in the context of storage enhancements that use the ext4 library.

Say we have a `container volume create --from path` command that creates and populates a filesystem in a single go. We'd need that command to know about this test and set the mount count to nonzero.

Are there other use cases where we want to setup uid/gid on a volume in instances other than this specific case (first mount, empty volume)?

For the sake of argument, what are our alternatives? Could the guest agent could perform this setup after mounting volumes and before starting the workload?

Also want to call out #729 which I don't see us doing in the immediate future but could have some interaction with this - if the volume is empty/first mount, what action(s) do we take and where?

radoxtech:
## Investigation Update - Issue May Be Resolved in 0.8.0

After testing on container version **0.8.0**, I can confirm that **the issue described in this PR no longer reproduces**.

---

### ✅ Verification Testing (Container 0.8.0)

I reproduced the exact test case from this PR on `container 0.8.0`:

**Dockerfile:**
```dockerfile
FROM alpine:latest
RUN addgroup -g 1000 appgroup && adduser -u 1000 -G appgroup -D appuser
USER appuser
```

**Commands:**
```bash
container volume create myvol
container run --rm -v myvol:/data test-nonroot sh -c \
  "echo 'test' > /data/file.txt && cat /data/file.txt && ls -la /data/"
```

**Result: ✅ Works without errors**
```
test
-rw-r--r-- 1 appuser appgroup 5 Jan 28 12:39 file.txt
```

---

### 🧪 Additional Tests (All Passing)

| Image | UID:GID | Status |
|-------|---------|--------|
| Alpine (latest & 3.14) | 1000:1000 | ✅ Works |
| Nginx | 33:33 | ✅ Works |
| Node.js | 1000:1000 | ✅ Works |
| Postgres | 999:999 | ✅ Works |
| Debian | 1000:1000 | ✅ Works |

Also tested:
- ✅ Pre-existing directories in images
- ✅ Direct writes to volume root
- ✅ Subdirectory creation

---

### 🔍 Possible Explanation

Between versions **0.5.0** (where the issue occurred) and **0.8.0** (where it works), there was a significant change to Virtualization Framework cache behavior:

**Commit:** [837aa5e](https://github.com/apple/container/commit/837aa5eb8324054e2592a48cb5c88bd2fd37998c) by @JaewonHur (Jan 12, 2026)

**What changed:**
- **File:** `Sources/ContainerResource/Container/Filesystem.swift`
- **Change:** `cache: CacheMode = .auto` → `cache: CacheMode = .on`
- **Intent:** Fix filesystem corruption (related to [#614](https://github.com/apple/container/issues/614) and [UTM PR #5919](https://github.com/utmapp/UTM/pull/5919))

**Theory:** The `.auto` mode may have caused inconsistent filesystem metadata between host and guest, leading to permission errors. The `.on` mode enforces consistent caching, potentially resolving this as a side effect.

> ⚠️ **Note:** I cannot definitively confirm this is the root cause - there were 137 commits between 0.5.0 and 0.8.0, and the actual fix may have been something else entirely.

---

### 📊 Timeline

| Version | Status | Notes |
|---------|--------|-------|
| **0.5.0** | ❌ Issue present | Permission denied errors with non-root users |
| **Jan 12, 2026** | 🔧 Possible fix | Cache mode change merged |
| **0.8.0** | ✅ Issue resolved | All tests passing |

PR #788, opened 2025-10-19
Added the split method for labeler workflow

Ronitsabhaya75:
## Type of Change

- [x] Added the split approach @katiewasnothere suggested using openssf
- [x] Added adtiional feature for colors

## Motivation and Context
This can basically Help teams know what changes are made in pr

## Testing
- [x] Tested locally

```
Currently this is for CLI only later we can add multiple
```

**Structure how workflow runs**
```
 Pull Request Opened (untrusted code) 
             ↓

**pr-label-analysis.yml (Read-only)**
- Analyzes files using native git commands
- No secrets, no write permissions
- No third-party actions
- Saves to artifact
             ↓

- Artifact Storage
- PR number + labels
             ↓

**pr-label-apply.yml (Write access)**
- Never touches PR code
- Only reads artifact data
- Uses GitHub official actions only
- Applies labels via github-script

             ↓

🏷️ Labels Applied!

### CODE REVIEW

File: .github/workflows/color-labels.yml [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Create or update labels
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const labels = [
+              { name: 'cli', color: '0E8A16', description: 'Changes to CLI components' },
```
katiewasnothere: Instead of trying to encode the different label categories into the workflow job itself, maybe instead we could use the PR number parameter to actions/labeler in the label-apply workflow like this: 

```
    - uses: actions/labeler@v6
      with:        
        pr-number: |
          1
          <PR number we got from the artifact download>
```

We should double check that the labeler doesn't try to read the untrusted code. But my understanding is that the action does not checkout the PR code, just analyzes based on the files that have been changed. 

Note: I think it would be unsafe to read the labeler's configuration file from the PR, we might have to settle for using the configuration in main. But we should look into that as well.

.github/workflows/color-labels.yml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - '.github/workflows/color-labels.yml'
+  workflow_dispatch:
+
+permissions:
+  issues: write
```
katiewasnothere: See https://github.com/apple/container/pull/781, the top level permissions should always be the least permissive possible and any additional permissions should be set at the job level.

.github/workflows/color-labels.yml
Ronitsabhaya75: Thanks for the suggestion, Katie! I've updated the workflows to use `actions/labeler@v5` with the PR number from the artifact as you recommended.

.github/workflows/color-labels.yml
Ronitsabhaya75: I changed the permissiion to read only

Ronitsabhaya75:
@katiewasnothere can you look at the changes I made please let me know if they are good

.github/workflows/color-labels.yml
```diff
@@ -0,0 +1,65 @@
+name: Color Labels
```
katiewasnothere: I don't think we need this. We can make them by hand for now.

Ronitsabhaya75: should i remvoe the colors file

katiewasnothere: Yes

.github/workflows/pr-label-apply.yml
```diff
@@ -0,0 +1,52 @@
...
+          PR_NUMBER=$(cat "${METADATA_DIR}/pr-number.txt")
+          echo "number=${PR_NUMBER}" >> $GITHUB_OUTPUT
+          echo "PR Number: ${PR_NUMBER}"
+      
+      - name: Apply labels using labeler
+        uses: actions/labeler@v5
+        with:
+          pr-number: ${{ steps.pr-number.outputs.number }}
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          configuration-path: .github/labeler.yml
```
katiewasnothere, line 54: Do we need to have checked out the repo to access this file?

Ronitsabhaya75: Yes, we need to checkout the repository to access the .github/labeler.yml configuration file. Without checkout, the workflow runner won't have access to any files from the repo, including the labeler configuration.

katiewasnothere: I don't see the repo being checked out in this workflow

Ronitsabhaya75: I think I accidentally removed it let me add it again. Thank you Katie for pointing that out.

Ronitsabhaya75:
@katiewasnothere can review the changes

Ronitsabhaya75:
@katiewasnothere could you please review this update code

katiewasnothere:
Hi @Ronitsabhaya75, I asked @jglogan to take a look at the PR as well for another set of eyes on the security of this change. I think he had some concerns about if we wanted to take this change, but I'll let him add his thoughts here.

Ronitsabhaya75:
> Hi @Ronitsabhaya75, I asked @jglogan to take a look at the PR as well for another set of eyes on the security of this change. I think he had some concerns about if we wanted to take this change, but I'll let him add his thoughts here.

thank you so much for letting me know

jglogan (approved):

PR #791, opened 2025-10-21
[networks]: Add creationDate field

### LINKED ISSUE #665: [Request]: Add Creation Date in Network Metadata

Similar to #302,

The creation date field will help describe the network(s), and will help with streamline the UX experience compared to other container tools.

Example from podman:

```
podman network inspect podman
[
     {
          "name": "podman",
          "id": "2f259bab93aaaaa2542ba43ef33eb990d0999ee1b9924b557b7be53c0b7a1bb9",
          "driver": "bridge",
          "network_interface": "podman0",
          "created": "2025-09-22T20:22:23.222759857+03:00",
          "subnets": [
               {
                    "subnet": "10.88.0.0/16",
                    "gateway": "10.88.0.1"
               }
          ],
          "ipv6_enabled": false,
          "internal": false,
          "dns_enabled": false,
          "ipam_options": {
               "driver": "host-local"
          },
          "containers": {}
     }
]
```

Compared to current details returned:

```
container network inspect default | jq .
[
  {
    "id": "default",
    "config": {
      "id": "default",
      "mode": "nat"
    },
    "state": "running",
    "status": {
      "address": "192.168.64.0/24",
      "gateway": "192.168.64.1"
    }
  }
]
```

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #665 

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerNetworkService/NetworkConfiguration.swift
+++ b/Sources/Services/ContainerNetworkService/NetworkConfiguration.swift
@@ -16,6 +16,7 @@
 
 import ContainerizationError
 import ContainerizationExtras
+import Foundation
 
 /// Configuration parameters for network creation.
 public struct NetworkConfiguration: Codable, Sendable, Identifiable {
@@ -24,6 +25,9 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
 
     /// The network type
     public let mode: NetworkMode
+
+    /// When the network was created.
+    public let creationDate: Date
 
     /// The preferred CIDR address for the subnet, if specified
     public let subnet: String?
@@ -39,6 +43,7 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
         labels: [String: String] = [:]
     ) throws {
         self.id = id
+        self.creationDate = Date()
         self.mode = mode
         self.subnet = subnet
         self.labels = labels
@@ -47,6 +52,7 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
 
     enum CodingKeys: String, CodingKey {
         case id
+        case creationDate
         case mode
         case subnet
         case labels
@@ -58,6 +64,7 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         id = try container.decode(String.self, forKey: .id)
+        creationDate = try container.decodeIfPresent(Date.self, forKey: .creationDate) ?? Date(timeIntervalSince1970: 0)
         mode = try container.decode(NetworkMode.self, forKey: .mode)
         subnet = try container.decodeIfPresent(String.self, forKey: .subnet)
         labels = try container.decodeIfPresent([String: String].self, forKey: .labels) ?? [:]
--- a/Sources/Services/ContainerNetworkService/NetworkState.swift
+++ b/Sources/Services/ContainerNetworkService/NetworkState.swift
@@ -53,4 +53,11 @@ public enum NetworkState: Codable, Sendable {
         case .running(let configuration, _): configuration.id
         }
     }
+
+    public var creationDate: Date {
+        switch self {
+        case .created(let configuration): configuration.creationDate
+        case .running(let configuration, _): configuration.creationDate
+        }
+    }
 }


saehejkang:
Not sure if I added the `createdAt` field correctly to the decoder, but I am getting an error when testing any network commands. 

```
Error: keyNotFound(CodingKeys(stringValue: "createdAt", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "running", intValue: nil), RunningCodingKeys(stringValue: "_0", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"createdAt\", intValue: nil) (\"createdAt\").", underlyingError: nil))
```

Maybe something to do with the encoding and decoding done in the `create` function [here](https://github.com/apple/container/blob/ae279105a20e86f7b5a804973184c4c4c6c2fcc9/Sources/ContainerClient/Core/ClientNetwork.swift#L42)? First time looking into the api/client side of things.

saehejkang:
Not sure the protocol with pinging maintainers, as I don't want to bloat anyone's notifications.

But @jglogan , I would love to get some 👀 on this. Mainly to check if I am even taking the right steps in what we want accomplish here.

jglogan:
@saehejkang I'll have a look next week.

jglogan:
@saehejkang I'll have a review of this done by the end of the day. The main thing I want to consider is whether what you have here follows a pattern that we can apply to the metadata for all our managed resources.

Sources/Services/ContainerNetworkService/NetworkState.swift
```diff
@@ -53,4 +53,11 @@ public enum NetworkState: Codable, Sendable {
         case .running(let configuration, _): configuration.id
         }
     }
+
```
jglogan: This pattern could get out of hand as we add more configuration attributes.

Perhaps we could have a `public var configuration: NetworkConfiguration?` that provides configuration if the state has that associated data (always would for now but it would give us leeway for states that don't)?

Sources/ContainerCommands/Network/NetworkList.swift
```diff
@@ -90,12 +90,14 @@ extension NetworkState {
 
 public struct PrintableNetwork: Codable {
     let id: String
+    let createdAt: Date
```
jglogan: We don't need this in PrintableNetwork if we're not printing it for now, right?

Let's leave it out here and not print the create date in tabular output, to start with.

Sources/Services/ContainerNetworkService/NetworkState.swift
saehejkang: I think I understand what you kinda mean? Do we want to update `id` as well?

```swift
public var configuration: NetworkConfiguration? {
   switch self {
      case .created(let configuration): configuration
      case .running(let configuration, _): configuration
   }
}

public var createdAt: Date? {
   configuration?.createdAt
}
```

Sources/Services/ContainerNetworkService/NetworkConfiguration.swift
```diff
@@ -58,6 +64,7 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         id = try container.decode(String.self, forKey: .id)
+        createdAt = try container.decode(Date.self, forKey: .createdAt)
```
saehejkang: https://github.com/apple/container/pull/791#issuecomment-3424700350

jglogan: Will this be backward-compatible? If I upgrade to this and have an existing container that doesn't have `createdAt`, serialization will probably fail?

`decodeIfPresent` defaulting to the epoch time probably makes sense here.

Sources/Services/ContainerNetworkService/NetworkConfiguration.swift
```diff
@@ -24,6 +25,9 @@ public struct NetworkConfiguration: Codable, Sendable, Identifiable {
 
     /// The network type
     public let mode: NetworkMode
+    
+    /// When the network was created.
+    public let createdAt: Date
```
jglogan: Naming not: I think I prefer `creationDate` for this, "At" feels more like a place than a time, while `Date` offers an indication of the actual type when looking at the code.

Sources/Services/ContainerNetworkService/NetworkState.swift
jglogan: Given that we just have these two for now, let's leave your change here as it currently is.

Sources/Services/ContainerNetworkService/NetworkConfiguration.swift
saehejkang: I was unsure of the point of serialization, but it is clear now.

jglogan (approved):

PR #792, opened 2025-10-21
Make container start idempotent

### LINKED ISSUE #772: [Bug]: container suddenly disappears
### I have done the following

### Steps to reproduce

If you run "container start somecontainer" twice and then "container stop somecontainer," the somecontainer will disappear from the list of all available containers.

### Current behavior

The container suddenly disappears.

### Expected behavior

show error without dissappear

### Environment

```markdown
- OS: macOS 26.0
- Xcode: 
- Container: Container CLI version 0.5
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

dcantah:
Fixes #772 

Today it fails in bootstrap the second go around, and we also have an error handler that automatically cleans up the container if bootstrap failed which is even worse. This change short circuits us first in the cli if the state is running when we get() the container, and also adds in a clause to bootstrap to just early return if we already have a client.

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerStart.swift
+++ b/Sources/ContainerCommands/Container/ContainerStart.swift
@@ -52,9 +52,24 @@ extension Application {
             }
             progress.start()
 
+            let detach = !self.attach && !self.interactive
             let container = try await ClientContainer.get(id: containerId)
+
+            // Bootstrap and process start are both idempotent and don't fail the second time
+            // around, however not doing an rpc is always faster :). The other bit is we don't
+            // support attach currently, so we can't do `start -a` a second time and have it succeed.
+            if container.status == .running {
+                if !detach {
+                    throw ContainerizationError(
+                        .invalidArgument,
+                        message: "attach is currently unsupported on already running containers"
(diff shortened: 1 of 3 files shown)


Sources/ContainerCommands/Container/ContainerStart.swift
```diff
@@ -52,9 +52,24 @@ extension Application {
...
             let container = try await ClientContainer.get(id: containerId)
+
+            // Bootstrap and process start are both idempotent and don't fail the second time
+            // around, however not doing an rpc is always faster :). The other bit is we don't
+            // support attach currently, so we can't do `start -a` a second time and have it succeed.
+            if container.status == .running {
+                if !detach {
+                    throw ContainerizationError(
+                        .invalidArgument,
+                        message: "attach is currently unsupported on already running containers"
```
realrajaryan, line 65: Can we also mention what to do instead here?

dcantah: There's not much to do 😅 we just don't support attach of any kind. What did you have in mind to write?

PR #793, opened 2025-10-21
Fix container build not using context-dir Dockerfile by default

### LINKED ISSUE #782: [Bug]: `container build` not using `context-dir`
### I have done the following

### Steps to reproduce

1. Run build with any other path: `container build ..`
2. If there's Dockerfile in the current path, it will build it. If there's not a Dockerfile, it will throw:
`Error: Dockerfile does not exist at path: Dockerfile`

### Current behavior

The `build` command does ignores the `context-dir` passed to it, it builds the Dockerfile it finds at the current path or throw an error if there is none.

### Expected behavior

`build` should use the argument path as the context directory instead of always using the current path.

### Environment

```markdown
- OS: 26.0.1
- Xcode: 26.0.1 (25A362)
- Container: container CLI version 0.5.0
```

### Relevant log output

```shell
Error: Dockerfile does not exist at path: Dockerfile
Usage: container build [<options>] [<context-dir>]
  See 'container build --help' for more information.
```

### Code of Conduct

0xEDU:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This PR proposes a solution that resolves #782.
If the -f flag is not provided, I prepend the context directory before the filename so that it matches Docker’s behavior and uses the Dockerfile in the context directory.
I also added tests for this behavior.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "ad946e609a69599b154a1855351540356d2e1b0ec0d8fa3e91c7cce46628f015",
+  "originHash" : "26b26b11a8c840723acebdf378a73b77950dc4c3dfeda984e6337a0ab4ef3f2b",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/containerization.git",
       "state" : {
-        "revision" : "056a909081181ab4e91f4143682a63f991a32584",
-        "version" : "0.11.0"
+        "revision" : "c7af5b8222551f07dc0b65b8488d5e6a386ba49a",
+        "version" : "0.12.1"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ import PackageDescription
 let releaseVersion = ProcessInfo.processInfo.environment["RELEASE_VERSION"] ?? "0.0.0"
 let gitCommit = ProcessInfo.processInfo.environment["GIT_COMMIT"] ?? "unspecified"
 let builderShimVersion = "0.6.1"
-let scVersion = "0.11.0"
+let scVersion = "0.12.1"
 
 let package = Package(
     name: "container",
--- a/Sources/ContainerClient/Core/ClientContainer.swift
+++ b/Sources/ContainerClient/Core/ClientContainer.swift
@@ -277,7 +277,7 @@ extension ClientContainer {
             guard let fds else {
                 throw ContainerizationError(
                     .internalError,
-                    message: "No log fds returned"
+                    message: "no log fds returned"
                 )
             }
             return fds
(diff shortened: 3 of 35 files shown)


Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -184,7 +184,8 @@ extension Application {
                     throw ValidationError("builder is not running")
                 }
 
-                let dockerfile = try Data(contentsOf: URL(filePath: file))
+                let filePath = file == "Dockerfile" ? contextDir + "/" + file : file
```
mdempsky: I believe `docker build -f Dockerfile dir` uses ./Dockerfile, not dir/Dockerfile.

I think to match that behavior, you'll have to change the flag behavior so you can distinguish no -f flag from an explicit `-f Dockerfile`.

Maybe change it to an optional string, and use `contextDir + "/Dockerfile"` as the default with it's absent.

And then two style suggestions:

1. Maybe add a helper method or getter so you don't need to duplicate this logic elsewhere.

2. Is string concatenation with "/" the right way to join paths in swift? But maybe it doesn't matter since this is a macOS-specific app anyway.

0xEDU:
I’ve applied the requested changes and am open to any additional suggestions or adjustments (also had a little quarrel with Git, hence the force pushes 😅).

Tests/CLITests/Subcommands/Build/CLIBuilderTest.swift
```diff
@@ -456,5 +456,120 @@ extension TestCLIBuildBase {
...
+        
+        @Test func testBuildDefaultDockerfileInContextDir() throws {
+            let tempDir: URL = try createTempDir()
+            let dockerfile: String =
+                """
+                FROM ghcr.io/linuxcontainers/alpine:3.20
+                RUN echo "default dockerfile" > /test.txt
+                """
+            let context: [FileSystemEntry] = [
+                .file("emptyFile", content: .zeroFilled(size: 1)),
```
katiewasnothere: We don't need this since it's not being used in the dockerfile

0xEDU:
I've addressed the test changes and fixed the formatting issue!

dcantah:
I realized right before opening https://github.com/apple/container/pull/812 that it fixes #782 as well 😅 my apologies. I should've remembered this was open

0xEDU:
No problem at all! Supersede this if yours fully covers the same case, happy to adjust or close mine if needed 👍

jglogan:
@0xEDU I merged #812 but please have a look at yours and see if there's anything from it that might still be useful to integrate, otherwise yeah go ahead and close it. Thank you for contributing and sorry for the mixup!

katiewasnothere:
@0xEDU Closing since the other changes have been merged. Feel free to open another PR if there's additional content your PR made that was not included in the merged changes, especially the tests :)

PR #796, opened 2025-10-21
Deps: Bump Containerization to 0.11.0

dcantah:
Renames the process config type and moves rosetta from being a container configuration value to being part of the vmms config.


realrajaryan (approved):
LGTM

Issue #641, opened 2025-09-18
[Request]: UX: Replace `--disable-progress-updates` with `--progress mode`.


jglogan:
### Feature or enhancement request details

Modes are:

- `none`: no progress output, replaces the `--disable-progress-updates` flag
- `ansi`: progress output using ANSI escape codes to produce clean, compact progress bar output

This makes it possible to have follow-on option values such as:

- `plain`: plain text output, one line per progress step with no ANSI codes
- `color`: allows differentiation between ANSI output with just cursor control, and color-coded output

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

karenheckel:
@jglogan @dkovba Hello! We are students at UT Austin working on issues for our virtualization class. If no one has taken this, we would love to work on it. Thank you!

saehejkang:
@karenheckel I noticed you are going through all the issues and commenting on them asking which is a good one to work on. I was going to come back to this one and wrap it up, but y'all can go ahead and work on this one! Looks like the two option values below need to be added. I will be active on this issue and help out as much as I can, if any questions arise. 

> plain: plain text output, one line per progress step with no ANSI codes
color: allows differentiation between ANSI output with just cursor control, and color-coded output

Could you also do me a favor and take a look at #855? I want to start collecting questions and things that new contributors have, so that it is easier for new developers to get up to speed on the codebase. Anything you have questions on (that is not already noted in ANY of the docs), just post them there.

Hook Em 🤘🏽

karenheckel:
@saehejkang Great thank you! We're looking for 3 issues to contribute to for our class, so this is a great issue to start with. We'll let you know if we have any questions!

Yes, we can definitely post any questions that come up to that discussion! Thanks again!

katiewasnothere:
@karenheckel Definitely feel free to pick up this issue! Thanks for checking in! Feel free to tag me for questions as well.

PhantomInTheWire:
hey @saehejkang @katiewasnothere @jglogan i have opened to a pr to push this forward could you please take a look? thanks.

nrthakur:
Hi @jglogan If no progress has been made on this, I'd love to pick it up

jglogan:
@nrthakur Were you thinking of adding additional progress modes?

I looked at the code and the work to address this issue was done by #808 (https://github.com/apple/container/blob/main/Sources/Services/ContainerAPIService/Client/Flags.swift#L330).

[...]

PR #808, opened 2025-10-25
[options]:  Replace --disable-progress-updates with --progress (none | ansi)

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Part 1 of #641

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerRun.swift
+++ b/Sources/ContainerCommands/Container/ContainerRun.swift
@@ -61,9 +61,9 @@ extension Application {
             let id = Utility.createContainerID(name: self.managementFlags.name)
 
             var progressConfig: ProgressConfig
-            if progressFlags.disableProgressUpdates {
-                progressConfig = try ProgressConfig(disableProgressUpdates: progressFlags.disableProgressUpdates)
-            } else {
+            switch self.progressFlags.progress {
+            case .none: progressConfig = try ProgressConfig(disableProgressUpdates: true)
+            case .ansi:
                 progressConfig = try ProgressConfig(
                     showTasks: true,
                     showItems: true,
--- a/Sources/ContainerCommands/Image/ImagePush.swift
+++ b/Sources/ContainerCommands/Image/ImagePush.swift
@@ -68,9 +68,9 @@ extension Application {
             let image = try await ClientImage.get(reference: reference)
 
             var progressConfig: ProgressConfig
-            if progressFlags.disableProgressUpdates {
-                progressConfig = try ProgressConfig(disableProgressUpdates: progressFlags.disableProgressUpdates)
-            } else {
+            switch self.progressFlags.progress {
+            case .none: progressConfig = try ProgressConfig(disableProgressUpdates: true)
+            case .ansi:
                 progressConfig = try ProgressConfig(
                     description: "Pushing image \(image.reference)",
                     itemsName: "blobs",
@@ -79,6 +79,7 @@ extension Application {
                     ignoreSmallSize: true
                 )
             }
+
             let progress = ProgressBar(config: progressConfig)
             defer {
                 progress.finish()
--- a/Tests/CLITests/Subcommands/Build/CLIRunBase.swift
+++ b/Tests/CLITests/Subcommands/Build/CLIRunBase.swift
@@ -22,28 +22,28 @@ class TestCLIRunBase: CLITest {
     var terminal: Terminal!
     var containerName: String = UUID().uuidString
 
-    var ContainerImage: String {
+    var containerImage: String {
         fatalError("Subclasses must override this property")
     }
 
-    var Interactive: Bool {
+    var interactive: Bool {
         false
     }
 
-    var Tty: Bool {
+    var tty: Bool {
         false
     }
 
-    var Entrypoint: String? {
+    var entrypoint: String? {
         nil
     }
 
-    var Command: [String]? {
+    var command: [String]? {
         nil
     }
 
-    var DisableProgressUpdates: Bool {
-        false
+    var progress: String {
+        "ansi"
     }
 
     override init() throws {
@@ -108,24 +108,23 @@ class TestCLIRunBase: CLITest {
             name,
         ]
 
-        if Interactive && Tty {
+        if interactive && tty {
             arguments.append("-it")
         } else {
-            if Interactive { arguments.append("-i") }
-            if Tty { arguments.append("-t") }
+            if interactive { arguments.append("-i") }
+            if tty { arguments.append("-t") }
         }
 
-        if DisableProgressUpdates {
-            arguments.append("--disable-progress-updates")
-        }
+        arguments.append("--progress")
+        arguments.append(progress)
 
-        if let entrypoint = Entrypoint {
+        if let entrypoint = entrypoint {
             arguments += ["--entrypoint", entrypoint]
         }
 
-        arguments.append(ContainerImage)
+        arguments.append(containerImage)
 
-        if let command = Command {
+        if let command = command {
             arguments += command
         }
         return try runInteractive(arguments: arguments)
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -79,7 +79,7 @@ container run [<options>] <image> [<arguments> ...]
 
 **Progress Options**
 
-*   `--disable-progress-updates`: Disable progress bar updates
+*   `--progress <type>`: Progress type (format: none|ansi) (default: ansi)
 
 **Examples**
 
@@ -395,7 +395,7 @@ Pulls an image from a registry. Supports specifying a platform and controlling p
 **Usage**
 
 ```bash
-container image pull [--debug] [--scheme <scheme>] [--disable-progress-updates] [--arch <arch>] [--os <os>] [--platform <platform>] <reference>
+container image pull [--debug] [--scheme <scheme>] [--progress <type>] [--arch <arch>] [--os <os>] [--platform <platform>] <reference>
 ```
 
 **Arguments**
@@ -405,7 +405,7 @@ container image pull [--debug] [--scheme <scheme>] [--disable-progress-updates] 
 **Options**
 
 *   `--scheme <scheme>`: Scheme to use when connecting to the container registry. One of (http, https, auto) (default: auto)
-*   `--disable-progress-updates`: Disable progress bar updates
+*   `--progress <type>`: Progress type (format: none|ansi) (default: ansi)
 *   `-a, --arch <arch>`: Limit the pull to the specified architecture
 *   `--os <os>`: Limit the pull to the specified OS
 *   `--platform <platform>`: Limit the pull to the specified platform (format: os/arch[/variant], takes precedence over --os and --arch)
@@ -417,7 +417,7 @@ Pushes an image to a registry. The flags mirror those for `image pull` with the 
 **Usage**
 
 ```bash
-container image push [--scheme <scheme>] [--disable-progress-updates] [--arch <arch>] [--os <os>] [--platform <platform>] [--debug] <reference>
+container image push [--scheme <scheme>] [--progress <type>] [--arch <arch>] [--os <os>] [--platform <platform>] [--debug] <reference>
 ```
 
 **Arguments**
@@ -427,7 +427,7 @@ container image push [--scheme <scheme>] [--disable-progress-updates] [--arch <a
 **Options**
 
 *   `--scheme <scheme>`: Scheme to use when connecting to the container registry. One of (http, https, auto) (default: auto)
-*   `--disable-progress-updates`: Disable progress bar updates
+*   `--progress <type>`: Progress type (format: none|ansi) (default: ansi)
 *   `-a, --arch <arch>`: Limit the push to the specified architecture
 *   `--os <os>`: Limit the push to the specified OS
 *   `--platform <platform>`: Limit the push to the specified platform (format: os/arch[/variant], takes precedence over --os and --arch)
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -204,11 +204,12 @@ public struct Flags {
     public struct Progress: ParsableArguments {
         public init() {}
 
-        public init(disableProgressUpdates: Bool) {
-            self.disableProgressUpdates = disableProgressUpdates
+        public enum ProgressType: String, ExpressibleByArgument {
+            case none
+            case ansi
         }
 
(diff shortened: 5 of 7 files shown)


Sources/ContainerCommands/Container/ContainerRun.swift
```diff
@@ -44,8 +44,8 @@ extension Application {
         @OptionGroup(title: "Registry options")
         var registryFlags: Flags.Registry
 
-        @OptionGroup(title: "Progress options")
-        var progressFlags: Flags.Progress
+        @Option(name: .long, help: ArgumentHelp("Progress type (format: none|plain)", valueName: "type"))
```
jglogan: Hi @saehejkang! Did you consider using an enum for this categorical value?

```swift
enum ProgressType: String, ExpressibleByArgument {
    case none
    case plain
}

@Option(name: .long, help: ArgumentHelp("Progress type (format: none|plain)", valueName: "type"))
var progress: ProgressType = .plain
```

jglogan (commented):
Let's look at the `ansi` aspect of this next week.

Right now the progress bar defaults to `ansi` style behavior (fine grained updates rewrite the same line over and over).

@dkovba would it be best to break this into two PRs?

- The first is this one, with `--progress none` and `--progress ansi` values in place of `--disable-progress-updates`.
- A follow-up PR adds support for `plain` progress to the `TerminalProgress` types.

Sources/ContainerCommands/Container/ContainerRun.swift
saehejkang: Yes, I considered using an `enum` but followed how the other `--progress` option was declared. I think an update to the other `--progress` option seems appropriate (found a small type in the command too)

EDIT: Made updates to use an `enum` in this case

saehejkang:
> Right now the progress bar defaults to ansi style behavior (fine grained updates rewrite the same line over and over).

So to be clear, if we want the progress bar to be disabled, the option would be `none`? Kind of like how it is set [here](https://github.com/apple/container/blob/f2b3cbdd73120e69d3aeec14d27b06a8745d18de/Sources/ContainerCommands/Image/ImagePull.swift#L84)?

however, the `else`, after that `if` statement is setting more options in the `ProgressConfig`. Would that be the `plain` or `ansi` option?

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -68,17 +69,20 @@ extension Application {
             let image = try await ClientImage.get(reference: reference)
 
             var progressConfig: ProgressConfig
-            if progressFlags.disableProgressUpdates {
-                progressConfig = try ProgressConfig(disableProgressUpdates: progressFlags.disableProgressUpdates)
-            } else {
+            switch self.progress {
+            case "none": progressConfig = try ProgressConfig(disableProgressUpdates: true)
+            case "plain":
```
dkovba: This switch repeats three times. Should we use a helper method for this? It can also be a method in the new `ProgressMode` enum.

Tests/CLITests/Subcommands/Build/CLIRunBase.swift
```diff
@@ -115,8 +115,10 @@ class TestCLIRunBase: CLITest {
             if Tty { arguments.append("-t") }
         }
 
-        if DisableProgressUpdates {
-            arguments.append("--disable-progress-updates")
+        if Progress == "none" {
+            arguments.append("--progress none")
+        } else {
+            arguments.append("--progress plain")
```
dkovba: This might be better:
```
arguments.append("--progress")
arguments.append(Progress)
```

Sources/ContainerCommands/Container/ContainerRun.swift
```diff
@@ -44,8 +44,8 @@ extension Application {
...
+        var progress: String = "plain"
```
dkovba: The default behavior when updates rewrite the same line over and over was proposed to be called "ansi". The word "plain" is reserved for printing certain progress updates in new lines.

dkovba:
> * The first is this one, with `--progress none` and `--progress ansi` values in place of `--disable-progress-updates`.
> * A follow-up PR adds support for `plain` progress to the `TerminalProgress` types.

Two PRs are fine. But the `plain` option shouldn't take much code to implement and could be done in the single PR. It might help to describe the expected behavior for each option in https://github.com/apple/container/issues/641.

jglogan:
@saehejkang @dkovba I've updated the #641 issue description for just `none` and `ansi`, and described other option values for follow-on PRs.

Sources/ContainerCommands/Image/ImagePush.swift
saehejkang: I don't agree that we should use a helper method for this. Correct me if I am wrong, but I am under the assumption that the values passed to the `ProgressConfig` differ for each command.  It makes no sense to have the helper method return the correct config because we would have to pass those values into the helper function? Seems redundant and more confusing than it needs to be.

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -30,8 +31,8 @@ extension Application {
         @OptionGroup
         var registry: Flags.Registry
 
-        @OptionGroup
-        var progressFlags: Flags.Progress
+        @Option(name: .long, help: ArgumentHelp("Progress type (format: none|plain)", valueName: "type"))
+        var progress: String = "plain"
```
dkovba: Should we use an enum to list supported options?

saehejkang: Pushing those updates now

jglogan:
@saehejkang That looks good. It looks like you'll need to do one more rebase onto main since some changes to the docs got merged today. After that we can get this merged.

jglogan (approved):

jglogan:
@saehejkang I went ahead and fixed that with the web editor, no need to put you through more of this rebase nonsense 😃 

@dkovba Please have a look now as it still requires your ✅

saehejkang:
> @saehejkang I went ahead and fixed that with the web editor, no need to put you through more of this rebase nonsense 😃 

I tend to click the merge with main button in PRs, so I am unsure if I should continue doing that? Is it just there should only be 1 merge with main commit (just don't click the button more than once)?

jglogan:
The commit history should just include your actual feature commit(s); there shouldn't be merge commits from `main` or any other branch in your PR's branch history.

If your PR branch conflicts with `main`, just fetch the latest `main` head from the upstream, merge or rebase it onto the `main` of your fork, and then rebase your PR branch against that.

dkovba:
After discussing the guidance for merge commits, we’ve decided that rebasing is not mandatory. We’ll review and accept PRs that include merge commits.

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -27,11 +28,16 @@ extension Application {
             abstract: "Push an image"
         )
 
+        enum ProgressType: String, ExpressibleByArgument {
+            case none
+            case ansi
+        }
+
```
dkovba: We should move `ProgressType` to a shared location (probably, `Sources/ContainerClient/Flags.swift`).

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -17,6 +17,7 @@
 import ArgumentParser
 import ContainerClient
 import Containerization
+import ContainerizationError
```
dkovba: Is this import used?

Tests/CLITests/Subcommands/Build/CLIRunBase.swift
```diff
@@ -42,8 +42,8 @@ class TestCLIRunBase: CLITest {
         nil
     }
 
-    var DisableProgressUpdates: Bool {
-        false
+    var Progress: String {
```
dkovba: From [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#general-conventions):
> Follow case conventions. Names of types and protocols are `UpperCamelCase`. Everything else is `lowerCamelCase`.

docs/command-reference.md
```diff
@@ -65,21 +65,20 @@ container run [<options>] <image> [<arguments> ...]
...
+  - `10.*.*.*`
+  - `192.168.*.*`
+  - `172.16.*.*` through `172.31.*.*`
+- The host ends with the machine's default container DNS domain (as defined in `DefaultsStore.Keys.defaultDNSDomain`, located [here](../Sources/ContainerPersistence/DefaultsStore.swift))
 
-        For internal/local registries, the client uses **HTTP**. Otherwise, it uses **HTTPS**.
+For internal/local registries, the client uses **HTTP**. Otherwise, it uses **HTTPS**.
 
-**Progress Options**
+**Progress options**
```
dkovba: "Progress Options" would be better for consistency with other command groups. This should also match `@OptionGroup()` in the code.

Sources/ContainerCommands/Container/ContainerRun.swift
```diff
@@ -44,8 +49,8 @@ extension Application {
         @OptionGroup(title: "Registry options")
         var registryFlags: Flags.Registry
 
-        @OptionGroup(title: "Progress options")
-        var progressFlags: Flags.Progress
+        @Option(name: .long, help: ArgumentHelp("Progress type (format: none|ansi)", valueName: "type"))
```
dkovba: @jglogan Should we keep `@OptionGroup(title: "Progress options")`?

dkovba (changes requested):

Sources/ContainerCommands/Container/ContainerRun.swift
jglogan: Sounds good to me - the advantage to the OptionGroup is it keeps the option name, value type, and help message DRY* across the different commands.

@saehejkang How about we restore the OptionGroup and Flags.Progress?

`* "don't repeat yourself"`

jglogan: And also, what Dmitry mentioned, we don't need three separate definitions of ProgressType. We should:

- Replace the nested ProgressType enums with a single definition in Flags.swift
- Use the following in each command:
    ```swift
    @OptionGroup(title: "Progress options")
    var progressFlags: Flags.Progress
    ````

saehejkang: > @saehejkang How about we restore the OptionGroup and Flags.Progress?

I actually pushed a commit a couple days back that restored that (it got lost with all my squashing and force pushing).
```swift
public struct Progress: ParsableArguments {
   public init() {}

   public enum ProgressType: String, ExpressibleByArgument {
      case none
      case ansi
   }
        
   @Option(name: .long, help: ArgumentHelp("Progress type (format: none|ansi)", valueName: "type"))
      public var progress: ProgressType = .ansi
   }
}
```
```swift
@OptionGroup(title: "Progress options")
var progress: Flags.Progress
```
```swift
switch self.progress.progress {
   case .none: progressConfig = try ProgressConfig(disableProgressUpdates: true)
   case .ansi:
      progressConfig = try ProgressConfig(
         showTasks: true,
         showItems: true,
         ignoreSmallSize: true,
         totalTasks: 2
      )
}
```

My qualms with this route is that there are other commands that have the `--progress` option, but they also have options like `tty` or `auto`. If we go the global flag route, should we not be consistent and use this flag EVERYWHERE else? The options in the `enum`, defaulted value, and argument help will be completely different to what is defined in the global flag file.

dkovba: There should be only two versions of the progress type enum. The new builder will likely be using the same progress bar and the same type. Having two for now is OK. CC @wlan0

saehejkang: Still kinda confused on this so I just pushed code for what I think we want

Tests/CLITests/Subcommands/Build/CLIRunBase.swift
saehejkang: are all the other variable declarations incorrect then 😅 ?

dkovba: Only those that are capitalized like type names (`UpperCamelCase`).

Sources/ContainerCommands/Container/ContainerRun.swift
dkovba: It seems `@Option(name: .long, help: ArgumentHelp("Progress type (format: none|ansi)", valueName: "type"))` should go into `Flags.Progress`.

Tests/CLITests/Subcommands/Build/CLIRunBase.swift
dkovba: Thanks for fixing the variable names! However, the change is incomplete and `test install-kernel integration` currently fails.

saehejkang: Just noticed the checks failed. I will get this fixed when I get a chance.

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -30,7 +30,7 @@ extension Application {
         @OptionGroup
         var registry: Flags.Registry
 
-        @OptionGroup
+        @OptionGroup(title: "Progress options")
```
dkovba: @jglogan Is the title not needed here?

jglogan: We only use the titles on `ContainerCreate`, `ContainerRun`, and `ContainerExec`; for all other commands we don't use title.

So yeah, I was a little overprescriptive here, sorry :(

https://github.com/apple/container/pull/808#discussion_r2479502776

To maintain this consistency, could you remove the option group title for ImagePush and ImagePull?

Looks like there are some test compile issues that cropped up but once these two things are done we should be good to go. Whew!

saehejkang:
I still have a question about how this option below will be updated (looks like the only other place we pass the flag in the BuildCommand).

```swift
@Option(name: .long, help: ArgumentHelp("Progress type (format: auto|plain|tty)]", valueName: "type"))
var progress: String = "auto"
```

Do we want this move to the global flags file? Would it be defined in the `Progress` struct? If so, would it not have to be named something different than `progress`? What would we want to name the `enum`?

dkovba:
At this moment, the builder is using a separate progress implementation. It can be different from the progress bar in the rest of the project.

dkovba (approved):
👍

jglogan:
Closes #641.

PR #813, opened 2025-10-27
fix: connect to backend immediately on channel active for server-first protocols(#794)

### LINKED ISSUE #794: [Bug]: TCP Port Forwarding Fails: Connection succeeds but data stream times out
### I have done the following

### Steps to reproduce

The following steps use a public mail server container to reliably reproduce the issue.

1.  On macOS 26, run the `marlonb/mailcrab:latest` container and map the SMTP port:
    ```sh
    container run --rm -p 1025:1025 marlonb/mailcrab:latest
    ```
2.  From the host terminal, confirm that the port is open and the TCP handshake succeeds using `nc`. **This step will report instant success.**
    ```sh
    nc -vz -4 localhost 1025
    ```
    *Successful Output:* `Connection to localhost port 1025 [tcp/blackjack] succeeded!`

3.  Next, attempt to communicate with the service using a protocol-aware client like `curl`. **This step will hang for 2 minutes and fail with a timeout.**
    ```sh
    # This command attempts to send a basic email via SMTP
    curl -v smtp://127.0.0.1:1025 --mail-from "test@example.com" --mail-rcpt "test@example.com" --upload-file /dev/null
    ```

### Current behavior

The `curl` command successfully establishes the TCP connection but then times out waiting for the initial server response, which never arrives. The connection is held open for 2 minutes before failing.

The exact verbose output is:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1:1025...
* Connected to 127.0.0.1 (127.0.0.1) port 1025
  0     0    0     0    0     0      0      0 --:--:--  0:01:59 --:--:--     0* server response timeout
  0     0    0     0    0     0      0      0 --:--:--  0:02:00 --:--:--     0
* Closing connection
curl: (28) server response timeout
```
This demonstrates that the initial connection is accepted (`* Connected to ...`), but the data proxy is failing, causing the timeout.

### Expected behavior

The `curl` command should successfully connect and communicate with the SMTP server inside the container, receiving a `220 mailcrab` welcome banner and subsequent protocol messages, and then exit cleanly.

### Environment

```markdown
- OS: 26.0.1 
- Container: 0.5.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Ronitsabhaya75:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
closes #794 

TCP port forwarding currently fails for server-first protocols (SMTP, FTP, SSH, PostgreSQL) because the backend connection is only established when the client sends data. 

**Current behavior in `ConnectHandler.swift`:**
```swift
func channelRead(context: ChannelHandlerContext, data: NIOAny) {
    if self.pendingBytes.isEmpty {
        self.connectToServer(context: context)  // Only connects on first data!
    }
}
```

This breaks any protocol where the server speaks first.

**The fix:**
Connect to the backend immediately when the channel becomes active, not when data arrives. This allows bidirectional communication from the start.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/SocketForwarder/ConnectHandler.swift
+++ b/Sources/SocketForwarder/ConnectHandler.swift
@@ -35,9 +35,6 @@ extension ConnectHandler: ChannelInboundHandler {
     typealias OutboundOut = ByteBuffer
 
     func channelRead(context: ChannelHandlerContext, data: NIOAny) {
-        if self.pendingBytes.isEmpty {
-            self.connectToServer(context: context)
-        }
         self.pendingBytes.append(data)
     }
 
@@ -45,6 +42,12 @@ extension ConnectHandler: ChannelInboundHandler {
         // Add logger metadata.
         self.log?[metadataKey: "proxy"] = "\(context.channel.localAddress?.description ?? "none")"
         self.log?[metadataKey: "server"] = "\(context.channel.remoteAddress?.description ?? "none")"
+    }
+
+    func channelActive(context: ChannelHandlerContext) {
+        self.log?.trace("frontend - channel active, connecting to backend")
+        self.connectToServer(context: context)
+        context.fireChannelActive()
     }
 }
 


liufeng-xiwo:
The problem also breaks connecting mysql service with port forwarding.

Ronitsabhaya75:
> The problem also breaks connecting mysql service with port forwarding.

can you go bit in detail please.

jglogan:
@Ronitsabhaya75 Thanks for the simplification+fix! Let's get this into the release.

@liufeng-xiwo I don't know whether I've seen this issue with mysql. If you still have a problem there, feel free to create another issue and provide the steps that would allow us to reproduce the problem you see.

jglogan (approved):

PR #827, opened 2025-10-31
[container-build]: Support inline Dockerfile from stdin

### LINKED ISSUE #727: [Request]: Support inline Dockerfile from stdin for container build

In some very common automation cases it would be great to support inlined Dockerfiles for container build.
For example:
```
$ echo -e "FROM alpine\nRUN echo 'Hello World'" | container build -t myimage -f - .
Error: Dockerfile does not exist at path: -
Usage: container build [<options>] [<context-dir>]
  See 'container build --help' for more information.
```

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #727

Tested variations of commands with `stdin` below

```
echo -e "FROM alpine\nRUN echo 'Hello World'" | container build -t myimage -f - .
```

```
container build -t myimage -f - . <<'EOF'
FROM alpine
RUN echo 'Hello World'
EOF
```

```
echo -e "FROM alpine\nRUN echo 'Hello World'" > Dockerfile.tmp
container build -t myimage -f Dockerfile.tmp .
rm Dockerfile.tmp
```

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/BuildCommand.swift
+++ b/Sources/ContainerCommands/BuildCommand.swift
@@ -208,18 +208,18 @@ extension Application {
                 }
 
                 let buildFileData: Data
-                let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("Dockerfile-\(UUID().uuidString)")
-                defer {
-                    try? FileManager.default.removeItem(at: tempFile)
-                }
-
                 // Dockerfile should be read from stdin
                 if file == "-" {
+                    let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("Dockerfile-\(UUID().uuidString)")
+                    defer {
+                        try? FileManager.default.removeItem(at: tempFile)
+                    }
+
                     guard FileManager.default.createFile(atPath: tempFile.path(), contents: nil) else {
                         throw ContainerizationError(.internalError, message: "unable to create temporary file")
                     }
 
-                    guard let outputHandle = try? FileHandle(forWritingTo: tempFile) else {
+                    guard let fileHandle = try? FileHandle(forWritingTo: tempFile) else {
                         throw ContainerizationError(.internalError, message: "unable to open temporary file for writing")
                     }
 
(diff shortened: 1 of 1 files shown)


Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -207,7 +207,33 @@ extension Application {
                     buildFilePath = resolvedPath
                 }
 
-                let buildFileData = try Data(contentsOf: URL(filePath: buildFilePath))
+                let buildFileData: Data
```
jglogan, line 210: Looks good to me.

Could you add a test for this? Just copy one of the simple tests in CLIBuilderTest and supply the Dockerfile via stdin instead of a file.

I think this will require creating a separate helper function to  https://github.com/apple/container/blob/main/Tests/CLITests/Subcommands/Build/CLIBuildBase.swift#L112, since there's nothing in that test support stuff that knows how to pipe stdin into `container build`.

Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -207,7 +207,33 @@ extension Application {
...
+                let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("TempDockerfile")
```
jglogan: FYI I finally got back to your #734 PR which is similar!

As with that one, do something like:

```swift
let tempFile = FileManager.default.temporaryDirectory.appendingPathComponent("Dockerfile-\(UUID().uuidString)")
```

Sources/ContainerCommands/BuildCommand.swift
saehejkang: We need to make changes in the `run` function of the `CLITest` file as well. There needs to be a way to add the data to the `stdin` pipe.

PR #832, opened 2025-11-03
[images]: reference image by their short digest value

### LINKED ISSUE #104: Add support for referencing images via their 'short sha'
Today if I have an image that is only referenced via a sha - its not easy for me to delete this image. For example

```
container i pull alpine@sha256:8a1f59ffb675680d47db6337b49d22281a139e9d709335b492be023728e11715

container i ls
NAME    TAG     DIGEST
alpine  <none>  8a1f59ffb675680d47db6337...

container i rm  alpine:8a1f59ffb675680d47db6337
Error: internalError: "failed to delete one or more images: ["alpine:8a1f59ffb675680d47db6337"]"
```

It is only if I use the `-q` flag with `image ls` is the full sha returned which can be used to perform the delete operation.
```
container i ls -q
alpine@sha256:8a1f59ffb675680d47db6337b49d22281a139e9d709335b492be023728e11715

container i rm  alpine@sha256:8a1f59ffb675680d47db6337b49d22281a139e9d709335b492be023728e11715
docker.io/library/alpine@sha256:8a1f59ffb675680d47db6337b49d22281a139e9d709335b492be023728e11715
Reclaimed 650.8 MB in disk space
```

It would make it easier we could use the short sha which is displayed during the output of `image ls` to refer to an image

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #104 

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Image/ImageDelete.swift
+++ b/Sources/ContainerCommands/Image/ImageDelete.swift
@@ -51,7 +51,13 @@ extension Application {
                     let notFound: [String] = []
                     return (found, notFound)
                 }
-                return try await ClientImage.get(names: options.images)
+
+                let (found, notFound) = try await ClientImage.get(names: options.images)
+
+                if !notFound.isEmpty {
+                    return try await ClientImage.getByPrefix(names: options.images)
+                }
+                return (found, notFound)
             }()
             var failures: [String] = notFound
             var didDeleteAnyImage = false
--- a/Sources/ContainerCommands/Image/ImageInspect.swift
+++ b/Sources/ContainerCommands/Image/ImageInspect.swift
@@ -36,7 +36,10 @@ extension Application {
 
         public func run() async throws {
             var printable = [any Codable]()
-            let result = try await ClientImage.get(names: images)
+            var result = try await ClientImage.get(names: images)
+            if !result.error.isEmpty {
+                result = try await ClientImage.getByPrefix(names: images)
+            }
             let notFound = result.error
             for image in result.images {
                 guard !Utility.isInfraImage(name: image.reference) else {
--- a/Sources/ContainerCommands/Image/ImageList.swift
+++ b/Sources/ContainerCommands/Image/ImageList.swift
@@ -41,11 +41,11 @@ extension Application {
 
     struct ListImageImplementation {
         static func createHeader() -> [[String]] {
-            [["NAME", "TAG", "DIGEST"]]
+            [["NAME", "TAG", "INDEX PREFIX"]]
         }
 
         static func createVerboseHeader() -> [[String]] {
-            [["NAME", "TAG", "INDEX DIGEST", "OS", "ARCH", "VARIANT", "SIZE", "CREATED", "MANIFEST DIGEST"]]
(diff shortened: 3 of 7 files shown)


jglogan:
@saehejkang What do you think about just prefix matching the image index, more like what Docker does. At the same time we could clean up the tablular output for the commands:

- When a user supplies an image ID to one of the image commands, the command first tries to match the ID as an image ref. If this fails, it tries to prefix match the ID against all image index hashes. A unique match selects the image, no match is treated as not found, just as today, and a multiple match is treated just as an image not found, but with a different error message.
- In the tabular presentation, we change the `DIGEST` (non-verbose) and `INDEX DIGEST` (verbose mode) fields to just be `INDEX PREFIX` for both modes, output the 12 digit prefix value, and drop the trailing ellipsis. For `MANIFEST DIGEST` (verbose), we change to `MANIFEST PREFIX`, output the 12 digit prefix and remove the ellipsis.

ContainerClient could implement a utility function to facilitate the prefix matching mechanism, but it's the command itself that should implement the policy, rather than building that fuzzy matching into the client calls. Does that make sense?

saehejkang:
> What do you think about just prefix matching the image index, more like what Docker does.

I will come back to this PR and evaluate this in the near future.

saehejkang:
> ContainerClient could implement a utility function to facilitate the prefix matching mechanism, but it's the command itself that should implement the policy, rather than building that fuzzy matching into the client calls.

Diving more into this, I think it makes the most sense to search for the image based on the `prefix` in both of the `ClientImage.get` functions. It allows for a single `get` call and a user can query for multiple images to `inspect/delete/etc` with both the ID, as well as the prefix if they would like.

> When a user supplies an image ID to one of the image commands, the command first tries to match the ID as an image ref. If this fails, it tries to prefix match the ID against all image index hashes.

Same idea as you stated above ^.

saehejkang:
@jglogan This is not ready for review yet 100% because I am still thinking about how error propagation should be done. I think the implementation that is in place for a command like `image inspect` should be reworked. 

For example, the `ClientImage.get` function returns an array of `strings` that should contain the "errors". However, this array will always only contain the name of the image itself that was not found. Finally, the `image inspect` command only checks if the array actually contains values (could be anything really).

```swift
if notFound.count > 0 {
   throw ContainerizationError(.notFound, message: "images: \(notFound.joined(separator: "\n"))")
}
```

This is under the assumption that EVERY single error for this command will always be something like `image not found`. 
****

When working on adding new errors like `digest not found` or `ambiguous digest prefix matches multiple images`, it got me 🤔 . I don't see the need to return an array of "errors", if they are simply `strings` with no real value. 

Some of the functions in the `ClientImage` are simply throwing the error as soon as it arises, so why don't we go that route? Or maybe we can actually add `ContainerizationErrors` to the array and then have the command read and throw them? Just thought I could maybe get more info on the design pattern and if starting a discussion here is worthwhile.

PR #839, opened 2025-11-03
Automatically clean up anonymous volumes for containers   with --rm flag

realrajaryan:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
Adds automatic cleanup of anonymous volumes when containers started with the `--rm` flag exit, preventing orphaned volumes from accumulating and consuming disk space.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/Volume.swift
+++ b/Sources/ContainerClient/Core/Volume.swift
@@ -62,9 +62,17 @@ extension Volume {
     /// Reserved label key for marking anonymous volumes
     public static let anonymousLabel = "com.apple.container.resource.anonymous"
 
+    /// Reserved label key for storing the container ID that created this volume
+    public static let createdByLabel = "com.apple.container.resource.created-by"
+
     /// Whether this is an anonymous volume (detected via label)
     public var isAnonymous: Bool {
         labels[Self.anonymousLabel] != nil
+    }
+
+    /// The container ID that created this volume
+    public var createdByContainerID: String? {
+        labels[Self.createdByLabel]
     }
 }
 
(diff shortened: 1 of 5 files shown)


Sources/Services/ContainerAPIService/Containers/ContainersService.swift
```diff
@@ -434,12 +434,17 @@ public actor ContainersService {
     }
 
     private func handleContainerExit(id: String, code: ExitStatus? = nil) async throws {
-        try await self.lock.withLock { [self] context in
+        let shouldCleanupVolumes = try await self.lock.withLock { [self] context in
```
dcantah, line 437: Curious why we don't just do the cleanup in handleContainerExit itself

realrajaryan: Once the lock is released if autoRemove was true, the container bundle has already been deleted by cleanup(), so we can't call getContainerCreationOptions() again to check the flag, the file is gone. So we need to capture the decision while the container still exists (inside the lock) then act on it outside the lock

dcantah: What I mean is if cleanup() is called we know that autoRemove was supplied, so why not just do it in there?

Sources/ContainerClient/Core/Volume.swift
```diff
@@ -62,10 +62,18 @@ extension Volume {
     /// Reserved label key for marking anonymous volumes
     public static let anonymousLabel = "com.apple.container.resource.anonymous"
 
+    /// Reserved label key for storing the container ID that created this volume
+    public static let createdByLabel = "com.apple.container.resource.created-by"
```
jglogan, line 66: My concern with this approach is that the ID of a container isn't really a unique ID across all time. In this case we might be able to get by with it but in general if you store the ID `test`, that could be one container on one day and a totally different container a week later.

Would I make more sense for us, when a container is removed in the API server, to quickly scan all containers for volume references, find all unreferenced volumes, and then prune those that are marked anonymous? Given what you have in `cleanupAnonymousVolumes()` this might be a simple change?

PR #846, opened 2025-11-04
Add --rosetta option for arm64 images

### LINKED ISSUE #391: [Request]: Support Enabling Rosetta on arm64 images

This is supported by Docker and Orbstack when Rosetta is enabled. Images like Ubuntu/Debian support adding and running across different architectures side by side. So fully emulating an x86-64 image seems to get some performance penalty and is redundant in this case since we have a better option to run a native arm64 kernel where x86-64 binaries can be run. Since containers already support Rosetta, letting Rosetta be exposed and available to it in the container will be relatively easy to achieve. 

### Code of Conduct

rgov:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
`container` automatically enables Rosetta for amd64 containers. This change allows Rosetta to be enabled for arm64 containers by passing the `--rosetta` flag. This allows native containers to benefit from being able to execute the occasional amd64 executable in niche situations, e.g., containers used for cross-compilation.

Resolves #391.

I considered adding a warning for passing `--rosetta` when it is not necessary, but there wasn't obvious infrastructure for conveying the message.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

I tested by downloading a statically linked x86 binary and running it inside an arm64 Alpine container created with the new `--rosetta` flag.

I attempted to write a test case but I hit [this issue](https://github.com/apple/container/discussions/700) with my Swift toolchain. An ideal test case might involve actually executing an amd64 binary within an arm64 container, but I didn't come up with a simple way to do so.

### DIFF
--- a/Sources/ContainerClient/Flags.swift
+++ b/Sources/ContainerClient/Flags.swift
@@ -184,6 +184,9 @@ public struct Flags {
         @Flag(name: [.customLong("rm"), .long], help: "Remove the container after it stops")
         public var remove = false
 
+        @Flag(name: .long, help: "Enable Rosetta in the container")
+        public var rosetta = false
+
         @Flag(name: .long, help: "Forward SSH agent socket to container")
         public var ssh = false
 
--- a/Sources/ContainerClient/Utility.swift
+++ b/Sources/ContainerClient/Utility.swift
@@ -213,8 +213,10 @@ public struct Utility {
             )
         }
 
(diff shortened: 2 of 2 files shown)


flukejones:
Thank you this. It is incredibly helpful for scenarios where I use buildroot and yocto at work.

dcantah:
Going to check this in. We can always revert if we find an issue or anyone has concerns on the ux.

PR #849, opened 2025-11-04
Reuse XPC connections across container operations

### LINKED ISSUE #666: [Request]: ClientContainer should not create a new xpc connection per request

Just wasteful, but the reason it does is because ClientContainer is somewhat overloaded in that it's used for both data (during list(), you get given a ClientContainer) and to talk to the daemon. It should only be used for the latter, and list/get should return ContainerSnapshot's.

### Code of Conduct

realrajaryan:
- Closes #666.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Currently, `ClientContainer` is overloaded, it's used for both data (during list(), you get given a ClientContainer) and to talk to the daemon. Each container operation creates a new XPC connection, which is pretty wasteful e.g. `container stop --all` with 10 containers creates 11 XPC connections (1 for list and 10 for individual stops)

To address this `list()`, `get()`, and `create()` will now return `ContainerSnapshot` instead of `ClientContainer`, and `ClientContainer` now accepts an optional `XPCClient` parameter for connection reuse so multiple operations can share the same connection. Commands that perform bulk operations now create a single shared `XPCClient` and pass it to all `ClientContainer` instances

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/ClientContainer.swift
+++ b/Sources/ContainerClient/Core/ClientContainer.swift
@@ -23,7 +23,7 @@ import Foundation
 import TerminalProgress
 
 public struct ClientContainer: Sendable, Codable {
-    static let serviceIdentifier = "com.apple.container.apiserver"
+    public static let serviceIdentifier = "com.apple.container.apiserver"
 
     /// Identifier of the container.
     public var id: String {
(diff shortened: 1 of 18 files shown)


Sources/ContainerClient/Core/ClientContainer.swift
```diff
@@ -43,16 +43,46 @@ public struct ClientContainer: Sendable, Codable {
...
+
     package init(configuration: ContainerConfiguration) {
         self.configuration = configuration
         self.status = .stopped
         self.networks = []
+        self.xpcClient = nil
     }
 
-    init(snapshot: ContainerSnapshot) {
+    public init(snapshot: ContainerSnapshot, xpcClient: XPCClient? = nil) {
```
jglogan, line 56: Having the caller provide a snapshot to a public method feels a little strange. I'd expect them to be able to create a ClientContainer using just a container ID.

Since we're now using snapshot as the return type for everything, do we really need to store configuration, status, and networks between calls?

dcantah:
Closed via https://github.com/apple/container/pull/1139

PR #851, opened 2025-11-05
Implement container stats

### LINKED ISSUE #824: [Request]: "container stats" to show container usage statistics

ContainerStatistics API has been added in https://github.com/apple/containerization/pull/318

Viewing CPU/memory usage in Activity Monitor is not very clear, as all containers are showing up as `Virtual Machine Service for container-runtime-linux`; and only via Open File Inspector it is possible to understand the containers name.

### Code of Conduct

dcantah:
Closes #824

This implements statistics gathering across the various components, but ultimately this is for implementing a new CLI command: `container stats`. This shows memory usage, cpu usage, network and block i/o and the number of processes in the container. The new command can inspect stats for 1-N containers and by default continuously updates in a `top` like stream.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerStats.swift
+++ b/Sources/ContainerCommands/Container/ContainerStats.swift
@@ -1,0 +1,262 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerClient
+import ContainerizationError
+import ContainerizationExtras
+import Foundation
+
+extension Application {
+    public struct ContainerStats: AsyncParsableCommand {
+        public static let configuration = CommandConfiguration(
+            commandName: "stats",
+            abstract: "Display resource usage statistics for containers")
+
+        @Argument(help: "Container ID or name (optional, shows all running containers if not specified)")
+        var containers: [String] = []
+
+        @Option(name: .long, help: "Format of the output")
+        var format: ListFormat = .table
+
+        @Flag(name: .long, help: "Disable streaming stats and only pull the first result")
+        var noStream = false
+
+        @OptionGroup
+        var global: Flags.Global
+
+        public init() {}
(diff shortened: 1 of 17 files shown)


Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -257,6 +257,41 @@ public actor SandboxService {
...
+        return try await self.lock.withLock { lock in
+            let containerInfo = try await self.getContainer()
+            let stats = try await containerInfo.container.statistics()
+
+            let containerStats = ContainerStats(
+                id: stats.id,
+                memoryUsageBytes: stats.memory.usageBytes,
+                memoryLimitBytes: stats.memory.limitBytes,
+                cpuUsageUsec: stats.cpu.usageUsec,
+                networkRxBytes: stats.networks.reduce(0) { $0 + $1.receivedBytes },
```
jglogan, line 281: Would it be more useful to only reduce these at the command line and make all of these fields arrays? If we did can the caller correlate the entries readily with the config?

Sources/ContainerCommands/Container/ContainerStats.swift
```diff
@@ -0,0 +1,262 @@
...
+            commandName: "stats",
+            abstract: "Display resource usage statistics for containers")
+
+        @Argument(help: "Container ID or name (optional, shows all running containers if not specified)")
+        var containers: [String] = []
+
+        @Option(name: .long, help: "Format of the output")
+        var format: ListFormat = .table
+
+        @Flag(name: .long, help: "Disable streaming stats and only pull the first result")
```
jglogan, line 35: How about a `stream` argument that's an enumeration of `follow` and `top` (which might not be 100% sensical now but becomes more so if we add arg-based or interactive sort).

Sources/ContainerCommands/Container/ContainerStats.swift
```diff
@@ -0,0 +1,262 @@
...
+        var noStream = false
+
+        @OptionGroup
+        var global: Flags.Global
+
+        public init() {}
+
+        public func run() async throws {
+            if format == .json || noStream {
```
jglogan, line 44: `json` + `top` doesn't make sense, but `json` + `follow` would give you JSONL/NDJSON output.

Sources/Services/ContainerSandboxService/SandboxService.swift
dcantah: I'm inclined to keep it for now, but we can always adjust as needed

Sources/ContainerCommands/Container/ContainerStats.swift
dcantah: Fwiw, I'd like to keep the current set as I have a bit grander plans for the UX eventually (a more fleshed out TUI experience). Currently this mimics the stats view that most other container engines offer, mostly as it's familiar.

jglogan (approved):

PR #870, opened 2025-11-12
Add createdAt and metadata to container, network, and volume resources

solventak:
Closes #791 , #665

## Type of Change
- [x] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Add createdAt metadata (generic metadata struct) for each Container, Volume, and Network as well as output in `ls` commands to display it.

Also fixes a bug when spawning `networkvmnet.service` where instead of initializing the network configuration from the stored information of `NetworkConfiguration` entity it would create a new configuration based on key metadata that was passed in through args.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/ClientContainer.swift
+++ b/Sources/ContainerClient/Core/ClientContainer.swift
@@ -22,7 +22,6 @@ import ContainerizationOCI
 import Foundation
 import TerminalProgress
 
-/// A client representation of a container managed by the container service.
 public struct ClientContainer: Sendable, Codable {
     static let serviceIdentifier = "com.apple.container.apiserver"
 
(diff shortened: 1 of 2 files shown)


solventak:
@jglogan for your consideration :)

jglogan:
@solventak Sorry, this one fell through the cracks an I didn't get around to reviewing it.

If you're still interested in moving this one forward, please take a look at #1020 as it gives us a more sustainable base for your work.

I'm thinking that our resource types (container, image, volume, network, and possibly others later; for example kernels) in ContainerResource could conform to an Identifiable-conforming ContainerResource protocol along the lines of what you have here. That protocol could have a `metadata` property that has `createdAt` and `labels`.

Let us know here if you're interested in tackling this.

PR #875, opened 2025-11-13
support partial matches for object IDs in commands

### LINKED ISSUE #758: [Request]: support for partial containers IDs in the CLI

Would be nice the CLI to be able to determine target container by partial ID like:

```
$ container stop A
```

for a container with id BBBA-123123...

or at least by prefix ABBB-....

### Code of Conduct

solventak:
Closes #758

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

In order to speed up control in the command line, partial ID matches are now supported for commands that take in IDs as an argument.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/ClientContainer.swift
+++ b/Sources/ContainerClient/Core/ClientContainer.swift
@@ -54,26 +54,6 @@ public struct ClientContainer: Sendable, Codable {
         self.status = snapshot.status
         self.networks = snapshot.networks
     }
-
-    public static func search(searches: [String]) async throws -> [ClientContainer] {
-        var containerIdsToStop = Set<String>()
-
-        for partialContainerId in searches {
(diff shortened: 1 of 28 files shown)


solventak:
Right now just containers are supported for multi-id argument commands:

```
container kill
container inspect
container delete
container stop
```

but I wanted to get some feedback before working on the rest of the commands (there's a lot of them)!

dcantah:
I think this feature should be in the APIServer (ContainerService) and not client logic. The server knows the IDs/what exists and can just get sent the shortened prefix and try and match against it. That way all the logic is in one spot and not dup'd across all of the CLI commands that deal with IDs.

solventak:
Great point! Thanks for the callout.

burtonrodman:
running v1.1.0...
`container rm <id>` not supporting partial id

Issue #425, opened 2025-08-02
[Bug]: .dockerignore is not respected during container build


samurai00:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1.  Create a project directory, for example, a Rust project which generates a large `target` directory.
2.  Create a `.dockerignore` file in the root of the project with the following content to exclude the build artifacts and IDE folders:
    ```
    target
    .idea
    .vscode
    ```
3.  Have a directory structure similar to this, where `target` is large (e.g., >400MB):
    ```shell
    $ dua
       4.10 KB .dockerignore
       4.10 KB .gitignore
       4.10 KB Cargo.toml
       4.10 KB Dockerfile
       ...
     409.96 MB target
     411.48 MB total
    ```
4.  Create a `Dockerfile` that uses a `COPY . .` instruction to copy the project source:
    ```dockerfile
    # Dockerfile
    FROM rust:alpine as builder
    WORKDIR /opt/src/my-project
    # This step is expected to exclude files from .dockerignore
    COPY . .
    RUN cargo build --release
    ...
    ```
5.  Run the build command:
    ```shell
    container build --tag my-app --file Dockerfile .
    ```

### Current behavior

The build process includes files and directories that should be ignored, resulting in an unnecessarily large build context.

In one instance, the build log shows a context size of over 400MB, even though the source code is less than 1MB and the `target` directory is in `.dockerignore`:
```
[internal] load build context                                                                                                3.8s
=> => transferring context: 405.05MB                                                                                            3.8s
```

The problem is more severe with very large ignored directories. For example, with a `target` directory of ~6.8 GB, the build fails immediately during the context creation phase, before any `Dockerfile` instructions are executed.

Project structure:
```shell
$ dua
   ...
  41.68 MB .git
   6.81 GB target
   6.85 GB total
```

Build command and error:
```shell
$ container build --tag zexus-axum-lite --file Dockerfile-axum-lite .
[+] Building 7.1s (2/3)
 => [resolver] fetching image...docker.io/library/rust:slim                                                                      0.0s
 => [resolver] fetching image...docker.io/library/debian:bookworm-slim                                                           0.0s
Error: failed to create entry
```

The error `failed to create entry` suggests that the tool cannot handle the massive build context, a problem that the `.dockerignore` file is specifically designed to prevent.

### Expected behavior

[...]

PR #879, opened 2025-11-14
Read .dockerignore files

torarvid:
## Type of Change
- [x] Bug fix for issue #425 
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
As the included test demonstrates, when a folder contains a .dockerignore file with pattern `to-be-ignored.txt`, the file `to-be-ignored.txt` is still included in the build context. Check out the first commit to see this test failing.

## Testing
- [ ] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/ContainerBuildTests/IgnoreSpecTests.swift
+++ b/Tests/ContainerBuildTests/IgnoreSpecTests.swift
@@ -145,6 +145,64 @@ enum TestError: Error {
         #expect(try !spec.shouldIgnore(relPath: "src/build", isDirectory: true))
     }
 
+    @Test func testIgnoreSpecHandlesNegationPatterns() throws {
+        let content = """
+            *.log
+            !important.log
+            """
+        let data = content.data(using: .utf8)!
+        let spec = IgnoreSpec(data)
+
+        // *.log should match all .log files
+        #expect(try spec.shouldIgnore(relPath: "debug.log", isDirectory: false))
+        #expect(try spec.shouldIgnore(relPath: "error.log", isDirectory: false))
+
+        // But !important.log should negate and include important.log
+        #expect(try !spec.shouldIgnore(relPath: "important.log", isDirectory: false))
+    }
+
+    @Test func testIgnoreSpecHandlesNegationWithDirectories() throws {
+        let content = """
+            foo
+            !foo/bar.txt
+            """
+        let data = content.data(using: .utf8)!
+        let spec = IgnoreSpec(data)
+
+        // foo should match the directory and its contents
+        #expect(try spec.shouldIgnore(relPath: "foo", isDirectory: true))
+        #expect(try spec.shouldIgnore(relPath: "foo/other.txt", isDirectory: false))
+
+        // But !foo/bar.txt should negate and include that specific file
+        #expect(try !spec.shouldIgnore(relPath: "foo/bar.txt", isDirectory: false))
+    }
+
+    @Test func testIgnoreSpecHandlesComplexNegation() throws {
+        let content = """
+            # Ignore all markdown files
+            **/*.md
+            # Except README files
+            !README.md
+            !**/README.md
+            # But ignore README in temp directories
+            temp/**/README.md
+            """
+        let data = content.data(using: .utf8)!
+        let spec = IgnoreSpec(data)
+
+        // *.md should match all markdown files
+        #expect(try spec.shouldIgnore(relPath: "notes.md", isDirectory: false))
+        #expect(try spec.shouldIgnore(relPath: "docs/guide.md", isDirectory: false))
+
+        // But !README.md should include README files
+        #expect(try !spec.shouldIgnore(relPath: "README.md", isDirectory: false))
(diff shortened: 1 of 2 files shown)


wlan0:
Thanks for the PR! 

High‑level, I like the approach:
	•	read `.dockerignore` once per walk call
	•	filter the `followPathsWalked` entries up front
	•	reuse the same filtered set both for the JSON response and for the tar writer

A few details I’d like us to tighten up:

 - We don’t currently support `!` negation patterns. In Docker’s `.dockerignore`, `!foo` will re‑include paths that were previously excluded. Right now a line starting with `!` is treated as a literal pattern, so this will behave differently from Docker.
 - Handling of `**` and leading `/` isn’t quite the same as Docker’s pattern matcher. In particular, patterns like `**/foo` are treated as two `*` globs rather than the “any number of directories” semantics that Docker uses, and leading `/ `on a pattern won’t match a relative path under the context root.

It would be good to extend the matcher now so it lines up more closely with Docker’s docs or atleast call it out in the docs

Sources/ContainerBuild/BuildFSSync.swift
```diff
@@ -136,6 +136,9 @@ actor BuildFSSync: BuildPipelineHandler {
         var entries: [String: Set<DirEntry>] = [:]
         let followPaths: [String] = packet.followPaths() ?? []
 
+        // Parse .dockerignore if present
+        let ignorePatterns = try parseDockerignore()
```
jglogan: resolve the path to the ignore file here and load the file into Data, and then create an IgnoreSpec(data)

Sources/ContainerBuild/BuildFSSync.swift
```diff
@@ -146,6 +149,12 @@ actor BuildFSSync: BuildPipelineHandler {
             }
 
             let relPath = try url.relativeChildPath(to: contextDir)
+
+            // Check if the file should be ignored
+            if try shouldIgnore(relPath, patterns: ignorePatterns, isDirectory: url.hasDirectoryPath) {
```
jglogan: this can become `ignoreSpec.shouldIgnore(relPath:isDirectory:)`?

Sources/ContainerBuild/BuildFSSync.swift
```diff
@@ -278,6 +287,62 @@ actor BuildFSSync: BuildPipelineHandler {
         return Array(globber.results)
     }
 
+    /// Parse .dockerignore file and return list of ignore patterns
```
jglogan: Extract these two methods into IgnoreSpec.swift.

Tests/ContainerBuildTests/IgnoreSpecTests.swift
```diff
@@ -0,0 +1,209 @@
+//===----------------------------------------------------------------------===//
```
jglogan, line 1: Rename to IgnoreSpecTests.swift

jglogan (changes requested):
Can you extract the two private functions you added to a dedicated type?

- Keeps BuildFSSync more concise and easier to comprehend
- Improves testability and reusablity.

EDIT: The specific comments I made in my review might not apply exactly after you rework for the changes Sid mentioned above, but it'd still be better to keep the ignore testing separate from the BuildFSSync implementation.

torarvid:
@wlan0 @jglogan Hello 👋 I have made an attempt now at improving the PR based on your comments. I decided to just go a regex route and move away from the Globber (didn't quite understand whether that's used for other stuff and so don't understand if it's fine to just change how it works).

Also added more tests for ** and ! patterns.

Let me know what you think. (Sorry if my response time is a bit slow, I don't have that much free time these days😊)

torarvid:
Hmmm. I'm trying this on a real project, and the problem now is that it's very slow (my project has some big-but-supposed-to-be-ignored-folders, like `node_modules`, `.mypy_cache` etc.

I guess this problem might be a bit bigger? Like, the way it is now it will still walk over all files inside of `node_modules/**/*` and test for ignoring, which makes little sense unless you were to have `!node_modules/important.txt` further down in .dockerignore 🤔

jglogan:
Have you considered looking at how it works in moby/patternmatcher and the underlying filepath.Match() implementation in the Go library?

polRk:
I strongly recommend against using "docker" in the file name; it's better to only support "Containerignore." Docker is a third-party, proprietary company, and supporting an open source tool for running custom logic is very bad. Recommended support for the .containerignore file only

aaronjsutton:
> Hmmm. I'm trying this on a real project, and the problem now is that it's very slow (my project has some big-but-supposed-to-be-ignored-folders, like `node_modules`, `.mypy_cache` etc.

FWIW: I built against this PR and also experience the performance drop in a non-trivial project. Beyond that, everything appears to work as expected. If I find some time this weekend, I'll try to dive a bit deeper into that.

jglogan:
apple/container-builder-shim#63 implements this in the builder shim.

Opened #1169 for Dockerfile-specific ignore files.

PR #899, opened 2025-11-19
feat: color and plain modes for --progress flag

PhantomInTheWire:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
relevant issue: https://github.com/apple/container/issues/641

Adds two new modes to `--progress`: **plain** and **color**.

* **plain**: Produces simple, line-by-line text output with *no ANSI escape codes*. Ideal for writing to log files or environments that don’t support terminal control sequences.

* **color**: Similar to the existing ANSI mode but with enhanced color usage. This mode provides a more expressive visual display. If it sounds good, we could consider making it the default in this or a follow up pr.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/TerminalProgress/ProgressConfig.swift
+++ b/Sources/TerminalProgress/ProgressConfig.swift
@@ -65,6 +65,10 @@ public struct ProgressConfig: Sendable {
     public let clearOnFinish: Bool
     /// The flag indicating whether to update the progress bar.
     public let disableProgressUpdates: Bool
+    /// The flag indicating whether to use colors.
+    public let color: Bool
+    /// The flag indicating whether to use plain output (no ANSI codes, one line per update).
+    public let plain: Bool
     /// Creates a new instance of `ProgressConfig`.
     /// - Parameters:
     ///   - terminal: The file handle for progress updates. The default value is `FileHandle.standardError`.
@@ -88,6 +92,8 @@ public struct ProgressConfig: Sendable {
     ///   - theme: The theme of the progress bar. The default value is `nil`.
     ///   - clearOnFinish: The flag indicating whether to clear the progress bar before resetting the cursor. The default is `true`.
     ///   - disableProgressUpdates: The flag indicating whether to update the progress bar. The default is `false`.
+    ///   - color: A flag indicating whether to enable ANSI color output. The default value is `false`.
+    ///   - plain: A flag indicating whether to force plain output with no control sequences. The default value is `false`.
     public init(
         terminal: FileHandle = .standardError,
         description: String = "",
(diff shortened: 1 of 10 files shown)


saehejkang:
I don't see anything out of the ordinary on these changes, but I also don't have permissions to approve PRs. You will have to wait for a maintainer to come back to this for a ✅ .

Sources/TerminalProgress/ProgressConfig.swift
```diff
@@ -65,6 +65,10 @@ public struct ProgressConfig: Sendable {
     public let clearOnFinish: Bool
     /// The flag indicating whether to update the progress bar.
     public let disableProgressUpdates: Bool
+    /// The flag indicating whether to use colors.
+    public let color: Bool
+    /// The flag indicating whether to use plain output (no ANSI codes, one line per update).
+    public let plain: Bool
```
katiewasnothere: Could we use an enum instead of these booleans to indicate the type of the output?

PhantomInTheWire: i was avoiding that since that would involve changing a widely used api would touch a lot unrelated lines but it does seem to make this a little bit more cleaner overall

PhantomInTheWire: refactored.

PhantomInTheWire:
@katiewasnothere @saehejkang  what do you guys think about making color mode the default mode instead of ansi since it visually looks a lot more appealing?

katiewasnothere:
@PhantomInTheWire From an accessibility standpoint, I think we should keep the default something without colors.

PhantomInTheWire:
sure i guess that makes sense, can you take a look again i have refactored with the requested change @katiewasnothere

jglogan:
> what do you guys think about making color mode the default mode instead of ansi since it visually looks a lot more appealing

#915 is open for using the standard buildkit environment variables to set color choices there.

Consider filing a follow-on issue for allowing users to set a persistent default for non-build `--progress` using a system property value.

So far our system properties are all backed by user defaults but we've discussed internally changing this to environment variables. That would allow users to set their global defaults in an rc file or profile, and then override them on a per-terminal basis.

PhantomInTheWire:
@jglogan sure i've opened up https://github.com/apple/container/issues/935 for now so it can be discussed and worked upon later

PhantomInTheWire:
@katiewasnothere quick ping in case this got buried

saehejkang:
@jglogan This PR can be closed, as it is a dupe of #1367 and #1384

katiewasnothere:
Closing this PR in favor of https://github.com/apple/container/pull/1367 and https://github.com/apple/container/pull/1384. @PhantomInTheWire Thank you for your work on this!

PR #805, opened 2025-10-25
Update documentation

dkovba:
Updates documentation according to the currently supported commands and displayed help strings.

PR #910, opened 2025-11-21
[docs]: add volume prune to command reference

saehejkang:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Looks like this got accidentally deleted in #805

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -797,13 +797,6 @@ container volume prune [--debug]
 
 No options.
 
-**Examples**
-
-```bash
-# remove all unused volumes
-container volume prune
-```
-
 ### `container volume list (ls)`
 
 Lists volumes.


docs/command-reference.md
```diff
@@ -783,6 +783,36 @@ container volume delete vol1 vol2 vol3
 container volume delete --all
 ```
 
+### `container volume prune`
```
realrajaryan, line 786: Please make it consistent with how we document other command references. 

Change the usage and options formatting, and remove the example output section.

docs/command-reference.md
```diff
@@ -783,6 +783,27 @@ container volume delete vol1 vol2 vol3
...
+**Options**
+
+No options.
+
+**Examples**
+
+```bash
+# remove all unused volumes
+container volume prune
+```
```
realrajaryan: should also get rid of this, doesn't add any value as it's the same as what's under usage

realrajaryan (changes requested):

realrajaryan (approved):

Issue #383, opened 2025-07-27
[Request]: Report detailed version information, user-friendly and structured.


jglogan:
### Feature or enhancement request details

The `--version` option of the container command reports short version information, but a richer version report should be available also. Output shall default to user friendly but a format option (consistent with options that already exist for other commands) shall allow the user to produce structured output for automation scripts.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

karenheckel:
@jglogan Hello! We are students at UT Austin working on issues for our virtualization class. If no one has taken this, we would love to work on it. Thank you!

karenheckel:
@jglogan Just checking in, can we help close this issue?

saehejkang:
@karenheckel I tend to comment on issues I am working on to let others know about its status. Since no one has commented on this issue in months + no PR has been opened, you should be good to work on this.

PR #911, opened 2025-11-21
feat: implement version sub command

fatelei:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
implement version sub command, give more info

image

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/Core/ClientHealthCheck.swift
+++ b/Sources/ContainerClient/Core/ClientHealthCheck.swift
@@ -43,9 +43,12 @@ extension ClientHealthCheck {
         guard let apiServerCommit = reply.string(key: .apiServerCommit) else {
             throw ContainerizationError(.internalError, message: "failed to decode apiServerCommit in health check")
         }
-        // Optional fields for newer servers
-        let apiServerBuild = reply.string(key: .apiServerBuild)
-        let apiServerAppName = reply.string(key: .apiServerAppName)
+        guard let apiServerBuild = reply.string(key: .apiServerBuild) else {
+            throw ContainerizationError(.internalError, message: "failed to decode apiServerBuild in health check")
+        }
+        guard let apiServerAppName = reply.string(key: .apiServerAppName) else {
+            throw ContainerizationError(.internalError, message: "failed to decode apiServerAppName in health check")
+        }
         return .init(
             appRoot: appRoot,
             installRoot: installRoot,
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -72,6 +72,12 @@ public struct Application: AsyncParsableCommand {
                 ]
             ),
             CommandGroup(
+                name: "Volume",
+                subcommands: [
+                    VolumeCommand.self
+                ]
+            ),
+            CommandGroup(
                 name: "Other",
                 subcommands: Self.otherCommands()
             ),
--- a/Sources/ContainerCommands/System/SystemCommand.swift
+++ b/Sources/ContainerCommands/System/SystemCommand.swift
@@ -31,9 +31,9 @@ extension Application {
                 SystemStart.self,
                 SystemStatus.self,
                 SystemStop.self,
-                VersionCommand.self,
+                SystemVersion.self,
             ],
             aliases: ["s"]
         )
-}
+    }
 }
--- a/Sources/ContainerCommands/System/Version.swift
+++ b/Sources/ContainerCommands/System/Version.swift
@@ -20,7 +20,7 @@ import ContainerVersion
 import Foundation
 
 extension Application {
-    public struct VersionCommand: AsyncParsableCommand {
+    public struct SystemVersion: AsyncParsableCommand {
         public static let configuration = CommandConfiguration(
             commandName: "version",
             abstract: "Show version information"
@@ -39,7 +39,7 @@ extension Application {
                 version: ReleaseVersion.version(),
                 buildType: ReleaseVersion.buildType(),
                 commit: ReleaseVersion.gitCommit() ?? "unspecified",
-                appName: "container CLI"
+                appName: "container"
             )
 
             // Try to get API server version info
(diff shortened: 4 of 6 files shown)


Sources/ContainerCommands/Application.swift
```diff
@@ -77,6 +77,12 @@ public struct Application: AsyncParsableCommand {
                     VolumeCommand.self
                 ]
             ),
+            CommandGroup(
```
jglogan: For the `container` CLI we've tried to avoid putting commands not related to container resources directly under `container`. Could you remove this command group, and make it so that the `version` command is a subcommand of `container system`?

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,62 @@
...
+        private func buildType() -> String {
+            #if DEBUG
+            return "debug"
+            #else
+            return "release"
+            #endif
+        }
+
+        private func printVersionTable(_ info: VersionInfo) {
+            print("\(info.appName) version \(info.version)")
```
jglogan: This isn't really tabular format. If we're going to have `table` and `json` format (which seems fine to me), the table format should use the `TableOutput` type that's used in all the resource.list() functions.

The other thing you could do for the table is try doing a ping() call on the API server and that would give you the API server release version if it's up (it should be the same but could differ if for example you did and upgrade and forgot to restart the API server, or had two installs and ran the wrong CLI executable). This could be an option (if the server responds) second row in the table.

If you do that, you could also add the `buildType` and `appName` values to what ping() returns for the API server.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,62 @@
...
+
+            switch format {
+            case .table:
+                printVersionTable(versionInfo)
+            case .json:
+                printVersionJSON(versionInfo)
+            }
+        }
+
+        private func buildType() -> String {
```
jglogan: How about relocating this into a public func in ReleaseVersion.swift?

You could then tweak ReleaseVersion.singleLine() to use ReleaseVersion.buildType() as well.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        private func printVersionJSON(_ info: VersionInfo) throws {
+            let data = try JSONEncoder().encode(info)
+            print(String(data: data, encoding: .utf8) ?? "{}")
+        }
+    }
+
+    public struct VersionInfo: Codable {
+        let version: String
+        let build: String
```
jglogan: Make this `buildType` so the meaning is a little more obvious.

docs/tutorial.md
```diff
@@ -54,6 +54,7 @@ USAGE: container [--debug] <subcommand>
 OPTIONS:
   --debug                 Enable debug output [environment: CONTAINER_DEBUG]
   --version               Show the version.
+  version                 Show detailed version information.
```
jglogan: This moves under `system`

saehejkang:
Looks like this PR closes #383

Sources/ContainerClient/Core/ClientHealthCheck.swift
```diff
@@ -43,6 +43,16 @@ extension ClientHealthCheck {
         guard let apiServerCommit = reply.string(key: .apiServerCommit) else {
             throw ContainerizationError(.internalError, message: "failed to decode apiServerCommit in health check")
         }
-        return .init(appRoot: appRoot, installRoot: installRoot, apiServerVersion: apiServerVersion, apiServerCommit: apiServerCommit)
+        // Optional fields for newer servers
+        let apiServerBuild = reply.string(key: .apiServerBuild)
```
jglogan: You can assume server and client are tightly coupled for now.

That is to say, go ahead and guard these assignments just like the others.

Sources/ContainerCommands/Application.swift
```diff
@@ -71,12 +71,6 @@ public struct Application: AsyncParsableCommand {
                     RegistryCommand.self,
                 ]
             ),
-            CommandGroup(
```
jglogan, line 74: This shouldn't be deleted, right?

Sources/ContainerCommands/System/SystemCommand.swift
```diff
@@ -31,8 +31,9 @@ extension Application {
                 SystemStart.self,
                 SystemStatus.self,
                 SystemStop.self,
+                VersionCommand.self,
```
jglogan: Rename to `SystemVersion`?

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerClient
+import ContainerVersion
+import Foundation
+
+extension Application {
+    public struct VersionCommand: AsyncParsableCommand {
```
jglogan: For consistency with other system subcommands, rename this file to `SystemVersion.swift` and the type to `SystemVersion`.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+        var global: Flags.Global
+
+        public init() {}
+
+        public func run() async throws {
+            let cliInfo = VersionInfo(
+                version: ReleaseVersion.version(),
+                buildType: ReleaseVersion.buildType(),
+                commit: ReleaseVersion.gitCommit() ?? "unspecified",
+                appName: "container CLI"
```
jglogan: What do you think about using the command names here - the one from `commandName` in Application here, and the one from `health. apiServerAppName` below? This should yield `container` and `container-apiserver` for the two entries.

Without the embedded whitespace it becomes a little easier to work with if someone does want to quickly grep/awk the tabular output.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public struct VersionInfo: Codable {
+        let version: String
+        let buildType: String
+        let commit: String
+        let appName: String
+    }
+
+    struct VersionJSON: Codable {
```
jglogan: We don't need this if outputting `[VersionInfo]`

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+                    version: health.apiServerVersion,
+                    buildType: health.apiServerBuild ?? "unknown",
+                    commit: health.apiServerCommit,
+                    appName: "container API Server"
+                )
+            } catch {
+                serverInfo = nil
+            }
+
+            switch format {
```
jglogan, line 61: Perhaps it'd be cleaner and more extensible to create an array containing client and optionally server, like this:

```swift
let versions = [cliInfo, serverInfo].compactMap { $0 }
```

We then pass `versions: [VersionInfo]` into both output calls.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+            if let server = server {
+                rows.append(["API Server", server.version, server.buildType, server.commit])
+            }
+            
+            let table = TableOutput(rows: rows)
+            print(table.format())
+        }
+
+        private func printVersionJSON(cli: VersionInfo, server: VersionInfo?) throws {
+            let output = VersionJSON(
```
jglogan: Here we simply output `versions` as a JSON array.

Sources/ContainerCommands/System/Version.swift
```diff
@@ -0,0 +1,109 @@
...
+            case .table:
+                printVersionTable(cli: cliInfo, server: serverInfo)
+            case .json:
+                try printVersionJSON(cli: cliInfo, server: serverInfo)
+            }
+        }
+
+
+        private func printVersionTable(cli: VersionInfo, server: VersionInfo?) {
+            var rows: [[String]] = [
```
jglogan: Here you can map each VersionInfo into a [String]

jglogan (commented):
@fatelei Thanks for the updates!

Just a couple of name changes for consistency, and a couple suggestions for simplifying the tabular and JSON presentation.

jglogan:
@fatelei Could you pull my conflict resolution commit, and then run `make fmt` and commit that? I tested your branch out locally, looks great thanks!

fatelei:
> @fatelei Could you pull my conflict resolution commit, and then run `make fmt` and commit that? I tested your branch out locally, looks great thanks!

done

PR #904, opened 2025-11-20
Implement container prune

tico88612:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Fixed #892

I'm a new contributor here. If there are any better ways to write it, please let me know.

Part of the function references content from PR #902. This PR will be rebased after that PR is merged.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

PR #914, opened 2025-11-22
[networks]: add prune command

### LINKED ISSUE #893: [Request]: Add `container network prune` command to remove unused networks

Add a `container network prune` subcommand to remove networks not connected to any containers.

Basic requirements:
- Remove all networks with no container connections
- Preserve default/system networks (e.g., bridge, host)
- Display list of removed network names

Example output:
```
Pruned networks:
some-network-1
some-network-2
```

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Closes #893 

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Tests/CLITests/TestCLINoParallelCases.swift
+++ b/Tests/CLITests/TestCLINoParallelCases.swift
@@ -125,6 +125,7 @@ class TestCLINoParallelCases: CLITest {
         #expect(alpineStillPresent, "expected image \(alpine) to remain")
     }
 
+    @available(macOS 26, *)
     @Test func testNetworkPruneNoNetworks() throws {
         // Ensure the testnetworkcreateanduse network is deleted
         // Clean up is necessary for testing prune with no networks
@@ -141,6 +142,7 @@ class TestCLINoParallelCases: CLITest {
         #expect(output.isEmpty, "should show no networks pruned")
     }
 
+    @available(macOS 26, *)
     @Test func testNetworkPruneUnusedNetworks() throws {
         let name = getTestName()
         let network1 = "\(name)_1"
@@ -180,6 +182,7 @@ class TestCLINoParallelCases: CLITest {
         #expect(!listAfter.contains(network2), "network2 should be pruned")
     }
 
+    @available(macOS 26, *)
     @Test(.disabled("https://github.com/apple/container/issues/953"))
     func testNetworkPruneSkipsNetworksInUse() throws {
         let name = getTestName()
@@ -234,6 +237,7 @@ class TestCLINoParallelCases: CLITest {
         #expect(!listAfter.contains(networkUnused), "unused network should be pruned")
     }
 
+    @available(macOS 26, *)
     @Test(.disabled("https://github.com/apple/container/issues/953"))
     func testNetworkPruneSkipsNetworkAttachedToStoppedContainer() async throws {
         let name = getTestName()


jglogan:
@saehejkang @suhasramanand

Is doing prune in the server needless complexity for `network prune`?

I'm asking this as a design question, not as an explicit criticism of the approach.

Our principal concern for these prune operations is consistency – if we run some arbitrary collection of operations on a set of related resources, we don't want to leave the system in an inconsistent state. Principally I think that's dangling references;  as an example, after a prune we should never have a case where a container refers to a non-existent network.

The network delete operation is already consistent on its own – a naive but functional implementation of prune would just to try deleting each network in turn regardless of what refers to it. Ideally there'd be a distinct EBUSY-like error so one could differentiate those (which could be silently ignored) versus other conditions.

I think the first, client-side-only, PR precomputed the set of unreferenced networks up front which would reduce the number of trips to the server that would return EBUSY – a simple and nice optimization.

Is there something that a server-side implementation of prune can do that can't be done by combining existing operations in ContainerClient? The guiding principle here is to keep the server side as slender and simple as possible, so that it's easier to reliably maintain.

It's probably worthwhile looking at what we need/want to do for other resources. For the image prune stuff there was a prune() operation for imagerefs all the way down in containerization, but we extracted that part out to container based on the same principle – the ImageStore only needs to ensure consistent pruning of content blobs, while pruning imagerefs can be dealt with in container.

Sources/Services/ContainerAPIService/Networks/NetworksService.swift
```diff
@@ -243,7 +243,41 @@ public actor NetworksService {
...
+
+            for network in networksToPrune {
+                do {
+                    try await self.store.delete(network.id)
+                    await self.removeNetworkState(for: network.id)
+
+                    prunedNames.append(network.id)
+                    self.log.info("Pruned network", metadata: ["name": "\(network.id)"])
+                } catch {
+                    self.log.error("Failed to prune network \(network.id): \(error)")
```
jglogan: Since we've locked the container list, the only errors here are real ones that need to make it back to the client. If we do decide to do a server-side implementation, we'd need to change this so the return data needs to indicate which networks were pruned and which failed (along with the respective errors).

jglogan (changes requested):

suhasramanand:
@jglogan I agree that a client-side approach would align better with the existing design pattern.

Looking at `ImagePrune` (in `Sources/ContainerCommands/Image/ImagePrune.swift`), it:
1. Lists images and containers on the client
2. Filters to determine which images to delete
3. Calls `ClientImage.delete()` for each one
4. Only uses server-side for `cleanupOrphanedBlobs()` (blob cleanup, not imageref pruning)

A client-side `network prune` would follow the same pattern:
- `ClientNetwork.list()` to get all networks
- `ClientContainer.list()` to determine which networks are in use
- Filter on the client to identify unreferenced networks
- Call `ClientNetwork.delete()` for each one, handling EBUSY-like errors gracefully

This keeps the server minimal and consistent with how image prune works. The `network delete` operation already ensures consistency, so the client-side approach should be sufficient.

The main benefit of server-side would be atomicity via `withContainerList`, but if `network delete` already handles consistency correctly, that may not be necessary.

saehejkang:
> Our principal concern for these prune operations is consistency

This may have been where the confusion first occurred, as a single design pattern does not seem to be used consistently. I took reference from the `volume prune` command, which is more of a server-side approach. Furthermore, the `image prune` command is mainly a client-side approach.

> Is there something that a server-side implementation of prune can do that can't be done by combining existing operations in ContainerClient?

There is nothing on the server-side implementation that can't be done by combining operations in the `ContainerClient`.

> It's probably worthwhile looking at what we need/want to do for other resources. 

Below is a list of the `prune` commands and the current/future implementation approach.

- `image prune` - client-side approach
- `volume prune` - server-side approach
- `container prune` - server-side approach (in progress #904)
- `network prune` - approach dependent on this discussion

Is there any reason why any of these resources would EVER need to use a server-side approach (besides image prune with the content blobs)?
***
Regarding the `network prune` command, we have two PRs, each proposing a valid approach. A decision on which one to merge is needed, and I will defer that to the maintainers. Finally, if we decide on the client-side approach, it wouldn’t be fair for me to simply make the changes here and then merge my PR.

jglogan:
> single design pattern does not seem to be used consistently

This. I've blocked out some time today to look at this (and the error handling bit I mentioned) from a broader perspective, and then will follow up here and we can discuss how to move forward.

jglogan:
OK, here are my thoughts after a little reflection:

- Let's keep the server APIs minimal, where we do the basic resource management operations reliably and consistently (focusing on getting contending operations on a single resource correct).
- The client is where we'll do composite operations, and operations on collections. The client shall be responsible for dealing with partial success - we should consistently report errors where operations could be performed on some resources and not others (making exceptions for things like prune where an operation may not be able to be performed because of a state change race).

The principal disadvantage that I see right now is performance. Our APIs operate on a single resource at once today, each one locking the container collection. This isn't really a new issue though: `container volume rm --all` would incur the same penalties.

@saehejkang and @suhasramanand What do you think?

saehejkang:
> Let's keep the server APIs minimal, where we do the basic resource management operations

This makes sense to me. Just to be clear, this is for get, create, delete. operations for a single resource?

> The client is where we'll do composite operations, and operations on collections

This also makes sense to me. Again, to be clear, this is for operations like `network prune`, as it would be going through the collection of containers/networks, and deleting unused networks?

***
1. Are we revisiting any commands (refer to my note [here](https://github.com/apple/container/pull/914#:~:text=for%20other%20resources.-,Below,-is%20a%20list)) and making the proper updates?
2. Question [here](https://github.com/apple/container/pull/914#:~:text=on%20this%20discussion-,Is,-there%20any%20reason)

> Our APIs operate on a single resource at once today, each one locking the container collection. 

3. Is this going to also be revisited in the future? Are we ever going to build APIs that operate on more than a single resource?

jglogan:
> 1. Are we revisiting any commands (refer to my note [here](https://github.com/apple/container/pull/914#:~:text=for%20other%20resources.-,Below,-is%20a%20list)) and making the proper updates?

Yes. Let's get `network prune` working this way, and then use it as a template for the `container prune` PR review, and we can rework `volume prune` so that all use the same pattern.

> 2. Question [here](https://github.com/apple/container/pull/914#:~:text=on%20this%20discussion-,Is,-there%20any%20reason)

Some reasons we might need a server-side approach:

* Performance, as mentioned before - if we're operating on a lot of resources, it could be that coalescing all the work under a single call could yield a decent speedup.
* Transactional operations - if we find that we need a compound operation that needs "all or nothing" semantics across some set of resources, perhaps that's better done in the server? I don't have a concrete example in mind for this though.

> > Our APIs operate on a single resource at once today, each one locking the container collection.
>
> 3. Is this going to also be revisited in the future? Are we ever going to build APIs that operate on more than a single resource?

There's nothing stopping us from that, or for moving compound operations like prune into the API server. The principal motivation for moving in the current direction (keeping things simple, and as you pointed out, consistent) for now is to start cleaning up the client code and get it in a good state for developers. Once that's done we can let time and experience guide how we enhance the client API and the underlying API server.

The thing that gets under my skin at present is that our client (the APIs and the core types) is dispersed across ContainerClient and Services. The dependencies between targets within the project are messier than they need to be. I want to change this so that ContainerClient contains all of the SDK material, such that you or I can look at the docc just for ContainerClient and find what we need to code against container.

There's a lot more developer documentation that could be done to support this. We still lack docs for our extension mechanisms (the plugin system in particular), for example.

saehejkang:
> Yes. Let's get network prune working this way, and then use it as a template for the container prune PR review, and we can rework volume prune so that all use the same pattern.

I would like to spearhead this initiative and complete it to fruition. Once `network prune` is wrapped up, I can go back and help review the `container prune` PR, and then work on updates to `volume prune`.

> cleaning up the client code and get it in a good state for developers

> I want to change this so that ContainerClient contains all of the SDK material, such that you or I can look at the doc just for ContainerClient and find what we need to code against container.

> There's a lot more developer documentation that could be done to support this.

I completely agree that cleanup is important and keeping things simple/consistent is important for quality. It is all coming full circle, as I remember asking a question in the discussions about how the SDK/ContainerClient works. I am sure that working on these updates will help me wrap my head around everything and I can work on adding some docs, hopefully in the near future.

***
Awesome points above and thank you for the all the explanations about the design/ your thought process!

> If we decide on the client-side approach, it wouldn’t be fair for me to simply make the changes here and then merge my PR.

How do we want to proceed with `network prune`?

jglogan:
@saehejkang Consider yourself signed up for these...I'll watch for an update on this PR, we can merge it and yep, go ahead and move to the next! Thank you.

Sources/Services/ContainerAPIService/Networks/NetworksService.swift
saehejkang: We are no longer going with the server-side implementation

Sources/ContainerCommands/Network/NetworkPrune.swift
```diff
@@ -0,0 +1,67 @@
...
+
+            var networksInUse = Set<String>()
+            for container in allContainers {
+                for network in container.configuration.networks {
+                    networksInUse.insert(network.network)
+                }
+            }
+
+            let networksToPrune = allNetworks.filter { network in
+                network.id != ClientNetwork.defaultNetworkName && !networksInUse.contains(network.id)
```
saehejkang, line 44: I purposefully left the check for if a network is running because the issue did not call for it. Furthermore, if the network is not being used by any containers, I feel should it not be pruned, no matter the state?

jglogan: > Furthermore, if the network is not being used by any containers, I feel should it not be pruned, no matter the state?

I'm not sure I understand this. If the network is not used by any containers, and it's not the default network, we should prune it, correct?

The code seems correct. I'm just confused about the reference to state. At present the network state model is simple - in the steady state it's running, or it doesn't exist.

If you're referring to container state, then yes, container state shouldn't matter. If the container exists (running or not) and references network `foo`, then network `foo` should not be pruned.

saehejkang: >  If the network is not used by any containers, and it's not the default network, we should prune it, correct?

Correct

> If you're referring to container state, then yes, container state shouldn't matter. If the container exists (running or not) and references network foo, then network foo should not be pruned.

I may have gotten the states mixed up 😅

jglogan:
@saehejkang Your side of this looks fine. I'm looking into an issue in vmnet where that one of your CLI tests found.

The problem exists whether or not we merge prune; it can be reproduced with the existing `network` and `run` commands. I'm going to do a bit more digging on this tomorrow but in the end we what we could do is disable the test for now and merge it, and create an issue to document the known problem.

If you have a moment, could capture some logs in a terminal like so:

```bash
log stream --debug --info --predicate 'subsystem = "com.apple.NetworkSharing"'
```

And then this set of commands which functions similarly to your prune test?

```bash
#! /bin/bash -x

container system stop
container system start
container network rm in-use unused
container network create in-use
container network create unused
container run --name in-use-server --network in-use --rm -d -p 127.0.0.1:8000:8000 python:slim python3 -m http.server --bind 0.0.0.0 8000
container ls
container network rm unused
container stop in-use-server
container network rm in-use
container run --name default-server --rm -d -p 127.0.0.1:8000:8000 python:slim python3 -m http.server --bind 0.0.0.0 8000 
sleep 3
curl http://localhost:8000
```

I see that the `curl` at the end fails, and the OS logs show that the container interface for `default-server` didn't get configured:

```console
2025-12-08 13:37:56.572274-0300 0x3a2c3    Debug       0x0                  38322  0    InternetSharing: [com.apple.NetworkSharing:daemon] netrbCreateInterface: network handle was passed in
2025-12-08 13:37:56.572233-0300 0x3a27c    Info        0x0                  38321  0    com.apple.Virtualization.VirtualMachine: (Netrb) [com.apple.NetworkSharing:framework.netrb] _NETRBNetworkStartVirtualMachineInterface: adding virtual interface to network 0x7eac48140
2025-12-08 13:37:56.572285-0300 0x3a2c3    Default     0x0                  38322  0    InternetSharing: [com.apple.NetworkSharing:daemon] netrbCreateInterface: no network related to id
2025-12-08 13:37:56.572315-0300 0x3a27c    Default     0x0                  38321  0    com.apple.Virtualization.VirtualMachine: (Netrb) [com.apple.NetworkSharing:framework.netrb] _NETRBNetworkStartVirtualMachineInterface_block_invoke_2: interface creation failed
```

saehejkang:
> but in the end we what we could do is disable the test for now

Which test is this? There are two tests that use the `doLongRun` function.

EDIT: I created an issue #953 and will be disabling both tests that use `prune` and `run` commands.

> could capture some logs

[network-logs.txt](https://github.com/user-attachments/files/24066965/network-logs.txt)

saehejkang:
#941 should be merged first before this one. I am moving `network prune` tests to the `TestCLINoParallelCases` file and there would have been merge conflicts :)

jglogan:
@saehejkang I merged #941 and am trying the build here, but we'll need the rebase for the serialized test changes now

Tests/CLITests/TestCLINoParallelCases.swift
```diff
@@ -124,4 +124,166 @@ class TestCLINoParallelCases: CLITest {
         let alpineStillPresent = try isImagePresent(targetImage: alpine)
         #expect(alpineStillPresent, "expected image \(alpine) to remain")
     }
+
+    @Test func testNetworkPruneNoNetworks() throws {
```
jglogan, line 129: Any test function that runs `container network` needs the `@available(macOS 26, *)` qualifier.

PR #917, opened 2025-11-28
Add content subcommand

malt3:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

This command can be used to interact with the content store XPC via simple commands. Since many container tools are written in Go, it's not easy to interact with the XPC service from other tools. Go has good support for unix domain sockets, but Mach ports (and XPC) are another story.

The content store is very useful to enable incremental image loading. The closest alternative ("container image load") requires materialization of all blobs in a tar file. With the content store, we can check which blobs are already available and only load missing ones. This will enable faster development workflows in tools like rules_img: https://github.com/bazel-contrib/rules_img.

A similar move is being made in Docker. The containerd content store used to be limited to the containerd socket. This upstream issue will expose the containerd content store directly via Docker: https://github.com/moby/moby/issues/44369

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -201,12 +201,14 @@ public struct Application: AsyncLoggableCommand {
         guard #available(macOS 26, *) else {
             return [
                 BuilderCommand.self,
+                ContentCommand.self,
                 SystemCommand.self,
             ]
         }
 
         return [
             BuilderCommand.self,
+            ContentCommand.self,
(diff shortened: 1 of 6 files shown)


malt3:
Marked as draft for now. Please let me know if this makes sense and what kind of tests you would like to see!

PR #921, opened 2025-11-30
[docs]: move callout regarding documentation revisions

### LINKED ISSUE #920: [Request]: Move callout regarding documentation revisions to the top of each Markdown doc.

For the time being we don't host the Markdown files in the docs directory anywhere and as such the only way to view the documents is to browse the repo. There are links on the README.md page but these go to the main branch and not the latest release. We do have a note to this effect with the links: https://github.com/apple/container?tab=readme-ov-file#next-steps.

However, it's easy to miss this, especially if you just drill down into the docs directory instead of accessing the docs through the README.md links, as with this example: https://github.com/apple/container/issues/916#issuecomment-3577784699.

As long as we're staying low-tech with the docs, we should move the note from the README.md file to just under the h1 in all of the [./docs](https://github.com/apple/container/tree/main/docs) files.

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Closes #920

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,6 @@ uninstall-container.sh -k
 - [Build and run](./BUILDING.md) `container` on your own development system.
 - View the project [API documentation](https://apple.github.io/container/documentation/).
 
-> [!IMPORTANT]
-> The links above are for the CURRENT BRANCH's documentation. To find documentation for official releases, find the target release on the [Release Page](https://github.com/apple/container/releases) and click the tag corresponding to your release version. 
->
-> Example: [release 0.4.1 tag](https://github.com/apple/container/tree/0.4.1)
-
 ## Contributing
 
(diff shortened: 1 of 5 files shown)


docs/README.md
```diff
@@ -0,0 +1,8 @@
+# Docs
```
saehejkang: I thought adding a `README.md` file in the `docs` directory with just that note only once is better. It could alleviate the need to worry about tech debt, and other notes/references about the `docs` can be added there in the future.

jglogan: I worry about discoverability though. This may fix the case where people nav to the docs directory, but if the follow the links from the top-level README section, they won't see it.

If we put it in each doc it's a bit harder to maintain but it's becomes very hard to not see :D

PR #929, opened 2025-12-04
Feat: enable configurable default disk capacity for native and builder containers

karenheckel:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
- Fixes apple#887
- This change adds support for specifying a default disk capacity. This allows users to define global storage defaults without needing to pass ```--storage``` each time they run a container. It aligns disk capacity behavior with the existing CPU and memory default configuration.
- Added ```rootfs.storage``` to the system property configuration.
- Added validation logic to ensure the user-specified capacity does not exceed available host space.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/RequestScheme.swift
+++ b/Sources/Services/ContainerAPIService/Client/RequestScheme.swift
@@ -40,7 +40,7 @@ public enum RequestScheme: String, Sendable {
     /// Returns the prescribed protocol to use while making a HTTP request to a webserver
     /// - Parameter host: The domain or IP address of the webserver
     /// - Returns: RequestScheme
-    package func schemeFor(host: String) throws -> Self {
+    public func schemeFor(host: String) throws -> Self {
         guard host.count > 0 else {
             throw ContainerizationError(.invalidArgument, message: "host cannot be empty")
         }


Sources/ContainerClient/Parser.swift
```diff
@@ -875,3 +879,28 @@ public struct Parser {
         }
     }
 }
+
+extension Parser {
+    /// Validates that the host has enough disk space for the requested storage.
+    public static func validateHostStorage(bytes: UInt64) throws {
+        let fileURL = URL(fileURLWithPath: "/")
```
jglogan: Would `[ApplicationRoot](https://github.com/apple/container/blob/dev/native-builder/Sources/ContainerPlugin/ApplicationRoot.swift#L30).url` make more sense here?

I'm not sure the benefit of this check outweighs the possible annoyances. Each container today assumes by default 512GB of storage, so it would be very easy to trigger this alert on systems with 1GB of SSD. Typical containers will only use a small fraction of the storage, so assuming that an average container requires all 512GB to run may be incorrect.

Sources/ContainerPersistence/DefaultsStore.swift
```diff
@@ -26,6 +26,8 @@ public enum DefaultsStore {
         case buildRosetta = "build.rosetta"
         case defaultDNSDomain = "dns.domain"
         case defaultBuilderImage = "image.builder"
+        case defaultBuilderStorage = "builder.storage"
```
jglogan: Why is it storage in one and capacity in another? How about `builder.storage` and `container.storage`?

Sources/ContainerCommands/System/Property/PropertySet.swift
```diff
@@ -74,6 +74,9 @@ extension Application {
                     throw ContainerizationError(.invalidArgument, message: "invalid CIDRv4 address: \(value)")
                 }
                 DefaultsStore.set(value: value, key: key)
+            case .defaultBuilderStorage, .defaultContainerStorage:
```
jglogan: The idea here is that we validate the input but store the user-friendly value?

Seems reasonable, it could be bypassed by someone using the `defaults` command so we should validate it separately when we read (see note above).

Sources/ContainerClient/Utility.swift
```diff
@@ -165,11 +165,26 @@ public struct Utility {
         var config = ContainerConfiguration(id: id, image: description, process: pc)
         config.platform = requestedPlatform
 
+        let effectiveStorage: String? = {
+            if let storage = resource.storage {
```
jglogan: Perhaps we should separately validate the cases of storage provided through the command vs provided through the user default, and throw two different invalid argument errors so that the user can see what to correct in the case of the user default (in the unlikely case that they set the user default incorrectly via the `defaults` command or via Swift code).

Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -81,6 +81,12 @@ extension Application {
         )
         var memory: String = "2048MB"
 
+        @Option(
```
jglogan: Could you create a CLI test for a regular container and one for the builder?

For a regular container you could do something along the lines of this (using one of the alpine images we have on ghcr.io for testing):

```
% container run -it --rm alpine:latest sh -c "df -m / | tail -1 | tr -s ' ' | cut -d' ' -f2"
516064                           
```

For the builder you'd start the builder, `exec` the test command, and then stop the builder.

jglogan (changes requested):
Thank you for contributing! I've left a couple of questions and minor change requests in the comments.

Sources/ContainerClient/Parser.swift
karenheckel: Thanks for the catch regarding sparse files! I originally implemented the check because the issue description mentioned "Validate user specified capacity against host available space." However, I agree that given the default 512 GB size and thin provisioning, it's too restrictive to enforce it like this. I'll go ahead and remove the function.

jglogan:
@karenheckel AFAICT I see one test failure:

```console
◇ Test testBuilderStorageFlag() started.
✘ Test testBuilderStorageFlag() recorded an issue at CLIBuilderLifecycleTest.swift:73:17: Expectation failed: (abs(reportedMiB - expectedMiB) → 511968) <= (tolerance → 409)
↳ expected root filesystem size ≈ 4096 MiB for storage 4096MB, got 516064 MiB
```

karenheckel:
@jplogan We have the plumbing set up for configuring the disk storage. However, we think that the underlying container runtime ignores the ```resources.storage``` field since the Containerization backend isn't wired to it yet, so the VM still boots with the default disk image size. A follow-up PR could extend the VM layer to actually size the rootfs block device according to ```resources.storage```.

If there is something we are not understanding correctly, please feel free to let us know!

karenheckel:
@jglogan We have attempted to provide backend support for the storage configuration support. [Here is that pull request.](https://github.com/apple/containerization/pull/447) The frontend (this PR) works, but the backend hasn't fully worked locally. If you have any suggestions, we would love to hear them. Thank you!

Issue #923, opened 2025-12-01
[Bug]: Running a kubernetes cluster inside a container crashes the container


arbhojntnx:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

- Download and Start container 
```
container system start
```
- Download and install kinc (Kubernetes in Container)
```
go install github.com/funkolab/kinc@latest
```
- Create a cluster
```
kinc create cluster
```

### Current behavior
The container crashes after a minute or so.

[container-runtime-linux-2025-12-01-100724.txt](https://github.com/user-attachments/files/23859876/container-runtime-linux-2025-12-01-100724.txt)

### Expected behavior

- A container is created with all the kind components and this environment persists

### Environment

```markdown
- OS: 26.1 (25B78)
- Xcode: Version 26.1.1 (17B100)
- Swift: swift-driver version: 1.127.14.1 Apple Swift version 6.2.1 (swiftlang-6.2.1.4.8 clang-1700.4.4.1)
- Container: container CLI version 0.6.0 (build: release, commit: a23bcf0)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

arbhojntnx:
Here is the complete cash report
[container-runtime-linux-2025-12-01-100724.txt](https://github.com/user-attachments/files/23860351/container-runtime-linux-2025-12-01-100724.txt)

Michaelgathara:
@arbhojntnx @jglogan Should be able to close this issue now

jglogan:
Closed by #930 and #957.

PR #930, opened 2025-12-05
Fix: Kubes Cluster in Container Crashing Container (IS#923)

Michaelgathara:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This fixed issue #923 

I fixed a race condition in `ConnectHandler.swift` where an asynchronous network connection could complete after the handler had already been removed from the pipeline.

This:
- Added a removed flag to track if the handler is still active.
- Updated the connection callback to check this flag; if the handler was removed while connecting, we now immediately close the new connection instead of trying to use it.

This prevents the EXC_BREAKPOINT crash in container-runtime-linux that occurred when kinc (Kubernetes in Container) created rapid connections. As we saw in the issue attached

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/SocketForwarder/ConnectHandler.swift
+++ b/Sources/SocketForwarder/ConnectHandler.swift
@@ -83,7 +83,7 @@ extension ConnectHandler {
                 case .success(let channel):
                     guard context.channel.isActive else {
                         self.log?.trace("backend - frontend channel closed, closing backend connection")
-                        channel.close(promise: nil)
+                        context.channel.close(promise: nil)
                         return
                     }
                     self.log?.trace("backend - connected")
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "ci"
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -111,6 +111,6 @@ jobs:
           tar xfz _site.tgz
 
       - name: Upload Artifact
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v4
         with:
           path: "./_site"
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "a20d52336b96991c23765c8301254aeff44ab33c0ccc7dee59191e50652f5070",
+  "originHash" : "64cb422fa3914611343af4301e317573002890fea7d174e550cc937a76571515",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/containerization.git",
       "state" : {
-        "revision" : "e8aff29be3b97afa18ccc256126466ae5e611bea",
-        "version" : "0.16.2"
+        "revision" : "9ba8267afbdff66e5ddce180312abdb41395292f",
+        "version" : "0.17.0"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ import PackageDescription
 let releaseVersion = ProcessInfo.processInfo.environment["RELEASE_VERSION"] ?? "0.0.0"
 let gitCommit = ProcessInfo.processInfo.environment["GIT_COMMIT"] ?? "unspecified"
 let builderShimVersion = "0.7.0"
-let scVersion = "0.16.2"
+let scVersion = "0.17.0"
 
 let package = Package(
     name: "container",
@@ -146,6 +146,7 @@ let package = Package(
             dependencies: [
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Containerization", package: "containerization"),
+                .product(name: "ContainerizationExtras", package: "containerization"),
                 .product(name: "ContainerizationOS", package: "containerization"),
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 "ContainerClient",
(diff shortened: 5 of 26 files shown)


jglogan:
@Michaelgathara Thanks! Could you set up commit signing and update your PR with signed commits?

See: https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests

Michaelgathara:
> @Michaelgathara Thanks! Could you set up commit signing and update your PR with signed commits?
> 
> See: https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests

Hi, sorry for missing that. Should be fixed now!

Sources/SocketForwarder/ConnectHandler.swift
```diff
@@ -79,6 +81,13 @@ extension ConnectHandler {
             .connect(to: serverAddress)
             .assumeIsolatedUnsafeUnchecked()
             .whenComplete { result in
+                guard !self.removed else {
```
jglogan: Can you explain the race a little more clearly? The sequence of events I think you're trying to guard against is:

1. `channelActive` event, ConnectHandler will be in the handler chain, this method gets called, and this block runs when the backend connection completes
2. this handler gets removed from the chain via `removeHandler()`
3. the backend connection setup completes successfully, and we call `glue()` to replace ConnectHandler with GlueHandler.

How does step 2 come about? The only way that I see `removeHandler()` getting called and `self.removed` being set to true is from `glue()`, which only happens if execution has proceeded past your check.

Michaelgathara: Yeah of course, from my understanding the race condition here occurs when the frontend channel closes before the backend completes as you mentioned. 

Where `removeHandler()` is called by NIO during pipeline teardown when a channel closes and not just from `glue()`. The sequence seems to be:

1. `channelActive` into `connectToServer()` starts async `connect()`
2. Frontend channel closes then NIO calls `removeHandler()` then `removed = true`
3. Async `connect()` completes then completion handler runs
4. Without guard: would call `glue()` with invalid context it crashes
5. With guard: detects `removed = true` and closes channel safely

and we see this within `ConnectHandler.swift` at:
```swift
private func glue(_ peerChannel: Channel, context: ChannelHandlerContext) {
    self.log?.trace("backend - gluing channels")

    let (localGlue, peerGlue) = GlueHandler.matchedPair()
    do {
        try context.channel.pipeline.syncOperations.addHandler(localGlue)
        try peerChannel.pipeline.syncOperations.addHandler(peerGlue)
        context.pipeline.syncOperations.removeHandler(self, promise: nil)  // Explicit removal (happy path)
    } catch {
        peerChannel.close(mode: .all, promise: nil)
        context.close(promise: nil)
    }
}
```

This would be the "happy path" in the sense that we explicitly remove the handler after successfully connecting. 

The **race condition** occurs when the channel closes _before_ this code runs. We can see this within `ConnectHandler.swift` in a couple places:

```swift
case .failure(let error):
    self.log?.error("backend - connect failed: \(error)")
    context.close(promise: nil)  // Closes frontend channel on error
    context.fireErrorCaught(error)
```

(in glue error handler)

```swift
} catch {
    // Close connected peer channel before closing our channel.
    peerChannel.close(mode: .all, promise: nil)
    context.close(promise: nil)  // Closes frontend channel on error
}
```

(in our guard)

```swift
guard !self.removed else {
    if case .success(let channel) = result {
        channel.close(promise: nil)  // Closes backend channel if handler was removed
    }
    return
}
```


So in terms of the fix here, it goes back to how NIO's RemovableChannelHandler protocol works. When a channel closes, NIO's internal pipeline teardown process:
1. Marks the channel as inactive
2. Calls `handlerRemoved()` or `removeHandler()` on all handlers in the pipeline
3. This happens **asynchronously** and can occur while async operations (like `connect()`) are still in progress

It seems that the `RemovableChannelHandler` protocol is specifically designed for handlers that need cleanup when removed, which is why I implemented it here.

Let me know if there is anything I missed or if you disagree with my understanding of the code here

jglogan: Thanks for clarifying that! Don't disagree, that's a good catch.

I'm wondering if we could do this without creating additional state? You mentioned that NIO marks the channel inactive at effectively the same time you're modifying `removed`, so could we just do something like this instead?

```swift
        ClientBootstrap(group: context.eventLoop)
            .connect(to: serverAddress)
            .assumeIsolatedUnsafeUnchecked()
            .whenComplete { result in
                switch result {
                case .success(let channel):
                    guard context.channel.isActive else {
                        self.log?.trace("backend - frontend channel closed, closing backend connection")
                        context.channel.close(promise: nil)
                        return
                    }
                    self.log?.trace("backend - connected")
                    self.glue(channel, context: context)
                case .failure(let error):
                    self.log?.error("backend - connect failed: \(error)")
                    context.close(promise: nil)
                    context.fireErrorCaught(error)
                }
            }
```

Michaelgathara: Yeah this is a much better solution, thank you!

Added this into the branch, kept the test as well

Michaelgathara:
@jglogan I noticed you added in the code to the `ConnectHandler.swift` file within https://github.com/apple/container/commit/4f887251582e63594cd3037f7b5790ab5c7111a5. I left this PR open for the test file. Happy to close if the test file is not needed

jglogan:
> I noticed you added in the code to the ConnectHandler.swift file within https://github.com/apple/container/commit/4f887251582e63594cd3037f7b5790ab5c7111a5. I left this PR open for the test file. Happy to close if the test file is not needed

I did? 😮

jglogan:
Oh jeez, sorry about that @Michaelgathara!

I wanted to merge it here. I was locally testing it in the tree where I was doing all that type work and forgot to back it out.

Let's definitely get the test merged so you get credited in the release!

Michaelgathara:
> Oh jeez, sorry about that @Michaelgathara!
> 
> I wanted to merge it here. I was locally testing it in the tree where I was doing all that type work and forgot to back it out.
> 
> Let's definitely get the test merged so you get credited in the release!

All good, it happens 😁

Thank you for the discussion, better solution and merge!

PR #941, opened 2025-12-08
[images]: refactor prune command

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Updates to `image prune` for consistency with how other `prune` commands are done. Added missing test cases as well for the command

Relates to the discussion from #914

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
+++ b/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
@@ -19,12 +19,7 @@ import ContainerizationOCI
 import Foundation
 import Testing
 
-@Suite(.serialized)
 class TestCLIImagesCommand: CLITest {
-    private func getTestName() -> String {
-        Test.current!.name.trimmingCharacters(in: ["(", ")"]).lowercased()
-    }
-
     @Test func testPull() throws {
         do {
             try doPull(imageName: alpine)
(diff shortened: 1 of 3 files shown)


Sources/ContainerCommands/Image/ImagePrune.swift
```diff
@@ -35,48 +35,49 @@ extension Application {
...
+                    prunedImages.append(image.reference)
+                    log.info("Pruned image", metadata: ["reference": "\(image.reference)"])
+                } catch {
+                    log.error("Failed to prune image \(image.reference): \(error)")
+                }
             }
 
             let (deletedDigests, size) = try await ClientImage.cleanupOrphanedBlobs()
 
+            for image in imagesToPrune {
```
saehejkang, line 69: Makes sense to just output the image/digest name that was pruned, but should we output what type was just pruned (like before)?

Sources/ContainerCommands/Image/ImagePrune.swift
```diff
@@ -35,48 +35,49 @@ extension Application {
...
 
-            for image in imagesToDelete {
-                try await ClientImage.delete(reference: image.reference, garbageCollect: false)
+            var prunedImages = [String]()
+
+            for image in imagesToPrune {
+                do {
+                    try await ClientImage.delete(reference: image.reference, garbageCollect: false)
+                    prunedImages.append(image.reference)
+                    log.info("Pruned image", metadata: ["reference": "\(image.reference)"])
```
jglogan: Don't log non-error conditions to stderr in commands.

Sources/ContainerCommands/Image/ImagePrune.swift
jglogan: Good question.

Maybe a small tweak on what we were doing before?

```
untagged digest1
untagged digest2
...
deleted ref1
deleted ref2
reclaimed-bytes n
```

jglogan: This is the current look:

```
 % container image prune 
Deleted images:
deleted: 0e183e7300eb11e319ed98842cf9672174b0197246e01f1baf6713af056e0f44
deleted: b808d7d5c394dafdff03dd9245076de2c0938d69ef448bc433e86004dd53c734
deleted: 5fb6c9156e4f9b766ab67dbf332b98a2f2800161f38923c19c56f63bb70644df
deleted: 3425d26b6a41af1924de81cba73a5b5ac39bd2d0e5d5d8e74157715f1b3a0374
deleted: ae0c6e66a433913d2b807fb1a14928ea02a764c292479457af4052deaffb3e55
```

Also wondering if your refs should include the `sha256:` prefix like:

```
 % container image prune 
deleted sha256:0e183e7300eb11e319ed98842cf9672174b0197246e01f1baf6713af056e0f44
deleted sha256:b808d7d5c394dafdff03dd9245076de2c0938d69ef448bc433e86004dd53c734
deleted sha256:5fb6c9156e4f9b766ab67dbf332b98a2f2800161f38923c19c56f63bb70644df
deleted sha256:3425d26b6a41af1924de81cba73a5b5ac39bd2d0e5d5d8e74157715f1b3a0374
deleted sha256:ae0c6e66a433913d2b807fb1a14928ea02a764c292479457af4052deaffb3e55
```

saehejkang: I am leaning towards how the other prune commands. The addition of the sha256 can help differentiate things without needing to specify it correct?

jglogan: > Also wondering if your refs should include the sha256: prefix like:

Let's leave this for a later fix. I think it'd be better if the fully qualified content IDs were coming back from the server

saehejkang: Should I omit the type then or do we want to keep that as well?

jglogan: Let's do the two element list – type and value separated by space – so we've got regular output.

Since it's not a plain list like others it'd still be nice for it to be something that's really easy to grep/awk/sed.

jglogan:
Ready to move out of draft?

saehejkang:
Looks like I am getting the same error mentioned [here](https://github.com/apple/container/pull/914#issuecomment-3634993125) for the `testImagePruneDanglingImages` test. It is again, a test that combines an existing `image` with a `run` command.

Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
```diff
@@ -19,7 +19,12 @@ import ContainerizationOCI
 import Foundation
 import Testing
 
+@Suite(.serialized)
```
saehejkang: This is necessary due to tests overlapping. The `prune` tests were running and accidentally deleting images that were unused.

```✘ Test testImagePruneUnusedImages() recorded an issue at TestCLIImagesCommand.swift:404:9: Expectation failed: (output → "untagged ghcr.io/linuxcontainers/alpine:testImageSaveAndLoadStdinStdout
untagged ghcr.io/linuxcontainers/alpine:testImageSaveAndLoad
untagged ghcr.io/linuxcontainers/alpine:3.18
untagged ghcr.io/linuxcontainers/alpine:testImageTag
untagged ghcr.io/containerd/busybox:testImageSaveAndLoad
...
```

jglogan: We have a serialized suite already (just added recently to deal with the concurrent download tests). Perhaps you could move the annotation and the prune tests to that suite as opposed to serializing the tests in this one, and then you should also be able to remove the `--no-parallel` from that suite in the Makefile.

jglogan (approved):

PR #944, opened 2025-12-08
Fix broken image integration tests.

### LINKED ISSUE #943: [Bug]: TestCLIImages current download tests break other image tests.
### I have done the following

### Steps to reproduce

The test for concurrent downloads removes the `ghcr.io/linuxcontainers/alpine:3.20` in between pulls with different concurrency values.

This will cause other image tests to randomly fail as they are also trying to work with this image.

### Current behavior

Image tests fail nondeterministically.

### Expected behavior

Image tests pass.

### Environment

```markdown
- OS: macOS 26
- Xcode: 26
- Container: 0.7.1
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

jglogan:
- Fixes #943.
- Use images other than alpine:3.20 for image concurrency test so as not to interfere with tests using that image.
- Rename test files to match suite names.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #943.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,7 @@ integration: init-block
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIVolumes || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIKernelSet || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIAnonymousVolumes || exit_code=1 ; \
-		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --no-parallel --filter TestNoParallelCases || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --no-parallel --filter TestCLINoParallelCases || exit_code=1 ; \
 		echo Ensuring apiserver stopped after the CLI integration tests ; \
 		scripts/ensure-container-stopped.sh ; \
 		exit $${exit_code} ; \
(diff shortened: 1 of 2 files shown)


Tests/CLITests/Subcommands/Run/TestCLIRunCommand.swift
```diff
@@ -513,20 +513,22 @@ class TestCLIRunCommand: CLITest {
...
-            #expect(success, "Request to \(url) failed after \(retries - retriesRemaining) retries")
+            try #require(success, "Request to \(url) failed after \(retries - retriesRemaining) retries")
             try doStop(name: name)
         } catch {
             Issue.record("failed to run container \(error)")
             return
         }
     }
 
+    /*
```
katiewasnothere: Did you mean to comment this test out?

Tests/CLITests/Subcommands/Images/TestCLIImages.swift
```diff
@@ -378,16 +378,13 @@ extension TestCLIImagesCommand {
...
         do {
-            try doPull(imageName: alpine, args: ["--max-concurrent-downloads", "1"])
-            let imagePresent = try isImagePresent(targetImage: alpine)
+            try doPull(imageName: alpine_concurrency1, args: ["--max-concurrent-downloads", "1"])
+            let imagePresent = try isImagePresent(targetImage: alpine_concurrency1)
             #expect(imagePresent, "Expected image to be pulled with maxConcurrentDownloads=1")
 
-            // Clean up
-            try? doRemoveImages(images: [alpine])
-
```
katiewasnothere, line 387: We're no longer guaranteeing that either of these images are not already pulled locally. And as it is right now, this test is not really validating that the value is respected, just that the command doesn't fail. If this test just cares about if the command can accept valid "max-concurrent-downloads" values, we can just try to pull the normal alpine image. 

We should probably add some unit tests at a different level. What do you think?

jglogan: In CI, the are guaranteed to be pulled locally in this test as long as nobody pulls those same images (which is why I used the symbols I did).

katiewasnothere (approved):

saehejkang:
Looks like this solves the issues that I am getting when writing tests for `image prune` 💯

Issue #460, opened 2025-08-06
Configure NAT66 for IPv6.


jglogan:
### Feature or enhancement request details

Once apple/containerization#445 is resolved we can perform the vmnet and interface configuration for NAT66 in containers.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
The basic network plumbing is complete, closing this issue and creating new issues for the remaining missing features.

PR #957, opened 2025-12-16
Use new IP/CIDR types from Containerization.

jglogan:
- Part of work for #460.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context
With CZ release 0.17.0, the IP and CIDR address types changed from String to IPv4Address and CIDRv4, respectively. This PR applies the corresponding adaptations to container.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Networks/NetworksService.swift
+++ b/Sources/Services/ContainerAPIService/Networks/NetworksService.swift
@@ -283,19 +283,24 @@ public actor NetworksService {
             serviceIdentifier,
         ]
 
-        if let subnet = (try configuration.subnet.map { try CIDRAddress($0) }) {
-            var existingCidrs: [CIDRAddress] = []
+        if let ipv4Subnet = (configuration.ipv4Subnet.map { $0 }) {
+            var existingCidrs: [CIDRv4] = []
             for networkState in networkStates.values {
                 if case .running(_, let status) = networkState {
-                    existingCidrs.append(try CIDRAddress(status.address))
+                    existingCidrs.append(status.ipv4Subnet)
                 }
             }
-            let overlap = existingCidrs.first { $0.overlaps(cidr: subnet) }
+            let overlap = existingCidrs.first {
+                $0.contains(ipv4Subnet.lower)
+                    || $0.contains(ipv4Subnet.upper)
+                    || ipv4Subnet.contains($0.lower)
+                    || ipv4Subnet.contains($0.upper)
+            }
             if let overlap {
-                throw ContainerizationError(.exists, message: "subnet \(subnet) overlaps an existing network with subnet \(overlap)")
+                throw ContainerizationError(.exists, message: "IPv4 subnet \(ipv4Subnet) overlaps an existing network with subnet \(overlap)")
             }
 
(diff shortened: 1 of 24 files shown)


Sources/Services/ContainerNetworkService/Attachment.swift
```diff
@@ -14,24 +14,58 @@
...
     /// The MAC address associated with the attachment (optional).
     public let macAddress: String?
 
-    public init(network: String, hostname: String, address: String, gateway: String, macAddress: String? = nil) {
+    public init(network: String, hostname: String, ipv4Address: CIDRv4, ipv4Gateway: IPv4Address, macAddress: String? = nil) {
         self.network = network
         self.hostname = hostname
-        self.address = address
-        self.gateway = gateway
+        self.ipv4Address = ipv4Address
```
jglogan, line 35: I don't think we need to worry about backward compatibility for Codable here since we're not storing these fields anywhere yet.

Sources/Services/ContainerSandboxService/SandboxService.swift
```diff
@@ -190,11 +190,10 @@ public actor SandboxService {
                 // a default /etc/hosts.
                 var hostsEntries = [Hosts.Entry.localHostIPV4()]
                 if !interfaces.isEmpty {
-                    let primaryIfaceAddr = interfaces[0].address
-                    let ip = primaryIfaceAddr.split(separator: "/")
+                    let primaryIfaceAddr = interfaces[0].ipv4Address
```
ajemory, line 193: Nitpick: I don't love the reference to interfaces[0]. Could make a struct to store interfaces[], but not much return on the effort -- definitely not blocking for this PR

Sources/Services/ContainerAPIService/Networks/NetworksService.swift
```diff
@@ -283,19 +283,24 @@ public actor NetworksService {
...
+        if let ipv4Subnet = (configuration.ipv4Subnet.map { $0 }) {
+            var existingCidrs: [CIDRv4] = []
             for networkState in networkStates.values {
                 if case .running(_, let status) = networkState {
-                    existingCidrs.append(try CIDRAddress(status.address))
+                    existingCidrs.append(status.ipv4Subnet)
                 }
             }
-            let overlap = existingCidrs.first { $0.overlaps(cidr: subnet) }
+            let overlap = existingCidrs.first {
```
ajemory, line 293: Wouldn't mind seeing this split into a separate function like it was previously, just to keep things clean -- not blocking

ajemory (approved):
Left a few nitpicky comments, but nothing blocking. Looks good!

Sources/Services/ContainerAPIService/Networks/NetworksService.swift
jglogan: Unfortunately the `overlaps(cidr:)` test got removed from the core CIDR types.

I'll extract this into an extension here when we add the IPv6 support.

Issue #956, opened 2025-12-16
[Bug]: env-file option does not support reading from named pipes


asgeirn:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

A common trick to protect secrets is to use named pipes as files to pass environment variables or other secrets to a process.

In bash and zsh, this is achieved by i.e. `--env-file <(echo ACCESS_KEY=secret)` -- and tools like 1Password do this for environment secrets.

This works just fine for podman, but `container run` throws the following error:

```
> ls -l .env
prw-------@ 1 asgeir.nilsen  staff  0 Dec 16 09:15 .env
> container run -it --rm --env-file .env containerimage:latest
Error: invalidArgument: "failed to read envfile at .env"
> container run -it --rm --env-file <(cat .env) containerimage:latest
Error: invalidArgument: "failed to read envfile at /dev/fd/11"
```

### Current behavior

Error: invalidArgument: "failed to read envfile at /dev/fd/11"

### Expected behavior

.env file is read as expected

### Environment

```markdown
- OS: macOS 26.2 (25C56)
- Xcode: xcode-select version 2416.
- Container: container CLI version 0.7.1 (build: release, commit: 420be74)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Bortnyak:
Hey. I want to contribute if nobody has already started working on it.

dcantah:
Closed via https://github.com/apple/container/pull/974

PR #964, opened 2025-12-17
Add support for reading env from named pipes

Bortnyak:
## Type of Change
- [x] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This is a fix for [issue#956](https://github.com/apple/container/issues/956)

`FileManager.default.contents(atPath:)` returns `nil` for named pipes (FIFOs) 
and process substitutions like `/dev/fd/XX` because:
1. It expects regular files with a known size
2. Named pipes are stream-based and block until data arrives

## Solution
Use `FileHandle(forReadingFrom:)` instead, which:
- Properly handles blocking I/O
- Works with named pipes, process substitutions, and regular files (mentioned in the [doc](https://developer.apple.com/documentation/foundation/filehandle))

## References
- https://stackoverflow.com/questions/60747062/ios-named-pipes-works-in-objc-does-not-work-in-swift-same-code

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs

## Result
```bash
-> % ./bin/container run -it --rm --env-file <(printf "VAR1=value1\nVAR2=value2\n") ubuntu:latest
root@725da974-da43-47cd-b86e-745265317ed9:/# echo $VAR2
value2
```

### DIFF
--- a/Sources/ContainerClient/Parser.swift
+++ b/Sources/ContainerClient/Parser.swift
@@ -110,17 +110,19 @@ public struct Parser {
         // This is a somewhat faithful Go->Swift port of Moby's envfile
         // parsing in the cli:
         // https://github.com/docker/cli/blob/f5a7a3c72eb35fc5ba9c4d65a2a0e2e1bd216bf2/pkg/kvfile/kvfile.go#L81
-        guard FileManager.default.fileExists(atPath: path) else {
-            throw ContainerizationError(
-                .notFound,
-                message: "envfile at \(path) not found"
-            )
-        }
 
-        guard let data = FileManager.default.contents(atPath: path) else {
+        let data: Data
+        do {
+            // Use FileHandle to support named pipes (FIFOs) and process substitutions
+            // like --env-file <(echo "KEY=value")
+            let fileHandle = try FileHandle(forReadingFrom: URL(fileURLWithPath: path))
+            defer { try? fileHandle.close() }
+            data = try fileHandle.readToEnd() ?? Data()
+        } catch {
             throw ContainerizationError(
                 .invalidArgument,
-                message: "failed to read envfile at \(path)"
+                message: "failed to read envfile at \(path)",
+                cause: error
             )
         }
 
--- a/Tests/ContainerClientTests/ParserTest.swift
+++ b/Tests/ContainerClientTests/ParserTest.swift
@@ -493,10 +493,12 @@ struct ParserTest {
         #expect {
             _ = try Parser.envFile(path: "/nonexistent/foo_bar_baz")
         } throws: { error in
-            guard let error = error as? ContainerizationError else {
+            guard let error = error as? ContainerizationError,
+                let cause = error.cause
+            else {
                 return false
             }
-            return error.description.contains("not found")
+            return String(describing: cause).contains("No such file or directory")
         }
     }
 
@@ -577,6 +579,41 @@ struct ParserTest {
             }
             return error.description.contains("no variable name")
         }
+    }
+
+    @Test
+    func testParseEnvFileFromNamedPipe() throws {
+        let pipePath = FileManager.default.temporaryDirectory
+            .appendingPathComponent("envfile-pipe-\(UUID().uuidString)")
+
+        // Create a named pipe (FIFO)
+        let result = mkfifo(pipePath.path, 0o600)
+        guard result == 0 else {
+            throw POSIXError(POSIXErrorCode(rawValue: errno) ?? .EPERM)
+        }
+        defer { try? FileManager.default.removeItem(at: pipePath) }
+
+        let group = DispatchGroup()
+
+        group.enter()
+        DispatchQueue.global().async {
+            do {
+                let handle = try FileHandle(forWritingTo: pipePath)
+                try handle.write(contentsOf: "SECRET_KEY=value123\n".data(using: .utf8)!)
+                try handle.close()
+            } catch {
+                Issue.record(error)
+            }
+            group.leave()
+        }
+
+        // Read from pipe (blocks until writer connects)
+        let lines = try Parser.envFile(path: pipePath.path)
+
+        // Wait for write to complete
+        group.wait()
+
+        #expect(lines == ["SECRET_KEY=value123"])
     }
 
     // MARK: Network Parser Tests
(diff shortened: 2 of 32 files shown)


Sources/ContainerClient/Parser.swift
```diff
@@ -110,14 +110,21 @@ public struct Parser {
...
-        }
 
-        guard let data = FileManager.default.contents(atPath: path) else {
+        let data: Data
+        do {
+            // Use FileHandle to support named pipes (FIFOs) and process substitutions
+            // like --env-file <(echo "KEY=value")
+            let fileHandle = try FileHandle(forReadingFrom: URL(fileURLWithPath: path))
+            defer { try? fileHandle.close() }
+            data = fileHandle.readDataToEndOfFile()
```
dcantah: This is deprecated. Let's use https://developer.apple.com/documentation/foundation/filehandle/readtoend()

Sources/ContainerClient/Parser.swift
```diff
@@ -110,14 +110,21 @@ public struct Parser {
...
+        } catch {
+            guard FileManager.default.fileExists(atPath: path) else {
```
dcantah: I would axe this, just edit the ContainerizationError below to:

```swift
throw ContainerizationError(
    .internalError,
    message: "failed to read envfile at \(path)",
    cause: error
)
```

Sources/ContainerClient/Parser.swift
Bortnyak: Fixed in https://github.com/apple/container/pull/964/commits/ccf552a401acb5c0c4217388018e0516a4df7306

Sources/ContainerClient/Parser.swift
Bortnyak: Fixed in https://github.com/apple/container/commit/ccf552a401acb5c0c4217388018e0516a4df7306

Bortnyak:
@dcantah Thank you for the review. I've added the requested changes. Please take a look.

Tests/ContainerClientTests/ParserTest.swift
```diff
@@ -579,6 +581,35 @@ struct ParserTest {
...
+        // Create a named pipe (FIFO)
+        let result = mkfifo(pipePath.path, 0o600)
+        guard result == 0 else {
+            throw POSIXError(POSIXErrorCode(rawValue: errno) ?? .EPERM)
+        }
+        defer { try? FileManager.default.removeItem(at: pipePath) }
+
+        // Write to pipe in background - must happen concurrently with read
+        // because pipes block until both ends are connected
+        async let writeTask: Void = {
```
dcantah: There's no point for this to be async, there's nothing async occurring in here.

dcantah:
This LGTM, but I'd like an integration test for this (e.g. try and spawn a container doing what's in the issue and verifying it works). Can be added to TestCLIRunCommand.swift

Tests/ContainerClientTests/ParserTest.swift
Bortnyak: > nothing async occurring in here


You are right, but this is kind of a workaround.
When marking writeTaks `async`, even if inside the closure there are no async tasks, this will create a child task that queues to an executor immediately (correct me if I'm wrong), and this gives the concurrency required by FIFO (both ends opened at the same time). Without that, the test deadlocks.
Another possible approach would be to use `DispatchQueue`, but it might be an unnecessary complication of things. 
If you know any other approach to test this, please suggest.
Thank you.

Bortnyak: To reproduce the deadlock:

```
@Test
func testParseEnvFileFromNamedPipe() throws {
    let pipePath = FileManager.default.temporaryDirectory
        .appendingPathComponent("envfile-pipe-\(UUID().uuidString)")

    let result = mkfifo(pipePath.path, 0o600)
    guard result == 0 else {
        throw POSIXError(POSIXErrorCode(rawValue: errno) ?? .EPERM)
    }
    defer { try? FileManager.default.removeItem(at: pipePath) }

    // This will hang here - waiting for a writer that never comes
    let lines = try Parser.envFile(path: pipePath.path)

    // Never reached
    let handle = try FileHandle(forWritingTo: pipePath)
    try handle.write(contentsOf: "SECRET_KEY=value123\n".data(using: .utf8)!)
    try handle.close()

    #expect(lines == ["SECRET_KEY=value123"])
}
```

dcantah: Sorry I should've put a bit more context. All the parser tests are non-async, and I'd prefer to just use Dispatch if we don't actually need any async. Something like this should work:

```swift
        let group = DispatchGroup()
        group.enter()
        DispatchQueue.global().async {
            do {
                let handle = try FileHandle(forWritingTo: pipePath)
                try handle.write(contentsOf: "SECRET_KEY=value123\n".data(using: .utf8)!)
                try handle.close()
            } catch {
                Issue.record(error)
            }
            group.leave()
        }

        let lines = try Parser.envFile(path: pipePath.path)
        group.wait()

```

Bortnyak: Thanks for example @dcantah,
added the changes + integration test
https://github.com/apple/container/pull/964/commits/5d66f3c4312584195f16d4d203e889ded312b564

dcantah:
You'll need to `make fmt` and re-push. Also it's complaining that the commits aren't signed (ssh, or gpg will work)

Bortnyak:
> You'll need to `make fmt` and re-push. Also it's complaining that the commits aren't signed (ssh, or gpg will work)

Done

dcantah:
@Bortnyak Still says some of these don't have signatures. Can you try squashing your commits down to one, signing that, and then pushing?

dcantah:
@Bortnyak I think the rebase was screwed up a bit, there's 32 files changed. Unrelated, but it also seems whatever email you're using for the commits is not assigned to your github account.

Bortnyak:
@dcantah I'm sorry, I messed things up while switching between laptops.
I've opened a new PR - https://github.com/apple/container/pull/974, let's close this one.

dcantah:
Fixed via https://github.com/apple/container/pull/974

PR #878, opened 2025-11-14
CLI: stop and delete should error on not found containers

dcantah:
Same as we'd done with kill.

```
➜  container git:(stop-kill-throw-error) ✗ ./bin/container stop foo bar baz
Warning! Running debug build. Performance may be degraded.
Error: internalError: "failed to stop container" (cause: "notFound: "container with ID baz not found"")
internalError: "failed to stop container" (cause: "notFound: "container with ID bar not found"")
internalError: "failed to stop container" (cause: "notFound: "container with ID foo not found"")
➜  container git:(stop-kill-throw-error) ✗ ./bin/container kill foo bar baz
Warning! Running debug build. Performance may be degraded.
Error: internalError: "failed to kill container" (cause: "notFound: "container with ID foo not found"")
internalError: "failed to kill container" (cause: "notFound: "container with ID bar not found"")
internalError: "failed to kill container" (cause: "notFound: "container with ID baz not found"")
➜  container git:(stop-kill-throw-error) ✗ ./bin/container delete foo bar baz
Warning! Running debug build. Performance may be degraded.
Error: internalError: "failed to delete container" (cause: "notFound: "container with ID foo not found"")
internalError: "failed to delete container" (cause: "notFound: "container with ID baz not found"")
internalError: "failed to delete container" (cause: "notFound: "container with ID bar not found"")
```

PR #979, opened 2025-12-22
Fix stop kill missing containers

zaineel:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
The `stop` and `kill` commands currently do not gracefully handle cases where container IDs are missing or not found. This PR adds proper error handling for missing containers and ensures validation tests run correctly.

Fixes #878

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerStop.swift
+++ b/Sources/ContainerCommands/Container/ContainerStop.swift
@@ -67,45 +67,53 @@ extension Application {
             let containers = try self.all
                 ? allContainers
                 : Self.containers(matching: containerIds, in: allContainers)
+            let set = Set<String>(containerIds)
+            let client = ContainerClient()
+            var containers = [ContainerSnapshot]()
+            if self.all {
+                containers = try await client.list()
+            } else {
+                containers = try await client.list().filter { c in
+                    set.contains(c.id)
+                }
+            }
 
             let opts = ContainerStopOptions(
                 timeoutInSeconds: self.time,
                 signal: try Signals.parseSignal(self.signal)
             )
-            let failed = try await Self.stopContainers(containers: containers, stopOptions: opts, log: log)
-            if failed.count > 0 {
-                throw ContainerizationError(
-                    .internalError,
-                    message: "stop failed for one or more containers \(failed.joined(separator: ","))"
-                )
-            }
+            try await Self.stopContainers(
+                client: client,
+                containers: containers,
+                stopOptions: opts
+            )
         }
 
-        static func stopContainers(containers: [ClientContainer], stopOptions: ContainerStopOptions, log: Logger) async throws -> [String] {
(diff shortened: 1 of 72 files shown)


zaineel:
@dcantah 
Could you please take a look at this PR? Thanks!

Sources/ContainerCommands/Container/ContainerStop.swift
```diff
@@ -45,25 +52,20 @@ extension Application {
...
-                containers = try await ClientContainer.list()
-            } else {
-                containers = try await ClientContainer.list().filter { c in
-                    set.contains(c.id)
-                }
-            }
+            let allContainers = try await ClientContainer.list()
+            let containers = try self.all
+                ? allContainers
+                : Self.containers(matching: containerIds, in: allContainers)
```
Ronitsabhaya75, line 69: I think we should handle ambiguous prefix matches here.

Currently, `allContainers.first(where:)` will silently pick the first matching container. If I have containers `abc` and `abd`, running `stop ab` might stop `abc` just because it appears first in the list.

It would be safer to filter all matches and throw an error if `matches.count > 1` (unless there's an exact match). This aligns with how Docker handles ambiguous prefixes.

Tests/ContainerCommandsTests/ContainerStopValidationTests.swift
```diff
@@ -0,0 +1,61 @@
...
+//===----------------------------------------------------------------------===//
+
+import ContainerClient
+import ContainerCommands
+import ContainerizationError
+import Testing
+
+@testable import ContainerCommands
+
+struct ContainerStopValidationTests {
```
Ronitsabhaya75, line 24: Could we add a test case for ambiguous prefixes?

We should verify that stop ab throws an error if both `abc` and `abd` exist. It's an important edge case to prevent accidental stops.

Sources/ContainerCommands/Container/ContainerStop.swift
```diff
@@ -21,6 +21,13 @@ import ContainerizationError
 import ContainerizationOS
 import Foundation
 
+package protocol ContainerIdentifiable {
```
Ronitsabhaya75, line 25: Refactoring this resolution logic into a shared helper is a great move. It keeps Stop and Kill consistent and makes testing way easier.

zaineel:
@Ronitsabhaya75 Added ambiguous-prefix handling with exact-match preference, plus tests for
  ambiguity and exact IDs. Ready for re‑review.

dcantah:
I think we can close this out as https://github.com/apple/container/pull/878 was merged. Thanks for trying to move this forward!

PR #980, opened 2025-12-22
Fix MAC address option typo in how-to documentation

claudeaceae:
## Summary
- Corrects the MAC address example command in the how-to guide to use the correct `--network` flag syntax instead of the incorrect `--mac-address` flag

## Details
The example on line 219 of `docs/how-to.md` was using `--mac-address 02:42:ac:11:00:02` which is incorrect. The correct syntax, as shown earlier in the same section (line 213), is `--network default,mac=02:42:ac:11:00:02`.

This was a documentation inconsistency where the example didn't match the documented syntax.

## Test plan
- [x] Verified the correct syntax is used earlier in the same section
- [x] Confirmed this matches the actual CLI implementation

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -216,7 +216,7 @@ container run --network default,mac=02:42:ac:11:00:02 ubuntu:latest
 To verify the MAC address is set correctly, run `ip addr show` inside the container:
 
 ```console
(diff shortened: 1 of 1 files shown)


claudeaceae:
Thanks for merging!

PR #987, opened 2025-12-23
Fix relative path resolution in entrypoint

### LINKED ISSUE #962: [Bug]: relative path in 'entrypoint' is not accepted
### I have done the following

(Sorry for not trying the `main` branch; I don't have a swift toolchain set up at the moment.)

(Note: this should possibly be categorized as a feature request rather than a bug; I wasn't sure.)

### Steps to reproduce

Run any image that users a relative path as an entrypoint, or explicitly specify a relative path in the `--entrypoint` argument.

As an example:

```
container run -w /usr/local/cargo/bin --entrypoint ./rustc rust:latest --version
```

### Current behavior

This error is printed:

```
Error: internalError: "failed to find target executable ./rustc"
```

### Expected behavior

This should run the `rustc` binary at `/usr/local/cargo/bin/rustc`, and print a version string. With `podman` or `docker`:

```
> podman run -w /usr/local/cargo/bin --entrypoint ./rustc rust:latest --version
rustc 1.92.0 (ded5c06cf 2025-12-08)
```

### Environment

```markdown
- OS: macOS 26.1
- Xcode: (not installed)
- Container: container CLI version 0.7.1 (build: release, commit: 420be74)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

ParkSeongGeun:
- Fixes #962 

## Type of Change
- [x] Bug fix (Coordinated with engine-side fix)
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Initially, this PR aimed to resolve relative paths within the CLI parser. However, following maintainer feedback (@jglogan), we've adopted a **"Cleaner Approach"** by moving the path resolution logic directly into the executor (`vmexec`) within the `apple/containerization` repository.

By handling the resolution at the engine level, we ensure that the OCI bundle configuration remains clean and consistent with industry standards (Docker/Podman), where the executor handles binary lookups relative to the `WORKDIR`.

**Changes in this PR:**
- **Reverted CLI-side resolution**: Keeps the `Parser.swift` logic simple, passing the raw entrypoint/cmd strings to the engine.
- **Added Passthrough Test**: Introduced `testProcessEntrypointRelativePathPassthrough` in `ParserTest.swift` to ensure that relative paths (e.g., `./uname`) are handed off to the executor without any unintended modifications by the CLI.

## Testing
- [x] **Unit Tests**: Ran `swift test --filter Passthrough` to confirm the CLI correctly passes relative paths untouched.
- [x] **Integration Tests**: Verified on **macOS 26.0** using a local build of `containerization` (via `swift package edit`).
  - **Scenario**: `container run --rm --workdir /bin --entrypoint ./uname alpine`
  - **Result**: Successfully executed and returned `Linux`. This confirms the end-to-end flow where the CLI passes `./uname` and the engine resolves it to `/bin/uname`.

### Coordination
The actual root-cause fix is submitted here: https://github.com/apple/containerization/pull/473

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "538f97f4c9123825536afb15abcaa4d9d865563e402bcb7fd26be69f55461c1d",
+  "originHash" : "b11a80bed0b568e35f4b538956344d483125e344c4931c362a2a31a2f7a8565b",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -16,7 +16,7 @@
       "location" : "https://github.com/apple/containerization.git",
       "state" : {
         "revision" : "452f354bac52ecbfe4a40b729880435a070c5a29",
-        "version" : "0.20.0"
+        "version" : "0.20.1"
       }
     },
     {
--- a/Package.swift
+++ b/Package.swift
@@ -23,7 +23,7 @@ import PackageDescription
 let releaseVersion = ProcessInfo.processInfo.environment["RELEASE_VERSION"] ?? "0.0.0"
 let gitCommit = ProcessInfo.processInfo.environment["GIT_COMMIT"] ?? "unspecified"
 let builderShimVersion = "0.7.0"
-let scVersion = "0.20.0"
+let scVersion = "0.20.1"
 
 let package = Package(
     name: "container",
--- a/Tests/ContainerAPIClientTests/ParserTest.swift
+++ b/Tests/ContainerAPIClientTests/ParserTest.swift
@@ -845,4 +845,22 @@ struct ParserTest {
             return error.description.contains("invalid property format")
         }
     }
+
+    // MARK: - Relative Path Passthrough Tests
+
+    @Test
+    func testProcessEntrypointRelativePathPassthrough() throws {
+        let processFlags = try Flags.Process.parse(["--cwd", "/bin"])
+        let managementFlags = try Flags.Management.parse(["--entrypoint", "./uname"])
+
+        let result = try Parser.process(
+            arguments: [],
+            processFlags: processFlags,
+            managementFlags: managementFlags,
+            config: nil
+        )
+
+        #expect(result.executable == "./uname")
+        #expect(result.workingDirectory == "/bin")
+    }
 }


Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -248,16 +248,45 @@ public struct Parser {
...
+                    result = entrypoint
+                }
             }
+
             if !arguments.isEmpty {
                 result.append(contentsOf: arguments)
             } else {
                 if let cmd = config?.cmd, !hasEntrypointOverride, !cmd.isEmpty {
-                    result.append(contentsOf: cmd)
+                    if let first = cmd.first, !first.hasPrefix("/") {
```
jglogan: Looks like we're repeating the same code in three places - extract method?

Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -248,16 +248,45 @@ public struct Parser {
...
-                result = [entrypoint]
+                if entrypoint.hasPrefix("/") {
+                    result = [entrypoint]
+                } else {
+                    let resolved = URL(fileURLWithPath: workingDir)
+                        .appendingPathComponent(entrypoint)
+                        .standardized
+                        .path
+                    result = [resolved]
+                }
```
jglogan: This doesn't cover all cases. Please add handling for bare commands:

```console
% container run --rm -w / --entrypoint ls alpine
Error: internalError: "failed to find target executable /ls"

% docker run --rm -w / --entrypoint ls alpine   
bin
dev
etc
```

Sources/Services/ContainerAPIService/Client/Parser.swift
ParkSeongGeun: @jglogan This is my mistake. I've refactored the path resolution logic into a private helper function called resolveExecutablePath

Sources/Services/ContainerAPIService/Client/Parser.swift
ParkSeongGeun: @jglogan Thank you for catching this case. 

I've updated the logic to ensure that bare commands (those without any slashes) are not resolved against the working directory.

ParkSeongGeun:
@jglogan Thanks for the thorough review

I've rebased the branch on the latest main and addressed all the feedback.

1. Refactored the path resolution logic into a helper function.
2. Added handling for bare commands to be resolved via PATH (executables without slashes are now ignored by the resolver).
3. Updated unit tests to cover these scenarios.

jglogan:
@ParkSeongGeun I think this works, but I'm curious as to the root cause for this error in your original failure case:

```console
% container run -it --rm -w /bin --entrypoint ./uname alpine:latest
Error: internalError: "failed to find target executable ./uname"
```

We might be able to address this down in containerization instead, such that we don't need to modify the path in the Bundle that gets generated. I feel that that is the cleaner approach.

Let me know if you're interested in looking at that. The bit of vmexec where the failure occurs is https://github.com/apple/containerization/blob/main/vminitd/Sources/vmexec/vmexec.swift#L74.

It might be as simple as pointing an AI assistant at a context of container, containerization, a failing command example, and this line of code and letting it crank out an answer for you. Or, troubleshoot it the old school way!

In either case see https://github.com/apple/container/blob/main/BUILDING.md for how to build `container` using a local copy of the containerization repo. I also noticed that the instructions for getting container to use your locally built init filesystem isn't there (you need this to see any changes that you make to vminitd or vmexec), so I filed #1030 to describe that procedure while I fix the docs.

ParkSeongGeun:
@jglogan I agree that fixing this problem in `containerization (vmexec)` is more cleaner approach.

I'll set up the local environment settings and wait for your guidance in #1030 on how to use a custom init filesystem. 

It's interesting to dive deeper into the root cause.

Thanks for your detailed explanation!

jglogan:
@ParkSeongGeun The missing instructions were in the PR. I just merged them into BUILDING.md.

TL;DR:

1. Check out a copy of containerization.
2. From your container project: `swift package edit --path /path/to/your/containerization containerization
3. `container system property set image.init vminit:latest`
4. In your container project: `make all install`
5. `container system stop ; container system start`

When you build container in step 4, the build runs `scripts/install-init.sh` that will do `make init` in your containerization project and install the resulting init image into your `container` content store.

The system property change tells container to use that local init filesystem image instead of the release image hosted on GHCR.

Don't forget to unedit the package and clear the system property to get back to using the normal containerization dependency and init fs.

ParkSeongGeun:
@jglogan

Apologies for the delay in getting back to you. I spent the last few days setting up a local development environment for both repositories and working through the guest VM build process to ensure the root cause fix was thoroughly verified.

As you suggested, I moved the path resolution logic directly into the executor (vmexec) in the apple/containerization repository. This ensures that the OCI bundle remains clean and the engine handles binary lookups relative to the working directory.

In this PR, I have
- Reverted the CLI-side normalization in Parser.swift to keep the logic focused on being a clean passthrough.
- Added a new unit test, testProcessEntrypointRelativePathPassthrough, in ParserTest.swift to ensure the CLI correctly hands off raw relative paths to the engine.

Verified the End-to-End flow on macOS 26.0 with the local engine build. 
- Running container run --workdir /bin --entrypoint ./uname alpine now successfully returns Linux.

I've submitted the engine-side fix here: https://github.com/apple/containerization/pull/473

jglogan:
@ParkSeongGeun Your test case should fail with the current 0.20.0, so add a change to 0.20.1 in Package.swift and then we should have a clean build.

ParkSeongGeun:
@jglogan Now I've updated Package.swift to version 0.20.1 and performed a targeted update of the Package.resolved file!

ParkSeongGeun:
@jglogan Thank you so much for your help!

I was interested in implementing virtualization with Swift, and I had a great experience with this contribution!

Issue #860, opened 2025-11-07
[Request]: Calculate/Include Full Size in JSON Output of images.


theetherGit:
### Feature or enhancement request details

> Right now I'm not sure it should be a bug or request.

Currently, in json format we have `descriptor` details and there is no full image size in json format while in table format we have calculated full image size in `size` field.

- List images with verbose flag

```sh 
container image ls -v
```
  **output**: 
  <img width="2087" height="135" alt="Image" src="https://github.com/user-attachments/assets/a2fdb2b7-3580-4821-a388-e99d68457df4"   />

- List images with json format
```sh
container image ls --format json
```
output:
<img width="2075" height="128" alt="Image" src="https://github.com/user-attachments/assets/60dffbb5-b818-4260-bd00-60ba553cbf40" />

```json
[
  {
    "reference": "docker.io/library/alpine:latest",
    "descriptor": {
      "digest": "sha256:4b7ce07002c69e8f3d704a9c5d6fd3053be500b7f1c69fc0d80990c2ad8dd412",
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "size": 9218
    }
  },
  {
    "reference": "docker.io/library/postgres:latest",
    "descriptor": {
      "digest": "sha256:41fc5342eefba6cc2ccda736aaf034bbbb7c3df0fdb81516eba1ba33f360162c",
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "size": 10229
    }
  },
  {
    "reference": "ghcr.io/apple/containerization/vminit:0.8.1",
    "descriptor": {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "size": 306,
      "digest": "sha256:666991717c0d26a211b400b09ad45ac84a9f64af8a08c2c64a54884e76d98954"
    }
  },
  {
    "reference": "ghcr.io/apple/containerization/vminit:0.9.1",
    "descriptor": {
      "size": 306,
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:6839bd6e230281df328571bd15929f8115303cba19737f3d888cc77a113e482d"
    }
  }
]
```

In json format, we get size from the `descriptor` instead of  sum of sizes from `descriptor `, `layers` and `manifest config`.

Code reference for how size of image is calculated in table format: https://github.com/apple/container/blob/main/Sources/ContainerCommands/Image/ImageList.swift#L83-L84

Expected:
In json format of images we should have full image size:

```json
{
[...]

Issue #709, opened 2025-10-03
[Bug]: Image size vs reclaimed space discrepancy


nigelpoulton:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. Pull or build an image
2. List images with verbose flag to see image size
3. Delete the image and reclaimed space will always be a lot more than the image size displayed in step 2

### Current behavior

Listing images shows the image size as ~110MB

```
$ container image ls -v
NAME         TAG   INDEX DIGEST                  OS      ARCH    SIZE
swift-demo   01    b130ac661a8a0488059ac4e6...   linux   arm64   110.7 MB
```

Deleting the image reclaims more space than the size of the image (reclaimed 446MB when the image was listed as only 110MB)

```
$ container image rm swift-demo:01
swift-demo:01
Reclaimed 446.4 MB in disk space
```

### Expected behavior

I would expect the reclaimed space to match the size of the image (as per `container image ls -v`) or at least be similar. Even if the value displayed in `container image ls` is compressed or something different, it's confusing that the two numbers are so different.

### Environment

```markdown
- OS: macOS 26.0.1 (25A362)
- Xcode: 26.0
- Container: 0.5.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #862, opened 2025-11-08
[image-list]: Add size field to json output

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #860

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

```
container image ls --format json | jq . 
```

```json
[
   "descriptor": {
      "digest": "sha256:320a7a4250aba4249f458872adecf92eea88dc6abd2d76dc5c0f01cac9b53990",
      "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
      "size": 988
    },
   "reference": "docker.io/library/python:3.9-slim-buster",
   "size": "42.8 MB"
]
```

PR #991, opened 2025-12-24
Fix image size calculation to show unpacked size

Bortnyak:
## Type of Change
- [x] Bug fix
- [X] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
https://github.com/apple/container/issues/709

Previously, `container image list -v` only showed compressed blob size. Now it shows:
Snapshot size (unpacked) **when available**
Compressed blob size **as fallback** when image hasn't been run yet

This better reflects actual storage usage and matches Docker's behavior of showing unpacked size.
Tests use >= comparisons because filesystems allocate space in fixed blocks (4KB on macOS), making exact byte counts unreliable.

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs

p.s.: if this doesn't match with the maintainers' vision of how image size should be represented, feel free to close this PR :)


Bortnyak:
Hey @dcantah, take a look at this PR please.

theetherGit:
@Bortnyak Does this PR fixes https://github.com/apple/container/issues/860 too?

Bortnyak:
> @Bortnyak Does this PR fixes #860 too?

No, but if the current one will be accepted, then the [pr](https://github.com/apple/container/pull/862) should be tweaked.

Bortnyak:
Hey @jglogan @dcantah 
Can someone take a look please?

jglogan:
@Bortnyak I'll try to review it next week. Thanks for the reminder, and sorry for the delay!

Could you rebase it PR to resolve the conflicts?

Bortnyak:
Thanks @jglogan.
Done

jglogan:
@Bortnyak Sorry, still owe this a review, I haven’t forgotten

Sources/ContainerCommands/Image/ImageList.swift
```diff
@@ -77,7 +77,9 @@ extension Application {
                     }
 
                     let created = config.created ?? ""
-                    let size = descriptor.size + manifest.config.size + manifest.layers.reduce(0, { (l, r) in l + r.size })
+                    let compressedSize = descriptor.size + manifest.config.size + manifest.layers.reduce(0, { (l, r) in l + r.size })
```
jglogan: See the implementation in #862, if that moves into the image client and we merge that, you could use the same logic for the OCI image size, and then you could extend it so it instead returns a struct containing say, ociImageSize and snapshotSize.

jglogan (commented):
@Bortnyak I think this looks fine. See my comment about maybe waiting for a common image size func in ClientImage from #862, and extending it for the snapshot size you're doing.

Sources/ContainerCommands/Image/ImageList.swift
saehejkang: Please refer to #1098 now :)

Bortnyak: Thanks to both of you. I'll wait until #1098 is merged and will use that logic

jglogan: @Bortnyak you should be good to go now!

jglogan:
@Bortnyak Sorry to drag this out, I think it'll need one last rebase to address the big change to the ContainerClient type (see #1139).

Sources/ContainerCommands/Image/ImageList.swift
Bortnyak: Hey @jglogan 

I've added the change consolidating size retrieval into a single method that returns both ociImageSize and snapshotSize in an ImageSizeInfo struct, following your suggestion.
Check the PR, please.

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -447,6 +452,25 @@ extension ClientImage {
...
+        let request = Self.newRequest(.snapshotSize)
+
+        try request.set(description: description)
+        try request.set(platform: platform)
+
+        let response = try await client.send(request)
+        return response.uint64(key: .imageSize)
+    }
+
+    /// Returns both OCI image size and snapshot size for the specified platform.
```
saehejkang, line 479: Let's update this comment follow the `docc` format style. You can follow how it is done in the `getFullImageSize` function in this file.

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -447,6 +452,25 @@ extension ClientImage {
         return fs
     }
 
+    public func getSnapshotSize(platform: Platform) async throws -> UInt64 {
```
saehejkang, line 468: nit: function docc updates

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -206,6 +206,15 @@ extension ImagesService {
         let img = try await self._get(description)
         return try await self.snapshotStore.get(for: img, platform: platform)
     }
+
+    public func getSnapshotSize(description: ImageDescription, platform: Platform) async throws -> UInt64 {
```
saehejkang, line 414: nit: function docc updates

saehejkang:
The `command-reference` docs could slightly be updated to add a note about the updates to the size that is outputted when running `container image list`.

https://github.com/apple/container/blob/main/docs/command-reference.md#container-image-list-ls:~:text=Image%20Management-,container,-image%20list%20(ls

saehejkang:
Are these the different types of image sizes that can be outputted?

- Snapshot Size = Unpacked Size = Full Size (which is outputted currently)
- Compressed Blob Size

I am thinking it might be worthwhile to output both the sizes in the `--verbose` output just so users can see how they differ.

Bortnyak:
> Are these the different types of image sizes that can be outputted?
> 
> * Snapshot Size = Unpacked Size = Full Size (which is outputted currently)
> * Compressed Blob Size
> 
> I am thinking it might be worthwhile to output both the sizes in the `--verbose` output just so users can see how they differ.

Hey @saehejkang 

Yup, those sizes are different: 
**Snapshot Size** - the unpacked/decompressed size of the image after its layers have been extracted into a snapshot;
**Compressed Blob Size** - the compressed size of the OCI image layers as stored on disk (this is what was downloaded/transferred);

Yes, I agree that it would be useful to output both in `--verbose` mode. I'll be more than happy to add this in a subsequent PR.

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -410,6 +410,15 @@ extension ImagesService {
...
+        self.log.info("ImagesService: \(#function) - description: \(description), platform: \(String(describing: platform))")
```
saehejkang, line 415: nit: this should be updated to follow patterns of the other `log` usages in this file, like [here](https://github.com/apple/container/blob/df25fe25bb722fb1637d976c687d8a74d5ef68da/Sources/Services/ContainerImagesService/Server/ImagesService.swift#L343)

saehejkang:
> Yes, I agree that it would be useful to output both in --verbose mode. I'll be more than happy to add this in a subsequent PR.

Sounds good!

saehejkang (approved):
This LGTM but needs a review from @jglogan

Issue #834, opened 2025-11-03
[Bug]: Stale nameserver in buildkit after network migration


vsarunas:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. Build a container that will cause buildkit to be started
2. container system stop and start had caused me network subset jump to happen; 192.168.64.0/24; then 192.168.65.0/24 and finally only a single network 192.168.66.0/24. Created for this https://github.com/apple/container/issues/835
3. While building again, buildit was started again and ended up with name server of `192.168.64.1`, ```container inspect buildkit | jq```:
```json
[
  {
    "status": "running",
    "configuration": {
      "ssh": false,
      "dns": {
        "nameservers": [
          "192.168.64.1"
        ],
        "searchDomains": [],
        "options": []
      },
// [...]
    "networks": [
      {
        "address": "192.168.66.3/24",
        "network": "default",
        "gateway": "192.168.66.1",
        "hostname": "buildkit"
      }
    ]
```

```
$ container network list
NETWORK  STATE    SUBNET
default  running  192.168.66.0/24
```

The end result is that container is having DNS failures and is not able to 

### Current behavior

Invalid stale name server is set

### Expected behavior

Avoid network changes during start/stop of system.
Update the name servers in the container if the network has migrated.

### Environment

```markdown
- OS: 26.0.1 (25A362)
- Xcode: 
- Container: container CLI version 0.6.0 (build: release, commit: a23bcf0)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #696, opened 2025-09-30
[Request]: Support a flag to force tcp DNS


byjrack:
### Feature or enhancement request details

So been seeing where UDP DNS queries are not getting connected back to the bridge network for containers.  If I do the same query via TCP using host, nslookup, dig, drill, whatever there are no issues.  I am pretty positive it's the device network controls my organization have out there, but can't find the specific one yet.  I can just see where a dns response via tcp the next "hop" is mdnsresponder but for udp the flow just ends.

I tried to update resolv.conf with `options use-vc` but it had no effect on the subsequent internal calls.  I need to test more, but been doing most tests using musl based images and not glibc which i believe is the issue there.

```
$ host google.com
;; communications error to 192.168.64.1#53: connection refused
;; communications error to 192.168.64.1#53: connection refused
;; no servers could be reached
$ host -T google.com
google.com has address 142.251.179.102
google.com has address 142.251.179.100
google.com has address 142.251.179.113
google.com has address 142.251.179.138
google.com has address 142.251.179.139
google.com has address 142.251.179.101
google.com has IPv6 address 2607:f8b0:4004:c09::71
google.com has IPv6 address 2607:f8b0:4004:c09::64
google.com has IPv6 address 2607:f8b0:4004:c09::8b
google.com has IPv6 address 2607:f8b0:4004:c09::66
google.com mail is handled by 10 smtp.google.com.
google.com has HTTP service bindings 1 . alpn="h2,h3"
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

byjrack:
And as some additional context no issues on Docker Desktop, Colima, or Rancher Desktop.  All those have used vpnkit or some variant of that project to try and avoid network control tripping hazards.  

I did spot that `--dns-option` was available which seemed to correlate with adding a `opts use-vc` to resolv.conf.  Still didn't fix things, but feels like maybe i can nudge dns to use TCP during init.

jglogan:
There is a bug in `--dns-option` down in the containerization layer that got fixed with apple/containerization#303.

I created #697 to pick up this fix, so you can test with main and see if `options use-vs` works.

byjrack:
So gave the unsigned build a shot.

It does fix the the `opts` to `options`, but tried a couple container images (musl and glibc) and all still used UDP for queries by default. Same experience i had when I added the option manually to resolv.conf.  I had wondered if it was init based, but I think I figured out the cause.

[...]

PR #998, opened 2025-12-27
fix(dns): improve DNS server reliability and compatibility

### LINKED ISSUE #817: [Bug]: Investigate DNS search path / resolver priority improvements.
### I have done the following

### Steps to reproduce

As pointed out by @austinlparker, there's a Tailscale issue that's open that identifies some possible ways for `container` to play more nicely when resolving plain hostnames.

The post in the issue containing the analysis is: https://github.com/tailscale/tailscale/issues/17096#issuecomment-3416342435

### Current behavior

Tailscale short hostnames (e.g., stan, kenny) to fail resolution, even though Tailscale MagicDNS is properly configured and full domain names work correctly.

### Expected behavior

MagicDNS should function properly.

### Environment

```markdown
- OS: macOS 26.1
- Xcode: Xcode 26
- Container: 0.6.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

swiftveteran:
Fixes #817, #696, #834

- Bind DNS server to network gateway addresses instead of only localhost
- Detect stale nameserver after network migration and refresh resolvers
- Add --dns-tcp flag to force TCP DNS queries for unreliable networks
- Inherit host search domains for Tailscale/VPN compatibility

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

The DNS server now binds to network gateway addresses instead of only localhost, allowing containers to resolve hostnames correctly across different network configurations. This also fixes stale nameserver detection after network migration by refreshing resolvers when the network changes. A new --dns-tcp flag forces TCP for DNS queries on unreliable networks, and host search domains are now inherited for compatibility with Tailscale and other VPN configurations.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerClient/HostDNSResolver.swift
+++ b/Sources/ContainerClient/HostDNSResolver.swift
@@ -95,6 +95,27 @@ public struct HostDNSResolver {
             .sorted()
     }
 
+    /// Returns the host system's DNS search domains from /etc/resolv.conf.
+    public static func getHostSearchDomains() -> [String] {
+        let resolvConfPath = "/etc/resolv.conf"
+        guard let content = try? String(contentsOfFile: resolvConfPath, encoding: .utf8) else {
+            return []
+        }
+
+        var searchDomains: [String] = []
+        for line in content.components(separatedBy: .newlines) {
+            let trimmed = line.trimmingCharacters(in: .whitespaces)
+            guard !trimmed.hasPrefix("#") else { continue }
+
+            if trimmed.hasPrefix("search ") {
+                let domains = trimmed.dropFirst(7).split(whereSeparator: { $0.isWhitespace })
+                searchDomains.append(contentsOf: domains.map { String($0) })
+            }
+        }
+
+        return searchDomains
+    }
+
     /// Reinitializes the macOS DNS daemon.
     public static func reinitialize() throws {
         do {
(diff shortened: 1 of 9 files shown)


Sources/ContainerClient/HostDNSResolver.swift
```diff
@@ -95,6 +95,27 @@ public struct HostDNSResolver {
             .sorted()
     }
 
+    /// Returns the host system's DNS search domains from /etc/resolv.conf.
+    public static func getHostSearchDomains() -> [String] {
```
jglogan, line 99: What's in `/etc/resolv.conf` isn't necessarily going to reflect the actual DNS configuration of the system.

You can see this in `resolv.conf` itself:

```
% more /etc/resolv.conf
#
# macOS Notice
#
# This file is not consulted for DNS hostname resolution, address
# resolution, or the DNS query routing mechanism used by most
# processes on this system.
#
# To view the DNS configuration used by this system, use:
#   scutil --dns
#
# SEE ALSO
#   dns-sd(1), scutil(8)
#
# This file is automatically generated.
#
search apple.com
nameserver <redacted>
nameserver <redacted>
```

jglogan:
It looks like this change is going to make it so that we'll have a listener for our scoped DNS service on port 2053 on each network gateway. Who will actually send requests to these listeners?

Issue #992, opened 2025-12-26
[Bug]: Cannot publish same port with both TCP and UDP


andreyrd:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Try to publish the same port number on UDP and TCP using, for example: `-p 53:53/udp -p 53:53/tcp`.

### Current behavior

Refuses to start. Get the error shown in the log section below.

### Expected behavior

It should allow forwarding TCP and UDP on the same port number. This is supported in Docker, Podman, etc and even by the underlying container backend it seems. I think it's just the [validation done here](https://github.com/apple/container/blob/main/Sources/ContainerClient/Utility.swift#L75) that needs to account for the protocol as well. I fixed this locally myself and tested, but I haven't touched Swift in years so my approach was pretty ugly, I changed the Set to be strings and just shoved the protocol and port into a string. This is where Swift tuples with Hashable members being Hashable would really come in handy.

### Environment

```markdown
- OS: 26.2 (25C56)
- Xcode: 26.2 (17C52)
- Container: 0.7.1
```

### Relevant log output

```shell
Error: invalidArgument: "host ports for different publish port specs may not overlap"
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1000, opened 2025-12-29
Fix port validation to allow same port for different protocols (#992)

iko1:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes: #992 

Port validation previously rejected valid configurations when the same port number was used for different protocols (TCP and UDP). For example:

`-p 1024:1024/udp -p 1024:1024/tcp`

Although this is a valid and common use case, the validation logic treated it as a conflict.

To fix this, I updated the validation key to include the protocol name. The validation now checks for overlapping port numbers only within the same protocol, rather than across all protocols.

This change enables binding the same port number for both TCP and UDP, aligning the validation behavior with real-world networking requirements.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/ContainerClientTests/UtilityTests.swift
+++ b/Tests/ContainerClientTests/UtilityTests.swift
@@ -127,7 +127,7 @@ struct UtilityTests {
             "8081:8081",
             "8081:8081/udp",
         ])
-        #expect(result.count == 4)
+        #expect(result.count == 6)
         try Utility.validPublishPorts(result)
     }
 }


Tests/ContainerClientTests/UtilityTests.swift
```diff
@@ -116,4 +116,16 @@ struct UtilityTests {
         }
 
     }
+
+    @Test
+    func testPublishPortsSamePortDifferentProtocols() throws {
+        let result = try Parser.publishPorts([
+            "8080:8080/tcp",
```
dcantah, line 123: Can you add a test case where the protocol isn't chosen so we can verify the default (tcp)? e.g.

"8081:8081",
"8081:8081/udp"

dcantah:
@iko1 Can you fix the build issues?

Tests/ContainerClientTests/UtilityTests.swift
iko1: Sure

iko1:
> @iko1 Can you fix the build issues?

I don't think the current build issue is related to my change.

jglogan:
@iko1 it's definitely not you. There's some weirdness in the OSS header checks due to the new year, and to the CI builds merging PRs against main before the header checks.

jglogan (approved):

jglogan (approved):
lgtm, thank you for the fix!

jglogan:
@iko1 it looks like you might need to push this again such that all commits have verified signatures

iko1:
> @iko1 it looks like you might need to push this again such that all commits have verified signatures

@jglogan Fixed, all commits in this PR are now re-signed.

PR #1007, opened 2026-01-02
Autoconfigure IPv6 DNS proxy.

### LINKED ISSUE #1004: [Request]: IPv6: autoconfigure DNS proxy.

When the sandbox service bootstraps the container VM, it autoconfigures the IPv4 DNS server address to use the vmnet proxy server at the gateway address of the first running NAT bridge found. We should also autoconfigure the IPv6 nameserver so that IPv6-only containers can function.

To do this for the IPv6 DNS proxy, it's necessary to wait for the NAT bridge to be fully materialized (at the point where the IPv6 address is assigned and secured as viewed in `ifconfig` or `scutil`). This only happens after the first interface is attached to the network, which only happens after the VM with that first interface has started.

Discovering the address isn't really straightforward today - it's necessary to use the `SystemConfiguration` framework to find the network whose IPv6 prefix matches the `container` network, registering for notifications and waiting until it appears in the case of the first container on the network.

As part of this work we should document that SLAAC only works on IPv6 subnets with a /64 prefix. For other networks the address must be configured manually, which we do not support at present (see #751).

### Code of Conduct

jglogan:
- Closes #1004.
- If DNS is not explicitly configured in the sandbox service, bootstrap the container and then find the IPv6 prefix for the first running network. Monitor the system configuration dynamic store until IPv6 is fully initialized for the network (this can take seconds for the first container attaching to the network), and then add the DNS proxy address for that network to the container DNS configuration.

TODO:
- [ ] After apple/containerization#466 is resolved, update the container DNS with the discovered proxy address.
- [ ] Add integration test.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Adds a feature needed for IPv6-only containers.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "538f97f4c9123825536afb15abcaa4d9d865563e402bcb7fd26be69f55461c1d",
+  "originHash" : "1e0a816e5d6688284f56a007aac9f1b5f5d8930740c3628048d8d405559a71cf",
   "pins" : [
     {
       "identity" : "async-http-client",
--- a/Sources/Services/ContainerSandboxService/Server/IPv6DNSLocator.swift
+++ b/Sources/Services/ContainerSandboxService/Server/IPv6DNSLocator.swift
@@ -1,0 +1,77 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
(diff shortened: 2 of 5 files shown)


jglogan:
Closing this PR, since we plan to handle IPv6 DNS autoconfig entirely in the guest agent.

PR #1008, opened 2026-01-02
CLI: Skip unresolvable digests

dcantah:
We error today, just skip them like we do in the --verbose codepath.


Sources/ContainerCommands/Image/ImageList.swift
```diff
@@ -133,7 +133,12 @@ extension Application {
             for image in images {
                 let processedReferenceString = try ClientImage.denormalizeReference(image.reference)
                 let reference = try ContainerizationOCI.Reference.parse(processedReferenceString)
-                let digest = try await image.resolved().digest
+                let digest: String
+                do {
```
jglogan, line 137: In the verbose path I thought the reason for silently continuing was because the image index has all the platforms, but not all the platform content hasn't been fetched. Here it feels like we're masking some kind of upstream error – perhaps we should log to stderr and continue at L140 so we can find out why this happens?

PR #1021, opened 2026-01-05
Forward SSL env variables to containerization

### LINKED ISSUE #305: [Bug]: failed to verify certificate with self signed certs
### I have done the following

### Steps to reproduce

- Docker registry that has self signed cert

### Current behavior

```
❯ container build -t test:test . --arch amd64
[+] Building 9.1s (4/13)
 => [resolver] fetching image...internal.cr/internal/base                                                                                                                                                                                                               0.0s
 => [internal] load .dockerignore                                                                                                                                                                                                                                             0.4s
 => => transferring context: 2B                                                                                                                                                                                                                                               0.4s
 => ERROR [linux/amd64  1/10] FROM internal.cr/internal/base:latest@sha256:cbc2d785b823ee730117c806ef73da8d6ce2aa774070e37221146041d66ee1bb                                                                                                                             0.1s
 => => resolve internal.cr/internal/base:latest@sha256:cbc2d785b823ee730117c806ef73da8d6ce2aa774070e37221146041d66ee1bb                                                                                                                                                 0.1s
 => CANCELED [internal] load build context                                                                                                                                                                                                                                    0.1s
 => => transferring context:                                                                                                                                                                                                                                                  0.1s
------
 > [linux/amd64  1/10] FROM internal.cr/internal/base:latest@sha256:cbc2d785b823ee730117c806ef73da8d6ce2aa774070e37221146041d66ee1bb:
------
Error: unknown (2): failed to solve: failed to load cache key: failed to do request: Head "https://internal.cr/v2/internal/base/manifests/sha256:cbc2d785b823ee730117c806ef73da8d6ce2aa774070e37221146041d66ee1bb": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

### Expected behavior

Use CA thats trusted by Keychain

### Environment

```markdown
- OS: 15.5 (24F74)
- Xcode: 15.5
- Container: container CLI version 0.2.0 (build: release, commit: 3b5c253)
```

### Relevant log output

```shell

```

### Code of Conduct

sebimarkgraf:
- Fixes #305.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This is a follow-up to https://github.com/apple/containerization/pull/402 allowing to easily use `container` with custom SSL proxies.
Especially in corporate setups, using an HTTPS Proxy is quite common for varying reasons and by following common coventions of env variables, `container` can be used easily without additional setup.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

Tagging @dcantah as he did the upstream PR review and this here is only the plumbing not the core change.

### DIFF
--- a/Sources/ContainerCommands/System/SystemStart.swift
+++ b/Sources/ContainerCommands/System/SystemStart.swift
@@ -72,7 +72,7 @@ extension Application {
 
             var env = PluginLoader.filterEnvironment(
                 env: ProcessInfo.processInfo.environment,
-                additionalAllowKeys: ["SSL_CERT_FILE", "REQUESTS_CA_BUNDLE", "CURL_CA_BUNDLE"],
+                additionalAllowKeys: ["SSL_CERT_FILE"],
             )
             env[ApplicationRoot.environmentName] = appRoot.path(percentEncoded: false)
             env[InstallRoot.environmentName] = installRoot.path(percentEncoded: false)


jglogan:
@sebimarkgraf If there isn't one, could you make an issue for this describing with examples what this enhances or fixes?

Also, please rebase onto the latest main as `ContainerClient` is gone, updating your imports if needed:

- `ContainerAPIClient` has the actual client code
- `ContainerResource` has the core types
- `ContainerSandboxServiceClient` has the sandbox client code (unlikely you're using this though)

sebimarkgraf:
I will rebase and update today.

This fixes https://github.com/apple/container/issues/305 by being able to provide custom certificates via the standard env variables.
I can write a more extensive feature issue if you think this is helpful.

jglogan:
Thanks! I'll update the PR description.

dcantah:
I agree with the above, sorry for approving the other two over in containerization. Those two seem odd looking back now, but the ssl_* ones are fairly standard across the board.

jglogan:
> We should only use SSL_CERT_FILE (Go honors that one)

Actually I think Go might only use that env variable on Linux.

@sebimarkgraf @dcantah let's use this issue to circle back on both this and apple/containerization#402.

Has this fix been verified in an environment that uses both a registry with a self-signed cert and registries that need the system trust store? Will the utility code only make it such the either a self-signed cert will be used or the system trust store, but not both?

```swift
public enum TLSUtils {

    public static func makeEnvironmentAwareTLSConfiguration() -> TLSConfiguration {
        var tlsConfig = TLSConfiguration.makeClientConfiguration()

        // Check standard SSL environment variables in priority order
        let customCAPath =
            ProcessInfo.processInfo.environment["SSL_CERT_FILE"]
            ?? ProcessInfo.processInfo.environment["CURL_CA_BUNDLE"]
            ?? ProcessInfo.processInfo.environment["REQUESTS_CA_BUNDLE"]

        if let caPath = customCAPath {
            tlsConfig.trustRoots = .file(caPath)
        }
        // else: use .default

        return tlsConfig
    }
}
```

sebimarkgraf:
Agree with the reasoning, included them as these are the ones I usually look for.

Reduced the list here to `SSL_CERT_FILE`.


Fix has only been tested in an environment where my proxy has a self-signed cert via `SSL_CERT_FILE`.
Let me try to create a script

jglogan:
Thanks! I don't have a problem with us getting this feature in place as the use case (I have a private/local registry with a self-signed certificate and want to use it without having to allow the cert to be used system-wide).

I wonder about the naming. `SSL_CERT_FILE` is used by all kinds of libraries (libssl, Go, various Rust TLS libraries) with varying meaning and applicability. It might make more sense to use a container-specific variable name that indicates the semantics.

If we chose `CONTAINER_CERT_FILE`, setting it would override the system trust store, which I believe is consistent with what apple/containerization#402 does. 

If we chose `CONTAINER_ADDITIONAL_CERT_FILE`, setting it would use both the system trust store and the additional cert, which would require a change to the TLS utility function you created.

There are of course other more complicated possibilities but let's make sure we (including @dcantah and anyone else interested!) all agree on the use case and then deliver the simplest feature that satisfies it.

PR #1023, opened 2026-01-05
CLI: Add 'container cp' command for file transfer

suyashbhawsar:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

This PR adds the `container cp` command to copy files and directories between containers and the local filesystem, similar to `docker cp`. This is a commonly requested feature that enables easier file transfer workflows with containers.

**Why is this change needed?**
- Users often need to copy configuration files, logs, or data between containers and the host
- Currently there's no built-in way to transfer files without mounting volumes
- This brings feature parity with Docker's `cp` command

## Implementation Details

The implementation uses tar archives for file transfer (same approach as Docker):
- `copyToContainer`: Local tar creates archive, streams to container tar for extraction
- `copyFromContainer`: Container tar creates archive, streams to local tar for extraction

**New files:**
- `CopyPath.swift` - Path parsing utility (distinguishes `container:path` vs local paths)
- `ContainerCopy.swift` - Main command implementation
- `TestCLICopy.swift` - 45 unit tests for edge cases

**Features:**
- Copy files/directories to and from running containers
- Rename during copy (different source/destination names)
- `-L/--follow-link` flag to dereference symlinks
- `-a/--archive` flag to preserve permissions (uid/gid)
- `-q/--quiet` flag to suppress progress output

**Usage examples:**
```bash
container cp mycontainer:/app/config.json ./config.json
container cp ./local-file.txt mycontainer:/tmp/file.txt
container cp -L mycontainer:/var/log/ ./logs/
```

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

**Unit tests (45 tests):**
- Path parsing for various formats (absolute, relative, container paths)
- Unicode filenames (Chinese characters, emoji)
- Special characters (spaces, brackets, quotes, etc.)
- Hidden files, symlinks, edge cases

**Integration tests performed:**
- Basic file copy (to/from container)
- Directory copy with nested structure
- Files with spaces, unicode, special characters
- Symlink handling with/without -L flag
- Error cases (non-existent paths, stopped containers)
- Rename during copy

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -75,6 +75,7 @@ let package = Package(
                 .product(name: "ContainerizationExtras", package: "containerization"),
                 .product(name: "ContainerizationOS", package: "containerization"),
                 "ContainerBuild",
+                "ContainerCommands",
                 "ContainerResource",
             ],
             path: "Tests/CLITests"
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -50,6 +50,7 @@ public struct Application: AsyncParsableCommand {
             CommandGroup(
                 name: "Container",
                 subcommands: [
+                    ContainerCopy.self,
                     ContainerCreate.self,
                     ContainerDelete.self,
                     ContainerExec.self,
--- a/Sources/ContainerCommands/Container/ContainerCopy.swift
+++ b/Sources/ContainerCommands/Container/ContainerCopy.swift
@@ -1,0 +1,458 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2025 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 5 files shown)


dcantah:
@suyashbhawsar Hi! We really appreciate this work, but I think the right place for this change (and it would require quite a bit more work) would be in https://github.com/apple/containerization. copying in and out should be on the core types and not rely on the container having tar installed, it should be done by the guest agent.

suyashbhawsar:
> @suyashbhawsar Hi! We really appreciate this work, but I think the right place for this change (and it would require quite a bit more work) would be in https://github.com/apple/containerization. copying in and out should be on the core types and not rely on the container having tar installed, it should be done by the guest agent.

Thanks, I'll take a look at the [containerization](https://github.com/apple/containerization) repo and will integrate this there

dcantah:
@suyashbhawsar We have quite a few design choices (swift does not have a native tar library, and we don't want to link any external libraries to the guest agent if we can avoid it) to work through on that end unfortunately. We want to encourage contribution, but this is probably not the best issue to start with :(. Let me know if there's something in here you'd like to tackle and I'd be happy to help. Just give me a ping

dcantah:
I'm going to close this out, I think we'll try and use this library to do the copies between the host<->guest after we're happy with the implementation. Once again, thanks for the work!

PR #1038, opened 2026-01-10
fix: use pax instead of tar for pkg payload extraction

manuschillerdev:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
It is common to have `gnu-tar` alongside other GNU tools installed and aliased for compatibility reasons.
However, this breaks the current make build. The following error happened when GNU tar tried to extract the pkg Payload (which is cpio+gzip, not tar) during `make all` > `install` target, extracting pkg Payload to `bin/` and `libexec/`:

```
gtar: This does not look like a tar archive
gtar: Skipping to next header
gtar: Exiting with failure status due to previous errors
```
The build succeeded, the pkg was created, but the extraction silently failed, leaving `bin/container` missing.

The proposed fix uses BSD-only binaries (no GNU equivalents that are commonly aliased), making the Makefile more portable.

Why `gunzip | cpio` instead of `tar`:
- macOS pkg Payloads are gzip-compressed cpio archives, not tar
- BSD tar (macOS default) can read them because libarchive auto-detects formats
- GNU tar (`gtar`) cannot
## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ install: installer-pkg
 	@if [ -z "$(SUDO)" ] ; then \
 		temp_dir=$$(mktemp -d) ; \
 		xar -xf $(PKG_PATH) -C $${temp_dir} ; \
-		(cd $${temp_dir} && tar -xf Payload -C "$(DEST_DIR)") ; \
+		(cd "$(DEST_DIR)" && pax -rz -f $${temp_dir}/Payload) ; \
 		rm -rf $${temp_dir} ; \
 	else \
 		$(SUDO) installer -pkg $(PKG_PATH) -target / ; \


Ronitsabhaya75:
@manuschillerdev This looks good, but switching the cd from `temp_dir` to `DEST_DIR` concerns me. What if `temp_dir` is relative? The old `cd $${temp_dir}` made the Payload reference safe. Could you verify `temp_dir` is always absolute, or restructure the pipe to be defensive?

manuschillerdev:
@Ronitsabhaya75 thx for the review!

I landed on `pax` as a final suggestion. It is a single cmd, has proper error propagation, is native to macOS, and is unlikely to be aliased. Would that be fine with you?

```bash
 # changes cd order, but safe since mktemp always returns absolute path 
 (cd "$(DEST_DIR)" && pax -rz -f $${temp_dir}/Payload)
```

other options considered, but rejected (see reasons in comments)

```bash
 # masks gunzip errors (pipe returns last command's exit status)
 (cd $${temp_dir} && gunzip -c Payload | (cd "$(DEST_DIR)" && cpio -idm))

 # not POSIX (bash-only)
 (cd $${temp_dir} && set -o pipefail && gunzip -c Payload | (cd "$(DEST_DIR)" && cpio -idm))

 # extra temp file, more complex than original
 (cd $${temp_dir} && gunzip -c Payload > Payload.cpio && (cd "$(DEST_DIR)" && cpio -idm) < Payload.cpio)
```

Ronitsabhaya75:
@manuschillerdev Great analysis! I like the `pax` solution it's cleaner and you're right that it avoids the pipe error masking issue. The mktemp absolute path guarantee makes sense too. Appreciate the thorough explanation of the alternatives you considered. This looks good!

jglogan:
We could solve this by defining a Makefile variable:

```Makefile
TAR := /usr/bin/tar
```
and then invoking the step using the variable instead of path-resolving the command.

That would ensure the macOS BSD tar is used.

Same would go for `pax`.

jglogan:
@manuschillerdev Thanks for the enhancement!

PR #730, opened 2025-10-07
cli: accept global flags after subcommands

cnaples79:
## Summary
- normalize global flags before parsing so  no longer errors
- keep passthrough arguments intact by respecting the  separator

## Rationale
The  flag is defined globally, but the help output implies it belongs to . When users pass it after the subcommand, ArgumentParser throws an "Unknown option" error. Moving global flags to the front preserves the intended UX while maintaining their global scope.

## Testing
- not run (Swift 6.2.0 required, local environment has 6.1.3)

Fixes #360

PR #1053, opened 2026-01-15
Reorder global args to start of list.

### LINKED ISSUE #360: [Bug]: "Error: Unknown option '--debug'. Did you mean '--debug'?"
### I have done the following

### Steps to reproduce

```shell
container run --debug ghcr.io/styrainc/regal:0.35.1
```

### Current behavior

```
Error: Unknown option '--debug'. Did you mean '--debug'?
Usage: container run [<options>] <image> [<arguments> ...]
  See 'container run --help' for more information.
```

### Expected behavior

`container run --help` tells me that the `run` command has a `--debug` flag. Trying to use that however gives me the error message in the issue title. It's a rather amusing error message, mind you! But I doubt that's intended.

If I do something like `--debug=foo` I'll get `Error: The option '--debug' does not take any value, but 'foo' was specified.` so evidently the `--debug` flag *is* recognized, it just doesn't seem to work when only provided as `--debug`.

My expectation would be that `--debug` works, or that the error message explains why it's not.

### Environment

```markdown
- OS: 15.5
- Xcode: 
- Container: 0.2.0
```

### Relevant log output

```shell

```

### Code of Conduct

jglogan:
Co-author: Chris Naples <cnaples79@gmail.com> (@cnaples79)

- Fixes #360.
- Adapts #730 to new dispatch mechanism.
- Gets rid of wonky error message when `--debug` is placed after the subcommand and before `--`.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
That error message is really vexing.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/CLI/ContainerCLI.swift
+++ b/Sources/CLI/ContainerCLI.swift
@@ -30,10 +30,4 @@ public struct ContainerCLI: AsyncParsableCommand {
     public static func main() async throws {
         try await Application.main()
     }
-
-    public func run() async throws {
-        var application = try Application.parse(arguments)
-        try application.validate()
-        try application.run()
-    }
 }
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -98,10 +98,11 @@ public struct Application: AsyncParsableCommand {
 
         let fullArgs = CommandLine.arguments
         let args = Array(fullArgs.dropFirst())
+        let normalizedArgs = Self.normalizeGlobalFlags(args)
 
         do {
             // container -> defaultHelpCommand
-            var command = try Application.parseAsRoot(args)
+            var command = try Application.parseAsRoot(normalizedArgs)
             if var asyncCommand = command as? AsyncParsableCommand {
                 try await asyncCommand.run()
             } else {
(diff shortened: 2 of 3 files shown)


Sources/CLI/ContainerCLI.swift
```diff
@@ -30,10 +30,4 @@ public struct ContainerCLI: AsyncParsableCommand {
     public static func main() async throws {
         try await Application.main()
     }
-
-    public func run() async throws {
-        var application = try Application.parse(arguments)
-        try application.validate()
-        try application.run()
-    }
```
dcantah, line 38: My brain is fried with all of the implicit main funcs being called and when `validate` actually runs for subcommands logic, but does this mean we don't run validate's logic for every command anymore now? This is where we were shoving in some "nice to run every time" logic:

```swift
public func validate() throws {
        // Not really a "validation", but a cheat to run this before
        // any of the commands do their business.
        let debugEnvVar = ProcessInfo.processInfo.environment["CONTAINER_DEBUG"]
        if self.global.debug || debugEnvVar != nil {
            log.logLevel = .debug
        }
        // Ensure we're not running under Rosetta.
        if try isTranslated() {
            throw ValidationError(
                """
                `container` is currently running under Rosetta Translation, which could be
                caused by your terminal application. Please ensure this is turned off.
                """
            )
        }
    }
```

Sources/ContainerCommands/Application.swift
```diff
@@ -251,4 +252,39 @@ extension Application {
...
+                }
+            }
+
+            remaining.append(argument)
+        }
+
+        return collectedGlobals + remaining
+    }
+
+    private static var globalFlags: Set<String> {
```
JaewonHur: How about using reflection for this?
```
let mirror = Mirror(reflecting: Flags.Global())
let properties = mirror.children.compactMap { $0.label }
```

This can collects globalFlag names.

PR #1054, opened 2026-01-15
Fix #1050: Handle XPC connection loss to update container state on ex…

### LINKED ISSUE #1050: [Bug]: Container is showing as running even after being killed
### I have done the following

### Steps to reproduce

Hi @jglogan !

While testing  #1041 using the kill script in https://github.com/apple/container/issues/614, I had noticed that state of the container is not updated if it is killed via `kill` or `launchctl bootout`.

Start a container:
```
container run --name alpine-sleep-$(date +%H%M) --detach alpine sleep infinity
```

```
container list
ID                 IMAGE                            OS     ARCH   STATE    ADDR             CPUS  MEMORY   STARTED
alpine-sleep-1648  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.2/24  4     1024 MB  2026-01-14T15:48:52Z
```

Kill:
```
launchctl list | grep "com.apple.container.*alpine-sleep" | awk '{print $1}' | xargs kill
```

Container is showing as running:
```
container list
ID                 IMAGE                            OS     ARCH   STATE    ADDR             CPUS  MEMORY   STARTED
alpine-sleep-1648  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.2/24  4     1024 MB  2026-01-14T15:48:52Z
```

Failing to be stopped:
```
2026-01-14T16:52:11+0100 error com.apple.container: [ContainerCommands] failed to stop container alpine-sleep-1648: internalError: "failed to stop container" (cause: "internalError: "failed to stop container alpine-sleep-1648 (cause: "interrupted: "XPC connection error: Connection invalid"")"")
Error: internalError: "stop failed for one or more containers alpine-sleep-1648"
```

Still showing as running:
```
 container list
ID                 IMAGE                            OS     ARCH   STATE    ADDR             CPUS  MEMORY   STARTED
alpine-sleep-1648  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.2/24  4     1024 MB  2026-01-14T15:48:52Z
```

State will be updated only after full system stop:
```
container system stop
container system start
```

### Current behavior

Container is shown as running; cannot be stopped and continues to be shown as running until container system stop.

### Expected behavior

Update container system state if the container exited outside of container.

### Environment

```markdown
- OS: macOS 26.2
- Xcode: 26.2
- Container: container CLI version 0.7.1-49-gdbec1db (build: release, commit: dbec1db)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Ronitsabhaya75:
…ternal termination

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Resolves #1050.

When containers are killed externally (via kill or launchctl bootout), the XPC connection loss causes the wait handler to throw an .interrupted error. This prevents cleanup and leaves containers stuck in running state.

Fix: Modified the wait handler in  ContainersService.swift to catch .interrupted errors and treat them as container exits (code 137), allowing normal cleanup to proceed.

```swift
let waitFunc: ExitMonitor.WaitHandler = {
    log.info("registering container \(id) with exit monitor")
    do {
        let code = try await client.wait(id)
        log.info("container \(id) finished in exit monitor, exit code \(code)")
        return code
    } catch let err as ContainerizationError where err.code == .interrupted {
        // XPC connection was lost (e.g., container killed externally).
        // Treat this as a container exit so cleanup can proceed.
        log.warning("container \(id) XPC connection lost, treating as external termination")
        return ExitStatus(exitCode: 137)  // SIGKILL
    }
}
```

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -363,10 +363,16 @@ public actor ContainersService {
                     let log = self.log
                     let waitFunc: ExitMonitor.WaitHandler = {
                         log.info("registering container \(id) with exit monitor")
-                        let code = try await client.wait(id)
-                        log.info("container \(id) finished in exit monitor, exit code \(code)")
-
-                        return code
+                        do {
+                            let code = try await client.wait(id)
+                            log.info("container \(id) finished in exit monitor, exit code \(code)")
+                            return code
+                        } catch let err as ContainerizationError where err.code == .interrupted {
+                            // XPC connection was lost (e.g., container killed externally).
+                            // Treat this as a container exit so cleanup can proceed.
+                            log.warning("container \(id) XPC connection lost, treating as external termination")
+                            return ExitStatus(exitCode: 137)  // SIGKILL
+                        }
                     }
                     try await self.exitMonitor.track(id: id, waitingOn: waitFunc)
 


Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -363,10 +363,16 @@ public actor ContainersService {
                     let log = self.log
                     let waitFunc: ExitMonitor.WaitHandler = {
                         log.info("registering container \(id) with exit monitor")
-                        let code = try await client.wait(id)
-                        log.info("container \(id) finished in exit monitor, exit code \(code)")
-
-                        return code
+                        do {
```
dcantah, line 366: I'd have to test this, but naively just reading it does this actually fix the issue? The only difference here seems to be we return instead of throwing, but `ExitMonitor`'s track method seems to handle this:

```swift
public func track(id: String, waitingOn: @escaping WaitHandler) async throws {
        guard let onExit = self.exitCallbacks[id] else {
            throw ContainerizationError(.invalidState, message: "ExitMonitor not setup for process \(id)")
        }
        guard self.runningTasks[id] == nil else {
            throw ContainerizationError(.invalidState, message: "already have a running task tracking process \(id)")
        }
        self.runningTasks[id] = Task {
            do {
                let exitStatus = try await waitingOn()
                try await onExit(id, exitStatus)
            } catch {
                self.log?.error("WaitHandler for \(id) threw error \(String(describing: error))")
                try? await onExit(id, ExitStatus(exitCode: -1))
            }
        }
    }
```

If `waitingOn` throws we just return a -1 exit code instead of your 137 (which I don't fully agree with) and that seems to be the only difference. I do like the extra log, but more curious if this actually solves the issue.

Ronitsabhaya75: @dcantah, thanks for the review! You're absolutely right - I see now that ExitMonitor.track() already catches errors from waitingOn() and calls onExit with exit code -1.

I honestly couldn't test this locally (no macOS 26 access), so I might have misunderstood the issue. Looking at the error handling you pointed, it seems like handleContainerExit should already be getting called, just with -1 instead of 137.

The only difference my change makes is:

- Explicitly detecting .interrupted errors

- Adding a warning log for visibil y

- Using exit code 137 instead of -1

If the existing error handling already works, then yeah, this PR might not actually fix #1050. Would you be able to test the scenario from the issue (killing a container externally) to see if it's already working or if there's a different root cause?

Happy to close this if it's not needed or rework it based on what you find!

dcantah: Yea, I don't think this will fix the source of the issue

PR #1057, opened 2026-01-16
Use configuration subnet and fallback to default subnet in allocation only plugin

katiewasnothere:
## Type of Change
- [x] Bug fix

## Motivation and Context
This PR changes the behavior of the Allocation Only vmnet network plugin to use the default subnet in the fallback case if no subnet was specified in the network configuration. This matches the behavior of `ReservedVmnetNetwork`. 

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
+++ b/Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
@@ -65,11 +65,11 @@ public actor AllocationOnlyVmnetNetwork: Network {
             ]
         )
 
-        let subnet = DefaultsStore.get(key: .defaultSubnet)
-        let subnetCIDR = try CIDRv4(subnet)
-        let gateway = IPv4Address(subnetCIDR.lower.value + 1)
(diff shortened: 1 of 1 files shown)


Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
```diff
@@ -65,11 +65,17 @@ public actor AllocationOnlyVmnetNetwork: Network {
             ]
         )
 
-        let subnet = DefaultsStore.get(key: .defaultSubnet)
-        let subnetCIDR = try CIDRv4(subnet)
-        let gateway = IPv4Address(subnetCIDR.lower.value + 1)
+        let defaultIPv4Subnet = try DefaultsStore.getOptional(key: .defaultSubnet).map { try CIDRv4($0) }
```
jglogan: Changing to `getOptional` means we will not fall back to `192.168.64.1/24` if neither `--subnet` is defined now the user default is set. This is going to cause startup to fail on Sequoia, unless the user has set the user default.

katiewasnothere: Fixed! Thanks

PR #1060, opened 2026-01-16
Fix unstable integration tests.

jglogan:
- TestCLIRunCommand now run so many tests concurrently that the API server gets swamped and tests randomly time out.
- The parallelism options on `swift test` only work for XCTest, not swift-testing.
- Work around this while retaining some parallelism (good for stress testing) by breaking the tests into two suites.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Most PRBs are breaking right now.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,7 +184,8 @@ integration: init-block
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIRunLifecycle || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIExecCommand || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLICreateCommand || exit_code=1 ; \
-		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIRunCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIRunCommand1 || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIRunCommand2 || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIStatsCommand || exit_code=1 ; \
 		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIImagesCommand || exit_code=1 ; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -180,19 +181,19 @@ integration: init-block
...
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIStatsCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIImagesCommand || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIRunBase || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIBuildBase || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIVolumes || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIKernelSet || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLIAnonymousVolumes || exit_code=1 ; \
+		$(SWIFT) test -c $(BUILD_CONFIGURATION) $(INTEGRATION_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLINoParallelCases || exit_code=1 ; \
```
iko1: The name suggests these tests should not run in parallel

Tests/CLITests/TestCLINoParallelCases.swift
```diff
@@ -29,6 +29,7 @@ class TestCLINoParallelCases: CLITest {
     @Test func testImageSingleConcurrentDownload() throws {
         // removing this image during parallel tests breaks stuff!
         _ = try? run(arguments: ["image", "rm", alpine])
+        defer { _ = try? run(arguments: ["image", "rm", "--all"]) }
```
iko1, line 36: Consider adding logging in case of failures

Makefile
jglogan: They are designed to run in parallel, and we do so to stress test the system. Unfortunately it appears that we have a suite that contains so many container run tests that the API server gets overloaded and XPC calls start timing out.

Tests/CLITests/TestCLINoParallelCases.swift
jglogan: It's a good idea but I think we'd want to build that into our test support code. We put together the integration tests and related support code early on and there's plenty of room for refinement. It's out of scope for this PR as we just need to get tests working again, but consider filing an issue describing the enhancement, and then knock out a PR!

PR #1067, opened 2026-01-17
Add --dns support to build command for custom DNS resolution

cympak2:
This change adds --dns parameter support to the build command, matching the existing functionality in the run command. This allows users to specify custom DNS nameservers (like Google DNS 8.8.8.8) when building container images, which is essential when the host machine uses a local DNS resolver (e.g., 127.0.0.2) that doesn't work inside containers.

Changes:
- Added --dns option to BuildCommand to accept DNS nameserver IP addresses
- Modified build logic to detect DNS configuration changes and recreate the builder container when DNS settings differ
- Passed DNS configuration to BuilderStart.start() method
- Updated builder container creation to use specified DNS nameservers or fall back to network gateway

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This change adds --dns parameter support to the build command, matching the existing functionality in the run command. This allows users to specify custom DNS nameservers (like Google DNS 8.8.8.8) when building container images, which is essential when the host machine uses a local DNS resolver (e.g., 127.0.0.2) that doesn't work inside containers.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -86,7 +86,7 @@ public struct Flags {
         public var nameservers: [String] = []
 
         @Option(
-            name: .long,
+            name: .customLong("dns-domain"),
             help: .init("Default DNS domain", valueName: "domain")
         )
         public var domain: String? = nil


Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -122,6 +122,12 @@ extension Application {
         @Option(name: .long, help: ArgumentHelp("Builder shim vsock port", valueName: "port"))
         var vsockPort: UInt32 = 8088
 
+        @Option(
```
jglogan: Could you see if it's possible to make an OptionGroup from the options in ManagementFlags:

```
  --dns <ip>              DNS nameserver IP address
  --dns-domain <domain>   Default DNS domain
  --dns-option <option>   DNS options
  --dns-search <domain>   DNS search domains
```

and then include that group in ManagementFlags and here?

Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -354,12 +361,9 @@ extension Application {
                         guard let dest = exp.destination else {
                             throw ContainerizationError(.invalidArgument, message: "dest is required \(exp.rawValue)")
                         }
-                        let result = try await ClientImage.load(from: dest.absolutePath(), force: false)
```
jglogan, line 361: Please rebase on to the latest main. You change needs to handle the updated `load` return value as shown in deleted L358-361 here.

Sources/ContainerCommands/Builder/BuilderStart.swift
```diff
@@ -44,6 +44,12 @@ extension Application {
         )
         var memory: String = "2048MB"
 
+        @Option(
```
jglogan: See comment on DNS options above.

Sources/ContainerCommands/BuildCommand.swift
cympak2: Done

Sources/ContainerCommands/BuildCommand.swift
cympak2: Done

Sources/ContainerCommands/Builder/BuilderStart.swift
cympak2: Done

Sources/ContainerCommands/Builder/BuilderStart.swift
```diff
@@ -103,6 +106,7 @@ extension Application {
                 let existingImage = existingContainer.configuration.image.reference
                 let existingResources = existingContainer.configuration.resources
                 let existingEnv = existingContainer.configuration.initProcess.environment
+                let existingDNS = existingContainer.configuration.dns?.nameservers ?? []
```
jglogan: Did you forget to plumb the additional args into the container DNS configuration?

I tried it out locally and `container create` / `container run` haven't regressed, but if I do this I only see the nameserver in the config file:

```bash
% container builder start --dns 8.8.8.8 --domain foo.com --dns-option foo,bar --dns-search foo.com,bar.com
% container exec buildkit cat /etc/resolv.conf
nameserver 8.8.8.8
%
```

cympak2: I believe I lack the necessary permissions to assign it to myself, so I’ve included it in this pull request.
% container delete buildkit
buildkit
% container builder start --dns 8.8.8.8 --domain foo.com --dns-option foo,bar --dns-search foo.com,bar.com
% container exec buildkit cat /etc/resolv.conf
nameserver 8.8.8.8
domain foo.com
search foo.com,bar.com
options foo,bar
%

Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -76,6 +76,34 @@ public struct Flags {
...
+        public init() {}
+
+        @Option(
+            name: .customLong("dns"),
+            help: .init("DNS nameserver IP address", valueName: "ip")
+        )
+        public var nameservers: [String] = []
+
+        @Option(
+            name: .long,
```
jglogan:
```suggestion
            name: .customLong("dns-domain"),
```

jglogan: Fixes an integration test failure.

jglogan:
Closes #1082.

PR #1076, opened 2026-01-22
feat: Moving bundle creation from ContainerService to SandboxService

### LINKED ISSUE #1046: [Request]: Move container bundle creation to SandboxService

Right now we're creating container bundles in `ContainersService`. Move this to the SandboxService to make it easier to support different container bundle types. 

### Code of Conduct

ajemory:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #1046 -- Right now we're creating container bundles in ContainersService. Move this to the SandboxService to make it easier to support different container bundle types.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -635,6 +635,8 @@ public actor ContainersService {
         }
 
         // Only try to deregister service if we have a valid config
+        // TODO: Change this so we don't have to reread the config
+        // possibly store the container ID to service label mapping
         if let config = config {
             let label = Self.fullLaunchdServiceLabel(
                 runtimeName: config.runtimeHandler,
@@ -715,7 +717,7 @@ public actor ContainersService {
         id == processID
     }
 
-    /// Get container configuration, either from existing bundle or from metadata
+    /// Get container configuration, either from existing bundle or from RuntimeConfiguration
     private static func getContainerConfiguration(at path: URL) throws -> ContainerConfiguration {
         let bundle = ContainerResource.Bundle(path: path)
         do {
--- a/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
+++ b/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
@@ -1315,7 +1315,7 @@ extension SandboxService {
         }
     }
 
-    /// Create bundle from metadata
+    /// Create bundle from RuntimeConfiguration
     private func createBundle() throws {
         do {
             let runtimeConfig = try RuntimeConfiguration.readRuntimeConfiguration(from: self.root)


dcantah:
Needs a `make fmt`

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -232,20 +232,25 @@ public actor ContainersService {
...
-                let imageFs = try await containerImage.getCreateSnapshot(platform: configuration.platform)
-                try bundle.setContainerRootFs(cloning: imageFs, readonly: configuration.readOnly)
-                try bundle.write(filename: "options.json", value: options)
+                try await sandboxClient.createBundle(
+                    configuration: configuration,
+                    kernel: kernel,
+                    options: options
+                )
 
                 let snapshot = ContainerSnapshot(
```
katiewasnothere, line 279: I think we can add the sandboxClient to the snapshot here and then remove the code in `bootstrap` that also makes the client. 

With the code today, do you know if the service gets re-registered or if it's just a noop? 

Might also consider adding the created bundle to the state struct.

JaewonHur: If we do not (re) register and make client in `bootstrap`,  then we may lost connection if we `stop` and `start` container again, as `ContainerStop` deregister the service and remove the client.

JaewonHur: This is tricky as we have to leave the service alive for every container created (even if they are not running).

jglogan:
@ajemory Could you rebase this onto main and push the PR again? As it stands hawkeye is freaking out on the header checks.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -232,20 +232,25 @@ public actor ContainersService {
...
+
+            try Self.registerService(
+                plugin: self.runtimePlugins.first { $0.name == configuration.runtimeHandler }!,
+                loader: self.pluginLoader,
+                configuration: configuration,
+                path: path
+            )
+
+            let runtime = configuration.runtimeHandler
+            let sandboxClient = try await SandboxClient.create(
```
jglogan: If we fail here, the service won't be unregistered, will it?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -232,20 +232,25 @@ public actor ContainersService {
             }
 
             let path = self.containerRoot.appendingPathComponent(configuration.id)
-            let systemPlatform = kernel.platform
-            let initFs = try await self.getInitBlock(for: systemPlatform.ociPlatform())
```
jglogan: you can remove the `func getInitBlock()` definition in this file now

Sources/ContainerResource/Container/Bundle.swift
```diff
@@ -109,6 +134,46 @@ extension Bundle {
...
+    public static func createFromMetadata(_ metadata: BundleMetadata) throws -> Bundle {
+        let bundle = try create(
+            path: metadata.path,
+            initialFilesystem: metadata.initialFilesystem,
+            kernel: metadata.kernel,
+            containerConfiguration: metadata.containerConfiguration
+        )
+
+        if let containerRootFs = metadata.containerRootFilesystem {
+            let readonly = metadata.containerConfiguration?.readOnly ?? false
```
katiewasnothere: Should the default for this be true?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -295,6 +297,13 @@ public actor ContainersService {
                     id: id,
                     runtime: runtime
                 )
+
+                // Create bundle if it doesn't exist
+                if !(await self.bundleExists(at: path)) {
+                    let metadataPath = Self.getMetadataPath(for: id)
+                    try await sandboxClient.createBundle(metadataPath: metadataPath)
+                }
```
katiewasnothere: Could we just have the SandboxService call bundle.create in bootstrap? We don't really need the bundle created in here right?

Sources/ContainerResource/Container/Bundle.swift
ajemory: I wanted to stay consistent with [setContainerRootFS](https://github.com/apple/container/pull/1076/changes/BASE..f812e9e1e7c6bd036bb1c6a9d7236b75571b4845#diff-009dc9a1bc0a8b6343b000ae58400c7fb7426baa9c188c5bc61e080e9dda9d32L125), which is currently false by default

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
ajemory: Just to clarify, you're suggesting to call bundle.create from SandboxService bootstrap instead of ContainersService bootstrap? Yeah, I think that's a better solution. I'll make the changes!

Sources/ContainerResource/Container/Bundle.swift
katiewasnothere: Sounds gtm thanks :)

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -235,17 +235,23 @@ public actor ContainersService {
...
 
-            let bundle = try ContainerResource.Bundle.create(
-                path: path,
-                initialFilesystem: initFs,
-                kernel: kernel,
-                containerConfiguration: configuration
-            )
+            let metadataPath = FileManager.default.temporaryDirectory
+                .appendingPathComponent("bundle-metadata-\(configuration.id).json")
+
```
katiewasnothere: Can we use `getMetadataPath` here instead?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -668,6 +695,43 @@ public actor ContainersService {
...
+            _ = try bundle.configuration
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    /// Get metadata file path for the given container ID
+    private static func getMetadataPath(for containerID: String) -> URL {
+        FileManager.default.temporaryDirectory
```
katiewasnothere: I think I'd prefer we just write the metadata to the bundle directory itself instead of a temporary location but also welcome to other thoughts on this. I think that way we just delete the bundle when the container is cleaned up and we're good to go. Plus then we know the file path in the SandboxService for free.

Sources/ContainerResource/Container/Bundle.swift
```diff
@@ -109,6 +134,46 @@ extension Bundle {
...
+        }
+
+        if let options = metadata.options {
+            try bundle.write(filename: "options.json", value: options)
+        }
+
+        return bundle
+    }
+
+    public static func writeMetadata(
```
JaewonHur: How about changing `writeMetadata` to `BundleMetadata.write`, and `readMetadata` to `BundleMetadata.load`?
Following the same pattern as `Bundle`.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -596,15 +599,39 @@ public actor ContainersService {
...
+        // Only try to deregister service if we have a valid config
+        if let config = config {
+            let label = Self.fullLaunchdServiceLabel(
+                runtimeName: config.runtimeHandler,
+                instanceId: id
+            )
+            try? ServiceManager.deregister(fullServiceLabel: label)
+        }
+
+        // Always try to delete the bundle directory, even if it's incomplete
```
JaewonHur: `bundle.delete()` and `FileManager.default.removeItem(at: path)` are doing the same task.

Once the metadata is moved to container root (i.e., bundle) directory, we can just clean up all together using `bundle.delete()`.

Sources/ContainerResource/Container/Bundle.swift
```diff
@@ -109,6 +140,50 @@ extension Bundle {
...
```
katiewasnothere: nit: this can probably be a method on BundleMetadata

Sources/ContainerResource/Container/Bundle.swift
```diff
@@ -109,6 +140,50 @@ extension Bundle {
...
+    ) throws {
+        // Ensure the parent directory exists
+        let directory = metadata.bundleMetadataPath.deletingLastPathComponent()
+        try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true)
+
+        let data = try JSONEncoder().encode(metadata)
+        try data.write(to: metadata.bundleMetadataPath)
+    }
+
+    public static func readMetadata(from bundlePath: URL) throws -> BundleMetadata {
```
katiewasnothere: nit: this can probably be a static function on the BundleMetadata type

Sources/ContainerResource/Container/Bundle.swift
```diff
@@ -109,6 +140,50 @@ extension Bundle {
...
+        let metadataPath = bundlePath.appendingPathComponent(BundleMetadata.bundleMetadataFileName)
+        guard FileManager.default.fileExists(atPath: metadataPath.path) else {
+            throw ContainerizationError(
+                .notFound,
+                message: "Bundle metadata file not found at path: \(metadataPath.path)"
```
katiewasnothere: nit: can we lowercase the first word of the error messages in all the errors we're adding in this PR?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -671,6 +688,36 @@ public actor ContainersService {
...
+        guard await bundleExists(at: path) else {
+            // Bundle doesn't exist, get config from metadata
+            let metadata = try ContainerResource.Bundle.readMetadata(from: path)
+            guard let config = metadata.containerConfiguration else {
+                throw ContainerizationError(.internalError, message: "Metadata missing container configuration")
+            }
+            return config
+        }
+        // Bundle exists, read config normally
+        let bundle = ContainerResource.Bundle(path: path)
```
katiewasnothere: Can we refactor this and the function above a bit to avoid needing to load the configuration twice (in Bundle's configuration var) in the case that the bundle already exists?

katiewasnothere (commented):
One small change and then a couple of nits. Ready to approve after that :)

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
ajemory: Fixed!

ajemory:
Everything is fixed now!

katiewasnothere (approved):

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -1300,4 +1299,38 @@ extension SandboxService {
...
+        let bundle = ContainerResource.Bundle(path: path)
+        do {
+            _ = try bundle.configuration
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    /// Create bundle from metadata
```
katiewasnothere: Nit: should we change this comment?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -625,15 +623,33 @@ public actor ContainersService {
...
+        }
+
+        // Only try to deregister service if we have a valid config
+        if let config = config {
+            let label = Self.fullLaunchdServiceLabel(
+                runtimeName: config.runtimeHandler,
+                instanceId: id
+            )
+            try? ServiceManager.deregister(fullServiceLabel: label)
+        }
```
katiewasnothere, line 646: I wonder if we should add a TODO or FIXME here? Maybe we should store the container ID to service label mapping so we don't need to try to reread the config? But I'm happy to have this be in a later PR or to discuss it more outside of this PR discussion first.

ajemory: I'll add a TODO for now, and get a fix in a separate PR :)

PR #1077, opened 2026-01-22
Add option to stop services in all launchd domains

p-linnane:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Services started from Terminal run in the `gui/$uid` launchd domain. When `ensure-container-stopped.sh` runs from a different context, `launchctl managername` may return a different domain, causing the script to check the wrong domain and miss running services. I noticed this after upgrading `container` via Homebrew.

~~This change queries all three domains directly instead of detecting the current context.~~

EDIT: Based on maintainer feedback, this was revised to introduce a getopt option that allows querying all domains explicitly, rather than doing so by default.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/scripts/ensure-container-stopped.sh
+++ b/scripts/ensure-container-stopped.sh
@@ -13,19 +13,59 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-domain_string=""
+ALL_DOMAINS=false
 
-launchd_domain=$(launchctl managername)
-
-if [[ "$launchd_domain" == "System" ]]; then
-  domain_string="system"
-elif [[ "$launchd_domain" == "Aqua" ]]; then
-  domain_string="gui/$(id -u)"
-elif [[ "$launchd_domain" == "Background" ]]; then
-  domain_string="user/$(id -u)"
-else
-    echo "Unsupported launchd domain. Exiting"
+usage() {
+    echo "Usage: $0 [-a] [-h]"
+    echo "Stop container services"
+    echo
+    echo "Options:"
+    echo "a     Stop container services in all launchd domains."
+    echo "h     Show this help message."
+    echo
     exit 1
-fi
+}
 
-launchctl list | grep -e 'com\.apple\.container\W' | awk '{print $3}' | xargs -I % launchctl bootout $domain_string/%
+while getopts ":ah" arg; do
+    case "$arg" in
+        a)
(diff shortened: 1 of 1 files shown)


jglogan:
@p-linnane that is interesting. Before we go with this approach, are there any alternatives? This seems to be a pretty big hammer.

So I understand correctly, this is what's happening when trying to shutdown a non-homebrew container setup with `container system stop` from Homebrew? Could you file an issue for this with repro steps? Thank you!

p-linnane:
For additional context, I'm a Homebrew maintainer and noticed this after our bump from 0.7.1 to 0.8.0. I observed the following due to the API changes:

```
container network ls --debug
Error: keyNotFound(CodingKeys(stringValue: "ipv4Subnet", intValue: nil), Swift.DecodingError.Context(codingPath: [_CodingKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "running", intValue: nil), RunningCodingKeys(stringValue: "_1", intValue: nil)], debugDescription: "No value associated with key CodingKeys(stringValue: \"ipv4Subnet\", intValue: nil) (\"ipv4Subnet\").", underlyingError: nil))
```

To clarify: this is entirely within Homebrew. The issue is that `ensure-container-stopped.sh` (called by `post_install` in the Homebrew formula) may run in a different launchd context than where the user originally started services via `container system start`.

Happy to file a separate issue if that would be helpful.

jglogan:
@p-linnane That's worth a new issue.

For this one, instead of removing the existing behavior, what if you added some getopt code to process a `-a` option (and `-h` for a usage message), and do the "all domains" stop if `-a` is specified?

p-linnane:
The problem I described (Homebrew's `post_install` running in a different launchd context) is exactly what this PR addresses. I've pushed a commit to add getopt code as requested. I tried to match the conventions as close to `uninstall-container.sh` as possible.

jglogan:
@p-linnane This looks nice, thanks!

Started the PRB and will approve/merge after.

p-linnane:
Cleaned things up to accurately reflect the implemented change.

jglogan:
@p-linnane Merged, hanks for the enhancement!

Issue #346, opened 2025-07-17
[Request]: Add ability to connect to a host service


zakhenry:
### Feature or enhancement request details

Hopefully this is just a case of adding documentation?

With docker I can run a webserver locally on `http://127.0.0.1:1420/ ` and run a container (e.g. `mcr.microsoft.com/playwright:v1.45.3-noble`) and connect to the server running locally *from the container* for the purposes of running tests against the local service.

With docker on macos I configure the url as `http://host.docker.internal:1420` which works, I tried `http://host.containers.internal:1420` & `http://host.container.internal:1420` but get `curl: (6) Could not resolve host: host.container.internal` (my first command run in container is `curl --fail --silent --show-error http://host.containers.internal:1420> /dev/null`)

Interestingly the `host.containers.internal` fails instantly, and `host.container.internal` take 11 seconds before failing.

Is there some other magic hostname that I should use with containers? Potentially of note I am on Sequoia 15.5 so can't try out the new networking options.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
@zakhenry thanks!

This is a known limitation; see https://github.com/apple/container/blob/main/docs/technical-overview.md#container-to-host-networking for a workaround, but it's something we do want to get to, so we'll use your issue for tracking the work. 

Making this work isn't too hard, until you need to make it usable across more than the default network which whenever `container` services are running. In terms of implementation approaches, one would be to:

1. Add a scoped DNS configuration that causes mDNSResponser to send queries for `host.docker.internal` to `container` (this sort of thing happens today when you run `sudo container system dns create test`).
2. Add a DNSHandler that resolves `host.docker.internal` to some dummy IP address, like something in the 172.16.0.0/20 block.
3. For each network, add a packet filter rule that redirects traffic from the network bridge to any port on the dummy address the corresponding port on 127.0.0.1 on the host.

The challenge that I see is that 1. and 3. require administrator privilege.

We could run a privileged helper for this but running a service this way requires a bit of attention to security, and it may not be acceptable in all environments.

For DNS we can limit the need for sudo to just creating the scoped resolver config in /etc/resolver, and HUPping mDNSResponder. It's not as simple for this, though, as networks come and go, so the user would need to be running sudo at both ends of a network's lifecycle. This isn't ideal for usability either.

debasisdwivedy:
MAC OS: **26.0.1**
Container version: container CLI version **0.6.0** (build: release, commit: a23bcf0)

[...]

PR #1078, opened 2026-01-22
Host connection

JaewonHur:
This PR enables connecting host's localhost port from container (#346).

It adds an option `--localhost <localhost>` to DNS create command, after which the packets heading ip address in container are redirected to localhost in host machine. Packet filter rule is added and deleted along with the creation and deletion of localhost domain.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Now users can access host's localhost services from container.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/System/DNS/DNSCreate.swift
+++ b/Sources/ContainerCommands/System/DNS/DNSCreate.swift
@@ -22,14 +22,14 @@ import ContainerizationExtras
 import Foundation
 
 extension Application {
-    public struct DNSCreate: AsyncParsableCommand {
+    public struct DNSCreate: AsyncLoggableCommand {
         public static let configuration = CommandConfiguration(
             commandName: "create",
             abstract: "Create a local DNS domain for containers (must run as an administrator)"
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var logOptions: Flags.Logging
 
         @Option(name: .long, help: "Set the ip address to be redirected to localhost")
         var localhost: String?
--- a/Sources/ContainerCommands/System/DNS/DNSDelete.swift
+++ b/Sources/ContainerCommands/System/DNS/DNSDelete.swift
@@ -21,7 +21,7 @@ import ContainerizationExtras
 import Foundation
 
 extension Application {
-    public struct DNSDelete: AsyncParsableCommand {
+    public struct DNSDelete: AsyncLoggableCommand {
         public static let configuration = CommandConfiguration(
             commandName: "delete",
             abstract: "Delete a local DNS domain (must run as an administrator)",
@@ -29,7 +29,7 @@ extension Application {
         )
 
         @OptionGroup
-        var global: Flags.Global
+        public var logOptions: Flags.Logging
 
         @Argument(help: "The local domain name")
         var domainName: String
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -16,7 +16,7 @@ jobs:
     name: Build and test the project
     if: github.repository == 'apple/container'
     timeout-minutes: 60
-    runs-on: [self-hosted, macos, sequoia, ARM64]
+    runs-on: [self-hosted, macos, tahoe, ARM64]
     permissions:
       contents: read
       packages: read
@@ -73,6 +73,10 @@ jobs:
           APP_ROOT=$(mktemp -d -p "${RUNNER_TEMP}")
           trap 'rm -rf "${APP_ROOT}"; echo Removing data directory ${APP_ROOT}' EXIT
           echo "Created data directory ${APP_ROOT}"
+          export NO_PROXY="${NO_PROXY},192.168.0.0/16,fe80::/10"
+          echo NO_PROXY=${NO_PROXY}
+          export no_proxy="${no_proxy},192.168.0.0/16,fe80::/10"
+          echo no_proxy=${no_proxy}
           make APP_ROOT="${APP_ROOT}" test install-kernel integration
         env:
           DEVELOPER_DIR: "/Applications/Xcode-latest.app/Contents/Developer"
--- a/.swift-format-nolint
+++ b/.swift-format-nolint
@@ -1,0 +1,68 @@
+{
+  "fileScopedDeclarationPrivacy" : {
+    "accessLevel" : "private"
+  },
+  "indentation" : {
+    "spaces" : 4
+  },
+  "indentConditionalCompilationBlocks" : false,
+  "indentSwitchCaseLabels" : false,
+  "lineBreakAroundMultilineExpressionChainComponents" : false,
+  "lineBreakBeforeControlFlowKeywords" : false,
+  "lineBreakBeforeEachArgument" : false,
+  "lineBreakBeforeEachGenericRequirement" : false,
+  "lineLength" : 180,
+  "maximumBlankLines" : 1,
+  "multiElementCollectionTrailingCommas" : true,
+  "noAssignmentInExpressions" : {
+    "allowedFunctions" : [
+      "XCTAssertNoThrow"
+    ]
+  },
+  "prioritizeKeepingFunctionOutputTogether" : false,
+  "respectsExistingLineBreaks" : true,
+  "rules" : {
+    "AllPublicDeclarationsHaveDocumentation" : false,
+    "AlwaysUseLowerCamelCase" : false,
+    "AmbiguousTrailingClosureOverload" : false,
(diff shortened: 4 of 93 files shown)


Sources/ContainerCommands/System/DNS/DNSCreate.swift
```diff
@@ -30,22 +31,52 @@ extension Application {
...
+            if let from = localhostIP {
+                let to = try! IPAddress("127.0.0.1")
+                do {
+                    try pf.createRedirectRule(from: from, to: to, domain: domainName)
+                } catch {
+                    _ = try resolver.deleteDomain(name: domainName)
+                    throw error
+                }
+            }
+            print(domainName)
```
jglogan, line 70:
```suggestion

            // Output the created resource ID after domain (and optional packet filter) configurations exist.
            // Just provide diagnostic messages for failed service restarts.
            print(domainName)
```

Sources/ContainerCommands/System/DNS/DNSDelete.swift
```diff
@@ -37,9 +38,9 @@ extension Application {
 
         public func run() async throws {
             let resolver = HostDNSResolver()
+            var localhostIP: IPAddress?
             do {
-                try resolver.deleteDomain(name: domainName)
-                print(domainName)
+                localhostIP = try resolver.deleteDomain(name: domainName)
```
jglogan, line 43: As we do the managed resource refactoring, let's plan to update the domain code to use the managed resource protocol, and then we'll use more conventional semantics for DELETE.

No need for changes in this PR.

Sources/ContainerCommands/System/DNS/DNSDelete.swift
```diff
@@ -49,6 +50,21 @@ extension Application {
...
+
+            guard let localhostIP else {
+                print(domainName)
+                return
+            }
+
+            let pf = PacketFilter()
+            try pf.removeRedirectRule(from: localhostIP, to: try! IPAddress("127.0.0.1"), domain: domainName)
+
+            do {
```
jglogan, line 62: The DNS manager makes it the responsibility of the caller to perform the reinitialize side effects for both create and delete.

For the packet filter, which is it the responsibility of the caller for create, and the callee for delete?

Tests/ContainerAPIClientTests/HostDNSResolverTest.swift
```diff
@@ -86,7 +86,7 @@ struct HostDNSResolverTest {
 
         let resolver = HostDNSResolver(configURL: tempURL)
         try resolver.createDomain(name: "foo.bar")
-        try resolver.deleteDomain(name: "foo.bar")
+        _ = try resolver.deleteDomain(name: "foo.bar")
```
jglogan, line 91: Why not check the output (did you check it in at least one test?)

Tests/ContainerAPIClientTests/PacketFilterTest.swift
```diff
@@ -0,0 +1,97 @@
...
+        #expect(!fm.fileExists(atPath: anchorURL.path))
+        let configText = try String(contentsOf: configURL, encoding: .utf8)
+        #expect(configText == "")
+    }
+
+    @Test
+    func testPacketFilterReinitialize() async throws {
+        let isAdmin = getuid() == 0
+        let pf = PacketFilter()
+        do {
```
jglogan: use `#expect(throws: Error) { ... }` or

```swift
expect {
    ...
} throws: { error in
    ...
}

File: docs/how-to.md [hunk 6]
```diff
@@ -198,6 +198,30 @@ Test access using `curl`:
 </body></html>
 ```
 
+## Access a host service from a container
+
+Create a DNS domain with `--localhost <ipv4-address>` to make a domain used by a container to access a host service. Any IPv4 address can be used as `<ipv4-address>`, which will be assigned to the domain name in container.
```
jglogan, line 203:
```suggestion
Create a DNS domain with `--localhost <ipv4-address>` to make a domain used by a container to access a host service. Any IPv4 address can be used as `<ipv4-address>`, which will be assigned to the domain name in container.

Choose an IP address that is least likely to conflict with any networks or reserved IP addresses in your environment. Reasonably safe address ranges include:

- The documentation ranges 192.0.2.0/24, 198.51.100.0/24, and 203.0.113.0/24.
- The 172.16.0.0/12 private range.
```

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    public func answer(query: Message) async throws -> Message? {
+        let question = query.questions[0]
+        var record: ResourceRecord?
+        switch question.type {
+        case ResourceRecordType.host:
+            if let ip = dns[question.name] {
+                record = HostRecord<IPv4>(name: question.name, ttl: ttl, ip: ip)
+            }
+        case ResourceRecordType.host6,
```
jglogan: Look at the corresponding code in the ContainerDNSHandler regarding AAAA requests. Should we reply with NODATA if we there's an IPv4 address but no IPv6 address?

Sources/Services/ContainerAPIService/Client/PacketFilter.swift
```diff
@@ -0,0 +1,198 @@
...
+
+        var lines = content.components(separatedBy: .newlines)
+        if !content.contains(redirectRule) {
+            lines.insert(redirectRule, at: lines.endIndex - 1)
+        }
+
+        try lines.joined(separator: "\n").write(toFile: anchorURL.path, atomically: true, encoding: .utf8)
+    }
+
+    public func removeRedirectRule(from: IPAddress, to: IPAddress, domain: String) throws {
```
jglogan, line 65: What happens if I create two localhost domains that use the same IP address?

Do we maintain the rules separately?

If so, does the packet filtering work ok with the redundant rules?

If we use one rule, what happens if we delete one of the domains?

JaewonHur: One pf rule pairs with one domain.

If two localhost domains with the same IP address are created, two pf rules are created as well.
PF follows the last one if it encounters multiple conflicting rules.
However, in this case, two pf rules are same, so I guess no problem.

If we delete one of the domain then, still the pf rule for other one remains.

Sources/ContainerCommands/System/DNS/DNSDelete.swift
JaewonHur: Packet filter the same, it is the responsibility of caller to reinitialize for both create and delete.

Sources/ContainerCommands/System/DNS/DNSCreate.swift
JaewonHur: What do you mean by `resource ID`? Is it in the context of ManagedResource?

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
JaewonHur: Fixed.

Tests/ContainerAPIClientTests/HostDNSResolverTest.swift
JaewonHur: Added test.

Tests/ContainerAPIClientTests/PacketFilterTest.swift
```diff
@@ -82,16 +82,9 @@ struct PacketFilterTest {
 
     @Test
     func testPacketFilterReinitialize() async throws {
-        let isAdmin = getuid() == 0
         let pf = PacketFilter()
-        do {
+        #expect(throws: ContainerizationError.self) {
```
JaewonHur, line 86: Not sure it's correct update..

Sources/ContainerCommands/System/DNS/DNSCreate.swift
jglogan: Yes but you suggest an interesting point. When we implement the ManagedResource stuff and if we did make domains a managed resource, the domain name property might be independent of its resource ID.

That's something for another PR though, the comment will be helpful to someone who's looking at the code for the first time.

JaewonHur: Yes, once we use resource ID, we should update this to use resource ID also.

Issue #1037, opened 2026-01-10
[Request]: Allow for port publishing / connectivity to container when "none" network is selected


matthewcmead:
### Feature or enhancement request details

Sometimes the purpose of using --network none is to isolate the container from being able to connect to external networks while connectivity from the host to the container would be useful.  Currently when using --network none, if ports are published with -p, there is no listening socket on the host and no connectivity to the container.

I'm requesting to change this to allow for host to container connectivity to work using port publishing when --network none is used.

There is a workaround for this, but it's somewhat of a hassle: --publish-socket can be used to connect a unix domain socket into the container and socat can be used on the host as well as the container to connect the published domain socket to ports on the host and in the container.  Perhaps this can serve as the mechanism to accomplish the connectivity in an automated way that entirely avoids activating the network stack with host connectivity for the container?

Thanks for the consideration.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
@matthewcmead If you feel like trying a quick hack, you could change the mode of the default network to `VMNET_HOST_MODE` and play around with that setup. The containers on that network can talk to each other and to the host, but can't go off host.

Grab the source, and in the main branch, make the change here: https://github.com/apple/container/blob/main/Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift#L113.

Clearly it doesn't give you the container-level granularity you want, but it's what could be delivered with a relatively simple contribution.

JaewonHur:
Let me look at it.

JaewonHur:
We have to options to support it.

1. Add a feature to make host-only restricted network and allow the host and containers connect under it.
2. After making container under default network, deleting gateway (i.e., `ip route del default`) leaves it unconnected, but still network under that subnet works. However, it's not a true isolation as the container can restore the external connection whenever it wants.

matthewcmead:
> We have to options to support it.
> 
> 1. Add a feature to make host-only restricted network and allow the host and containers connect under it.
> 2. After making container under default network, deleting gateway (i.e., `ip route del default`) leaves it unconnected, but still network under that subnet works. However, it's not a true isolation as the container can restore the external connection whenever it wants.

[...]

PR #1079, opened 2026-01-22
Isolated network

JaewonHur:
This gives `--internal` flag when creating a network, which will be used only for host internal network. Containers under the same subnet can connect each other (and the host), but not the external network (#1037).

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Network/NetworkCreate.swift
+++ b/Sources/ContainerCommands/Network/NetworkCreate.swift
@@ -31,7 +31,7 @@ extension Application {
         @Option(name: .customLong("label"), help: "Set metadata for a network")
         var labels: [String] = []
 
-        @Flag(name: .customLong("internal"), help: "Restrict external access to the network")
+        @Flag(name: .customLong("internal"), help: "Restrict to host-only network")
         var hostOnly: Bool = false
 
         @Option(
--- a/Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
+++ b/Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
@@ -39,8 +39,8 @@ extension NetworkVmnetHelper {
         @Option(name: .shortAndLong, help: "Network identifier")
         var id: String
 
-        @Flag(name: .long, help: "Restrict external access to the network")
-        var hostOnly: Bool = false
+        @Option(name: .long, help: "Network mode")
+        var mode: NetworkMode = .nat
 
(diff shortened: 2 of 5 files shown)


JaewonHur:
vmnet issue I guess. Checked the tests succeed in the latest MacOS version.

Sources/ContainerCommands/Network/NetworkCreate.swift
```diff
@@ -31,6 +31,9 @@ extension Application {
         @Option(name: .customLong("label"), help: "Set metadata for a network")
         var labels: [String] = []
 
+        @Flag(name: .customLong("internal"), help: "Restrict external access to the network")
```
jglogan: Wouldn't this be prohibiting external access by containers on the network?

This reads like things outside the network cannot access containers on the network, but the host is outside the network and it can access it, right?

Let's consider how the "network mode" (NAT or host-only today, but maybe someday we'll have access to "host mode" too) fits in with the changes on #1081.

Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
```diff
@@ -39,6 +39,9 @@ extension NetworkVmnetHelper {
         @Option(name: .shortAndLong, help: "Network identifier")
         var id: String
 
+        @Flag(name: .long, help: "Restrict external access to the network")
```
jglogan: Use a `--mode` flag here that transforms String to NetworkMode to program the mode directly.

Same comment on the help text as for the top level CLI.

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
```diff
@@ -32,7 +32,7 @@ public actor AllocationOnlyVmnetNetwork: Network {
         configuration: NetworkConfiguration,
         log: Logger
     ) throws {
-        guard configuration.mode == .nat else {
+        guard configuration.mode == .nat || configuration.mode == .hostOnly else {
```
jglogan: Reject host-only mode here. You can never configure a host-only network with this one.

Sources/ContainerCommands/Network/NetworkCreate.swift
JaewonHur: Updated the help message.

I think not much conflict with #1081?

Issue #1048, opened 2026-01-14
[Request]: Support the option to use different network plugins


katiewasnothere:
### Feature or enhancement request details

Right now the NetworksService only allows us to use a single network plugin during the lifetime of `container`. See [here](https://github.com/apple/container/blob/dbec1db03e0b67afda56dd07cc152304b2ffd53e/Sources/Services/ContainerAPIService/Server/Networks/NetworksService.swift#L58). We should support the ability to choose different network plugins during the lifetime of `container`. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
Closed by #1151.

PR #1081, opened 2026-01-23
Support using multiple different network plugins

katiewasnothere:
## Type of Change
- [x] New feature  
- [x] Breaking change

## Motivation and Context
* To support this, we need to change the architecture of communicating with network service instances, which is a breaking change. Now to communicate with the network service instance, you must first communicate with the NetworksService via the APIServer so you can accurately determine which network plugin was used to manage that network instance.
* This follows a similar model to what the we do for the SandboxClient and SandboxService.

Closes https://github.com/apple/container/issues/1048

Future changes:
- Docs changes for the new fields on NetworkCreate 
- Better UX for determining the default network plugin and default network configurations 
- Additional network plugins and better testing 

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/ContainerCommands/Network/NetworkCreate.swift
+++ b/Sources/ContainerCommands/Network/NetworkCreate.swift
@@ -45,6 +45,12 @@ extension Application {
             })
         var ipv6Subnet: CIDRv6? = nil
 
+        @Option(name: .long, help: "Set the plugin to use to create this network.")
+        var plugin: String = "container-network-vmnet"
+
+        @Option(name: .long, help: "Set the variant of the network plugin to use.")
+        var pluginVariant: String?
+
         @OptionGroup
         var global: Flags.Global
 
@@ -60,7 +66,8 @@ extension Application {
                 mode: .nat,
                 ipv4Subnet: ipv4Subnet,
                 ipv6Subnet: ipv6Subnet,
-                labels: parsedLabels
+                labels: parsedLabels,
+                pluginInfo: NetworkPluginInfo(plugin: self.plugin, variant: self.pluginVariant)
             )
             let state = try await ClientNetwork.create(configuration: config)
             print(state.id)
--- a/Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
+++ b/Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
@@ -71,11 +71,17 @@ extension NetworkVmnetHelper {
                 log.info("configuring XPC server")
                 let ipv4Subnet = try self.ipv4Subnet.map { try CIDRv4($0) }
                 let ipv6Subnet = try self.ipv6Subnet.map { try CIDRv6($0) }
+                let pluginInfo = NetworkPluginInfo(
+                    plugin: NetworkVmnetHelper._commandName,
+                    variant: self.variant.rawValue
+                )
+
                 let configuration = try NetworkConfiguration(
                     id: id,
                     mode: .nat,
                     ipv4Subnet: ipv4Subnet,
                     ipv6Subnet: ipv6Subnet,
+                    pluginInfo: pluginInfo
                 )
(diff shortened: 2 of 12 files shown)


Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
```diff
@@ -71,11 +71,17 @@ extension NetworkVmnetHelper {
                 log.info("configuring XPC server")
                 let ipv4Subnet = try self.ipv4Subnet.map { try CIDRv4($0) }
                 let ipv6Subnet = try self.ipv6Subnet.map { try CIDRv6($0) }
+                let pluginInfo = NetworkPluginInfo(
```
jglogan, line 74: Could we replace all our existing config options with a `--config network-entity.json`?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -147,8 +147,11 @@ public actor SandboxService {
             var interfaces: [Interface] = []
             for index in 0..<config.networks.count {
                 let network = config.networks[index]
-                let client = NetworkClient(id: network.network)
-                let (attachment, additionalData) = try await client.allocate(hostname: network.options.hostname, macAddress: network.options.macAddress)
+                let (attachment, additionalData) = try await ClientNetwork.allocate(
```
jglogan, line 150: Instead of this (and exposing the network helper APIs on the API server), could we move the IP address allocation and cleanup to the API server?

This would solve an existing problem because we'd then more reliably clean up IP addresses.

Sources/ContainerCommands/Network/NetworkCreate.swift
```diff
@@ -45,6 +45,12 @@ extension Application {
             })
         var ipv6Subnet: CIDRv6? = nil
 
+        @Option(name: .long, help: "Set the plugin to use to create this network.")
```
jglogan, line 48: At the user level, we may want to just have the notion of "driver", with the convention that it maps to `container-network-{driver-id}`

Sources/Helpers/NetworkVmnet/NetworkVmnetHelper+Start.swift
katiewasnothere: Definitely. I think this is something we want to do moving forward, but I'd prefer to have it in a later PR.

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
katiewasnothere: @jglogan and I have discussed this further. We can't move the network allocation into the ContainersService (aka the APIServer) as it is today because we'd be creating a circular dependency between the NetworksService and ContainersService, each of which would hold a reference to the other. Overall this seems like an area we're still discussing and want to investigate more. So we will not be addressing this in this PR. 

Here's some of the options we were thinking about: 
- weak references for needed services within a service object (aka weak reference to NetworksService in ContainersService)
- pass in closures to the different services (aka a closure for getting the list of containers from the ContainersService, for example)
- making XPC connections to the other services (aka xpc connection to NetworksService from ContainersService) and (assuming swift supports this) it takes the fast path since the two ends are running in the same process AND add semantics that a valid XPC connection from another service implies the called service's appropriate lock is being held.

Sources/ContainerCommands/Network/NetworkCreate.swift
katiewasnothere: I don't mind calling this arg something like "driver". But I'm hesitant on the mapping to `container-network-{driver-id}`. For me personally, it's less typing to do that, but I feel like it's less clear and less searchable. What do you think?

jglogan: We can leave it like this for now, and give it some thought. Changing it a little while later isn't going to break anything.

PR #1101, opened 2026-01-28
Launch a service with waitForDebugger if specified

JaewonHur:
This PR enables launching a service with `waitForDebugger` flag if the service label matches a given env variable `CONTAINER_DEBUG`.

To enable this feature, user starts API server with `CONTAINER_DEBUG=<debug target> container system start`. Then, the service whose label matches the `<debug target>` will wait for a debugger to be attached.

The logic matches the prefix of label separated by `.`, so that multiple service labels can be matched.

For example, `CONTAINER_DEBUG=com.apple.container.container-runtime-linux` will match any launchd service `com.apple.container.container-runtime-linux.<container-id>`.

If this is good, let me work more to stop timeout handling if `CONTAINER_DEBUG` is provided---i.e., currently system fails often as attaching debugger manually exceeds the timeout easily.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Enable attaching debugger to individual service.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Containers/TestCLIExec.swift
+++ b/Tests/CLITests/Subcommands/Containers/TestCLIExec.swift
@@ -122,9 +122,15 @@ class TestCLIExecCommand: CLITest {
             do {
                 _ = try doExec(name: name, cmd: ["sleep", "infinity"])
             } catch CLIError.executionFailed(let message) {
-                #expect(message.contains("is not running"))
+                // There's no nice way to check fail reason here
+                #expect(message.contains("is not running"), "expected container is not running if exec failed")
             }
-            #expect(try getContainerStatus(name) == "stopped")
+
+            // Give time for the exec (or start) error handling settles down
+            sleep(1)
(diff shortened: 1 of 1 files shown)


Sources/ContainerPlugin/LaunchPlist.swift
```diff
@@ -18,6 +18,8 @@
 import Foundation
 
 public struct LaunchPlist: Encodable {
+    static let debugTarget = "CONTAINER_DEBUG"
```
jglogan: This is going to be infrequently used enough that a more verbose name would be better.

```
CONTAINER_DEBUG_LAUNCHD_LABEL
```

We could do `CONTAINER_DEBUG_LAUNCHD_LABELS` as a comma separated list to debug multiple plugins?

PR #1108, opened 2026-01-29
Warn user when a named volume is auto-created

poetryofcode:
Fix applied: Implemented the TODO at                                      
  Sources/Services/ContainerAPIService/Client/Utility.swift:358 — warn the  
  user when a named volume is auto-created.                                 
                                                                            
  Previously, when a user ran a container referencing a named volume that   
  didn't exist, the volume was silently created. Now it prints:             
  Warning: volume "myvolume" not found, auto-created                        
                                                                            
  This only triggers for named volumes (not anonymous ones), matching       
  Docker's behavior of informing users about implicit resource creation.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Utility.swift
+++ b/Sources/Services/ContainerAPIService/Client/Utility.swift
@@ -21,6 +21,7 @@ import ContainerizationError
 import ContainerizationExtras
 import ContainerizationOCI
 import Foundation
+import Logging
 import TerminalProgress
 
 public struct Utility {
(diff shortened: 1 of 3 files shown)


Sources/Services/ContainerAPIService/Client/Utility.swift
```diff
@@ -355,7 +357,9 @@ public struct Utility {
             volume = try await ClientVolume.inspect(parsed.name)
         }
 
-        // TODO: Warn user if named volume was auto-created
+        if wasCreated && !parsed.isAnonymous {
```
jglogan, line 375: @poetryofcode For output that should go to stderr, use `log.warning`. You'll need to pass the logger in from the calling command.

Just use `named volume was automatically created` as the message, and use structured logging: `metadata: [ "name": "\(parsed.name)" ]

Look at other log messages in ContainerCommands for examples of this pattern.

Thank you!

jglogan (changes requested):

jglogan:
@poetryofcode Could you configure commit signing, rebase your changes onto main, and push this change again?

See https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests for details on signing.

PR #1109, opened 2026-01-29
feat: Add runtime flag to container create and run commands (#1049)

### LINKED ISSUE #1049: [Request]: Support requesting runtime types in container commands

Let's say someone wants to make a different SandboxService for creating containers. Right now, container commands like `container create`, etc do not have an option to select the runtime type to use for the container you're creating. 

### Code of Conduct

iko1:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #1049 
-  Add --runtime option for Flags.Management
-  Apply runtime handler in containerConfigFromFlags
-  Update command-reference.md doc

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -22,7 +22,7 @@ jobs:
       packages: read
     steps:
       - name: Checkout repository
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
 
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -8,6 +8,35 @@ on:
     types: [opened, reopened, synchronize]
 
 jobs:
+  verify-signatures:
+    name: Verify commit signatures
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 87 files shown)


jglogan (commented):
@iko1 Is there a use case for exposing this flag?

I'd think we'd want to expose it once another runtime helper exists somewhere, and that that helper needs to be invoked explicitly via a flag like this.

iko1:
> @iko1 Is there a use case for exposing this flag?
> 
> I'd think we'd want to expose it once another runtime helper exists somewhere, and that that helper needs to be invoked explicitly via a flag like this.

@jglogan , thanks for the review

 If @katiewasnothere  has a concrete use case for exposing it, I’m happy to update the PR accordingly; otherwise I’ll go ahead and close this and revisit when such a use case comes up.

jglogan:
@iko1 I'll check with @katiewasnothere

katiewasnothere:
@iko1 Hi! One example use case for this would be if someone wanted to make a FreeBSD runtime plugin for `container` (see [this open issue](https://github.com/apple/containerization/issues/226)) or a runc runtime plugin. It would be nice for us to be able to start and manage those containers through the same CLI commands as opposed to needing a CLI plugin to do so.

PR #1110, opened 2026-01-29
Set mtu to old Containerization default

dcantah:
The old mtu default in containerization was 1280 to account for some alpine/musl images that have issues with
1500. This changed in the last couple tags to be modifiable, but the new default is the standard 1500. Ideally we eventually allow supplying the mtu to be used when you create a network (or possibly a container creation setting), but for now just default in here back to what CZ used to use.


jglogan (approved):
Thanks @dcantah. In my local tests I was no longer seeing alpine fail with the larger MTU; failures with alpine were what prompted the change in the first place.

I'm not sure though if that has to do with a change to the environment I'm running in, because older alpine versions also work, which seems to rule out the possibility that this was an alpine issue that was fixed recently.

For now this will prevent a possible regression. We can drop 1280 and use the default once we add an mtu option for `--network` and we have a system property that allows users to override the hardcoded default.

PR #1120, opened 2026-01-31
Add resource.role label to builder container

realrajaryan:
Add resource.role label to builder container


jglogan (approved):
Thanks, Raj!

PR #1135, opened 2026-02-02
Fix usage of "clean up" as verb

DePasqualeOrg:
"clean up" is a verb, and "cleanup" is a noun. I've fixed function names and comments accordingly. I didn't mark the incorrect function name as deprecated, since deprecation notices are not desired in [this related PR](https://github.com/apple/containerization/pull/507#pullrequestreview-3741684153).

CI tests will fail until https://github.com/apple/containerization/pull/507 is merged.

PR #1138, opened 2026-02-03
[registry]: refactor function calls with new parameters

saehejkang:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context

- Updates to function calls with new parameters from the `containerization` project

## Depends On

https://github.com/apple/containerization/pull/502
https://github.com/saehejkang/containerization/pull/1

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Registry/Login.swift
+++ b/Sources/ContainerCommands/Registry/Login.swift
@@ -56,9 +56,9 @@ extension Application {
                 }
                 password = String(decoding: passwordData, as: UTF8.self).trimmingCharacters(in: .whitespacesAndNewlines)
             }
-            let keychain = KeychainHelper(id: Constants.keychainID)
+            let keychain = KeychainHelper(securityDomain: Constants.keychainID)
             if username == "" {
-                username = try keychain.userPrompt(domain: server)
+                username = try keychain.userPrompt(hostname: server)
             }
             if password == "" {
                 password = try keychain.passwordPrompt()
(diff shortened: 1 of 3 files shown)


jglogan:
@saehejkang you can now use 0.25.0, the containerization build will need to complete before you can run integration tests (need the vminit image pushed)

saehejkang:
> you can now use 0.25.0

Should I just make the changes to use the new version in this PR? Do I just update the version and run the builds? That should update the package resolve file automatically?

jglogan:
Yes, if you want to be explicit about it, update the version in `Package.swift` and then run:

```bash
swift package update containerization
```

But just running the build should work as well.

I think you're going to get errors though because @DePasqualeOrg made a change as well – I should have split those into 0.25.0 and 0.26.0 tags.

I could just merge your commits and bypass the build failures, but what might be better is for you to add https://patch-diff.githubusercontent.com/raw/apple/container/pull/1135.patch as an extra commit here and then fix up anything that needs fixing (but there shouldn't be if both PRs are individually correct).

And then for this commit, just add:

```
Co-authored-by: Anthony DePasquale <anthony@depasquale.org>
```

and we can close #1135 after merging this one.

saehejkang:
Resolved with #1161

PR #1139, opened 2026-02-03
CLI: Rework ClientContainer

dcantah:
ClientContainer was honestly extremely awkward. It could only be created by passing either a ContainerConfiguration, or a Snapshot that had to be obtained from calling a static method on the type itself. The type also did not store a connection, so every single method got a new xpc connection to the APIServer. This change aims to rework this type to be just a generic client, that is *not* a client for one specific container, but for any.

- Rename to ContainerClient
- Have list() return [ContainerSnapshot]
- Create a connection in the constructor
- Change all the callsites to use the new API
- Small, somewhat related, change to logs API in the APIServer. Now that we don't need to call get() to grab a client anymore which was typically what did "does this container exist" logic and gave a nice error message, I added a small check in the APIServer to see if the container exists and return mostly the same error message.

### DIFF
--- a/Sources/ContainerCommands/Builder/BuilderStatus.swift
+++ b/Sources/ContainerCommands/Builder/BuilderStatus.swift
@@ -16,6 +16,7 @@
 
 import ArgumentParser
 import ContainerAPIClient
+import ContainerResource
 import ContainerizationError
 import ContainerizationExtras
 import Foundation
@@ -42,7 +43,8 @@ extension Application {
 
         public func run() async throws {
             do {
-                let container = try await ClientContainer.get(id: "buildkit")
+                let client = ContainerClient()
+                let container = try await client.get(id: "buildkit")
                 try printContainers(containers: [container], format: format)
             } catch {
                 if error is ContainerizationError {
@@ -59,7 +61,7 @@ extension Application {
             [["ID", "IMAGE", "STATE", "ADDR", "CPUS", "MEMORY"]]
         }
 
-        private func printContainers(containers: [ClientContainer], format: ListFormat) throws {
+        private func printContainers(containers: [ContainerSnapshot], format: ListFormat) throws {
             if format == .json {
                 let printables = containers.map {
                     PrintableContainer($0)
@@ -88,7 +90,7 @@ extension Application {
     }
 }
 
-extension ClientContainer {
+extension ContainerSnapshot {
     fileprivate var asRow: [String] {
         [
             self.id,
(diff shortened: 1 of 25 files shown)


Sources/ContainerCommands/Builder/BuilderStatus.swift
```diff
@@ -59,12 +61,9 @@ extension Application {
...
         }
 
-        private func printContainers(containers: [ClientContainer], format: ListFormat) throws {
+        private func printContainers(containers: [ContainerSnapshot], format: ListFormat) throws {
             if format == .json {
-                let printables = containers.map {
-                    PrintableContainer($0)
-                }
-                let data = try JSONEncoder().encode(printables)
+                let data = try JSONEncoder().encode(containers)
```
realrajaryan: why are we not wrapping it in `PrintableContainer` anymore? `ContainerList` still uses it, so this creates inconsistent JSON output formats across commands

Sources/ContainerCommands/Container/ContainerInspect.swift
```diff
@@ -34,10 +35,9 @@ extension Application {
         var containerIds: [String]
 
         public func run() async throws {
-            let objects: [any Codable] = try await ClientContainer.list().filter {
+            let client = ContainerClient()
+            let objects: [any Codable] = try await client.list().filter {
                 containerIds.contains($0.id)
-            }.map {
```
realrajaryan, line 39: same as what I mentioned above

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -446,8 +446,11 @@ public actor ContainersService {
         self.log.debug("\(#function)")
 
         // Logs doesn't care if the container is running or not, just that
-        // the bundle is there, and that the files actually exist.
+        // the bundle is there, and that the files actually exist. We do
+        // first try and ge tthe container state so we get a nicer error message
```
realrajaryan: > ge tthe

nit: typo

Sources/ContainerResource/Container/ContainerSnapshot.swift
```diff
@@ -14,13 +14,24 @@
...
+
+    /// Identifier of the container.
+    public var id: String {
+        configuration.id
+    }
+
+    /// Configured platform for the container.
+    public var platform: ContainerizationOCI.Platform {
+        configuration.platform
+    }
```
realrajaryan, line 34: nit: missing newline

dcantah:
@realrajaryan Fixed!

PR #1152, opened 2026-02-04
CI: Add signed commit check

dcantah:
This requirement of ours often goes unnoticed because nothing yells at you openly about it. Lets fail CI to make it more obvious.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -8,6 +8,35 @@ on:
     types: [opened, reopened, synchronize]
 
 jobs:
+  verify-signatures:
+    name: Verify commit signatures
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check all commits are signed
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          commits=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/commits --paginate)
+          unsigned_commits=""
+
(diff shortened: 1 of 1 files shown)


Ronitsabhaya75:
Hey @dcantah, I was thinking we could use a separate .yaml workflow that runs automatically just to check if commits are signed. That way, we don't need to include this step in the main merge build/pipeline.

.github/workflows/pr-build.yml
```diff
@@ -8,6 +8,33 @@ on:
...
+            if [ "$verified" != "true" ]; then
+              unsigned_commits="$unsigned_commits  - $sha by $author\n"
+            fi
+          done < <(echo "$commits" | jq -r '.[] | "\(.sha)|\(.commit.author.name)|\(.commit.verification.verified)"')
+
+          if [ -n "$unsigned_commits" ]; then
+            echo "::error::The following commits are not signed:"
+            echo -e "$unsigned_commits"
+            echo ""
+            echo "Please sign your commits. See: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits"
```
jglogan:
```suggestion
            echo "Please sign your commits. See:"
            echo  "  - https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests"
            echo  "  - https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits"
```

jglogan (approved):

PR #1160, opened 2026-02-05
Remove @Suite(.serialized) workaround by fixing network cleanup with retries

### LINKED ISSUE #1124: [Bug]: Integration test failures - TestCLINetwork
### I have done the following

### Steps to reproduce

We were seeing test flakes with the functions in TestCLINetwork, where a single function in the suite fails intermittently in CI builds.

I've not been able to reproduce the issue locally.

It also seemed that this problem started occurring after `testIsolatedNetwork` was added.

Currently we are working around the issue by annotating the test suite with `@Suite(.serialized)`. When we have time we should identify the root cause, fix it, and remove the serialization.

### Current behavior

It seems that most often the failing test is `testIsolatedNetwork`, but other functions fail at times.

### Expected behavior

Tests should reliably pass when run in parallel, unless there are proper bugs in the application code.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main (prior to #1114)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Ronitsabhaya75:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #1124

`TestCLINetwork` was failing intermittently in CI when tests ran in parallel, requiring the `@Suite(.serialized)` workaround to force sequential execution.

**Root Cause:** Tests used `defer { try? ... }` which silently swallowed cleanup errors. When network deletion failed (e.g., container still attached), resources leaked and caused conflicts in parallel tests.

**The fix:** 
- Added `doNetworkDeleteWithRetry()` and `doStopWithRetry()` with exponential backoff (500ms, 1s, 2s)
- Fixed cleanup order: stop containers *before* deleting networks
- Replaced silent `try?` with proper error handling that logs failures to CI
- Removed unsafe force unwraps (`throw error!` → `throw error ?? fallback`)

```swift
// Before: Silent failures, wrong order
defer { _ = try? run(arguments: networkDeleteArgs) }  // Runs second
defer { try? doStop(name: name) }  // Runs first (LIFO)
```
```swift
// After: Proper order, visible errors
defer {
    do {
        try doStopWithRetry(name: name)
        try doNetworkDeleteWithRetry(name: name)
    } catch {
        print("Test cleanup failed for '\(name)': \(error)")
    }
}
```

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
+++ b/Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
@@ -58,7 +58,7 @@ class TestCLINetwork: CLITest {
                 } catch {
                     print("Container stop threw error for '\(name)': \(error)")
                 }
-                
+
                 do {
                     let deleteResult = try run(arguments: ["network", "rm", name])
                     if deleteResult.status != 0 {
@@ -231,7 +231,7 @@ class TestCLINetwork: CLITest {
         if result.status != 0 {
             throw CLIError.executionFailed("command failed: \(result.error)")
         }
-        
+
         defer {
             // Proper cleanup order: stop container first, then delete network
             do {
@@ -245,7 +245,7 @@ class TestCLINetwork: CLITest {
             } catch {
                 print("Container stop threw error for '\(name)': \(error)")
             }
-            
+
             do {
                 let deleteResult = try run(arguments: ["network", "rm", name])
                 if deleteResult.status != 0 {
@@ -258,7 +258,7 @@ class TestCLINetwork: CLITest {
                 print("Network cleanup threw error for '\(name)': \(error)")
             }
         }
-        
+
         let port = UInt16.random(in: 50000..<60000)
         try doLongRun(
             name: name,


Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
```diff
@@ -48,17 +47,19 @@ class TestCLINetwork: CLITest {
                 throw CLIError.executionFailed("command failed: \(result.error)")
             }
             defer {
-                _ = try? run(arguments: networkDeleteArgs)
+                do {
+                    try doStopWithRetry(name: name)
```
jglogan: In real code, we don't retry cleanup errors, so I don't think we should do it here to get things to pass.

Perhaps you can use this PR to ascertain why we see errors when cleaning up on network tests? Perhaps just change `_ = try? run(arguments: networkDeleteArgs)` to `let result = try run(arguments: networkDeleteArgs)` and on an error, log the name, error, and stdout, stderr, and status

jglogan (approved):

Tests/CLITests/Subcommands/Networks/TestCLINetwork.swift
Ronitsabhaya75: You're right that retries hide the real issue. I'll update this to just add diagnostic logging instead, so that way we can see exactly why cleanup is failing.

jglogan:
@Ronitsabhaya75 needs a `make fmt`

PR #1163, opened 2026-02-05
CLI: Forego logging errors on multi-container operations

dcantah:
Instead of logging errors, and then additionally throwing an error stating what containers couldn't be stopped/killed/deleted, let's just concatenate the errors and throw the single error.

### DIFF
--- a/Sources/ContainerCommands/AggregateError.swift
+++ b/Sources/ContainerCommands/AggregateError.swift
@@ -1,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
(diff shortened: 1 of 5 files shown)


Sources/ContainerCommands/AggregateError.swift
```diff
@@ -0,0 +1,32 @@
+//===----------------------------------------------------------------------===//
```
jglogan, line 1: Would we ever have to deal with aggregate errors below the command level? If so, would this be better defined along with `container`-specific error types in ContainerResource?

dcantah: For now I don't see much use, it's mostly a cli presentation helper

PR #1164, opened 2026-02-05
chore: Replace force-unwrapped String(data:encoding:) with String(decoding:as:)

manojmahapatra:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update
- [x] Chore

## Motivation and Context
Use the non-optional String(decoding:as:) initializer for converting JSON-encoded Data to String. This is safe for UTF-8 and eliminates force unwraps.

## Testing
No new tests needed for this change. It's a 1:1 API swap with identical behavior - String(decoding:as:) produces the same output as String(data:encoding:)! for valid UTF-8.


jglogan (approved):
Thanks! Didn't realize we still had those force-unwraps!

Issue #1019, opened 2026-01-05
[Request]: container image commit


sungkim11:
### Feature or enhancement request details

Can I save a changed container as an image?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
There is no feature like `container commit` to save a container as an image today.

PR #1172, opened 2026-02-06
Support exporting current container to image

JaewonHur:
This PR adds a command to export container to image (#1019).

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Users can export their container to image.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Containers/TestCLIExport.swift
+++ b/Tests/CLITests/Subcommands/Containers/TestCLIExport.swift
@@ -26,11 +26,21 @@ class TestCLIExportCommand: CLITest {
         let name = getTestName()
         try doLongRun(name: name, autoRemove: false)
         defer {
+            try? doStop(name: name)
             try? doRemove(name: name)
         }
 
         let mustBeInImage = "must-be-in-image"
         _ = try doExec(name: name, cmd: ["sh", "-c", "echo \(mustBeInImage) > /foo"])
+
+        _ = try doExec(name: name, cmd: ["sh", "-c", "mkdir -p /parent/child"])
+        let hardlinkMustRemain = "hardlink-must-remain"
+        _ = try doExec(name: name, cmd: ["sh", "-c", "echo \(hardlinkMustRemain) > /parent/child/bar"])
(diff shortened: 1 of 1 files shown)


Sources/ContainerCommands/ExportCommand.swift
```diff
@@ -0,0 +1,67 @@
...
+import Foundation
+import TerminalProgress
+
+extension Application {
+    public struct ExportCommand: AsyncLoggableCommand {
+        public init() {}
+        public static var configuration: CommandConfiguration {
+            var config = CommandConfiguration()
+            config.commandName = "export"
+            config.abstract = ""
```
ajemory: Add a help message to .abstract

Sources/ContainerCommands/ExportCommand.swift
```diff
@@ -0,0 +1,67 @@
...
+import ContainerAPIClient
+import ContainerizationError
+import Foundation
+import TerminalProgress
+
+extension Application {
+    public struct ExportCommand: AsyncLoggableCommand {
+        public init() {}
+        public static var configuration: CommandConfiguration {
+            var config = CommandConfiguration()
```
ajemory: nit: it's cleaner to pass the config options as parameters here instead of initializing config and setting them after

Sources/ContainerCommands/ExportCommand.swift
```diff
@@ -0,0 +1,67 @@
...
+            config.commandName = "export"
+            config.abstract = ""
+            config._superCommandName = "container"
```
ajemory: I would move this entire file into ContainerCommands/Container, rename it ContainerExport and remove _superCommandName (Moving it to Image instead of Container could also make sense). This will also set helpNames to -h and --help by default, so you can remove that as well

JaewonHur:
@jglogan This PR has minimal test writing a file before exporting a container and checking it remains correct in the exported image.
Do you have any edge cases to be checked ? e.g., volume mount should not be preserved..?

ajemory (approved):

PR #1175, opened 2026-02-07
APIServer: Add support for filtering to list rpc

dcantah:
This is not intended to be used to support `--filter` or similar on the CLIs list yet, it's solely to clean up our rather awkward use of `ContainerClient.list()` today in the CLI. The list RPC simply returns all of the containers we have created. Because of this, for a LOT of our commands we filter to what we need client side, which feels like a waste.. This change introduces a filter struct that we can provide an array of container IDs, labels, and the status of the containers to filter the `list()` output from.

This additionally, because it was killing (pun not intended) me and I was already having to change this area for the `list()` additions, changes container kill slightly to return an error if you try and kill a container that doesn't exist.

### DIFF
--- a/Sources/ContainerResource/Container/ContainerListFilters.swift
+++ b/Sources/ContainerResource/Container/ContainerListFilters.swift
@@ -1,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// Filters for listing containers.
+public struct ContainerListFilters: Sendable, Codable {
+    /// Filter by container IDs. If non-empty, only containers with matching IDs are returned.
+    public var ids: [String]
+    /// Filter by container status.
+    public var status: RuntimeStatus?
+    /// Filter by labels. All specified labels must match.
+    public var labels: [String: String]
+
+    /// No filters applied. Will return all containers.
+    public static let all = ContainerListFilters()
+
+    public init(
+        ids: [String] = [],
+        status: RuntimeStatus? = nil,
+        labels: [String: String] = [:]
+    ) {
+        self.ids = ids
+        self.status = status
+        self.labels = labels
+    }
+}
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -106,10 +106,33 @@ public actor ContainersService {
         return results
     }
 
-    /// List all containers registered with the service.
-    public func list() async throws -> [ContainerSnapshot] {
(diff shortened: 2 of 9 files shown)


Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -106,10 +106,33 @@ public actor ContainersService {
...
+    /// List containers matching the given filters.
+    public func list(filters: ContainerListFilters = .all) async throws -> [ContainerSnapshot] {
         self.log.debug("\(#function)")
-        return self.containers.values.map { $0.snapshot }
+
+        return self.containers.values.compactMap { state -> ContainerSnapshot? in
+            let snapshot = state.snapshot
+
+            if !filters.ids.isEmpty {
+                guard filters.ids.contains(snapshot.id) else {
```
manojmahapatra, line 117: **Question**: I think this will be a breaking change for anyone using short IDs. We will probably need;
```Swift
guard filters.ids.contains(where: { snapshot.id == $0 || snapshot.id.hasPrefix($0) }) else {
        return nil
}
```

Sources/ContainerResource/Container/ContainerListFilters.swift
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// Filters for listing containers.
+public struct ContainerListFilters: Sendable, Codable {
+    /// Filter by container IDs. If non-empty, only containers with matching IDs are returned.
+    public var ids: [String]
```
manojmahapatra, line 22: **Suggestion**: We should consider using strongly-typed `ContainerID` instead of `[String]` for the `ids` field. This would prevent accidentally passing wrong ID types.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
dcantah: We don't support prefix matching/short IDs today, but this change is directly preceding a change I have to add this :)

Sources/ContainerResource/Container/ContainerListFilters.swift
dcantah: The IDs today are fundamentally just strings, so there's not much else I can pass at the moment

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
manojmahapatra: Got it. I saw `guard let container = allContainers.first(where: { $0.id == containerId || $0.id.starts(with: containerId) })` in ContainerStats and thought there was already a prefix matching exists today.

dcantah: Ahh, I might've added some crummy half-support to stats. The change I have in mind is the prefix matching is done in the server as well so we don't need to filter client side

manojmahapatra: makes sense, thank you!

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -106,10 +106,33 @@ public actor ContainersService {
...
-    public func list() async throws -> [ContainerSnapshot] {
+    /// List containers matching the given filters.
+    public func list(filters: ContainerListFilters = .all) async throws -> [ContainerSnapshot] {
         self.log.debug("\(#function)")
-        return self.containers.values.map { $0.snapshot }
+
+        return self.containers.values.compactMap { state -> ContainerSnapshot? in
+            let snapshot = state.snapshot
+
+            if !filters.ids.isEmpty {
```
jglogan, line 116: nit: not necessarily for this PR - should `ContainerListFilters` implement the filter test?

Then it's easier to write unit tests for the filtering logic.

Issue #158, opened 2025-06-12
[Request]: Add auto-start functionality on boot and container auto-start.


Phfebb:
### Feature or enhancement request details

Does it not have the auto-start function upon booting up?

(edit: see also related #286 for restart policy for individual containers, the features complement one another so both would be nice to have)

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

x8x:
Meanwhile this works:

1. Create your custom container
```
container system start
container run -d --name nginx nginx
```

2. Create your 'auto-start' file, ex: `~/Library/LaunchAgents/user.nginx.plist`
```
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN"
    "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
	<dict>
		<key>Label</key>
		<string>user.nginx</string>
		<key>ProgramArguments</key>
		<array>
		    <string>/usr/bin/script</string>
			<string>-q</string>
			<string>/dev/null</string>
			<string>/bin/sh</string>
			<string>-c</string>
			<string>/usr/local/bin/container system start && sleep 3 && /usr/local/bin/container start nginx</string>
		</array>
		<key>RunAtLoad</key>
		<true />
		<key>KeepAlive</key>
		<false />
		<key>StandardOutPath</key>
		<string>/tmp/container.log</string>
		<key>StandardErrorPath</key>
		<string>/tmp/container.log</string>
	</dict>
</plist>
```

3. Reboot

jglogan:
This should probably be split into separate issues as they are distinct features:

- Add restart policy for containers (#286).
- Start `container` automatically when the system boots.
- Start `container` automatically when I log in.

jglogan:
@Phfebb OK if I modify the title to just "Add auto-start functionality on boot" since container restart policy is covered by #286?

Phfebb:
> OK if I modify the title to just "Add auto-start functionality on boot" since container restart policy is covered by #286?

ok

goranche:
I liked the temporary "solution" @x8x provided [above](#issuecomment-3018438587), but wanted something a bit more flexible, so expanded on it.

I wrote a [script](https://github.com/goranche/apple-container/blob/goranche/launch-on-boot/scripts/launch-on-boot.sh) that starts the container system up, and then iterates over containers that have a label named "restart" with the value "boot". 
[...]

PR #1176, opened 2026-02-07
[system]: add LaunchAgent for container auto-start on user login

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Relates to #158

## Testing

```bash
# Enable launch agent
container system launch-agent enable
# Check the registered label exists
launchctl list | grep container.launchagent
```

Open up your `System Settings` -> `General` -> `Login Items and Extensions` and you should see the `launch-container` background activity. You can now stop your `container` system (or leave it running), log out of your account, log back in, wait for everything to load up, and the `container` system should already be started!

```bash
# Disable launch agent
container system launch-agent disable
# Check the registered label is removed
launchctl list | grep container.launchagent
```
Open up your `System Settings` -> `General` -> `Login Items and Extensions` and you should no longer see (might have to wait a little bit for it to register) the `launch-container` background activity. Now if you log out of your account, log back in, wait for everything to load up, the `container` system WILL NOT be running (as expected).

- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/scripts/launch-container.sh
+++ b/scripts/launch-container.sh
@@ -25,6 +25,28 @@ PATH=$PATH:/usr/local/bin
 
 # Check that required commands are installed
 command -v container >/dev/null 2>&1 || error "container CLI is not installed"
+command -v expect >/dev/null 2>&1 || error "expect is not installed"
 
 # Start the container system
-container system start || error "Failed to start container system"
+expect <<EOF
+set timeout 30
+
+spawn container system start
+
+expect {
+    -re "Install.*recommended.*kernel.*" {
+        send "Y\r"
+        exp_continue
+    }
+    eof
+}
+
+catch wait result
+exit [lindex \$result 3]
+EOF
+
+START_EXIT_CODE=$?
+
+if [ $START_EXIT_CODE -ne 0 ]; then
+    error "container system start failed with exit code $START_EXIT_CODE"
+fi


docs/command-reference.md
```diff
@@ -1223,3 +1223,30 @@ container system property clear dns.domain
...
+```
+
+### `container system login-launch enable`
+
+Enables the `container` to be launched automatically on login.
+
+**Usage**
+
+```bash
+container system login-launch enable [--debug]
```
manojmahapatra: **Question**: The docs has `--debug` flag for login-launch enable/disable, but the code doesn't use logOptions - the flag is accepted but ignored.

docs/command-reference.md
```diff
@@ -1223,3 +1223,30 @@ container system property clear dns.domain
...
+### `container system login-launch disable`
+
+Disables the `container` to be launched automatically on login.
+
+**Usage**
+
+```bash
+container system login-launch disable [--debug]
+```
+**Options**
```
manojmahapatra, line 1274: Suggestion: It has "No options" which contradicts the `[--debug]` shown in usage - this should be fixed either way.

Sources/ContainerCommands/System/LoginLaunch/LoginLaunchDisable.swift
```diff
@@ -0,0 +1,45 @@
...
+            abstract: "Disable the container to launch on user login"
+        )
+
+        @OptionGroup
+        public var logOptions: Flags.Logging
+
+        public init() {}
+
+        public func run() async throws {
+            let logger = Logger(label: "LoginLaunchDisable", factory: { _ in StderrLogHandler() })
```
manojmahapatra: Suggestion: This logger instance is redundant and not required. `AsyncLoggableCommand` has already a `log` and can be used here instead.

Sources/ContainerCommands/System/LoginLaunch/LoginLaunchEnable.swift
```diff
@@ -0,0 +1,46 @@
...
+            abstract: "Enable the container to launch on user login"
+        )
+
+        @OptionGroup
+        public var logOptions: Flags.Logging
+
+        public init() {}
+
+        public func run() async throws {
+            let logger = Logger(label: "LoginLaunchEnable", factory: { _ in StderrLogHandler() })
```
manojmahapatra: same feedback as above.

docs/command-reference.md
saehejkang: The "Options" below are omitted in the `command-reference` because they exist for every command.
```
  -h, --help              Show help information.
```

Unless there was a fourth option, only then that option would be added to the docs.

docs/command-reference.md
saehejkang: The command accepts a `--debug` flag because `logOptions` is required by the `AsyncLoggableCommand` protocol. A large refactor is needed, based on work done in #1066, to standardize logging across all commands. The flag doesn’t have any visible effect yet because debug-level logs still need to be added in the appropriate places (#507). It is all in a weird place, so that is why `--debug` is still accepted, but not fully functional at the moment.

Sources/ContainerCommands/System/LoginLaunch/LoginLaunchEnable.swift
saehejkang: same comment as above

Sources/ContainerCommands/System/LoginLaunch/LoginLaunchDisable.swift
saehejkang: Again, everything is all over the place. Take a look at #642 for some context on this.

docs/command-reference.md
manojmahapatra: Got it. So here it what's happening w/ the changes IIUC;
(a) The --debug flag is accepted (via logOptions)
(b) But it's never used since LoginLaunch(Enable/Disable) creates a hardcoded StderrLogHandler() instead of respecting the debug flag.
(c) As suggested `log` from `AsyncLoggableCommand` would respect the flag.

docs/command-reference.md
manojmahapatra: Ahh, I see. So it's a known tech debt.

docs/command-reference.md
manojmahapatra: Please disregard this.

Issue #563, opened 2025-08-31
[Bug]: Passing --rm flag to `run` masks cause of error


anderseknert:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Compare the output of the below command with and without the `--rm` flag.
```
> container run --rm openpolicyagent/opa:1.8.0-static version
```

### Current behavior

```
> container run --rm openpolicyagent/opa:1.8.0-static version
Error: interrupted: "internalError: "failed to start container" (cause: "interrupted: "XPC connection error: Connection interrupted"")
Ensure container system service has been started with `container system start`."
```

With `--rm`, the command fails instantaneously.

The logs provide no further insight:
```
2025-08-31 12:43:11.903878+0200 0x6b89b    Info        0x0                  7969   0    container-apiserver: [com.apple.container:APIServer] Handling container 6b1f948b-9a1d-47ab-946d-dab24912b9c1 exit. Code -1
2025-08-31 12:43:11.930672+0200 0x6c031    Error       0x0                  7969   0    container-apiserver: [com.apple.container:APIServer] xpc client handler connection error Connection invalid
```

Running the same command without `--rm`, the cause of the error is provided:
```
> run openpolicyagent/opa:1.8.0-static version
Error: internalError: "failed to start container" (cause: "internalError: "failed to start process (cause: "internal error (13): create managed process: internalError: "Cannot find User '1000' in passwd file."")"")
```

Without `--rm`, the command fails after 30 seconds or so.

Logs are also helpful (although not needed as the cause is printed to the console):
```
2025-08-31 12:48:27.430922+0200 0x6c10a    Info        0x0                  7969   0    container-apiserver: [com.apple.container:APIServer] Handling container fc075355-45e7-4158-b7cb-ffb7283c26a7 exit. Code -1
2025-08-31 12:48:27.431287+0200 0x6d08e    Error       0x0                  8699   0    container-runtime-linux: [com.apple.container:RuntimeLinuxHelper] handler for com.apple.container.sandbox/start threw error internalError: "failed to start process" (cause: "internal error (13): create managed process: internalError: "Cannot find User '1000' in passwd file."") [uuid=fc075355-45e7-4158-b7cb-ffb7283c26a7]
```

### Expected behavior

[...]

PR #1190, opened 2026-02-10
Add container copy/cp command for host-container file transfer

### LINKED ISSUE #232: [Request]: Add `container cp` command to copy files into and out of running containers

Many container runtimes offer a `cp` command which can copy files into and out of running containers.   

Usually the `cp` command has a few of main modes of operation:
* copy a single file in or out of the container
* recursively copy a directory in or out of the container
* create a tar archive of a path in the container and stream it out to the host, or receive a tar stream from the host and unpack it in the container

There are also often options to control the details of how symlinks and permissions are handled.

[Swift SDK Generator](https://github.com/swiftlang/swift-sdk-generator/) currently relies on the tar streaming mode to extract Swift SDKs from container images: [Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator.swift#L122](https://github.com/swiftlang/swift-sdk-generator/blob/af1bea6204fbd43ebd14ad62a6e05233b8ba2d30/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator.swift#L122)

It's possible to work around the lack of `cp` by `exec`ing `tar` in the container image and piping the output to `tar` on the host, however this isn't a general solution because the container might not have a `tar` binary. 

### Code of Conduct

simone-panico:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
Closes #232.

Adds the `container copy` (aliased as `cp`) command to copy files between a running container and the local filesystem.

I saw [#1023](https://github.com/apple/container/pull/1023) and the feedback from @dcantah — the previous attempt relied on tar being installed inside the container. 
This implementation takes the recommended approach:
file transfers go through the guest agent via the existing `copyIn`/`copyOut` methods on the core `Containerization`, with no dependency on container tooling.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs


Sources/ContainerCommands/Container/ContainerCopy.swift
```diff
@@ -0,0 +1,81 @@
...
+
+extension Application {
+    public struct ContainerCopy: AsyncLoggableCommand {
+        enum PathRef {
+            case local(String)
+            case container(id: String, path: String)
+        }
+
+        static func parsePathRef(_ ref: String) -> PathRef {
+            if let colonIdx = ref.firstIndex(of: ":") {
```
JaewonHur: You may want to use `ref.components(separatedBy: ":")`.

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
```diff
@@ -297,6 +297,43 @@ public struct ContainerClient: Sendable {
         return fh
     }
 
+    /// Copy a file from the host into the container.
+    public func copyIn(id: String, source: String, destination: String, mode: UInt32 = 0o644) async throws {
```
JaewonHur: It'd be better to get `source` and `destination` as `URL` type and convert them to string here.

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
```diff
@@ -297,6 +297,43 @@ public struct ContainerClient: Sendable {
...
+            throw ContainerizationError(
+                .internalError,
+                message: "failed to copy into container \(id)",
+                cause: error
+            )
+        }
+    }
+
+    /// Copy a file from the container to the host.
+    public func copyOut(id: String, source: String, destination: String) async throws {
```
JaewonHur: Same as `copyIn`

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -468,6 +468,24 @@ public actor ContainersService {
         }
     }
 
+    /// Copy a file from the host into the container.
+    public func copyIn(id: String, source: String, destination: String, mode: UInt32) async throws {
+        self.log.debug("\(#function)")
+
+        let state = try self._getContainerState(id: id)
+        let client = try state.getClient()
```
JaewonHur, line 774: Need to check container is running.
It's a bit different as container requires the agent in VM to be running to get the file copied.
(docker does not)

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -468,6 +468,24 @@ public actor ContainersService {
...
+        try await client.copyIn(source: source, destination: destination, mode: mode)
+    }
+
+    /// Copy a file from the container to the host.
+    public func copyOut(id: String, source: String, destination: String) async throws {
+        self.log.debug("\(#function)")
+
+        let state = try self._getContainerState(id: id)
```
JaewonHur, line 782: Need to check container is running.

JaewonHur:
Related to #232

docs/how-to.md
```diff
@@ -60,6 +60,31 @@ total 4
 %
 </pre>
 
+## Copy files between host and container
```
JaewonHur: This might not be needed here.

Sources/Services/ContainerSandboxService/Client/SandboxClient.swift
```diff
@@ -275,6 +275,39 @@ extension SandboxClient {
         }
     }
 
+    public func copyIn(source: String, destination: String, mode: UInt32) async throws {
+        let request = XPCMessage(route: SandboxRoutes.copyIn.rawValue)
+        request.set(key: SandboxKeys.sourcePath.rawValue, value: source)
+        request.set(key: SandboxKeys.destinationPath.rawValue, value: destination)
+        request.set(key: SandboxKeys.fileMode.rawValue, value: UInt64(mode))
```
JaewonHur, line 291: Why UInt64 here?

simone-panico: Because the `set` Method in `XPCMessage.swift` expects a UInt64

simone-panico: Do you want to handle it differently? I looked at the code again and couldn't find a better option

Sources/ContainerCommands/Container/ContainerCopy.swift
```diff
@@ -0,0 +1,84 @@
...
+extension Application {
+    public struct ContainerCopy: AsyncLoggableCommand {
+        enum PathRef {
+            case local(String)
+            case container(id: String, path: String)
+        }
+
+        static func parsePathRef(_ ref: String) -> PathRef {
+            let parts = ref.components(separatedBy: ":")
+            if parts.count == 2 {
```
JaewonHur: How about this?

```
let parts = ref.components(separatedBy: ":")
switch parts.count {
case 1:
    return .local(ref)
case 2 where !parts[0].isEmpty && parts[1].starts(with: "/"):
    return .container(id: parts[0], path: parts[1])
default:
    throw ContainerizationError(.invalidArgument, message: "invalid path given: \(ref)")
}
```

JaewonHur:
⏺ `CopyOut`/`CopyIn` Scenarios: Current vs Docker Behavior

  Scenario 1: `copy(in/out) src dst` (**`src` must exist**)

  - File, Directory → Non-existing:
    - Docker (desired): save as `dst`
    - Current: File is saved as `dst`, Directory is not supported.

  - File → File
    - Docker (desired): replace `dst`
    - Current: replace `dst`

  - Directory → File
    - Docker (desired): error
    - Current: Directory is not supported

  - File, Directory → Directory
    - Docker (desired): save under `dst`
    - Current: _**File returns error**_, Directory is not supported.

  Scenario 2: `copy(in/out) src dst/` (**`dst` must be a directory if existing**)

  - File → Non-existing:
    - Docker (desired): error no such directory
    - Current: _**`dst` is created and `src` file is saved under `dst`**_.

  - Directory → Non-existing:
    - Docker (desired): save as `dst`
    - Current: Directory is not supported.

  - File, Directory → Directory
    - Docker (desired): save under `dst`
    - Current: File is saved under `dst`. Directory is not supported.

  Scenario 3: `copy(in/out) src/ dst` (**src must be an existing directory**)

  - Directory → Non-existing:
    - Docker (desired): save as `dst`
    - Current: Directory is not supported.

  - Directory → File
    - Docker (desired): error
    - Current: Directory is not supported.

  - Directory → Directory
    - Docker (desired): save under `dst`
    - Current: Directory is not supported.

  Scenario 4: `copy(in/out) src/ dst/` (**src must be an existing directory & dst must be a directory if existing**)
 
  - Directory → Non-existing:
    - Docker (desired): save as `dst`
    - Current: Directory is not supported.

  - Directory → Directory
    - Docker (desired): save under `dst`
    - Current: Directory is not supported.

JaewonHur:
Hi @simone-panico
Above I summarized the desired behavior and current behavior.

The bold italic phrases are what is different from desired behavior (i.e., docker) now, and can be addressed.

Below is general comments.

1. Copying directory is not supported yet, we need to handle it later.
2. If `dst/`, it would be better to check in command line if `dst` is directory if it exists.

Sources/ContainerCommands/Container/ContainerCopy.swift
```diff
@@ -0,0 +1,84 @@
...
+
+        @Argument(help: "Destination path (container:path or local path)")
+        var destination: String
+
+        public func run() async throws {
+            let client = ContainerClient()
+            let srcRef = Self.parsePathRef(source)
+            let dstRef = Self.parsePathRef(destination)
+
+            switch (srcRef, dstRef) {
```
JaewonHur, line 64: We need a sanitization logic here to check local path.

1. If `localPath` is source
  a. Check if file or directory at `localPath` exists.
  b. If `localPath` ends with `/`, check it's directory.

2. If `localPath` is dest
  a. If `localPath` ends with `/`, need to check if it's not existing or a directory.

Then, we need to extend `destURL` correctly following the comment I left.

JaewonHur:
Hi @simone-panico 
Could you try this PR with apple/containerization#474, which supports copying directories.

You may need to build this project with local containerization ([BUILDING.md](https://github.com/apple/container/blob/main/BUILDING.md#develop-using-a-local-copy-of-containerization)), and wire up `SandboxService`.

dcantah:
@JaewonHur the change above does not allow copying directories. It adds a tar reader/writer that will be used to support this, but it's not in the change itself.

simone-panico:
If you’re interested, I could try to create a Directory Copier within `containerization`. It would also be a great challenge for me

simone-panico:
@JaewonHur @dcantah, I created [apple/containerization#543](https://github.com/apple/containerization/issues/543) for clarity. 

As soon as [apple/containerization#474](https://github.com/apple/containerization/pull/474) is merged I think I could start with implementing the directory copying. What do you think?

JaewonHur:
Hi @simone-panico, sorry for the delay!

We have an internal discussion how to better support `container cp`, and want to pause this for now. We will definitely revisit this PR once every thing becomes clear! Thanks!

simone-panico:
Hey Guys, I experimented a bit with the [ContainerizationTar](https://github.com/apple/containerization/pull/474) and got a running version for Copying Directories [#563](https://github.com/apple/containerization/pull/563/changes). It would be great if you could give me a short Feedback if that is the right way to go

JaewonHur:
Hi @simone-panico, sorry for the late reply!

We finally merged support copying directory in containerization (apple/containerization#571).

Previously, we did custom implementation of tar reader/writer (apple/containerization#474) for vminitd (which runs in linux VM) as the swift linux sdk didn't include libarchive.
However, they are finally included, and we are back using it so that we can copy directories using tar.

You would be able to check it under apple/containerization#571, and we can continue this PR to add full support for `container copy` :)

Sources/Services/ContainerAPIService/Client/XPC+.swift
```diff
@@ -124,14 +124,19 @@ public enum XPCKeys: String {
     case volumeContainerId
 
     /// Container statistics
-    case statistics
+    case statisticshttps://github.com/apple/container/pull/1190/conflict?name=Sources%252FHelpers%252FAPIServer%252FAPIServer%252BStart.swift&ancestor_oid=53175b91c6a1d7019854596f41660d0e45600ba0&base_oid=155d6c3448d8a05053ad3494b8ab58ef619757b2&head_oid=f5de594190fdf80b039e02950c732a203ddaeca0
```
JaewonHur: Oops sorry! It was wrongly introduced while merging the branch.
Could you fix it later?

JaewonHur:
Here is test script.
You can run either `./run_copy_test.sh docker` or `./run_copy_test.sh container`.

[run_copy_tests.sh](https://github.com/user-attachments/files/25879578/run_copy_tests.sh)

JaewonHur:
Current test result

```
=========================================
Test Results Summary
=========================================
Total : 70
Passed: 54
Failed: 16

Failed tests:
  ✗ CopyOut S2: Directory → Non-existing/: succeeds
  ✗ CopyOut S2: Directory → Non-existing/: dst created
  ✗ CopyOut S2: Directory → Non-existing/: dst/file1.txt
  ✗ CopyOut S4: Directory/ → Non-existing/: succeeds
  ✗ CopyOut S4: Directory/ → Non-existing/: dst created
  ✗ CopyOut S4: Directory/ → Non-existing/: dst/file1.txt
  ✗ CopyIn S1: File → Existing directory: succeeds
  ✗ CopyIn S1: File → Existing directory: dst/file.txt exists
  ✗ CopyIn S1: Directory → Existing directory: dst/dir exists
  ✗ CopyIn S1: Directory → Existing directory: dst/dir/file1.txt
  ✗ CopyIn S2: File → Non-existing/: errors
  ✗ CopyIn S2: Directory → Non-existing/: dst/file1.txt
  ✗ CopyIn S3: Directory/ → Existing directory: dst/dir exists
  ✗ CopyIn S3: Directory/ → Existing directory: dst/dir/file1.txt
  ✗ CopyIn S3: Directory/ → Existing directory: dst/dir/sub exists
  ✗ CopyIn S4: Directory/ → Non-existing/: dst/file1.txt
```

JaewonHur:
`./run_copy_test.sh container` has some failing cases.

simone-panico:
> `./run_copy_test.sh container` has some failing cases.

Yes, I already saw it, but Thank you very much 😄 
I haven't had a chance to dig into it yet, but I believe this is a deeper issue. I think there is also a bug in the containerization implementation, but not 100% sure. 
I will get back to you :)

JaewonHur:
Thank you! :)

simone-panico:
Hey @JaewonHur, all the tests are passing now :) 
As mentioned, I had to make some changes in [586#containerization](https://github.com/apple/containerization/pull/586) to get everything working. All the changes I made to Containerization were done before to the [grpc-swift-2 upgrade](https://github.com/apple/containerization/pull/578), since they wouldn't have been compatible with container otherwise. Once [1308#Container](https://github.com/apple/container/issues/1308) is done and merged I'll adapt the changes and run the tests again.

JaewonHur:
Hi @simone-panico thank you so much for all your hard work!

I looked into the code, but my small concern is whether it is the best approach to modify vminitd API. Current vminitd's `copy(src, dst)` API, looks to me, has clear semantics: create `dst` if non-existing, overwrite if both `src` and `dst` are files, extract to `dst` if both are directories, otherwise error. Adding additional parameters here might make it unclear for later implementation of new vminitd.

On the other hand, I see the problem is we cannot stat whether a given path in guest container is file, directory or non-existing.

We might need more discussion. One possible other direction would be to rely on exception handling. For example, given `container cp foo guest:/bar/`, first try `copy(foo, guest:/bar/foo)`, and if it fails, then try `copy(foo, guest:/bar)` if `foo` is directory.

@dcantah @jglogan Do you have any thoughts?

simone-panico:
> Hi @simone-panico thank you so much for all your hard work!
> 
> I looked into the code, but my small concern is whether it is the best approach to modify vminitd API. Current vminitd's `copy(src, dst)` API, looks to me, has clear semantics: create `dst` if non-existing, overwrite if both `src` and `dst` are files, extract to `dst` if both are directories, otherwise error. Adding additional parameters here might make it unclear for later implementation of new vminitd.
> 
> On the other hand, I see the problem is we cannot stat whether a given path in guest container is file, directory or non-existing.
> 
> We might need more discussion. One possible other direction would be to rely on exception handling. For example, given `container cp foo guest:/bar/`, first try `copy(foo, guest:/bar/foo)`, and if it fails, then try `copy(foo, guest:/bar)` if `foo` is directory.
> 
> @dcantah @jglogan Do you have any thoughts?

Yeah, I completely see your point, it's not the most elegant solution. I think handling it with the retry approach is also not the best solution, because if it starts creating and writing files and then fails we would need to first delete everything and then recreate. 

Another option could be by making a `Stat RPC` that gives back two Bools (`exists` and `is_directory`). Like that we could handle the logic in `Container` but get the relevant Information we need. Something like the [Unix stat syscall](https://www.ibm.com/docs/en/i/7.6.0?topic=ssw_ibm_i_76/apis/stat.html).

JaewonHur:
> if it starts creating and writing files and then fails we would need to first delete everything and then recreate.

Could you describe more how it fails in the middle of writing files? I thought vminitd's `copy` would fail before starting any persistent change if initial conditions don't meet.

> Another option could be by making a Stat RPC that gives back two Bools (exists and is_directory). Like that we could handle the logic in Container but get the relevant Information we need. Something like the [Unix stat syscall](https://www.ibm.com/docs/en/i/7.6.0?topic=ssw_ibm_i_76/apis/stat.html).

Yes, we might modify current `copy` RPC or add an RPC for `stat`.

simone-panico:
> `container cp foo guest:/bar/`, first try `copy(foo, guest:/bar/foo)`, and if it fails, then try `copy(foo, guest:/bar)` if `foo` is directory.

I initially thought the first `copy(foo, guest:/bar/foo)` would create parent directories as a side effect, making rollback necessary. But actually we can do `copy(foo, guest:/bar/foo, createParents: false)` on the first call, so it fails.  
My Bad ;)


> Yes, we might modify current `copy` RPC or add an RPC for `stat`.

Sure, just let me know what the team decides and I'll implement it :)

JaewonHur:
Thanks a lot! :)
Will be in touch soon!

JaewonHur:
Hi @simone-panico we'd like to keep the API change in vminitd as small as possible.

Would it be able to achieve it without changing vminitd?

simone-panico:
> Hi @simone-panico we'd like to keep the API change in vminitd as small as possible.
> 
> Would it be able to achieve it without changing vminitd?

Hey, yeah sure, I'll try it :)

vxnkjyffmq-code (approved):

JaewonHur:
Hi @simone-panico 

Following is my pseudo code.

Given `cp src dst`
1. If `dst` is directory
  * copy `src` to `dst/src`
2. If `dst` is file
  * `src` must be file, and `dst` is replaced with `src`.
3. If `dst` is not existing
  * If `dst` has trailing `/`, and `src` is file, error
  * else, just create `dst` and write `src`

Based on it, I thought how the logic would be if we can get an error root cause from the failed copy (going retry approach).

Given `cp src dst`, we perform the following logic (`copy` in the following I mean vminitd's `copy`)
```
if copy src to dst/src succeeds,
    done
else
    if the error indicates dst = file
        if src = file
            copy src to dst
        else
            error
    else if the error indicates dst = non-existing
        if src = dir
            copy src to dst
        else
            if dst ends with "/"
                error
            else
                copy src to dst
    else
        error
```

It's so ugly lol. But, without something like `stat` RPC, I cannot come out with something better.
(actually we are using that error message instead of `stat`)

Oh but this logic only works for `copyIn` scenario as we need to stat `src`.
For `copyOut`

```
If dst = directory
    copy src to dst/src
else
    if dst ends with "/"
        copy src to dst
        then, check if src is file, and yes, then remove and error
    else
        copy src to dst (and let it error if those type are different)
```

LOL it's even more ugly.


How do you think?

simone-panico:
Hey @JaewonHur, Thank you for the Pseudo.
Yes, it's not the prettiest . I'd still highly recommand adding a `Stat RPC`. 
That said, I went ahead and implemented it using the exception handling approach (like your Pseudo) and all the Tests are passing :)

I also found a bug that I had to fix in [590#containerization](https://github.com/apple/containerization/pull/590), so that PR needs to be merged first. 

If you and your team change your mind about the `Stat RPC`, I'd be happy to do that!

JaewonHur:
@simone-panico Thanks for all your hard work :)

Give me a few more time til we decide the best direction!

simone-panico:
Hey @JaewonHur, I hope you're doing great. 
Do you have any updates?

JaewonHur:
Hi @simone-panico sorry for delay 🥲
I made a PR for supporting `stat` in `vminitd` (apple/containerization#614), but waiting there for adding test.

Once I merge that PR, you can continue working on `container copy`, and I bet this will be the final!
Sorry again for taking your time.

Let me ping you as soon as I merge that PR!

JaewonHur:
Hi @simone-panico!

I just merged apple/containerization#614, which adds `stat` RPC to vminitd.

It takes absolute path (outside container environment) and returns stat result of the file system object under that path.
So, now you should be able to stat whether the copy source or destination is file or directory.

This shall help making the logics clean.
Thank you for your patience, and let's finish it finally!

simone-panico:
Hey @JaewonHur 
No worries, and thank you for the `Stat RPC`. I've started looking into it but haven't finished yet. I'll get to it as soon as I can 😄

JaewonHur:
@simone-panico Thank you!
You might want to change `copyIn` and `copyOut` methods of `LinuxContianer` in containerization to use `stat` RPC.

simone-panico:
Hey @JaewonHur, sorry this took so long. I had to finish my finals.
I’ve now implemented the `Stat` RPC in [apple/containerization#727](https://github.com/apple/containerization/pull/727) and updated this PR to use it for `copyIn`.
Every copy test is passing now: 70/70. 😄

JaewonHur:
@simone-panico Not at all :) I really appreciate for all your hard work!

Package.swift
```diff
@@ -47,7 +47,7 @@ let package = Package(
         .library(name: "TerminalProgress", targets: ["TerminalProgress"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/containerization.git", exact: Version(stringLiteral: scVersion)),
+        .package(path: "~/projects/container/apple-container/containerization"),
```
JaewonHur: Could you clean up these? and `Package.resolved` should be cleaned up as well.

Once the PR in containerization is merged and released, we'd bump containerization project version in container, and can merge this PR.

JaewonHur:
@simone-panico Could you do `make fmt` and push again.

Once the containerization PR is merged, we should bump that in Package.swift, and able to merge this PR.

simone-panico:
> This one also doesn't seem to be signed correctly. Could you please sign this as well?

Should be correct now, right?

JaewonHur:
Yes. Looks good.

JaewonHur:
Hi! I tagged latest containerization to `0.32.2`.

Here could you update `scVersion` in `Package.swift` to `0.32.2`, then `swift package update containerization`?

The `Package.resolved` file should have only the `containerization` package updated (and the `originHash` at the top).

JaewonHur (approved):

JaewonHur:
@simone-panico Finally merged :) Thank you so much for your hard work and patience!

jglogan:
@simone-panico Definitely, thanks for the patience and the effort you've put in. It's nice to have this feature in place!

PR #1193, opened 2026-02-11
CI observability enhancements.

jglogan:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

- Adds a a `--log-root` option to `swift system start`, propagating the value as `CONTAINER_LOG_ROOT` to services for logging to files instead of the OS log facility. This is not a "production" capability as it neither merges nor rotates logs.
- Currently we don't collect logs on CI builds, and we don't have permission to run the `log` command there. The PR adds `--log-root` to the CI test phase, archives the results, and uploads the archive as an artifact.
- Use FilePath from swift-system for the log root. Foundation URL is a bit of a footgun for filesystem paths, so unless we identify a showstopper, we should incrementally transition to this type everywhere except where we really need network URLs.
- Output the hostname of the CI runner at the start of the test phase so we can identify runner-specific issues where they exist.
- Fix formatting for log messages with multiple metadata items, and fix unstructured messages on instances that weren't found using `grep -r 'log\.' Sources`.
- Adds command reference documentation for `--log-root`.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerLog/ServiceLogger.swift
+++ b/Sources/ContainerLog/ServiceLogger.swift
@@ -19,22 +19,39 @@ import SystemPackage
 
 /// Common logging setup for application services.
 public struct ServiceLogger {
-    /// Create a logger
+    /// Set up the logging system and create a root logger.
+    ///
+    /// - Parameters:
+    ///   - label: A unique identifier for the application.
+    ///   - category: An identifier for the application subsystem.
+    ///   - metadata: Metadata to include for all messsages. A message
+    ///     specific value for a duplicate key overrides these values.
+    ///   - debug: Enable debug logging.
+    ///   - logPath: If supplied, create log files under the named
+    ///     directory. Otherwise, log to the OS log facility.
+    /// - Returns: The root logger.
     public static func bootstrap(
-        category: String,
         label: String = "com.apple.container",
+        category: String,
         metadata: [String: String] = [:],
         debug: Bool,
         logPath: FilePath?
     ) -> Logger {
+        // Select the log handler and bootstrap logging.
         LoggingSystem.bootstrap { label in
             if let logPath {
-                if let handler = try? FileLogHandler(label: label, path: logPath) {
+                if let handler = try? FileLogHandler(
+                    label: label,
+                    category: category,
+                    path: logPath
+                ) {
                     return handler
                 }
             }
             return OSLogHandler(label: label, category: category)
         }
+
+        // Configure log level and metadata.
         var log = Logger(label: label)
         if debug {
             log.logLevel = .debug
@@ -42,6 +59,16 @@ public struct ServiceLogger {
         for (key, value) in metadata {
             log[metadataKey: key] = "\(value)"
         }
+
+        // Log an error if for some reason FileLogHandler init failed.
+        if let logPath, log.handler as? OSLogHandler != nil {
+            log.error(
+                "unable to initialize FileLogHandler, using OSLogHandler",
+                metadata: [
+                    "logPath": "\(logPath)"
+                ])
+        }
+
         return log
     }
 }
--- a/Sources/Helpers/APIServer/APIServer+Start.swift
+++ b/Sources/Helpers/APIServer/APIServer+Start.swift
@@ -142,7 +142,12 @@ extension APIServer {
                     */
                 }
             } catch {
-                log.error("helper failed", metadata: ["name": "\(commandName)", "error": "\(error)"])
+                log.error(
+                    "helper failed",
+                    metadata: [
+                        "name": "\(commandName)",
+                        "error": "\(error)",
+                    ])
                 APIServer.exit(withError: error)
             }
         }
@@ -224,6 +229,7 @@ extension APIServer {
             let svc = HealthCheckHarness(
                 appRoot: appRoot,
                 installRoot: installRoot,
+                logRoot: logRoot,
                 log: log
             )
             routes[XPCRoute.ping] = svc.ping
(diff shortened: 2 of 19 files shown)


Sources/Helpers/APIServer/APIServer+Start.swift
```diff
@@ -39,13 +41,19 @@ extension APIServer {
         @Flag(name: .long, help: "Enable debug logging")
         var debug = false
 
+        @Option(name: .customLong("log-path"), help: "Log to the named file, instead of the macOS log facility")
```
jglogan: Do we need this?

ajemory (approved):

Ronitsabhaya75:
@jglogan  can you have look at #1167  I have added retry logic for api server so that we can atleast 5 tries during install kernel.

I saw CI has same failure. so i added retry logic instead of 1 time trying install kernel added 5 times.

lemme know if this fix can help

jglogan:
@Ronitsabhaya75 Retrying around undiagnosed problems is not the answer, and in this case you can retry a million times and it still won't work.

Ronitsabhaya75:
> @Ronitsabhaya75 Retrying around real problems is not the answer, and in this case you can retry a million times and it still won't work.

let me see what can potential issue for installing-kernel can be. thats where CI has been failing all time.

jglogan:
I am committing a likely fix. The problem has nothing to do with installing the kernel.

Ronitsabhaya75:
```Error: internalError: "failed to get a response from apiserver: internalError: "XPC timeout for request to com.apple.container.apiserver/ping""
make: *** [install-kernel] Error 1
tests failed with status: 2
Removing data directory /Users/runner/actions-runner/_work/_temp/tmp.Moka3ycaZe
Error: Process completed with exit code 2.
```


yeah time out issue (wording mistake)

Ronitsabhaya75:
@jglogan do you think this is good idea adding endpoint that captures the system state like running containers, 
network status, service health at any point. 

then calling it automatically before operation when CI is failing ( like install kernel stage).

this can give us snapshot of before and after for comparision.

Ronitsabhaya75:
Hey @jglogan, I was thinking about this last night after our conversation. You're right that retries won't fix this but I think I might have an idea about what's actually going wrong.

I'm wondering if the install-kernel timeout is actually a race condition. Like, we're trying to install the kernel before the API server is fully ready to accept XPC connections. The service might be "started" but not fully initialized yet.

Instead of retrying or just increasing the timeout, what if we add a proper readiness check? And actually, your new JSON status output makes this really easy to implement:

```make
install-kernel: start-system
	@echo "Waiting for system to be ready..."
	@for i in 1 2 3 4 5; do \
		if $(CONTAINER) system status --output json | jq -e '.apiServerRunning == true' > /dev/null 2>&1; then \
			echo "API server ready"; \
			break; \
		fi; \
		echo "Waiting for API server (attempt $$i/5)..."; \
		sleep 2; \
	done
	$(CONTAINER) kernel install
```
sample output
```make
$ make install-kernel
Waiting for system to be ready...
Waiting for API server (attempt 1/5)...
Waiting for API server (attempt 2/5)...
Waiting for API server (attempt 3/5)...
Waiting for API server (attempt 4/5)...
Waiting for API server (attempt 5/5)...
Warning: API server did not become ready, attempting install anyway...
Installing kernel extension...
Error: Failed to connect to API server (timeout)
```

jglogan:
> I'm wondering if the install-kernel timeout is actually a race condition. Like, we're trying to install the kernel before the API server is fully ready to accept XPC connections. The service might be "started" but not fully initialized yet.

Perhaps. But the API server accepted an XPC connection at `ClientHealthCheck.ping` of SystemStart before trying to make the XPC connection through the `installDefaultKernel` call further down?

Sources/Helpers/RuntimeLinux/RuntimeLinuxHelper+Start.swift
```diff
@@ -42,13 +43,16 @@ extension RuntimeLinuxHelper {
...
         var machServiceLabel: String {
             "\(Self.label).\(uuid)"
         }
 
         func run() async throws {
-            let commandName = Self._commandName
-            let log = RuntimeLinuxHelper.setupLogger(debug: debug, metadata: ["uuid": "\(uuid)"])
+            let commandName = RuntimeLinuxHelper._commandName
+            let logPath = logRoot.map { $0.appending("\(commandName)-\(uuid).log") }
+            let log = ServiceLogger.bootstrap(category: "NetworkVmnetHelper", metadata: ["uuid": "\(uuid)"], debug: debug, logPath: logPath)
```
JaewonHur: Should be `RuntimeLinuxHelper`.

Sources/ContainerLog/ServiceLogger.swift
```diff
@@ -0,0 +1,48 @@
...
+    public static func bootstrap(
+        category: String,
+        label: String = "com.apple.container",
+        metadata: [String: String] = [:],
+        debug: Bool,
+        logPath: FilePath?
+    ) -> Logger {
+        LoggingSystem.bootstrap { label in
+            if let logPath {
+                if let handler = try? FileLogHandler(label: label, path: logPath) {
```
ajemory: Will the user be aware of a fallback to OS logs here in case of failure?

jglogan: Logging isn't set up and this is in a service context, so I don't see a good way to communicate it.

We could try to make note of the condition here and then log a warning above L46, but that would still go into the OS log; it wouldn't be immediately visible to the user.

ajemory: Makes sense. I'm not seeing a good way to go about it either. Not blocking for this PR, but maybe something to think about

ajemory (approved):

PR #1119, opened 2026-01-30
Feat: add container registry list

tico88612:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Supported apple/containerization#502

Fixed #1088

Simple test in here, but without login or logout feature.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

PR #1195, opened 2026-02-11
Feat: add RegistryResource

tico88612:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Follow up: https://github.com/apple/container/pull/1119#discussion_r2747978356

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Registry/RegistryResource.swift
+++ b/Sources/ContainerResource/Registry/RegistryResource.swift
@@ -87,20 +87,20 @@ public struct RegistryResource: ManagedResource {
     ///   - hostname: The registry hostname (also used as the resource ID)
     ///   - username: The username for authentication
     ///   - creationDate: The time the resource was created
-    ///   - modifiedDate: The time the resource was last modified
+    ///   - modificationDate: The time the resource was last modified
     ///   - labels: Optional key-value labels for metadata (default: empty dictionary)
     public init(
         hostname: String,
         username: String,
         creationDate: Date,
-        modifiedDate: Date,
+        modificationDate: Date,
         labels: [String: String] = [:]
     ) {
         self.id = hostname
         self.name = hostname
         self.username = username
         self.creationDate = creationDate
-        self.modificationDate = modifiedDate
+        self.modificationDate = modificationDate
         self.labels = labels
     }
 }
@@ -114,7 +114,7 @@ extension RegistryResource {
             hostname: registryInfo.hostname,
             username: registryInfo.username,
             creationDate: registryInfo.createdDate,
-            modifiedDate: registryInfo.modifiedDate
+            modificationDate: registryInfo.modifiedDate
         )
     }
 }
--- a/Tests/ContainerResourceTests/RegistryResourceTests.swift
+++ b/Tests/ContainerResourceTests/RegistryResourceTests.swift
@@ -92,7 +92,7 @@ struct RegistryResourceTests {
         #expect(jsonString.contains("\"name\""), "JSON should contain name field")
         #expect(jsonString.contains("\"username\""), "JSON should contain username field")
         #expect(jsonString.contains("\"creationDate\""), "JSON should contain creationDate field")
-        #expect(jsonString.contains("\"modifiedDate\""), "JSON should contain modifiedDate field")
+        #expect(jsonString.contains("\"modificationDate\""), "JSON should contain modificationDate field")
         #expect(jsonString.contains(hostname), "JSON should contain the hostname")
         #expect(jsonString.contains(username), "JSON should contain the username")
     }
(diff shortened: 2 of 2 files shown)


saehejkang:
Once this is merged, I'll take a stab at the refactoring/creation of the registry client.

Tests/ContainerResourceTests/RegistryResourceTests.swift
```diff
@@ -0,0 +1,200 @@
...
+
+import Foundation
+import Testing
+
+@testable import ContainerResource
+@testable import ContainerizationOS
+
+struct RegistryResourceTests {
+
+    // Mock RegistryInfo for testing
```
jglogan: nit: you could go with `createRegistryInfo` here and drop `mock` everywhere else as it's not really a mock you're creating, it is a real `RegistryInfo`.

Sources/ContainerResource/Registry/RegistryResource.swift
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerizationOS
+import Foundation
+
+public struct RegistryResource: ManagedResource {
+    public var id: String
```
jglogan: Can these be `let` or `var { get { } }`?

Resources that are immutable should be easier to reason about. AFAIK there's no need to modify a resource after it's created.

Sources/ContainerResource/Registry/RegistryResource.swift
```diff
@@ -0,0 +1,63 @@
...
+    public let username: String
+
+    public var creationDate: Date
+
+    public var modifiedDate: Date
+
+    public var labels: [String: String]
+
+    public static func nameValid(_ name: String) -> Bool {
+        let pattern = #"^([a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?\.)*[a-zA-Z0-9]([a-zA-Z0-9\-]*[a-zA-Z0-9])?(:[0-9]+)?$"#
```
jglogan: How did you arrive at this?

I'm wondering about impact of the difference between this def and that of the OCI library in container at:  https://github.com/apple/containerization/blob/main/Sources/ContainerizationOCI/Reference.swift#L83.

Should they be the same? Which is the better definition? Is there one still better?

Sources/ContainerResource/Registry/RegistryResource.swift
```diff
@@ -0,0 +1,63 @@
...
+public struct RegistryResource: ManagedResource {
+    public var id: String
+
+    public var name: String
+
+    public let username: String
+
+    public var creationDate: Date
+
+    public var modifiedDate: Date
```
jglogan: Let's use `modificationDate` for consistency. At some later point when we're breaking CZ APIs for some other reason we could change `RegistryInfo`.

tico88612: Fixed.

Sources/ContainerResource/Registry/RegistryResource.swift
tico88612: Fixed.

Tests/ContainerResourceTests/RegistryResourceTests.swift
tico88612: Fixed.

Sources/ContainerResource/Registry/RegistryResource.swift
tico88612: This only validates the domain and IP, but doesn't consider IPv6. My thought is to keep it as consistent as possible with the containerization results.

Sources/ContainerResource/Registry/RegistryResource.swift
```diff
@@ -0,0 +1,73 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerizationOS
+import ContainerizationOCI
+import Foundation
+
+public struct RegistryResource: ManagedResource {
```
saehejkang, line 25: nit: maybe wouldn't hurt to add comments about each field (to follow the docc format)

tico88612: Done.

jglogan (approved):

tico88612:
I will update the test

tico88612:
@jglogan Fixed.

PR #1201, opened 2026-02-12
[containers]: add system-start flag for auto-start of containers

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [x] Breaking change
- [x] Documentation update

## Motivation and Context

Relates to #158

Adds a simple flag that can be set on `container create` and `container run` to auto-start any container that is stopped after `container system stop`.

## Testing

1. Create a container or run a container with the `--system-start` flag set
2. Run `container ls` to see that the flag is `enabled`
3. Run `container system stop`
4. Run `container system start` and the logs should output the containers that were started up again
5. Run `container ls` to see that the container is `running`

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/System/SystemStart.swift
+++ b/Sources/ContainerCommands/System/SystemStart.swift
@@ -100,10 +100,6 @@ extension Application {
 
             try ServiceManager.register(plistPath: plistURL.path)
 
-            // Start all the containers that have system-start enabled
-            log.info("starting containers", metadata: ["startTimeoutSeconds": "\(Self.startTimeoutSeconds)"])
-            try await startContainers()
-
             // Now ping our friendly daemon. Fail if we don't get a response.
             do {
                 print("Verifying apiserver is running...")
@@ -119,10 +115,13 @@ extension Application {
                 try? await installInitialFilesystem()
             }
 
-            guard await !kernelExists() else {
-                return
+            if await !kernelExists() {
+                try? await installDefaultKernel()
             }
-            try await installDefaultKernel()
+
+            // Start all the containers that have system-start enabled
+            log.info("starting containers", metadata: ["startTimeoutSeconds": "\(Self.startTimeoutSeconds)"])
+            try await startContainers()
         }
 
         /// Starts all containers that are configured to automatically start on system start.
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -303,8 +303,7 @@ public actor ContainersService {
             // We've already bootstrapped this container. Ideally we should be able to
             // return some sort of error code from the sandbox svc to check here, but this
             // is also a very simple check and faster than doing an rpc to get the same result.
-            // However, if it marked for system-start, continue with bootstrap.
-            if state.client != nil && state.snapshot.createOptions?.systemStart != true {
+            if state.client != nil {
                 return
             }
 


Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -230,6 +230,12 @@ public struct Flags {
 
         @Option(name: .long, help: "Set the runtime handler for the container (default: container-runtime-linux)")
         public var runtime: String?
+
+        @Flag(
+            name: .customLong("systemstart"),
+            help: "Automatically start the container when the container system starts"
+        )
+        public var systemStart: Bool = false
```
saehejkang, line 238: I neglected to use `autoStart` because `containers` are automatically spun up with `run` and `create`. It will cause confusion, as that is a totally different process from what this flag should inherently be doing.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -300,7 +303,8 @@ public actor ContainersService {
             // We've already bootstrapped this container. Ideally we should be able to
             // return some sort of error code from the sandbox svc to check here, but this
             // is also a very simple check and faster than doing an rpc to get the same result.
-            if state.client != nil {
+            // However, if it marked for system-start, continue with bootstrap.
+            if state.client != nil && state.snapshot.createOptions?.systemStart != true {
```
JaewonHur: Could you explain rationale behind this?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -83,9 +83,11 @@ public actor ContainersService {
             do {
                 let bundle = ContainerResource.Bundle(path: dir)
                 let config = try bundle.configuration
+                let options = try bundle.options
```
JaewonHur: It throws exception on legacy container, which does't have `CreateOption.json` file.

Sources/ContainerCommands/System/SystemStart.swift
```diff
@@ -89,6 +100,10 @@ extension Application {
 
             try ServiceManager.register(plistPath: plistURL.path)
 
+            // Start all the containers that have system-start enabled
+            log.info("starting containers", metadata: ["startTimeoutSeconds": "\(Self.startTimeoutSeconds)"])
+            try await startContainers()
```
JaewonHur: This might be invoked before the API server finishes bootstrapping.

Sources/ContainerCommands/System/SystemStart.swift
```diff
@@ -110,6 +125,49 @@ extension Application {
             try await installDefaultKernel()
         }
 
+        /// Starts all containers that are configured to automatically start on system start.
+        ///
+        /// - Throws: `AggregateError` containing all errors encountered during container startup.
+        private func startContainers() async throws {
+            let client = ContainerClient()
+            let containers = try await client.list()
+            let systemStartContainers = containers.filter { $0.createOptions?.systemStart == true && $0.status != .running }
```
JaewonHur, line 133: Should be `$0.status == .stopped`?

Sources/ContainerCommands/Container/ContainerList.swift
```diff
@@ -50,7 +50,7 @@ extension Application {
         }
 
         private func createHeader() -> [[String]] {
-            [["ID", "IMAGE", "OS", "ARCH", "STATE", "ADDR", "CPUS", "MEMORY", "STARTED"]]
+            [["ID", "IMAGE", "OS", "ARCH", "STATE", "ADDR", "CPUS", "MEMORY", "STARTED", "SYSTEM-START"]]
```
JaewonHur, line 53: I'm not sure it's the best to print "SYSTEM-START" here..

Sources/ContainerCommands/System/SystemStart.swift
```diff
@@ -110,6 +125,49 @@ extension Application {
             try await installDefaultKernel()
         }
 
+        /// Starts all containers that are configured to automatically start on system start.
+        ///
+        /// - Throws: `AggregateError` containing all errors encountered during container startup.
+        private func startContainers() async throws {
```
JaewonHur, line 130: We might be able to move starting containers to the API server, near `loadAtBoot`? Then, we can avoid changing `ContainerSnapshot`.

Sources/ContainerCommands/System/SystemStart.swift
saehejkang: either or in this case

Sources/ContainerCommands/Container/ContainerList.swift
saehejkang: I was mulling on the idea of adding this to the print output. I feel users should have the ability to know if a container will be started automatically on system startup. It might be worthwhile adding a `--verbose` option (like image list), where this field will be added to the output (default output will not included it).

Sources/ContainerCommands/System/SystemStart.swift
saehejkang: I can take a look at this but `ContainerSnapshot` was changed just so the `SYSTEM-START` field could be outputted.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
saehejkang: Whoops. That was a check I added just in case to test something and I forgot to remove it.

Sources/ContainerCommands/System/SystemStart.swift
JaewonHur: Oh got it. We need to change `ContainerSnapshot` anyway.

Sources/ContainerCommands/System/SystemStart.swift
saehejkang: Hmmm. Interestingly enough, I tested it by waiting until the API server was bootstrapped/kernel install and the function never got invoked. The logic will need to be updated a tiny bit so this function is invoked correctly.

saehejkang:
Because these changes need to sort of "simulate" a system stop/start, I am thinking of making the following changes to the `Makefile` for tests. Thought about this for a while and tried different things, but below is the only solution that I makes sense.

```Makefile
exit_code=0; \
# Run setup tests (create containers with --system-start)
echo "Running setup tests..." ; \
$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestCLICreateCommand || exit_code=1 ; \

# Stop the system
echo "Stopping container system..." ; \
bin/container system stop && sleep 3 && scripts/ensure-container-stopped.sh ; \

# Start the system again
echo "Starting container system..." ; \
bin/container system start $(SYSTEM_START_OPTS) ; \

# Run verification tests (check containers are running)
echo "Running verification tests..." ; \
$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter TestSystemStart || exit_code=1 ; \

# Rest of the tests below (omitted for readability)
```

Sources/ContainerResource/Container/ContainerCreateOptions.swift
```diff
@@ -16,11 +16,13 @@
 
 public struct ContainerCreateOptions: Codable, Sendable {
```
JaewonHur, line 17: @jglogan @dcantah How do you think merging this into `ContainerConfiguration.label`?

For now, there's no way to inform the users of this create options.
`label` might be a good choice to expose it without drastic changing?

docs/command-reference.md
```diff
@@ -68,6 +68,7 @@ container run [<options>] <image> [<arguments> ...]
 *   `-v, --volume <volume>`: Bind mount a volume into the container
 *   `--virtualization`: Expose virtualization capabilities to the container (requires host and guest support)
 *   `--runtime`: Set the runtime handler for the container (default: container-runtime-linux)
+*   `--systemstart`: Automatically start the container when the container system starts
```
JaewonHur, line 71: `--system-start` would be better.

Sources/ContainerCommands/System/SystemStart.swift
JaewonHur: NVM this comment, I was testing AI based PR review yesterday and it corrupted my brain :)

saehejkang:
@JaewonHur Take a look at the comment [above](https://github.com/apple/container/pull/1201#issuecomment-3895068437) and let me know if you see a better way to handle this. I would love to get your input!

docs/command-reference.md
saehejkang: Agreed. Will make that change before this gets merged.

JaewonHur:
Sorry for the delay! We are still thinking about how `ContainerConfiguration.labels` should be used versus `ContainerCreateOptions`.

Regarding the test, it feels somewhat unusual to me if a single test case is performed by multiple subsequent commands.

How about adding a new (self-contained) test case under `Tests/CLITests/Subcommands/System`?
There, we can start API server, create system-start container, stop API server, and restart again.

saehejkang:
> Sorry for the delay! We are still thinking about how ContainerConfiguration.labels should be used versus ContainerCreateOptions.

All good and no worries.

> How about adding a new (self-contained) test case under Tests/CLITests/Subcommands/System?
There, we can start API server, create system-start container, stop API server, and restart again.

That was my initial thought, but I think I was having issues with executing `container system start` and `container system stop` there. I don't think the test suites are configured to properly start/stop the system, as that seems to be handled by the install/uninstall scripts (used in the makefile when running integration tests). 

I was thinking that since this is the first "feature" that relies on the system, we build out a single new test suite. This test suite will be ran after an intentional system stop/start (using the scripts again in the makefile). Maybe I am not being clear enough, but it is also hard to explain my thought process in a text box lol.

JaewonHur:
I agree with your last sentence :)
A new test suite for "features" that rely on system operations, which can run after all integration tests are done (and system is stopped as usual).

saehejkang:
> A new test suite for "features" that rely on system operations, which can run after all integration tests are done (and system is stopped as usual).

💯 

-----

I will wait to make any more updates on this PR until a direction is chosen on the either using labels or create options.

JaewonHur:
Hi @saehejkang 
Please take a look at #1258.

I thought it'd be better to generally support `restart` functionality first, and extend that to `system-start` as you made here.
So, I created an initial PR for describing the implementation plan for `restart` feature.

It would be really good if you can contribute to that first :)
Please leave any comment (in that thread) if you feel weird or somewhat unclear.

PR #1203, opened 2026-02-12
Add init methods for all the members of the Flags type

adityaramani:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Authors of CLI plugins for container will be able to reuse the container flags defined in the CLI package, instead of having to duplicate them in their project

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -22,12 +22,38 @@ public struct Flags {
     public struct Logging: ParsableArguments {
         public init() {}
 
+        public init(debug: Bool) {
+            self.debug = debug
+        }
+
         @Flag(name: .long, help: "Enable debug output [environment: CONTAINER_DEBUG]")
         public var debug = false
     }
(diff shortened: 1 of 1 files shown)


Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -127,6 +165,54 @@ public struct Flags {
     public struct Management: ParsableArguments {
         public init() {}
 
+        public init(
+            arch: String = Arch.hostArchitecture().rawValue,
```
katiewasnothere: Just a general swift question I think? Since everything in the second init here has a default value, which init function gets called if you provide no parameters?

adityaramani: I think the init with the no parameters gets called, atleast thats what I read online. To make it simpler, I removed the default values from the init methods

PR #1204, opened 2026-02-12
Check unused imports

JaewonHur:
Add `scripts/check-unused-imports.sh` which checks the unused imports in modified swift source files.
Good to be used before making a pull request after modifying or creating lots of swift source files.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Developers can check and refine unused imports using this script, before making a PR.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,2 @@
+analyzer_rules:
+  - unused_import
--- a/scripts/check-unused-imports.sh
+++ b/scripts/check-unused-imports.sh
@@ -26,11 +26,8 @@ BUILD_LOG=${PROJECT_ROOT}/.build/build.log
 LINT_LOG=${PROJECT_ROOT}/.build/swiftlint.log
 
 echo 'Building `container` with verbose flag'
-if [ ! -f ${BUILD_LOG} ];
-then
-    make -C ${PROJECT_ROOT} clean
-    make -C ${PROJECT_ROOT} BUILD_OPTIONS="-v &>${BUILD_LOG}"
-fi
+make -C ${PROJECT_ROOT} clean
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -55,13 +55,13 @@ all: init-block
 build:
 	@echo Building container binaries...
 	@$(SWIFT) --version
-	@$(SWIFT) build -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION)
+	@$(SWIFT) build -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) $(BUILD_OPTIONS)
```
JaewonHur, line 57: Not sure it's a good approach.

saehejkang:
Thoughts on updating the `make fmt` command to run this script? Maybe even the possibility of it making the formatting changes, ie. actually remove the unused imports?

JaewonHur:
Hi @saehejkang I'm not sure it might be merged to `make fmt` :)
It requires rebuilding the project with verbose flag, which is quite slow.

Makefile
jglogan: For the pre-commit hook I added a separate swift-format config file for format-only checking. Could we use `swift format` in lint mode to do this, and then perhaps it's just a matter of adjusting the existing lint rule file?

```Makefile
swift-fmt-check:
	@echo Applying the standard code formatting...
	@$(SWIFT) format lint --recursive --strict --configuration .swift-format-nolint $(SWIFT_SRC)
```

JaewonHur: It seems `swift-format` does not have a rule for detecting unused imports.

PR #1184, opened 2026-02-09
Move --debug flag after `apiserver start`

JaewonHur:
Fix starting api server with debug flag (#1183)

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
API server currently doesn't start with `container system start --debug` as it sets the command `/usr/local/bin/container-apiserver --debug start` which returns unknown options error.

The command should be `/usr/local/bin/container-apiserver start --debug`.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

PR #1205, opened 2026-02-13
Fix persistent CI failures.

jglogan:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
- CI build is failing because runners don't have an /etc/resolver directory, causing the directory monitor to fail. This occurs while the install-kernel make target is running, so it appears that kernel download is failing when it's just that the API server is dying. Created #1207 for the issue.
- Fixing the initial scan for the directory just moves the problem to the filesystem watch loop, likely because we're not testing the result of `open()` for errors.
- Right now the priority is getting CI running and PRs merged, so just commenting out the realhost DNS server setup.
- Also seeing that under some conditions it can take quite a while for launchd to start the helper for the default network (8 seconds or more). With the 10 second health check timeout after API server registration, this means that some CI runs can exhibit this failure mode. Added a `--timeout` option to SystemStart and set a 60 second timeout for install-kernel and integration Makefile targets.
- Fixed a bug where `--debug` was being placed in the wrong location in the api server startup args.
- Disabled all network CLI tests due to container bootstrap errors when trying to run the container immediately after creating the network. The slow network helper launch could be the reason behind the failures that drove us to serialize these tests. Filed #1206 for this issue.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Helpers/APIServer/APIServer+Start.swift
+++ b/Sources/Helpers/APIServer/APIServer+Start.swift
@@ -113,7 +113,7 @@ extension APIServer {
                     group.addTask {
                         let localhostResolver = LocalhostDNSHandler(log: log)
                         try localhostResolver.monitorResolvers()
-
+                    
                         let nxDomainResolver = NxDomainResolver()
                         let compositeResolver = CompositeResolver(handlers: [localhostResolver, nxDomainResolver])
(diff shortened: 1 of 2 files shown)


JaewonHur:
This resolves #1184.

Issue #1189, opened 2026-02-10
[Bug]: The SSH flag for running containers doesn't update when the variable changes.


vibbix:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. Use a setup that requires some kind of forwarding agent (like 1Password).
2. Have a SSH Config like this:
```
Host *
	IdentityAgent "~/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock"
```
3. In terminal:
```
❯❯ ~  12:58 echo $SSH_AUTH_SOCK
/private/tmp/com.apple.launchd.sOMZW8wQ8o/Listeners
❯❯ ~  12:59 export SSH_AUTH_SOCK="$HOME/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock"
❯❯ fix_docker_build git:(main) 13:00 container run -it --rm --ssh --uid 0 alpine:latest sh
/ # apk add openssh-client
(1/6) Installing openssh-keygen (10.2_p1-r0)
(2/6) Installing ncurses-terminfo-base (6.5_p20251123-r0)
(3/6) Installing libncursesw (6.5_p20251123-r0)
(4/6) Installing libedit (20251016.3.1-r0)
(5/6) Installing openssh-client-common (10.2_p1-r0)
(6/6) Installing openssh-client-default (10.2_p1-r0)
Executing busybox-1.37.0-r30.trigger
OK: 12.8 MiB in 22 packages
/ # ssh-add -l
The agent has no identities.
/ # ^C

/ # exit
❯❯ fix_docker_build git:(main)  13:00 echo $SSH_AUTH_SOCK
/Users/vibbix/Library/Group Containers/2BUA8C4S2C.com.1password/t/agent.sock
❯❯ fix_docker_build git:(main) 13:00 container run -it --rm --ssh --uid 0 alpine:latest sh
❯❯ fix_docker_build git:(main)  13:01 container run -it --rm --volume "${SSH_AUTH_SOCK}:/run/host-services/ssh-auth.sock" --env "SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock" --uid 0 alpine:latest sh

/ # apk add openssh-client
(1/6) Installing openssh-keygen (10.2_p1-r0)
(2/6) Installing ncurses-terminfo-base (6.5_p20251123-r0)
[...]

PR #1214, opened 2026-02-14
fix: --ssh flag for running containers (issue #1189)

chavezMac:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]
-Issue pointed out by @vibbix, when attempting to use the  `--ssh` flag in the `container run ...` command the value of `SSH_AUTH_SOCK` was not persisting and would result in a "The agent has no identities." as noted in the original issue. 

Following the [how-to documentation](https://github.com/apple/container/blob/cf9b335aa1236f06c7bb30901c44024cababc2cc/docs/how-to.md#mount-your-host-ssh-authentication-socket-in-your-container), future users may run into this issue as well.

**Root cause:** The runtime helper is started by launchd (per-container plist), so it sees launchd’s environment, not the shell that ran `container run`. The client’s `SSH_AUTH_SOCK` (e.g. from 1Password) was never passed into the container config, so the wrong socket was being used or no socket was used.

**Fix:** The client now captures the caller’s `SSH_AUTH_SOCK` when `--ssh` is set and stores it in the container config. The sandbox service resolves the host socket with precedence: config → runtime env → launchctl, and mounts that path into the container so the correct agent is used across stop/logout/login/restart.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Container/ContainerConfiguration.swift
+++ b/Sources/ContainerResource/Container/ContainerConfiguration.swift
@@ -49,8 +49,6 @@ public struct ContainerConfiguration: Sendable, Codable {
     public var virtualization: Bool = false
     /// Enable SSH agent socket forwarding from host to container.
     public var ssh: Bool = false
-    /// Optiional preferred SSH agent socket path captured from client-side environment
-    public var sshAuthSocketPath: String? = nil
     /// Whether to mount the rootfs as read-only.
     public var readOnly: Bool = false
     /// Whether to use a minimal init process inside the container.
@@ -73,7 +71,6 @@ public struct ContainerConfiguration: Sendable, Codable {
         case runtimeHandler
         case virtualization
         case ssh
-        case sshAuthSocketPath
         case readOnly
         case useInit
     }
@@ -105,7 +102,6 @@ public struct ContainerConfiguration: Sendable, Codable {
         runtimeHandler = try container.decodeIfPresent(String.self, forKey: .runtimeHandler) ?? "container-runtime-linux"
         virtualization = try container.decodeIfPresent(Bool.self, forKey: .virtualization) ?? false
         ssh = try container.decodeIfPresent(Bool.self, forKey: .ssh) ?? false
-        sshAuthSocketPath = try container.decodeIfPresent(String.self, forKey: .sshAuthSocketPath)
         readOnly = try container.decodeIfPresent(Bool.self, forKey: .readOnly) ?? false
         useInit = try container.decodeIfPresent(Bool.self, forKey: .useInit) ?? false
     }
--- a/Sources/ContainerResource/Container/ContainerConfiguration.swift
+++ b/Sources/ContainerResource/Container/ContainerConfiguration.swift
@@ -49,6 +49,8 @@ public struct ContainerConfiguration: Sendable, Codable {
     public var virtualization: Bool = false
     /// Enable SSH agent socket forwarding from host to container.
     public var ssh: Bool = false
+    /// Optiional preferred SSH agent socket path captured from client-side environment
+    public var sshAuthSocketPath: String? = nil
     /// Whether to mount the rootfs as read-only.
     public var readOnly: Bool = false
     /// Whether to use a minimal init process inside the container.
@@ -71,6 +73,7 @@ public struct ContainerConfiguration: Sendable, Codable {
         case runtimeHandler
         case virtualization
         case ssh
+        case sshAuthSocketPath
         case readOnly
         case useInit
     }
@@ -102,6 +105,7 @@ public struct ContainerConfiguration: Sendable, Codable {
         runtimeHandler = try container.decodeIfPresent(String.self, forKey: .runtimeHandler) ?? "container-runtime-linux"
         virtualization = try container.decodeIfPresent(Bool.self, forKey: .virtualization) ?? false
         ssh = try container.decodeIfPresent(Bool.self, forKey: .ssh) ?? false
+        sshAuthSocketPath = try container.decodeIfPresent(String.self, forKey: .sshAuthSocketPath)
         readOnly = try container.decodeIfPresent(Bool.self, forKey: .readOnly) ?? false
         useInit = try container.decodeIfPresent(Bool.self, forKey: .useInit) ?? false
     }
--- a/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
+++ b/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
@@ -113,7 +113,8 @@ public struct ContainerClient: Sendable {
     }
 
     /// Bootstrap the container's init process.
-    public func bootstrap(id: String, stdio: [FileHandle?]) async throws -> ClientProcess {
+    /// - Parameter sshAuthSocketPath: Optional path to the current shell's SSH agent socket, supplied at bootstrap time when SSH forwarding is enabled.
+    public func bootstrap(id: String, stdio: [FileHandle?], sshAuthSocketPath: String? = nil) async throws -> ClientProcess {
         let request = XPCMessage(route: .containerBootstrap)
 
         for (i, h) in stdio.enumerated() {
@@ -134,6 +135,9 @@ public struct ContainerClient: Sendable {
 
         do {
             request.set(key: .id, value: id)
+            if let sshAuthSocketPath {
+                request.set(key: .sshAuthSocketPath, value: sshAuthSocketPath)
+            }
             try await xpcClient.send(request)
             return ClientProcessImpl(containerId: id, xpcClient: xpcClient)
         } catch {
--- a/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
+++ b/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
@@ -113,8 +113,8 @@ public struct ContainerClient: Sendable {
     }
 
     /// Bootstrap the container's init process.
-    /// - Parameter sshAuthSocketPath: Optional path to the current shell's SSH agent socket, supplied at bootstrap time when SSH forwarding is enabled.
-    public func bootstrap(id: String, stdio: [FileHandle?], sshAuthSocketPath: String? = nil) async throws -> ClientProcess {
+    /// - Parameter dynamicEnv: Optional start-time environment overrides passed through bootstrap.
+    public func bootstrap(id: String, stdio: [FileHandle?], dynamicEnv: [String: String] = [:]) async throws -> ClientProcess {
         let request = XPCMessage(route: .containerBootstrap)
 
         for (i, h) in stdio.enumerated() {
@@ -135,8 +135,9 @@ public struct ContainerClient: Sendable {
 
         do {
             request.set(key: .id, value: id)
-            if let sshAuthSocketPath {
-                request.set(key: .sshAuthSocketPath, value: sshAuthSocketPath)
+            if !dynamicEnv.isEmpty {
+                let encodedDynamicEnv = try JSONEncoder().encode(dynamicEnv)
+                request.set(key: .dynamicEnv, value: encodedDynamicEnv)
             }
             try await xpcClient.send(request)
             return ClientProcessImpl(containerId: id, xpcClient: xpcClient)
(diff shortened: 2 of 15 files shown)


Sources/ContainerResource/Container/ContainerConfiguration.swift
```diff
@@ -49,6 +49,8 @@ public struct ContainerConfiguration: Sendable, Codable {
     public var virtualization: Bool = false
     /// Enable SSH agent socket forwarding from host to container.
     public var ssh: Bool = false
+    /// Optiional preferred SSH agent socket path captured from client-side environment
+    public var sshAuthSocketPath: String? = nil
```
JaewonHur: Sorry for the confusion. I found this much more complex.

If we simply put `sshAuthSocketPath` in `ContainerConfiguration`, then it would not work if the ssh auth socket path changes while container is stopped.

We need a way to convey current shell's `$SSH_AUTH_SOCK` to the API server on container bootstrap, but there seems no way today.

Let me discuss it further! Thanks!

jglogan: Does this variable do anything that just `-v ${sshAuthSocketPath}:/run/host-services/ssh-auth.sock -e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock` doesn't already do?

jglogan (changes requested):
@chavezMac @JaewonHur The analysis makes sense but not the fix. Storing it in the container configuration is no different than setting the environment in the configuration.

I think what you have is most of the way there, but what if we got rid of the the `sshAuthSocketPath` configuration variable and the existing grab of the environment variable in the sandbox server, and instead do something like:

- CLI - no change to `--ssh` variable
- Client - update bootstrap to pass the current value of SSH_AUTH_SOCKET to the API server using an `.sshAuthSocketPath` XPC key with value type `String?` (for API compatibility with older clients, we're starting to consider this)
- API server - extract `sshAuthSocketPath` from the XPC request if present and pass it through to the bootstrap call to the sandbox
- sandbox - extract `sshAuthSocketPath` from the XPC request, and if present and `--ssh` is true, set the environment variable (replacing the existing env lookup code which uses the launch env)

jglogan:
Rather than special casing `sshAuthSocketPath` it might be generally useful to be able to pass dynamic (start-time) environment overrides from the user. This isn't something that should get surfaced in the CLI but could be useful for these specific cases where the container really needs something from the environment.

The alternative approach would be for the user to be able to specify a path statically that refers to a symlink they create.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -378,7 +378,8 @@ public actor ContainersService {
     }
 
     /// Bootstrap the init process of the container.
-    public func bootstrap(id: String, stdio: [FileHandle?]) async throws {
+    /// - Parameter sshAuthSocketPath: Optional path to the current shell's SSH agent socket, supplied by the client at bootstrap time when SSH forwarding is enabled.
+    public func bootstrap(id: String, stdio: [FileHandle?], sshAuthSocketPath: String? = nil) async throws {
```
jglogan: @chavezMac I had a thought on this to future-proof the API.

What do you think about changing it to:

```swift
public func bootstrap(id: String, stdio: [FileHandle?], dynamicEnv: [String:String] = [:]) async throws {...}
```

We'd not expose this up to the CLI (`container start`) as generally, the env variables are baked in at create time. With this API we can address any other cases like SSH_AUTH_SOCK without breaking changes.

Do you think it'd work cleanly with what you have?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -54,13 +54,46 @@ public actor SandboxService {
...
+        case bootstrap = "bootstrap"
+    }
+
+    private static func isUnixSocket(path: String) -> Bool {
+        (try? File.info(path).isSocket) ?? false
+    }
+
+    /// Resolves the host path for SSH agent socket forwarding. Uses only the path passed by the
+    /// client at bootstrap time; the sandbox does not read SSH_AUTH_SOCK from its own environment
+    /// or launchctl (the client is responsible for passing the current value).
```
JaewonHur: Why are we not using the default one?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
chavezMac: I do like the idea of a dynamicEnv API as a future-proof idea to support other bootstrap‑time env overrides without breaking changes. If we see more use cases beyond SSH, I’d be happy to prototype that in a separate PR.

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
chavezMac: I believe the initial bug was stemming from the use of the default one. We only use the path the client passes at bootstrap so forwarding matches the shell that ran the container run/start command.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
jglogan: The problem with not doing it now is that it guarantees there will be a breaking change.

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
JaewonHur: I believe it still needs to fallback to default one if there is not `SSH_AUTH_SOCK` provided from command line. We would not want to surprise the users relying on it.

jglogan: If no SSH_AUTH_SOCK is defined at the time the container is started, there is no fallback and there should be no ssh socket forwarding.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
chavezMac: I apologize for the delay in this request, I have been busy with other work. I plan on getting an updated PR for this by end of day, Friday!

chavezMac: Updated PR! @jglogan

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
```diff
@@ -134,6 +135,10 @@ public struct ContainerClient: Sendable {
 
         do {
             request.set(key: .id, value: id)
+            if !dynamicEnv.isEmpty {
```
JaewonHur, line 138: It might be better just keep it simple: encode `dynamicEnv` always (assuming it's not that performance heavy operation).

Sources/Services/ContainerAPIService/Server/Containers/ContainersHarness.swift
```diff
@@ -55,7 +55,13 @@ public struct ContainersHarness: Sendable {
             )
         }
         let stdio = message.stdio()
-        try await service.bootstrap(id: id, stdio: stdio)
+        let dynamicEnv: [String: String] =
```
JaewonHur, line 58: Here the same?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -398,7 +398,8 @@ public actor ContainersService {
     }
 
     /// Bootstrap the init process of the container.
-    public func bootstrap(id: String, stdio: [FileHandle?]) async throws {
+    /// - Parameter dynamicEnv: Optional start-time environment overrides passed from the client.
+    public func bootstrap(id: String, stdio: [FileHandle?], dynamicEnv: [String: String] = [:]) async throws {
```
JaewonHur, line 402: Don't need default value for `dynamicEnv`.

Sources/Services/ContainerSandboxService/Client/SandboxClient.swift
```diff
@@ -77,7 +77,8 @@ public struct SandboxClient: Sendable {
 
 // Runtime Methods
 extension SandboxClient {
-    public func bootstrap(stdio: [FileHandle?], allocatedAttachments: [AllocatedAttachment]) async throws {
+    /// - Parameter dynamicEnv: Optional start-time environment overrides passed from the API service.
+    public func bootstrap(stdio: [FileHandle?], allocatedAttachments: [AllocatedAttachment], dynamicEnv: [String: String] = [:]) async throws {
```
JaewonHur, line 81: Don't need default value.

Sources/Services/ContainerSandboxService/Client/SandboxClient.swift
```diff
@@ -96,6 +97,11 @@ extension SandboxClient {
             }
         }
 
+        if !dynamicEnv.isEmpty {
```
JaewonHur, line 100: Might be better to keep it simple?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -75,13 +75,78 @@ public actor SandboxService {
...
+    private static func resolveSSHAuthSocketHostPath(config: ContainerConfiguration, bootstrapOverridePath: String? = nil) -> (path: String, source: SSHAuthSocketSource)? {
```
JaewonHur, line 90: Why do we need to return `source` here? I found this only being used in L228?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -916,17 +1023,21 @@ public actor SandboxService {
...
-        config: ContainerConfiguration
+        config: ContainerConfiguration,
+        bootstrapOverridePath: String? = nil,
+        dynamicEnv: [String: String] = [:]
     ) throws {
         let process = config.initProcess
 
         czConfig.process.arguments = [process.executable] + process.arguments
-        czConfig.process.environmentVariables = process.environment
+        czConfig.process.environmentVariables = Self.mergedEnvironmentVariables(
```
JaewonHur, line 1033: I think `dynamicEnv` might not to be merged here for container's env variables?
I guess the reason we pass `dynamicEnv` from CLI side is to guide the server-side logic from the user-predicted shell environment.
If a user wants to feed env variables to the container they explicitly specify those using `--env` flags.

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -144,6 +209,36 @@ public actor SandboxService {
...
             var config = try bundle.configuration
+            // Extract dynamic env overrides from the XPC request; when SSH forwarding is enabled,
+            // SSH_AUTH_SOCK from this map provides the host socket path to mount.
+            let dynamicEnv: [String: String] =
+                if let data = message.dataNoCopy(key: SandboxKeys.dynamicEnv.rawValue) {
+                    try JSONDecoder().decode([String: String].self, from: data)
+                } else {
+                    [:]
+                }
+            let bootstrapSshAuthPath = dynamicEnv[Self.sshAuthSocketEnvVar]
```
JaewonHur, line 220: It's a bit complicated both `bootstrapSshAuthPath` and `dynamicEnv` are flowing around. Could we remain only `dynamicEnv` and use it in `Self.resolveSSHAuthSocketHostPath(config: config, dynamicEnv: dynamicEnv)` to figure out the `SSH_AUTH_SOCKET` path?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
JaewonHur, line 90: `resolveSSHAuthSocketHostPath` and `sshAuthSocketHostURL` are seemingly for same purpose?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
JaewonHur: Nit comment, `hostEnv` might be the better name for `dynamicEnv`.

jglogan:
@chavezMac please see #1420

PR #1215, opened 2026-02-14
[registry]: add RegistryKeychainClient to the main client library

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Introduces the `RegistryKeychainClient` to the main API client library. The client does not talk to the server and is more so an extraction of keychain/registry logic from the commands.

## Depends On

Waiting for #1195 to be merged, before marking this ready for review.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Registry/RegistryLogin.swift
+++ b/Sources/ContainerCommands/Registry/RegistryLogin.swift
@@ -18,6 +18,7 @@ import ArgumentParser
 import ContainerAPIClient
 import ContainerPersistence
 import ContainerPlugin
+import ContainerResource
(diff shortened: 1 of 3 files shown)


Sources/ContainerCommands/Registry/RegistryLogin.swift
saehejkang, line 83: Neglected to name the new client `RegistryClient`, because it is already defined like so in the `containerization` project

PR #1216, opened 2026-02-14
[networks]: add decoder/encoder to Attachment and NetworkStatus

### LINKED ISSUE #1196: Attachment and NetworkStatus lack backward-compatible decoding (unlike NetworkConfiguration)
## Summary

The `Attachment` and `NetworkStatus` structs in `ContainerResource` do not handle backward-compatible JSON decoding when communicating with older versions of `container-apiserver`. This causes `keyNotFound` decoding errors when a client built against v0.9.0 of this package talks to the v0.7.0 `container-apiserver` that ships with macOS 26.

`NetworkConfiguration` already handles this correctly with fallback decoding — `Attachment` and `NetworkStatus` should do the same.

## Background

Between v0.7.0 and v0.9.0, several JSON field names were renamed:

| Struct | v0.7.0 field | v0.9.0 field |
|--------|-------------|-------------|
| `Attachment` | `address` (String) | `ipv4Address` (CIDRv4) |
| `Attachment` | `gateway` (String) | `ipv4Gateway` (IPv4Address) |
| `NetworkStatus` | `address` (String) | `ipv4Subnet` (CIDRv4) |
| `NetworkStatus` | `gateway` (String) | `ipv4Gateway` (IPv4Address) |
| `NetworkConfiguration` | `subnet` (String?) | `ipv4Subnet` (CIDRv4?) |

`NetworkConfiguration` already has a custom `init(from decoder:)` that tries `ipv4Subnet` first and falls back to `subnet`:

```swift
// NetworkConfiguration.swift (already correct)
let subnetText =
    try container.decodeIfPresent(String.self, forKey: .ipv4Subnet)
    ?? container.decodeIfPresent(String.self, forKey: .subnet)
```

`Attachment` and `NetworkStatus` use the default synthesized `Codable` conformance, which only accepts the new field names and fails with `keyNotFound` when it encounters the old ones.

## Impact

Any downstream project (e.g., [socktainer](https://github.com/socktainer/socktainer)) that builds against `apple/container` v0.9.0 but runs on a stock macOS 26 system (which ships `container-apiserver` v0.7.0) will crash when decoding network data.

Error:
```
keyNotFound(CodingKeys(stringValue: "ipv4Address", intValue: nil),
  Swift.DecodingError.Context(codingPath: [..., "networks", ...],
  debugDescription: "No value associated with key \"ipv4Address\""))
```

## Suggested Fix

Add custom `init(from decoder:)` to `Attachment` and `NetworkStatus`, matching the pattern already used in `NetworkConfiguration`.

For `Attachment` (`Sources/ContainerResource/Network/Attachment.swift`):

```swift
public init(from decoder: Decoder) throws {
    let container = try decoder.container(keyedBy: CodingKeys.self)
    self.network = try container.decode(String.self, forKey: .network)
    self.hostname = try container.decode(String.self, forKey: .hostname)

    if let addr = try? container.decode(CIDRv4.self, forKey: .ipv4Address) {
        self.ipv4Address = addr
    } else {
        self.ipv4Address = try container.decode(CIDRv4.self, forKey: .address)
    }

    if let gw = try? container.decode(IPv4Address.self, forKey: .ipv4Gateway) {
        self.ipv4Gateway = gw
    } else {
        self.ipv4Gateway = try container.decode(IPv4Address.self, forKey: .gateway)
    }

    self.ipv6Address = try container.decodeIfPresent(CIDRv6.self, forKey: .ipv6Address)
    self.macAddress = try container.decodeIfPresent(MACAddress.self, forKey: .macAddress)
}

private enum CodingKeys: String, CodingKey {
    case network, hostname, ipv4Address, ipv4Gateway, ipv6Address, macAddress
    case address, gateway // v0.7.0 field names
}
```

Same pattern for `NetworkStatus` (`Sources/ContainerResource/Network/NetworkState.swift`).

## Related

- socktainer/socktainer#181 — downstream breakage caused by this inconsistency

saehejkang:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context

Adds backward-compatible JSON decoding/encoding fallback for `Attachment` and `NetworkStatus` to prevent `keyNotFound` errors when communicating with older `container-apiserver` versions.

Temporary fix but closes #1196

## Testing

Tests ✅ based on comment [here](https://github.com/apple/container/issues/1196#issuecomment-3902754548)


Sources/ContainerResource/Network/NetworkState.swift
```diff
@@ -39,6 +39,43 @@ public struct NetworkStatus: Codable, Sendable {
         self.ipv4Gateway = ipv4Gateway
         self.ipv6Subnet = ipv6Subnet
     }
+
+    enum CodingKeys: String, CodingKey {
+        case ipv4Subnet
+        case ipv4Gateway
+        case ipv6Subnet
+        // TODO: retain for deserialization compatability for now, remove later
```
manojmahapatra: typo: compatability -> compatibility

Sources/ContainerResource/Network/Attachment.swift
```diff
@@ -47,4 +47,49 @@ public struct Attachment: Codable, Sendable {
...
     }
+
+    enum CodingKeys: String, CodingKey {
+        case network
+        case hostname
+        case ipv4Address
+        case ipv4Gateway
+        case ipv6Address
+        case macAddress
+        // TODO: retain for deserialization compatability for now, remove later
```
manojmahapatra: typo: compatability -> compatibility.

Sources/ContainerResource/Network/Attachment.swift
```diff
@@ -47,4 +47,49 @@ public struct Attachment: Codable, Sendable {
...
+    }
+
+    /// Create a configuration from the supplied Decoder, initializing missing
+    /// values where possible to reasonable defaults.
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        network = try container.decode(String.self, forKey: .network)
+        hostname = try container.decode(String.self, forKey: .hostname)
+        if let address = try? container.decode(CIDRv4.self, forKey: .ipv4Address) {
```
manojmahapatra, line 75: Suggestion: would be nice to add some tests for legacy payload decoding.

manojmahapatra, line 75: Suggestion: I think optional `try?` for fallback approach will swallow decode errors. A better approach would be fallback only on keyNotFound and throw error for malformed payloads. You could use something akin to;

```Swift
 private static func decodeOrFallback<T: Decodable>(
      _ type: T.Type,
      from container: KeyedDecodingContainer<CodingKeys>,
      primary: CodingKeys,
      fallback: CodingKeys
  ) throws -> T {
      do {
          // return primary
      } catch let DecodingError.keyNotFound(key, _) where key.stringValue ==
  primary.stringValue {
          // return fallback
      } catch {
          // throw error
      }
  }

// callsite will be much cleaner

ipv4Address = try Self.decodeOrFallback(
      CIDRv4.self,
      from: container,
      primary: .ipv4Address,
      fallback: .address
  )
```

saehejkang: I see the point about being more precise with error handling, but in this case I think it is a bit over-engineered for what we need right now. I also agree that swallowing all decode errors via `try?` isn’t ideal, but at the moment we don’t have meaningful error propagation wired up (managed resource work). Would prefer to just keep what we have here.

saehejkang: This might warrant an issue (bug or request) to be added on later.  I want to get these changes merged soon because they are blocking `socktainer` users from using the latest version of `container`.

Sources/ContainerResource/Network/Attachment.swift
saehejkang: good catch

Sources/ContainerResource/Network/NetworkState.swift
saehejkang: good catch

saehejkang:
@jglogan bringing this to your inbox as it a blocker for socktainer and should get merged soon

saehejkang:
@JaewonHur ready for your ✅ again

JaewonHur:
Looks like some conflicts with main..?

saehejkang:
> Looks like some conflicts with main..?

should be fixed now I think

saehejkang:
@JaewonHur bump 😄

Sources/ContainerResource/Network/Attachment.swift
```diff
@@ -51,4 +51,49 @@ public struct Attachment: Codable, Sendable {
...
+    enum CodingKeys: String, CodingKey {
+        case network
+        case hostname
+        case ipv4Address
+        case ipv4Gateway
+        case ipv6Address
+        case macAddress
+        // TODO: retain for deserialization compatibility for now, remove later
+        case address
+        case gateway
```
JaewonHur, line 65: Sorry for the delay.
Seems `mtu` field is added.
Let me approve and merge it once it is done!

saehejkang: I just pushed the changes that should fix the ci issues. Completely forgot about the `mtu` field that was just added.

vxnkjyffmq-code:
Appreciate it

Issue #1212, opened 2026-02-14
[Bug]: Environment variables can be duplicated on run


natevw:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

`container run` does not fully override environment variables of the underlying image.

1. provide a default `ENV` variable in a Dockerfile (e.g. <https://github.com/alpine-docker/ollama/blob/e92f203364ee4f5dafc00115cc23fc68a5bfb8a4/Dockerfile#L17-L18>)
2. attempt to override it when running the image e.g. `container run -e OLLAMA_HOST=ollama.local:11434 alpine/ollama ps`
3. the executable does not behave as though the variable is set (in this case it complains "Error: could not connect to ollama server, run 'ollama serve' to start it")

### Current behavior

The reason can be seen by inspecting the container's environment via e.g. `container run -e OLLAMA_HOST=ollama.local:11434 --entrypoint=/usr/bin/env alpine/ollama` which reveals:

```
PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin
SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
OLLAMA_HOST=0.0.0.0
OLLAMA_HOST=ollama.local:11434
HOME=/root
```
so when the normal entrypoint calls some equivalent of `getenv("OLLAMA_HOST")` it receives the value of the *first* entry rather than the last.

Note that this behavior is masked by any intervening shell. For example running `container run -e OLLAMA_HOST=ollama.local:11434 --entrypoint=/bin/sh alpine/ollama -c env` outputs the actually-desired outcome instead:

```
SHLVL=1
HOME=/root
OLLAMA_HOST=ollama.local:11434
SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt
PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin
PWD=/
```

I'm assuming `sh` (or perhaps any intervening process) must have the effect of cleaning up its environment and removing the duplicates.

### Expected behavior

The output of `docker  run -e OLLAMA_HOST=ollama.local:11434 --entrypoint=/usr/bin/env alpine/ollama` is simply:

```
PATH=/usr/local/sbin:/usr/local/bin:/usr/bin:/usr/sbin:/sbin:/bin
HOSTNAME=d64b91a7b258
TERM=xterm
[...]

PR #1218, opened 2026-02-14
Fix: Environment variables can be duplicated on run 

simone-panico:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
As reported in [1212](https://github.com/apple/container/issues/1212), environment variables defined in a Dockerfile (ENV) were not being overridden when passed via -e at runtime, resulting in duplicate entries in the container's environment.

**Root cause:** `Parser.allEnv()` simply merged image, env-file, and user-provided environment variables without deduplication. Since `getenv()` returns the first match, the image default took precedence over the user override.

**Fix:** `allEnv()` now deduplicates by key, with later sources overriding earlier ones

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,3 +3,14 @@ cli:
     - any-glob-to-any-file:
       - 'Sources/CLI/**'
       - 'Sources/ContainerCommands/**'
+
+documentation:
+  - changed-files:
+    - any-glob-to-any-file:
+      - '**/*.md'
+      - 'docs/**'
+
+ci:
+  - changed-files:
+    - any-glob-to-any-file:
+      - '.github/**'
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -71,15 +71,37 @@ jobs:
       - name: Test the container project
         run: |
           APP_ROOT=$(mktemp -d -p "${RUNNER_TEMP}")
+          LOG_ROOT="${APP_ROOT}/logs"
           trap 'rm -rf "${APP_ROOT}"; echo Removing data directory ${APP_ROOT}' EXIT
-          echo "Created data directory ${APP_ROOT}"
+          echo "created data directory: ${APP_ROOT}"
+          echo "hostname: $(hostname)"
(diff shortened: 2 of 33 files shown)


Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -98,13 +98,25 @@ public struct Parser {
...
+        var combined: [String] = []
+        combined.append(contentsOf: Parser.env(envList: imageEnvs))
         for envFile in envFiles {
             let content = try Parser.envFile(path: envFile)
-            output.append(contentsOf: content)
+            combined.append(contentsOf: content)
+        }
+        combined.append(contentsOf: Parser.env(envList: envs))
+
+        var seen = [String: Int]()
```
JaewonHur: Hi @simone-panico!
Thank you for submitting this PR :)

One minor thing.
Could we do this more functional way?
My suggestion is first converting `combined` to `Map[String:String]` using `combined.reduce`, and then returning the reconstructed array of `"<key>=<value>"`.

Tests/ContainerAPIClientTests/ParserTest.swift
```diff
@@ -552,6 +552,42 @@ struct ParserTest {
...
     }
+    
+    @Test
+    func testAllEnvUserOverridesImage() throws {
+        let result = try Parser.allEnv(
+            imageEnvs: ["FOO=fromimage", "BAR=kept"],
+            envFiles: [],
+            envs: ["FOO=fromuser"]
+        )
+        #expect(result == ["FOO=fromuser", "BAR=kept"])
```
JaewonHur: By doing so, you would have to do something like `#expect(Set(result) == Set(["FOO=fromuser", "BAR=kept"]))` as `Map` doesn't preserve order between keys (which is totally fine!).

JaewonHur:
@simone-panico Could you run `make fmt` and commit again?

Issue #1047, opened 2026-01-14
[Request]: Support loading network configuration from file


katiewasnothere:
### Feature or enhancement request details

Today our network plugin will take in a set of parameters for configuring the network we're trying to make. Instead, we could load the network configuration from a fixed path in the container bundle. This would allow us to be more flexible in what options we send to different network plugins. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1227, opened 2026-02-18
Support loading network configuration from file

apoorvdarshan:
## Summary
- Adds a `--config` option to `container-network-vmnet start` that accepts a JSON network configuration file, closing #1047
- Introduces `NetworkConfigurationFile` as a `Codable` input model that converts into the current `NetworkConfiguration` schema
- Keeps CLI flags as overrides when both CLI values and a config file are provided
- Supports current network `plugin` / `options` fields, including the vmnet `variant` option

## Test plan
- [x] Added unit tests for minimal and complete config decoding, invalid subnets, and invalid names
- [x] Existing `NetworkConfigurationTest` coverage continues to pass
- [x] `swift test --filter NetworkConfigurationTest` passes (7/7 tests)

### DIFF
--- a/Sources/ContainerResource/Network/NetworkConfiguration.swift
+++ b/Sources/ContainerResource/Network/NetworkConfiguration.swift
@@ -149,3 +149,59 @@ private struct _LegacyPluginInfo: Codable {
     let plugin: String
     let variant: String?
 }
+
+/// Intermediate representation for loading network configuration from a JSON file.
+///
+/// Only `mode` is required. All other fields have sensible defaults.
+/// The network name (ID) is provided separately via the CLI argument, not the file.
+public struct NetworkConfigurationFile: Codable, Sendable {
+    /// The network type (e.g. "nat", "hostOnly").
+    public let mode: NetworkMode
+
+    /// The preferred CIDR address for the IPv4 subnet, if specified.
+    public let ipv4Subnet: String?
+
+    /// The preferred CIDR address for the IPv6 subnet, if specified.
+    public let ipv6Subnet: String?
+
+    /// Key-value labels for the network.
+    public let labels: [String: String]?
+
+    /// The network plugin that manages this network.
+    public let plugin: String?
+
+    /// Plugin-specific options for this network.
+    public let options: [String: String]?
+
+    public init(
+        mode: NetworkMode,
+        ipv4Subnet: String? = nil,
+        ipv6Subnet: String? = nil,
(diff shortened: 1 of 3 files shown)


jglogan:
@apoorvdarshan Thank you for the contribution!

@katiewasnothere Was the `--config` option in #1047 meant for `container network create` or for `container-network-vmnet start`? I thought it was for the latter.

katiewasnothere:
@jglogan https://github.com/apple/container/issues/1047 is for adding --config as an option on `container-network-vmnet start`

saehejkang:
I was drafting this comment when I noticed things were changing a bit. However, I believe the feedback is still valid and may be worth discussing.

-----

Instead of making the flags mutually exclusive, I think we should check whether a CLI option is explicitly set and prefer it over the value from the config file when both are provided (intentionally or accidentally)? It seems to be a more common and flexible pattern that improves usability.

apoorvdarshan:
Thanks for the feedback!

@katiewasnothere Good catch — I've moved the `--config` option from `container network create` to `container-network-vmnet start` as intended by #1047.

@saehejkang Agreed, the override approach is better. I've updated the implementation so that CLI flags override config file values when both are provided, rather than making them mutually exclusive. `--mode` and `--variant` are now optionals so we can detect when they're explicitly set vs defaulted.

jglogan:
@apoorvdarshan Thanks! I think we'll need until a little later next week to work through some design stuff related to this. We may need a change or two but what you have is in the direction of where things are going.

On these helper executables, we're wanting to eliminate the options in favor of either a `--config` option or maybe a `--path` option to an entity directory (like the container bundle directory), and then the config would be at a conventional path inside.

This helps make it so that the plugin-specific details of runtimes and networks are the concern of the CLI and service plugins, and those details are opaque to the containers and networks services in the API server; they just need to create the entity directories that the plugins understand.

apoorvdarshan:
@jglogan Thanks for the context! That makes sense — having helper executables take a `--config` or `--path` option instead of individual flags keeps plugin-specific details out of the API server. Happy to iterate on this once the design is finalized next week.

apoorvdarshan:
Rebased this onto the current `main` and ported the config-file support to the newer network model (`name`, `ResourceLabels`, `plugin`, and `options`) and the moved vmnet plugin source. CLI values still override file values. `swift test --filter NetworkConfigurationTest` passes all 7 focused tests.

PR #1234, opened 2026-02-19
Fix directory watcher

### LINKED ISSUE #1207: [Bug]: Directory monitor for localhost DNS fails if no resolver directory.
### I have done the following

### Steps to reproduce

In CI, many runners don't have an /etc/resolver directory set up, which causes the local DNS task to fail on API server start.

This results in an error during the test phase of the git action; the`install-kernel` make target fails on `container system start`.

The directory monitor must transition between two operational states:

- when the directory is not present, remove all localhost DNS entries, and poll until the directory is present
- when the directory is present, process any localhost DNS files and then monitor for changes

### Current behavior

tests crash on CI while installing kernel

### Expected behavior

tests run as expected

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: 0.9.0.2
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

JaewonHur:
Current DirectoryWatcher fails if `/etc/resolver` does not exist. This PR fixes DirectoryWatcher to handle non-existing `/etc/resolver` directory. If that directory does not exist, it first watches `/etc` directory to check if `/etc/resolver` directory is created later. Once it detects new `/etc/resolver` directory, it starts watching new DNS resolver files there.

This PR also fixes to log the exception thrown by API server's tasks.

Resolves #1207

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
In case of non-existing `/etc/resolver` starting API server fails.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Helpers/APIServer/LocalhostDNSHandler.swift
+++ b/Sources/Helpers/APIServer/LocalhostDNSHandler.swift
@@ -22,36 +22,37 @@ import DNS
 import DNSServer
 import Foundation
 import Logging
+import Synchronization
 
 actor LocalhostDNSHandler: DNSHandler {
     private let ttl: UInt32
     private let watcher: DirectoryWatcher
 
-    nonisolated(unsafe) private var dns: [String: IPv4]
+    private let dns: Mutex<[String: IPv4]>
 
     public init(resolversURL: URL = HostDNSResolver.defaultConfigPath, ttl: UInt32 = 5, log: Logger) {
         self.ttl = ttl
 
         self.watcher = DirectoryWatcher(directoryURL: resolversURL, log: log)
-        self.dns = [:]
+        self.dns = Mutex([:])
     }
 
     public func monitorResolvers() async {
         await self.watcher.startWatching { fileURLs in
-            var dns: [String: IPv4] = [:]
+            var dns: [String: String] = [:]
             let regex = try Regex(HostDNSResolver.localhostOptionsRegex)
 
             for file in fileURLs.filter({ $0.lastPathComponent.starts(with: HostDNSResolver.containerizationPrefix) }) {
                 let content = try String(contentsOf: file, encoding: .utf8)
 
                 if let match = content.firstMatch(of: regex),
-                    let ipv4 = IPv4(String(match[1].substring ?? ""))
+                    let ipv4 = (match[1].substring.map { String($0) })
                 {
                     let name = String(file.lastPathComponent.dropFirst(HostDNSResolver.containerizationPrefix.count))
                     dns[name + "."] = ipv4
                 }
             }
-            self.dns = dns
+            self.dns.withLock { $0 = dns.compactMapValues { IPv4($0) } }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/ContainerOS/DirectoryWatcher.swift
```diff
@@ -15,46 +15,48 @@
 //===----------------------------------------------------------------------===//
 
 import ContainerizationError
+import ContainerizationOS
```
jglogan, line 18: I don't think this should go in DNSServer.

Right now LocalDNSHandler depends on ContainerAPIClient and ContainerDNSHandler holds a reference on networksService and thus depends on ContainerAPIService.

Is it possible to rework ContainerDNSHandler so that it doesn't use an injected dependency and instead uses the API client like LocalDNSHandler does?

Then we the dependencies simplified, we could create a `Sources/Services/ContainerAPIService/Library` or something like it and move this and the DNS handlers into it.

Sources/DNSServer/DirectoryWatcher.swift
```diff
@@ -15,46 +15,48 @@
 //===----------------------------------------------------------------------===//
 
 import ContainerizationError
+import ContainerizationOS
 import Foundation
 import Logging
 
 public class DirectoryWatcher {
```
jglogan: add docc for class and methods

Sources/DNSServer/DirectoryWatcher.swift
```diff
@@ -66,19 +68,64 @@ public class DirectoryWatcher {
...
 
-    deinit {
-        guard let source else {
+    public func startWatching(handler: @escaping ([URL]) throws -> Void) throws {
+        guard source == nil else {
+            throw ContainerizationError(.invalidState, message: "already watching on \(directoryURL.path)")
+        }
+
+        let parent = directoryURL.deletingLastPathComponent().resolvingSymlinksInPathWithPrivate()
+        guard parent.isDirectory else {
```
jglogan: We should document this function to mention that it only works if the parent directory exists.

What would the tradeoff be if we just polled for the presence of the parent directory? Some extra latency when going from polling to watching, but can watch anything under any condition.

JaewonHur: Moved to use polling based approach.

Sources/ContainerOS/DirectoryWatcher.swift
JaewonHur: Added `Sources/ContainerOS` for low level stuffs and moved `DirectoryWatcher.swift` under it.

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
```diff
@@ -15,18 +15,19 @@
...
 import Foundation
 import Logging
 
-class LocalhostDNSHandler: DNSHandler {
+actor LocalhostDNSHandler: DNSHandler {
     private let ttl: UInt32
     private let watcher: DirectoryWatcher
 
-    private var dns: [String: IPv4]
+    nonisolated(unsafe) private var dns: [String: IPv4]
```
jglogan: `nonisolated(unsafe)` says that the type is protected against concurrent access some other way? Is that the case here?

JaewonHur: Updated to use mutex. Thanks!

PR #1242, opened 2026-02-21
fix(BuildFSSync): use readlink(2) to avoid autofs stall when resolving symlinks to /home/ paths

### LINKED ISSUE #1233: [Bug]: deleting a symlink to a /home/<user> folder inside container causes autofs delay
### I have done the following

### Steps to reproduce

I have a symlink in the container created by software that looks like:

 latest -> /home/scion/.claude/debug/c3091f3d-0324-4310-b268-32c53d4af2a5.txt

When I go to remove this symlink, or delete the container, I get about a 5-10 second delay

This is whether I do any kind of unlink, mv, rm  etc causes this delay

My looking into this blames "quirk of macOS's autofs (automounter) subsystem."

I haven't been able to find any way to avoid this

### Current behavior

5-10 second delay on container removal

### Expected behavior

no delay

### Environment

```markdown
- OS: 
- Xcode: 
- Container:
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Ronitsabhaya75:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #1233.

When a container has a virtiofs bind-mount mapping a host directory to `/home/<user>` inside the container, and a symlink inside that mount points to an absolute path (e.g.` /home/scion/.claude/debug/foo.txt`), any operation that causes the symlink target to be resolved through the macOS VFS layer triggers a 5–10 second stall.

Root Cause: `URL.resolvingSymlinksInPath()` calls `realpath(3)` under the hood, which asks the kernel to fully resolve the symlink target. For paths under `/home/`, macOS's `autofs` automounter intercepts the lookup and times out waiting for a mount that never comes.

The fix: Replace `resolvingSymlinksInPath()` with a `rawSymlinkTarget()` helper that uses `Darwin.readlink(2)` to read the raw string stored in the symlink inode without ever touching the target path in the filesystem, so `autofs` is never triggered.

```swift
fileprivate func rawSymlinkTarget() -> String? {
    var buf = [CChar](<repeating: 0, count: Int(PATH_MAX>))
    let len = Darwin.readlink(self.path(percentEncoded: false), &buf, buf.count - 1)
    guard len > 0 else { return nil }
    return buf.prefix(len).withUnsafeBytes { String(decoding: $0, as: UTF8.self) }
}

```

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerBuild/BuildFSSync.swift
+++ b/Sources/ContainerBuild/BuildFSSync.swift
@@ -152,12 +152,15 @@ actor BuildFSSync: BuildPipelineHandler {
             entries[parentPath, default: []].insert(entry)
 
             if url.isSymlink {
-                let target: URL = url.resolvingSymlinksInPath()
-                if self.contextDir.parentOf(target) {
-                    let relPath = try target.relativeChildPath(to: self.contextDir)
-                    let entry = DirEntry(url: target, isDirectory: target.hasDirectoryPath, relativePath: relPath)
-                    let parentPath: String = try target.deletingLastPathComponent().relativeChildPath(to: self.contextDir)
-                    entries[parentPath, default: []].insert(entry)
+                if let rawTarget = url.rawSymlinkTarget() {
+                    let target = URL(fileURLWithPath: rawTarget, relativeTo: url.deletingLastPathComponent())
+                        .standardized
+                    if self.contextDir.parentOf(target) {
+                        let relPath = try target.relativeChildPath(to: self.contextDir)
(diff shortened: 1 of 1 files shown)


dcantah:
This does not seem like it will help the issue linked, nor has there been any reports of resolving these links causing slowdowns in the build surface itself

PR #1243, opened 2026-02-21
Add one missing line of variant plumbing.

jglogan:
- Forgot to pass variant into the plugin registration call in #1239.


jglogan:
Closing, and #1255 reverts the previous change. This was fine for testing but the variant needs to be in a persistent runtime configuration file so that the value is available for `container start`.

PR #1244, opened 2026-02-22
cli: Add --init to run/create

### LINKED ISSUE #1225: [Bug]: Orphaned processes are never reaped (no --init support)
### I have done the following

### Steps to reproduce

```sh
container run --name reapro -d alpine:latest sleep 3600

# Create 5 orphans: parent forks child, parent exits, child sleeps 1s then exits
container exec reapro sh -c '
for i in 1 2 3 4 5; do
  sh -c "sh -c \"sleep 1\" &" &
  wait
done'

sleep 3

# All 5 are now permanent zombies
container exec reapro sh -c '
for p in /proc/[0-9]*/status; do
  grep -q "^State:.*Z" "$p" 2>/dev/null && echo "zombie: $(echo $p | cut -d/ -f3)"
done'

container rm -f reapro
```

### Current behavior

All 5 orphaned processes become permanent zombies parented to PID 1. They are never reaped and accumulate over the container's lifetime.

```
zombie: 8
zombie: 10
zombie: 12
zombie: 14
zombie: 16
```

The user's entrypoint runs directly as PID 1 inside the container's PID namespace. It never calls `wait()`, so orphaned children that exit become permanent zombies.

### Expected behavior

Orphaned processes should be reaped. Docker solves this with `--init`, which injects `tini` as PID 1 to reap children. `container` has no equivalent flag.

### Environment

```markdown
- OS: macOS 26.2 (25C56)
- Xcode: 26.2 (17C52)
- Container: container CLI version 0.9.0 (build: release)
```

### Relevant log output

```shell
$ container exec reapro sh -c '
for p in /proc/[0-9]*/status; do
  grep -q "^State:.*Z" "$p" 2>/dev/null && echo "zombie: $(echo $p | cut -d/ -f3)"
done'
zombie: 8
zombie: 10
zombie: 12
zombie: 14
zombie: 16
```

### Code of Conduct

dcantah:
Closes #1225

Add a flag to signify that we'd like to run a minimal init process that can reap zombie processes. The actual support for this is in the Containerization library so the plumbing here is very simple.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -188,6 +188,7 @@ public struct Flags {
             runtime: String?,
             ssh: Bool,
             tmpFs: [String],
+            useInit: Bool,
             virtualization: Bool,
             volumes: [String]
         ) {
@@ -213,6 +214,7 @@ public struct Flags {
             self.runtime = runtime
             self.ssh = ssh
             self.tmpFs = tmpFs
+            self.useInit = useInit
             self.virtualization = virtualization
             self.volumes = volumes
         }
@@ -238,6 +240,15 @@ public struct Flags {
         )
         public var entrypoint: String?
 
+        @Flag(name: .customLong("init"), help: "Run an init process inside the container that forwards signals and reaps processes")
+        public var useInit = false
+
+        @Option(
+            name: .long,
+            help: .init("Use a custom init image instead of the default", valueName: "image")
+        )
+        public var initImage: String?
+
         @Option(
             name: .shortAndLong,
             help: .init("Set a custom kernel path", valueName: "path"),
@@ -247,12 +258,6 @@ public struct Flags {
             }
         )
         public var kernel: String?
-
-        @Option(
-            name: .long,
-            help: .init("Use a custom init image instead of the default", valueName: "image")
-        )
-        public var initImage: String?
 
         @Option(name: [.short, .customLong("label")], help: "Add a key=value label to the container")
         public var labels: [String] = []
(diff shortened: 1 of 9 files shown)


Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -318,6 +318,9 @@ public struct Flags {
         @Option(name: [.customLong("volume"), .short], help: "Bind mount a volume into the container")
         public var volumes: [String] = []
 
+        @Flag(name: .long, help: "Run an init process inside the container that forwards signals and reaps processes")
+        public var `init` = false
```
jglogan: Could you add init in order? And relocate `--runtime` as well?

```console
MANAGEMENT OPTIONS:
  -a, --arch <arch>       Set arch if image can target multiple architectures (default: arm64)
  --cidfile <cidfile>     Write the container ID to the path provided
  -d, --detach            Run the container and detach from the process
  --dns <ip>              DNS nameserver IP address
  --dns-domain <domain>   Default DNS domain
  --dns-option <option>   DNS options
  --dns-search <domain>   DNS search domains
  --entrypoint <cmd>      Override the entrypoint of the image
  -k, --kernel <path>     Set a custom kernel path
  --init-image <image>    Use a custom init image instead of the default
  -l, --label <label>     Add a key=value label to the container
  --mount <mount>         Add a mount to the container (format: type=<>,source=<>,target=<>,readonly)
  --name <name>           Use the specified name as the container ID
  --network <network>     Attach the container to a network (format: <name>[,mac=XX:XX:XX:XX:XX:XX])
  --no-dns                Do not configure DNS in the container
  --os <os>               Set OS if image can target multiple operating systems (default: linux)
  -p, --publish <spec>    Publish a port from container to host (format: [host-ip:]host-port:container-port[/protocol])
  --platform <platform>   Platform for the image if it's multi-platform. This takes precedence over --os and --arch
  --publish-socket <spec> Publish a socket from container to host (format: host_path:container_path)
  --read-only             Mount the container's root filesystem as read-only
  --rm, --remove          Remove the container after it stops
  --rosetta               Enable Rosetta in the container
  --ssh                   Forward SSH agent socket to container
  --tmpfs <tmpfs>         Add a tmpfs mount to the container at the given path
  --virtualization        Expose virtualization capabilities to the container (requires host and guest support)
  -v, --volume <volume>   Bind mount a volume into the container
  --runtime <runtime>     Set the runtime handler for the container (default: container-runtime-linux)
```

Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -318,6 +318,9 @@ public struct Flags {
         @Option(name: [.customLong("volume"), .short], help: "Bind mount a volume into the container")
         public var volumes: [String] = []
 
+        @Flag(name: .long, help: "Run an init process inside the container that forwards signals and reaps processes")
```
jglogan: Why not customLong("init") and `useInit` for the field name, and no more backticks?

jglogan (approved):

jglogan (approved):

jglogan (approved):

Issue #286, opened 2025-07-01
[Request]: Add support for --restart flag for run


egernst:
### Feature or enhancement request details

As a container user, I can use the --restart option on a container run or container create command so that my containers automatically stay running as long as container runs.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

turbosnail11:
+1 for this.  I think this is more important than Docker-Compose (which I also always use) because there is no simple workaround for this. If containers aren't resistant to crashing then they simply can't be used.

ronguest:
+1 again, this seems essential so hopefully it can be added soon

0777888212:
Hello

null

tcarlson25:
Any update on this?

saehejkang:
I will take a look into this, as it seems to be highly requested

PR #1258, opened 2026-02-24
Add support for container restart policy

JaewonHur:
This PR initiates the implementation for adding restart feature (#286), which can be extended to support auto-start later (#158, #1201). The first milestone is to implement restart policies `no`, `on-failure`, and `always` (which behaves similar to [docker-restart-policy](https://docs.docker.com/engine/containers/start-containers-automatically/)).

To support restart policy, following persistent and runtime state variables should be added:
* In `ContainerCreateOption` (persistent)
  * `RestartPolicy`: remember restart policy across the container life cycle.
* In `ContainerState` (run-time)
  * `StoppedState(startError: Error?, exitStatus: ExitStatus?)`: determine restart logic when container exits
  * `manualStopped: Bool`: remember if the container is stopped by user
    * `True` on first container creation, `False` on `loadAtBoot`
  * `backOff: Duration?`: remember backoff timer.
    * Default to `nil`
    * Set 100ms on first non-manual exit (and then doubled until 10 seconds)
    * Set `nil` on `startError` or `manualStopped`

A new state should be added in container life cycle:
* `restarting`: restarted containers first go to this state and waits for `backOff` to start.

Implementation steps are described below:

- [x] **[ContainersService]** Add state variables for restart policy
- [x] **[ContainerResource, ContainerCommand]** Add restart policy and expose command line.
- [ ] **[ContainersService]** Add `restarting` state, where the restarting containers first visit, and proceeds to `bootstrap` (after `backOff`) then `startProcess`.
- [ ] **[ContainersService]** Implement restart logic in `handleContainerExit`
   * [restart == no]
        * Idle (stay at stopped).
    * [restart == on-failure]
        * startError || manualStopped || exitStatus == nil
            * Idle
        * exitStatues != nil && exitStatus != 0
            * Restart
    * [restart == always]
        * startError || manualStopped
            * Idle
        * Else
            * Restart
- [ ] Add stability call, which marks the container to be stable 10 seconds after `running`---i.e., resetting the `backOff`.
    * It might be something like `stabilityMonitor` (similar to `ExitMonitor`) that registers callback on successful `startProcess`.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [X] Breaking change
- [ ] Documentation update

## Motivation and Context
Users can setup auto-restarting containers using this feature.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


saehejkang:
@JaewonHur Busy week and I hope to get to review this here soon. I saw your comment and I am all for following the direction we take in this PR. Though it might be worthwhile to get started on the discussion on how we want to handle test cases and maybe work together on that piece?

JaewonHur:
Hi @saehejkang, sorry for moving forward on my own.

I just did minimal implementation today, but I'll definitely put our name together here.
This PR now just contains prototype implementation, and anything can be changed! Please take a look :)

Further things to do might be:
1. supporting `--restart systemStart` (or make `always` to start on system start?)
2. Add restart counter for `onFailure`.
3. Also, testing for restarting container across system can be added.

Sources/ContainerResource/Container/ContainerCreateOptions.swift
```diff
@@ -14,13 +14,32 @@
...
+
+    public static let `default` = ContainerCreateOptions(autoRemove: false, restartPolicy: .no)
+
+    enum CodingKeys: String, CodingKey {
+        case autoRemove
+        case restartPolicy
     }
 
-    public static let `default` = ContainerCreateOptions(autoRemove: false)
+    public init(from decoder: Decoder) throws {
```
saehejkang, line 39: Why was the decoder not being used before when it was just `autoRemove`?

Sources/ContainerResource/Container/ContainerStatus.swift
```diff
@@ -0,0 +1,33 @@
...
+
+/// Runtime status for a sandbox or container.
+public enum ContainerStatus: String, CaseIterable, Sendable, Codable {
+    /// The object is in an unknown status.
+    case unknown
+    /// The object is currently stopped.
+    case stopped
+    /// The object is waiting to be restarted.
+    case restarting
+    /// The object is currently bootstrapped.
```
saehejkang, line 27: What does the `bootstrapped` status mean? The container or sandbox is `stopped` but ready to be `started`?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -130,6 +170,85 @@ public actor ContainersService {
         return results
     }
 
+    public func runRestartScheduler() throws {
```
saehejkang, line 173: Never thought I would see a `scheduler` again after OS in college 😆.

saehejkang (commented):
I went over things as best as I could for now. Lots of changes that I have not gotten a chance to dig deeper into (the scheduler). Don't worry about taking over and wrapping this up!

saehejkang:
> supporting --restart systemStart (or make always to start on system start?)

Solid foundation work was done with the restart policy and I am sure it could follow the same pattern?

> Add restart counter for onFailure.

What would be the use case for this? Could it be used to automatically stop trying to restart a container after a certain count?

> Also, testing for restarting container across system can be added.

Hopefully we can use something like #1176 to be a sort of final e2e test of all the policies.

Sources/ContainerResource/Container/ContainerStatus.swift
JaewonHur: `bootstrapped` means (by itself) the container is bootstrapped (i.e., `bootstrap` function has been successfully executed on that container).

Before this update, the state was same `stopped` whether it is bootstrapped or not. The only difference was `SandboxClient` is allocated or not---I personally felt that somewhat weird, but left there.

But after adding `restarting` state, it becomes more weird, since container state should go `restarting` (waiting for backOff), then `stopped` (after bootstrap`) then `running`.

BTW, now we split the `RuntimeStatus` to `ContainerStatus` and `SandboxStatus`, so it only applies to containers.

JaewonHur: While writing this comment, I found that I missed setting the container `restarting` in scheduler lol.

Sources/ContainerResource/Container/ContainerCreateOptions.swift
JaewonHur: I actually have no idea :)
Maybe, wanted to keep the code as simple as possible at that time?

JaewonHur:
> Solid foundation work was done with the restart policy and I am sure it could follow the same pattern?

Thanks, if you make PR to my branch, let me merge it :)

> What would be the use case for this? Could it be used to automatically stop trying to restart a container after a certain count?

Yes. Docker supports that functionality (https://docs.docker.com/engine/containers/start-containers-automatically/)

saehejkang:
@JaewonHur I have not forgotten about this PR. Just another busy week and I hope to come back to this and #1201 in the near future.

Sources/ContainerResource/Container/ContainerStatus.swift
saehejkang: > bootstrapped means (by itself) the container is bootstrapped (i.e., bootstrap function has been successfully executed on that container).

I understand that this has to do with the `SandboxService` and `SandboxClient`. Except, I have yet to understand what is exactly the point of those two things?

> BTW, now we split the RuntimeStatus to ContainerStatus and SandboxStatus, so it only applies to containers.

So it seems like `Containers` and the `Sandbox` are two separate entities with different capabilities. Goes back to my question above and how they differentiate.

JaewonHur: `Container` and `Sandbox` are separate entities with different lifecycle (though they are tightly coupled).

`Sandbox` is more for managing the lifecycle of VM.

`Sandbox` is only alive after the `Container` successfully executed `bootstrap`.

saehejkang:
@JaewonHur Back at it and here to help wrap this up! I know you are very busy so feel free to assign me to anything from the implementation steps. I think it is best to wrap up `restart` before I work on completing `auto-start`.

tonycoco:
This is also quite nice when NFS/CIFS/SMB mounts fail and a container needs to restart for connection purposes.

stephenlclarke:
I reviewed this again from the downstream [`stephenlclarke/container-compose`](https://github.com/stephenlclarke/container-compose) side alongside [apple/container#286](https://github.com/apple/container/issues/286). I do not want to open a competing restart-policy PR while this one is active.

The latest fork work is now split into narrow, PR-shaped branches rather than the older integration commits:

- `stephenlclarke/container` branch `restart-policy-create-options`, commit `c5668c19d139b1aeb7e2529cb1dedd01fb4532c1`: create-time restart policy model, `ContainerCreateOptions.restartPolicy`, Docker-compatible `container run/create --restart`, and `--rm` conflict validation.
- `stephenlclarke/container` branch `restart-policy-runtime`, commit `b41bb830db708bc839c94e01c8a75c7fecbe3db0`: runtime restart scheduling, manual-stop suppression, cancellation, retry limits, backoff, and stable-run reset.
- `stephenlclarke/container` branch `restart-policy-timing`, commit `8b1eff72481fa497328414e0483a08c768826f1a`: optional retry delay and successful-run window controls.
- `stephenlclarke/container-compose` branch `compose-restart-policy-mapping`, commit `06cc3220c03b8ce0ab2eaf83ed81c08aca7f74b4`: downstream service `restart` and Deploy `restart_policy` mapping, plus a local-only Docker Compose V2 parity script that is intentionally not wired into CI.

I think the clean review stack is:

1. Create-time API and CLI contract only.

   This establishes the public shape without restarting containers yet. It covers:

   - `no`
   - `always`
   - `unless-stopped`
   - `on-failure`
   - `on-failure:<max-retries>`

   It also keeps old serialized create options decoding as the default `no` policy and rejects `--rm` with any non-default restart policy before container creation.

2. Runtime restart behavior.

   This then applies the stored policy when the init process exits. It covers manual-stop suppression, `on-failure` retry limits, `always` / `unless-stopped`, backoff, stable-run reset, and cancellation when a container is stopped, killed, deleted, or manually restarted.

3. Timing controls.

   This follows after the base behavior is accepted and adds optional retry delay / successful-run window fields. That lets downstream tools map Compose Deploy `restart_policy.delay` and `restart_policy.window` without putting Compose-specific behavior into `apple/container`.

Two details I would keep explicit in the first slice:

- the CLI syntax should stay Docker-compatible (`on-failure`, not Swift enum spelling such as `onFailure`);
- `on-failure:0` should mean unlimited retries, not zero retries. The fork now documents this in code as Docker/Moby behavior; Colima delegates this to Docker Engine, and Moby treats `MaximumRetryCount == 0` as unlimited.

For [`stephenlclarke/container-compose`](https://github.com/stephenlclarke/container-compose), the direct upstream dependency is just the generic `apple/container` primitive: a stored restart policy, then runtime behavior, then optional timing. Compose precedence rules, one-off `compose run` behavior, Deploy-model normalization, and local Docker Compose parity checks stay downstream.

Would Apple prefer this PR branch to be refreshed into that smaller stacked shape, or would a new first PR for only the create-time API/CLI contract be easier to review?

PR #1262, opened 2026-02-25
feat: Added container commit feature

i-nick:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Docker commit comes very handy with containers' images.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/Helpers/APIServer/APIServer+Start.swift
+++ b/Sources/Helpers/APIServer/APIServer+Start.swift
@@ -125,7 +125,7 @@ extension APIServer {
                             log.error("could not initialize resolver monitor", metadata: ["error": "\(error)"])
                             throw error
                         }
-                    
+
                         let nxDomainResolver = NxDomainResolver()
                         let compositeResolver = CompositeResolver(handlers: [localhostResolver, nxDomainResolver])
                         let hostsQueryValidator = StandardQueryValidator(handler: compositeResolver)
@@ -270,6 +270,7 @@ extension APIServer {
             routes[XPCRoute.containerStats] = harness.stats
             routes[XPCRoute.containerDiskUsage] = harness.diskUsage
             routes[XPCRoute.containerExport] = harness.export
+            routes[XPCRoute.containerCommit] = harness.commit
 
             return service
         }
--- a/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
+++ b/Sources/Services/ContainerAPIService/Client/ContainerClient.swift
@@ -164,12 +164,17 @@ public struct ContainerClient: Sendable {
     }
 
     /// Stop the container and all processes currently executing inside.
-    public func stop(id: String, opts: ContainerStopOptions = ContainerStopOptions.default) async throws {
+    public func stop(
+        id: String,
+        opts: ContainerStopOptions = ContainerStopOptions.default,
+        preserveOnStop: Bool = false
+    ) async throws {
         do {
             let request = XPCMessage(route: .containerStop)
             let data = try JSONEncoder().encode(opts)
             request.set(key: .id, value: id)
             request.set(key: .stopOptions, value: data)
+            request.set(key: .preserveOnStop, value: preserveOnStop)
 
             try await xpcClient.send(request)
         } catch {
@@ -333,6 +338,23 @@ public struct ContainerClient: Sendable {
             throw ContainerizationError(
                 .internalError,
                 message: "failed to export container",
+                cause: error
+            )
+        }
+    }
+
+    public func commit(id: String, archive: URL, live: Bool) async throws {
+        let request = XPCMessage(route: .containerCommit)
+        request.set(key: .id, value: id)
+        request.set(key: .archive, value: archive.absolutePath())
+        request.set(key: .commitLive, value: live)
+
+        do {
+            try await xpcClient.send(request)
+        } catch {
+            throw ContainerizationError(
+                .internalError,
+                message: "failed to commit container",
                 cause: error
             )
         }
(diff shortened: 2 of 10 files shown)


Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -884,13 +900,50 @@ public actor ContainersService {
...
+        case .stopping:
+            throw ContainerizationError(.invalidState, message: "container is stopping")
+        default:
+            throw ContainerizationError(.invalidState, message: "container is in an unknown state")
+        }
+
+        let path = self.containerRoot.appendingPathComponent(id)
+        let bundle = ContainerResource.Bundle(path: path)
+        let rootfs = bundle.containerRootfsBlock
+        try EXT4.EXT4Reader(blockDevice: FilePath(rootfs)).export(archive: FilePath(archive))
```
jglogan, line 923: I suspect this won't work well on a live disk image that doing any amount of I/O.

Could you revise this to do something like:

1. Check volume capabilities to ensure that we can clone (add this in a utility type down in ContainerOS).
2. Clone the disk file to a temp file in the same directory and defer remove it.
3. Run the extraction of the disk image on the cloned file.

The extraction needs to run in a task so that it doesn't block the server.

jglogan (changes requested):

JaewonHur:
Hi @i-nick, thank you for contributing to container project!

We might need some refactoring regarding this PR and #1265.

* We need updating `container export` to export tar file only as like what docker does.

* We need `container commit` as it is, but I feel like most of the server side code is the same as that for the `export`.
Instead of introducing XPC endpoint for `commit`, could we reuse (or slightly modify) that for `export`?

On the other hand, it might be worthwhile to discuss more on `restart` feature (and also `--live`). In some cases, it would be better to just abort and alert user when they run `container commit` on running containers as we don't know how the containers should be shutdown.

@jglogan  any thought?

jglogan:
I think the approach @i-nick has here is an interesting one, and could possibly be exploited for export on a running container.

The risk are:

- How well the read paths work for ContainerizationEXT4 on filesystems that have seen a lot of arbitrary activity. We've principally used the write paths for creating the container disk images from the OCI content.
- How well a snapshotted image works in general. If the APFS snapshot is a true point in time copy, then I believe any clone should be in a recoverable state when it's used as an image for another container. And the `--live` option prints a disclaimer, so if it doesn't work 1/10 times, the user was warned!

As a side note, Virtualization Framework also has [pause/resume](https://developer.apple.com/documentation/virtualization/vzvirtualmachine/pause(completionhandler:)) functionality, though I'm not sure it would improve commit/export beyond what a snapshot offers.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
i-nick: Sure, my intention for --live was a best effort copy. But yes, I will revise the PR. Thanks for the guidance!

Sources/Helpers/APIServer/APIServer+Start.swift
```diff
@@ -270,6 +270,7 @@ extension APIServer {
             routes[XPCRoute.containerStats] = harness.stats
             routes[XPCRoute.containerDiskUsage] = harness.diskUsage
             routes[XPCRoute.containerExport] = harness.export
+            routes[XPCRoute.containerCommit] = harness.commit
```
JaewonHur, line 273: Since `container export` and `container commit` share most of the functionalities, I believe we can have only one XPC for them. Could we leave `XPCRoute.containerExport` only, and have them support `live` feature as you did?

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
```diff
@@ -164,12 +164,17 @@ public struct ContainerClient: Sendable {
     }
 
     /// Stop the container and all processes currently executing inside.
-    public func stop(id: String, opts: ContainerStopOptions = ContainerStopOptions.default) async throws {
+    public func stop(
```
JaewonHur, line 167: The container might not cleanly exit as a user might have different logic for stopping their containers. It'd better to just abort if a user requests `container commit` without `--live` option, but the container is running.

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
```diff
@@ -337,4 +342,21 @@ public struct ContainerClient: Sendable {
             )
         }
     }
+
+    public func commit(id: String, archive: URL, live: Bool) async throws {
```
JaewonHur, line 346: Let's merge it to `export`.

Sources/Services/ContainerAPIService/Client/ContainerClient.swift
JaewonHur: After adding the basic functionality, we can figure out a better way to support it.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -884,13 +900,50 @@ public actor ContainersService {
         try EXT4.EXT4Reader(blockDevice: FilePath(rootfs)).export(archive: FilePath(archive))
     }
 
-    private func handleContainerExit(id: String, code: ExitStatus? = nil) async throws {
+    public func commitRootfs(id: String, archive: URL, live: Bool) async throws {
```
JaewonHur, line 903: We might be able to merge it to `exportRootfs` also.

jglogan:
@i-nick Just got back from a bunch of other stuff and saw you closed this.

If you're still interested in moving this forward, I'm wondering if creating two separate issues and doing commit in stages would make the most sense:

- The first issue would cover adding `--live` to `container export` so that we can export an OCI layout tar file from a running container.
- The second issue would add the `container commit` subcommand that could use the same business logic a `container export` to create the a temporary tar file, and then use the business logic from `container image load` to load it.

PR #1269, opened 2026-02-26
Use Sendable DNS types.

### LINKED ISSUE #1268: [Request]: Eliminate external DNS library dependencies.

We simply aren't using the [async DNS client library](https://github.com/swift-server/async-http-client.git), and can just drop it.

The [DNS record library](https://github.com/Bouke/DNS.git) we depend on is unmaintained and does not have Sendable types, so it's very hard to use with Swift 6. We should create our own record types, using the bit fiddling ops we use for ContainerizationNetlink.

### Code of Conduct

jglogan:
- Closes #1268.
- The types we were using weren't very usable with Swift 6 structured concurrency.
- Implements just the subset of records that we use.
- Use notImplemented instead of formatError for unknown record types.
- Use pure actor for LocalhostDNSHandler now that we have sendable types.
- Use DNSName as key for table lookups in LocalhostDNSHandler and HostTableResolver.
- Utilize dot-suffixed domain names everywhere in the lookup chain.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/DNSServer/Records/DNSName.swift
+++ b/Sources/DNSServer/Records/DNSName.swift
@@ -22,7 +22,7 @@ import Foundation
 /// For example, "example.com" becomes: `[7]example[3]com[0]`
 public struct DNSName: Sendable, Hashable, CustomStringConvertible {
     /// The labels that make up this name (e.g., ["example", "com"]).
-    public var labels: [String]
+    public private(set) var labels: [String]
 
     /// Creates a DNS name representing the root (empty label list).
     public init() {


Sources/DNSServer/Records/DNSName.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    /// The wire format size of this name in bytes.
+    public var size: Int {
+        // Each label: 1 byte length + label bytes, plus 1 byte for null terminator
+        labels.reduce(1) { $0 + 1 + $1.utf8.count }
+    }
+
+    /// The fully-qualified domain name with trailing dot.
+    public var description: String {
+        labels.isEmpty ? "." : labels.joined(separator: ".") + "."
```
manojmahapatra, line 94: Question: some lookup paths appear to use plain string (no normalization), will this change cause any issue? For example in `AttachmentAllocator`;
```
// store
func allocate(hostname: String).. 

// lookup
func lookup(hostname: String) async throws -> UInt32? {
        hostnames[hostname]
}
```

jglogan: @manojmahapatra Thanks for looking at this! I converted it to draft.

This one was done by telling Claude to create the types we needed using the bit-fiddling code that ContainerizationNetlink uses, and I pushed it primarily to back the change up and get something started for the issue.

I still need to review the work myself and sanity check it, but I'll bet I would have missed this!

I'll have a look through the relevant RFCs as I review this myself. This isn't an area I'm super expert in so if you've got experience and suggestions I'd welcome them!

manojmahapatra: I'm not an expert in this area either, but I reviewed the PR in detail against current behavior. Overall this is moving in the right direction. 

I see two gaps worth addressing here:

- Hostname normalization consistency (e.g. `foo` vs `foo.`) across parse/store/lookup paths. (additional focused unit tests will be nice)
- Compression-pointer safety in `DNSName` decode to avoid malformed packet loops. Something like this maybe;
```Swift
// Calculate pointer offset from message start
                let pointerLocation = offset
                let pointer = Int(length & 0x3F) << 8 | Int(buffer[offset + 1])
                let pointerTarget = messageStart + pointer
                guard pointerTarget < pointerLocation else {
                    throw DNSBindError.unmarshalFailure(type: "DNSName", field: "compression pointer not prior")
                }
                offset = pointerTarget
```

jglogan:
@manojmahapatra I've addressed what you called out, and also we're now folding to lowercase when creating domain name records.

The larger change (requiring a trailing dot on all the DNS APIs) is in the first follow-up commit, and the smaller one for case folding and compression pointer recursion preventions is in the second one.

Finally got myself out from under some other obligations which've kept me busy elsewhere. I'll look over your apple/containerization#577 early tomorrow. Thank you again for your scrutiny on this PR!

Sources/Services/ContainerAPIService/Server/Networks/NetworksService.swift
```diff
@@ -339,8 +339,12 @@ public actor NetworksService {
     }
 
     /// Perform a hostname lookup on all networks.
-    public func lookup(hostname: String) async throws -> Attachment? {
-        try await self.stateLock.withLock { _ in
+    ///
+    /// - Parameter dnsHostname: A DNS-format hostname, optionally with a trailing dot
+    ///   (e.g. `"example.com."` or `"example.com"`). The trailing dot is stripped before lookup.
+    public func lookup(dnsHostname: String) async throws -> Attachment? {
```
manojmahapatra: Question: Looks like it's a public API break due to the function rename? Do we need a shim?

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -14,14 +14,23 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-import DNS
+import ContainerizationExtras
 
 /// Handler that uses table lookup to resolve hostnames.
+///
+/// All keys in `hosts4` must be canonical DNS names — fully-qualified with a
+/// trailing dot (e.g. `"example.com."`). This matches the canonical form used
```
manojmahapatra: 👍

Sources/DNSServer/Records/DNSName.swift
```diff
@@ -0,0 +1,146 @@
...
+    /// Creates a DNS name from an array of labels.
+    public init(labels: [String] = []) {
+        self.labels = labels
+    }
+
+    /// Creates a DNS name from a dot-separated string (e.g., "example.com.").
+    public init(_ string: String) {
+        // Remove trailing dot if present, then split
+        let normalized = string.hasSuffix(".") ? String(string.dropLast()) : string
+        self.labels = normalized.isEmpty ? [] : normalized.split(separator: ".").map { String($0).lowercased() }
```
manojmahapatra: Question: So here labels are lowercases when parsing names. However, `HostTableResolver` does not normalize keys on init, which I _think_ might be an issue. can you please confirm?

manojmahapatra:
@jglogan overall the changes looks good, just two questions.

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -14,14 +14,23 @@
...
-    public let hosts4: [String: IPv4]
+    public let hosts4: [String: IPv4Address]
     private let ttl: UInt32
 
-    public init(hosts4: [String: IPv4], ttl: UInt32 = 300) {
+    /// Creates a resolver backed by a static IPv4 host table.
+    ///
+    /// - Parameter hosts4: A dictionary mapping fully-qualified domain names (with trailing dot)
+    ///   to IPv4 addresses. Keys without a trailing dot will not match wire-decoded queries.
+    /// - Parameter ttl: The TTL in seconds to set on answer records.
```
manojmahapatra: nitpick: The TTL in seconds to set on answer records. Defaults to 300.

Sources/Services/ContainerAPIService/Server/Networks/NetworksService.swift
jglogan: Ah, my bad. With the docc description we don't really need to change the parameter name.

Tests/DNSServerTests/RecordsTests.swift
```diff
@@ -0,0 +1,443 @@
...
+        @Test("Serialize A record")
+        func serializeARecord() throws {
+            let ip = try IPv4Address("10.0.0.1")
+            let record = HostRecord(name: "test.com.", ttl: 300, ip: ip)
+            var buffer = [UInt8](repeating: 0, count: 64)
+
+            let endOffset = try record.appendBuffer(&buffer, offset: 0)
+
+            // Should have written: name + type(2) + class(2) + ttl(4) + rdlen(2) + rdata(4)
+            #expect(endOffset > 0)
```
katiewasnothere: Can we check that the end offset is > 4 instead?

Tests/DNSServerTests/RecordsTests.swift
```diff
@@ -0,0 +1,443 @@
...
+            #expect(buffer[ipStart + 3] == 1)
+        }
+
+        @Test("Serialize AAAA record")
+        func serializeAAAARecord() throws {
+            let ip = try IPv6Address("::1")
+            let record = HostRecord(name: "test.com.", ttl: 300, ip: ip)
+            var buffer = [UInt8](repeating: 0, count: 64)
+
+            let endOffset = try record.appendBuffer(&buffer, offset: 0)
```
katiewasnothere, line 384: Should we validate the endOffset here as well?

Tests/DNSServerTests/RecordsTests.swift
```diff
@@ -0,0 +1,443 @@
...
+            var buffer = [UInt8](repeating: 0, count: 64)
+            _ = try upper.appendBuffer(&buffer, offset: 0)
+
+            var parsed = DNSName()
+            _ = try parsed.bindBuffer(&buffer, offset: 0)
+            #expect(parsed.labels == ["example", "com"])
+        }
+
+        @Test("Reject forward compression pointer")
+        func rejectForwardCompressionPointer() throws {
```
katiewasnothere, line 225: Can we add a positive test case for compression pointers?

katiewasnothere (approved):
Couple test requests, otherwise LGTM

Sources/DNSServer/Records/DNSName.swift
jglogan: Thanks for catching that. We're not using HostTableResolver yet, but it would have been broken. I reworked the case folding and the HostTableResolver a little to do things correctly, and a little more cleanly.

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -14,30 +14,44 @@
...
-import DNS
+import ContainerizationExtras
 
 /// Handler that uses table lookup to resolve hostnames.
+///
+/// Keys in `hosts4` are normalized to `DNSName` on construction, so lookups
+/// are case-insensitive and trailing dots are optional.
 public struct HostTableResolver: DNSHandler {
-    public let hosts4: [String: IPv4]
+    public let hosts4: [DNSName: IPv4Address]
```
jglogan, line 24: summary:

- Replace `IPv4` with our `IPv4Address` everywhere.
- Use `DNSName` for keys to validate and normalize hostnames for lookup.

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -14,30 +14,44 @@
...
+    /// - Parameter ttl: The TTL in seconds to set on answer records (default is 300).
+    /// - Throws: `DNSBindError.invalidName` if any key is not a valid DNS name.
+    public init(hosts4: [String: IPv4Address], ttl: UInt32 = 300) throws {
+        self.hosts4 = try Dictionary(uniqueKeysWithValues: hosts4.map { (try DNSName($0.key), $0.value) })
         self.ttl = ttl
     }
 
     public func answer(query: Message) async throws -> Message? {
-        let question = query.questions[0]
+        guard let question = query.questions.first else {
```
jglogan, line 40: summary: updated all handlers to ensure `questions[0]` is present, and if not, defer to the next handler

This would result in a `NXDOMAIN` when the last handler returns nil, when a `.formatError` would be more appropriate. The typical usage, though, is to put a `StandardQueryValidator` at the front of the chain which ensures that exactly one question is present and returns a `.formatError` otherwise.

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -14,30 +14,44 @@
...
+            return nil
+        }
+        let n = question.name.hasSuffix(".") ? String(question.name.dropLast()) : question.name
```
jglogan, line 43: summary: for the DNS names that come in on the wire, we don't perform hostname validation - we just split into labels and form a `DNSName`, and look them up. If they have weird characters or are otherwise invalid hostnames, they'll just miss as L35 ensures that our keys are always valid hostnames.

Sources/DNSServer/Handlers/HostTableResolver.swift
```diff
@@ -48,28 +62,11 @@ public struct HostTableResolver: DNSHandler {
             }
             // If hostname doesn't exist, return nil which will become NXDOMAIN
             return nil
-        case ResourceRecordType.nameServer,
```
jglogan, line 51: summary: all unhandled question types in handlers should return `.notImplemented`

Sources/DNSServer/Records/DNSBindError.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// Errors that can occur during DNS message serialization/deserialization.
+public enum DNSBindError: Error, CustomStringConvertible {
```
jglogan, line 18: summary: extracted types from `Bindable.swift` into individual files

Sources/DNSServer/Records/DNSName.swift
```diff
@@ -0,0 +1,204 @@
...
+/// A DNS name encoded as a sequence of labels.
+///
+/// DNS names are encoded as: `[length][label][length][label]...[0]`
+/// For example, "example.com" becomes: `[7]example[3]com[0]`
+public struct DNSName: Sendable, Hashable, CustomStringConvertible {
+    /// The labels that make up this name (e.g., ["example", "com"]).
+    public var labels: [String]
+
+    /// Creates a DNS name representing the root (empty label list).
+    public init() {
```
jglogan, line 28: summary: used to building DNS names label by label from wire data

Sources/DNSServer/Records/UInt8+Binding.swift
```diff
@@ -0,0 +1,72 @@
...
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+// TODO: This copies some of the Bindable code from Containerization,
+// but we can't use Bindable as it presumes a fixed length record.
+// We can look at refining this later to see if we can use some common
+// bit fiddling code everywhere.
+
+extension [UInt8] {
```
jglogan, line 24: summary: removed Bindable.swift, extracting types into individual files, this code is brought over from containerization

Sources/DNSServer/DNSServer+Handle.swift
```diff
@@ -27,23 +27,32 @@ extension DNSServer {
         outbound: NIOAsyncChannelOutboundWriter<AddressedEnvelope<ByteBuffer>>,
         packet: inout AddressedEnvelope<ByteBuffer>
     ) async throws {
-        let chunkSize = 512
-        var data = Data()
+        // RFC 1035 §2.3.4 limits UDP DNS messages to 512 bytes. We don't implement
+        // EDNS0 (RFC 6891), and this server only resolves host A/AAAA queries, so a
+        // legitimate query will never approach this limit. Reject oversized packets
+        // before reading to avoid allocating memory for malformed or malicious datagrams.
+        let maxPacketSize = 512
```
jglogan, line 34: summary: added for safety

Sources/DNSServer/DNSServer+Handle.swift
```diff
@@ -64,21 +73,48 @@ extension DNSServer {
 
             self.log?.debug("serializing response")
             responseData = try response.serialize()
+        } catch let error as DNSBindError {
```
jglogan, line 76: summary: we now differentiate between structural errors in DNS queries (`.formatError`) and unsupported query values (`.notImplemented`)

Sources/DNSServer/DNSServer+Handle.swift
```diff
@@ -64,21 +73,48 @@ extension DNSServer {
...
+                id: id,
+                type: .response,
+                returnCode: returnCode,
+                questions: [],
+                answers: []
+            )
+            responseData = try response.serialize()
         } catch {
-            self.log?.error("error processing message from \(query): \(error)")
+            let rawId = data.count >= 2 ? data[0..<2].withUnsafeBytes { $0.load(as: UInt16.self) } : 0
```
jglogan, line 98: summary: our "500" path

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
```diff
@@ -28,73 +28,65 @@ actor LocalhostDNSHandler: DNSHandler {
     private let ttl: UInt32
     private let watcher: DirectoryWatcher
 
-    private let dns: Mutex<[String: IPv4]>
+    private var dns: [DNSName: IPv4Address]
```
jglogan, line 31: summary: no need for Mutex, actor isolation is fine here, and use DNSName keys to ensure hostname validity

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
```diff
@@ -28,73 +28,65 @@ actor LocalhostDNSHandler: DNSHandler {
...
+        await self.watcher.startWatching { [weak self] fileURLs in
+            var dns: [DNSName: IPv4Address] = [:]
             let regex = try Regex(HostDNSResolver.localhostOptionsRegex)
 
             for file in fileURLs.filter({ $0.lastPathComponent.starts(with: HostDNSResolver.containerizationPrefix) }) {
                 let content = try String(contentsOf: file, encoding: .utf8)
 
                 if let match = content.firstMatch(of: regex),
-                    let ipv4 = (match[1].substring.map { String($0) })
+                    let ipv4 = (match[1].substring.flatMap { try? IPv4Address(String($0)) })
```
jglogan, line 49: summary: `flatMap` should be used here since `substring` produces an optional

Sources/Helpers/APIServer/LocalhostDNSHandler.swift
```diff
@@ -28,73 +28,65 @@ actor LocalhostDNSHandler: DNSHandler {
...
         }
     }
 
-    nonisolated public func answer(query: Message) async throws -> Message? {
-        let question = query.questions[0]
+    public func answer(query: Message) async throws -> Message? {
+        guard let question = query.questions.first else {
+            return nil
+        }
+        let n = question.name.hasSuffix(".") ? String(question.name.dropLast()) : question.name
```
jglogan, line 66: Same pattern as HostTableHandler, don't need to do hostname validation on the wire data.

Sources/Services/ContainerAPIService/Client/HostDNSResolver.swift
```diff
@@ -51,10 +51,10 @@ public struct HostDNSResolver {
 
         let dnsPort = localhost == nil ? "2053" : "1053"
         let options =
-            localhost == nil
-            ? ""
-            : HostDNSResolver.localhostOptionsRegex.replacingOccurrences(
-                of: #"\((.*?)\)"#, with: localhost!.description, options: .regularExpression)
+            localhost.map {
```
jglogan, line 54: summary: remove force-unwrap

Issue #1277, opened 2026-02-26
[Bug]: `container run -it <invalid executable>` hangs


JaewonHur:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

`container run -it --rm debian:bookworm dsfsdfsf`

`container exec -it <name> foobarbaz` hang also.

### Current behavior

It hangs the terminal.

### Expected behavior

It should not hang.

### Environment

```markdown
- OS: 
- Xcode: 
- Container:
```

### Relevant log output

```shell
2026-02-26 15:33:52.688216-0800 0xd3814    Error       0x0                  13571  0    container-apiserver: [com.apple.container:APIServer] route handler threw an error [error=internalError: "failed to start process xxx in container xxx" (cause: "internalError: "failed to start process (cause: "internal error (13): startProcess: failed to start process: internalError: "vmexec error: internalError: "failed to find target executable sdfdsfsdf""")"")] [route=containerStartProcess]
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1278, opened 2026-02-27
Fix container start hang on invalid executable

JaewonHur:
This PR fixes container hang when starting an invalid executable (#1277).

Current `ProcessIO.handleProcess` has races between `process.wait()` and `process.start()`. In `process.wait`, `SandboxService.wait` registers and block on a continuation so that later exit of that `LinuxContainer` can resume it.

Even if `SandboxService.wait` checks the container state, there can be a race so that `SandboxService.startInitProcess` fails earlier with no continuation to resume, then `SandboxService.wait` registers the continuation, which forever hang the thread.

The root cause is because `SandboxService.waiters` is not updated atomically, so this PR adds mutex to protect updating `SandboxService.waiters` after checking it's in the correct state.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Starting a container with invalid executable hangs the terminal.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Run/TestCLIRunLifecycle.swift
+++ b/Tests/CLITests/Subcommands/Run/TestCLIRunLifecycle.swift
@@ -112,4 +112,15 @@ class TestCLIRunLifecycle: CLITest {
         let status = try getContainerStatus(name)
         #expect(status == "stopped")
     }
+
+    @Test func testRunInvalidExcutable() async throws {
+        let name = getTestName()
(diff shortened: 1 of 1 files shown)


dcantah:
This feels like a design/implementation failure on `wait`'s part. Is there any way we can fix that instead?

dcantah:
Oh never mind this PR is now entirely different from my first viewing LOL

Issue #913, opened 2025-11-21
[Request]: Consistent and useful multi-platform semantics for `pull`, `push`, `save`, `load`, `build`.


jglogan:
### Feature or enhancement request details

We've had #455 open to address #83 for quite a while now. Before we merge it (and celebrate that Alpine doesn't take a minute to pull) we wanted to make sure that we settle on the semantics for the multi-platform UX for all the image store operations.

This issue will be the umbrella for the whole of the work. We can close it once everything's merged / closed:

- [ ] #83
- [ ] Consistent, useful multi-platform semantics for `image push`
- [ ] Consistent, useful multi-platform semantics for `image save`
- [ ] Consistent, useful multi-platform semantics for `image load`
- [ ] Consistent, useful multi-platform semantics for `build`
- [ ] #874 (we can close and include in the `image save` issue once we've defined what we want)
- [ ] #712 (we can close and include in the `image save` issue once we've defined what we want)
- [x] #1252 (supports amd64-only development workflow)
- [ ] Complete a how-to.md section that summarizes how multi-platform works

Let's discuss alternatives here. We're taking "consistent" means something that's simple and easy to reason about while avoiding foolish consistency, and "useful" to mean something that avoids surprises and works well for the most common use cases.

We suppose that the most common use case is a development loop using native (arm64) containers.

However, we should be able to support an amd64-only development loop. You would use this if you're building and pushing your cloud deployment images on your local system (instead of using a CI system that builds/pushes amd64 or amd64+arm64), or if you really need to validate amd64 operation locally full time.

#### `image pull` and `image push`

Pull and push semantics are informed by #83:

- Without any options, the commands use the native platform (`linux/arm64`)
- The `--platform` option selects a single, specific platform
- The `--all-platforms` option selects all platforms

#### `image save` and `image load`

We have two options:

- Identical semantics as `image pull` and `image push`
- Default to saving and loading all platforms, instead of native

The former is definitely more consistent. Is the latter more useful?

- Time to read/write multi-platform archives is less of an issue.
- It may make more sense to write all platforms by default because if a user built a multi-platform image, they've indicated an intention that maybe we shouldn't undo on a `save`.

#### amd64 only development

For the amd64-only development loop, the proposal is to use an environment variable `CONTAINER_DEFAULT_PLATFORM` that functions similarly to `DOCKER_DEFAULT_PLATFORM` which if defined, overrides the native platform as the default for `pull`, `push`, `build`, `create`, and `run`.

[...]

PR #1286, opened 2026-03-03
Add fallback to CONTAINER_DEFAULT_PLATFORM for image architecture

### LINKED ISSUE #1252: [Request]: Choose platform with CONTAINER_DEFAULT_PLATFORM and DOCKER_DEFAULT_PLATFORM.

See #913 for the broader scope.

`container` commands that support `--platform` should check environment variables for a platform setting when `--platform` isn't explicitly provided.

`CONTAINER_DEFAULT_PLATFORM` should be checked first, followed by `DOCKER_DEFAULT_PLATFORM`.

### Code of Conduct

Willjianger9:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #1252 (part of #913).

Commands that accept `--platform` currently require it to be passed explicitly every time. Users working consistently with a non-native platform (e.g. `linux/amd64` on Apple Silicon) have no way to set a default, unlike Docker's `DOCKER_DEFAULT_PLATFORM`.

This adds support for the `CONTAINER_DEFAULT_PLATFORM` environment variable as a fallback when `--platform` is not explicitly provided.

## Changes

- **New file `DefaultPlatform.swift`** — centralized logic for reading, validating, and resolving the environment variable with clear precedence rules.
- **Wired into all commands with `--platform`:** `run`, `create`, `image pull`, `image push`, `image save`, and `build`.
- **Precedence:** explicit `--platform` flag → `--os`/`--arch` flags → `CONTAINER_DEFAULT_PLATFORM` → native host architecture.
- **User feedback:** prints a notice to stderr when the environment variable is active (e.g. `Notice: Using platform "linux/amd64" from CONTAINER_DEFAULT_PLATFORM`).
- **Help text:** all `--platform` options now show `[environment: CONTAINER_DEFAULT_PLATFORM]`.
- **Error handling:** invalid values in the environment variable produce a clear error message naming the variable.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

20 unit tests added in `DefaultPlatformTests.swift` covering:
- Environment variable parsing (valid platforms, empty, missing, invalid)
- Precedence for `resolve()` (pull/push/save): explicit flags override env var
- Precedence for `resolveWithDefaults()` (run/create): env var overrides os/arch defaults
- Invalid env var is skipped when explicit `--platform` is provided
- All 134 pre-existing unit tests pass with no regressions

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -106,7 +106,7 @@ jobs:
 
       - name: Upload logs if present
         if: always()
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: container-test-logs
           path: container-logs.tar.gz
@@ -114,14 +114,14 @@ jobs:
           if-no-files-found: ignore
 
       - name: Save documentation artifact
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: api-docs
           path: "./_site.tgz"
           retention-days: 14
 
       - name: Save package artifacts
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: container-package
           path: ${{ github.workspace }}/outputs
@@ -138,7 +138,7 @@ jobs:
         uses: actions/configure-pages@v5
 
       - name: Download a single artifact
-        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
+        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
         with:
           name: api-docs
 
--- a/.github/workflows/pr-label-analysis.yml
+++ b/.github/workflows/pr-label-analysis.yml
@@ -21,7 +21,7 @@ jobs:
           echo "${PR_NUMBER}" > ./pr-metadata/pr-number.txt
       
       - name: Upload PR metadata as artifact
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: pr-metadata-${{ github.event.pull_request.number }}
           path: pr-metadata/
--- a/.github/workflows/pr-label-apply.yml
+++ b/.github/workflows/pr-label-apply.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Download PR metadata artifact
-        uses: actions/download-artifact@37930b1c2abaa49bbe596cd826c3c89aef350131 # v7
+        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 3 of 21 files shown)


jglogan:
@Willjianger9 Thanks for the contribution! Could you make sure that your commit is signed and verified? See https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests for details.

Willjianger9:
Thanks for bringing that to my attention! I've signed the commit and force-pushed so it should now show as verified. Please let me know if there's anything else that needs to be addressed 😊

Sources/Services/ContainerAPIService/Client/DefaultPlatform.swift
```diff
@@ -0,0 +1,101 @@
...
+                message: "invalid platform \"\(value)\" in \(environmentVariable) environment variable",
+                cause: error
+            )
+        }
+    }
+
+    /// Prints a notice to stderr indicating that the environment variable is being used.
+    public static func printNotice(_ platform: ContainerizationOCI.Platform) {
+        let message = "Notice: Using platform \"\(platform.description)\" from \(environmentVariable)\n"
+        FileHandle.standardError.write(Data(message.utf8))
```
jglogan: A couple of things here:

- For the CLI we manage stderr output through a Logger, and use structured logging. Furthermore, if it's something other than `container` that's using the APIs, stderr output may not be desired. What do you think about the following?
  - Make `printNotice` private, and add a `log: Logger` parameter and log the result at warn level. Have a look at the logging statements in ContainerCommand for structured logging examples. The log message should contain no interpolated values.
  - For the other functions, add a `log: Logger?` parameter and move the `printNotice` invocations in `resolve` and `resolveWithDefaults` into `fromEnvironment`.
- Add docc for the public func parameters and return values.

Willjianger9: Thanks for the feedback! I do agree that using a Logger makes the API much more flexible for non-CLI clients and cleans up the stderr output, which is something I overlooked in my changes. I'll get started on updating the logging structure, adjusting the parameters, and adding the missing docc comments

Ronitsabhaya75:
@Willjianger9 @jglogan can you help me understand here:

I noticed that `BuildCommand` manually calls `fromEnvironment()` inline, whereas the other commands (`ImagePull`/`Push`/`Save`) delgating to `DefaultPlatform.resolve()`. Should `BuildCommand` also use a centralized helper so the precedence logic doesn't drift out of sync?

`resolve()` currently uses the raw `Platform(from:)` OCI initializer, while `resolveWithDefaults()` uses `Parser.platform(from:)`. Since the original code used `Parser.platform()`, could this cause any subtle behavioral changes?

I saw some tests cases were missing
   * A test for a platform with a variant (like `linux/arm/v7`) via the environment variable. 
   * A test for `resolve()` where an invalid env var is provided but `--platform` is explicit.

Willjianger9:
@Ronitsabhaya75 Thanks for the feedback!

Here is my thinking on these (and @jglogan, feel free to jump in if I missed any context):
- BuildCommand vs. others: BuildCommand needs to produce a `Set<Platform>` from an array of strings, while the other commands only need a single `Platform?`, which is why it can't use `resolve()` directly. However, it does use the centralized `DefaultPlatform.fromEnvironment()` to read the env var, so the validation logic won't drift. The precedence is still  `explicit --platform > CONTAINER_DEFAULT_PLATFORM > --os/--arch`
- Parser logic: `Parser.platform(from:)` is just a thin wrapper that calls try `.init(from: platform)`, so it’s functionally identical to using `Platform(from:)`. There shouldn't be any subtle behavioral changes, and I just kept them separate to preserve the original behavior of the specific callsites (pull/push/save vs. run/create)

I totally agree on the tests, I have gone ahead and added two missing test cases

Ronitsabhaya75:
@Willjianger9 Thanks for the explanation!

That all makes sense.

You're completely right about `Set<Platform>` vs `Platform?`. Since it still goes through `fromEnvironment()` to parse the env var, the validation logic is safely centralized.

Thanks for confirming that `Parser.platform(from:)` is just a thin wrapper around the OCI initializer. Preserving the original behavior of the local call sites is definitely the safest move here.

jglogan:
> * BuildCommand vs. others: BuildCommand needs to produce a `Set<Platform>` from an array of strings, while the other commands only need a single `Platform?`

this.

jglogan:
@Willjianger9 Been hitting consistent CI failures for this PR, pretty sure it's not your changes.

I'm going to merge main into the PR to see if the most recent test changes will help. Apologies for the delay on this.

PR #1289, opened 2026-03-04
Fix race condition on SandboxService.waiters

### LINKED ISSUE #1277: [Bug]: `container run -it <invalid executable>` hangs

JaewonHur:
This PR fixes #1277.

`SandboxService.waiters` had a consistency issue (not exactly race).
`SandboxService.wait` XPC can be executed on arbitrary `id`, and it will hang forever if no other handler resumes it. Without knowing this internal, the high level entity can run into this issue, and deadlock.

This PR simplifies the mental model: **`SandboxService.waiters[id]: ExitWaiter(continuations, exitCode)` can only be in three states: i) non-existing, ii) existing with nil `exitCode`, and iii) existing with concrete `exitCode`.**

**If it is non-existing, no handler has been registered to resume it later. If existing with nil `exitCode`, It is guaranteed the registered `continuations` will be resumed later with a concrete `exitCode`. Finally, if already a concrete `exitCode`, a handler has been registered, and already resumed (with that `exitCode`).**

Thus, `SandboxService.wait` should return immediately if `waiters[id]` is non-existing or existing with a concrete `exitCode` (as no handler will resume it later). It should only block when `waiters[id]` is existing with nil `exitCode` as it is guaranteed to be resumed later. By doing so, we can guarantee there is no deadlock at all.

For that this PR does followings:
1. Introduce `ExitMonitor` class to updates `continuations` and `exitCode` all together atomically. Initially, `state` variable saved the `exitCode`, but it cannot be tied with `continuations` as they are protected by different primitives (i.e., lock and actor).
2. Gather `waiters` related operations into a single actor method, guaranteeing those are performed atomically under actor protection---i.e., we actually don't need Mutex here.
3. Ensure initialized `waiters` are released (i.e., resumed) later (under any possible circumstances).
4. Move `process.wait` after `process.start` in `io.handleProcess` to run `SandboxService.wait` only after the `waiters[id]` is initialized.

By doing fourth step, we can guarantee `SandboxService.wait` can meet only one of two following `ExitMonitor` state: i) existing with nil `exitCode`, or ii) existing with concrete `exitCode` (in case the process exited too early). In both cases, `exitCode` is preserved and returned.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
+++ b/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
@@ -44,7 +44,7 @@ public actor SandboxService {
     private var container: ContainerInfo?
     private let monitor: ExitMonitor
     private let eventLoopGroup: any EventLoopGroup
-    private var waiters: [String: [CheckedContinuation<ExitStatus, Never>]] = [:]
+    private var waiters: [String: ExitWaiter] = [:]
     private let lock: AsyncLock = AsyncLock()
     private let log: Logging.Logger
     private var state: State = .created
@@ -53,6 +53,27 @@ public actor SandboxService {
 
     private static let sshAuthSocketGuestPath = "/run/host-services/ssh-auth.sock"
     private static let sshAuthSocketEnvVar = "SSH_AUTH_SOCK"
+
+    class ExitWaiter {
+        public var exitCode: Int32? = nil
+        public var continuations: [CheckedContinuation<ExitStatus, Never>] = []
+
+        public func register(_ cc: CheckedContinuation<ExitStatus, Never>) {
+            continuations.append(cc)
+        }
+
+        public func doExit(code: Int32) {
+            for cc in continuations {
+                cc.resume(returning: ExitStatus(exitCode: code))
+            }
+
+            exitCode = code
+        }
+
+        public func exited() -> Bool {
+            exitCode != nil
+        }
+    }
 
     private static func sshAuthSocketHostUrl(config: ContainerConfiguration) -> URL? {
         if config.ssh, let sshSocket = Foundation.ProcessInfo.processInfo.environment[Self.sshAuthSocketEnvVar] {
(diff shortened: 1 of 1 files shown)


Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -598,39 +619,11 @@ public actor SandboxService {
...
-            return reply
-        }
-
         let exitStatus = await withCheckedContinuation { cc in
             // Is this safe since we are in an actor? :(
-            if !self.addWaiter(id: id, cont: cc) {
-                cc.resume(returning: ExitStatus(exitCode: -1))
+            let (added, exitCode) = self.addWaiter(id: id, cont: cc)
+            if !added {
+                cc.resume(returning: ExitStatus(exitCode: exitCode ?? -1))
```
dcantah, line 626: Neat idea

PR #1290, opened 2026-03-04
Reap auto remove containers if there are running on system start

JaewonHur:
This PR reaps already running containers with auto remove flag on container system start.

Container API server shutdown ideally should stop every running container, and reap the containers with auto remove flag. However, crash or unexpected exit of API server leaves the containers running. In such case, next system start should stop running containers (i.e., launchd services) and reap the containers with auto remove flag.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Preserve the system to be in a consistent state after unexpected failure.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "7258f8fc7fdb646c316258bdf7b3a7d7b2cf9f75f28521c25b6914d426d595c0",
+  "originHash" : "040dd2e2c8649defb737f900e4032d270935f0af91ebf4a87d64391ecd4ea40b",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -15,8 +15,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/containerization.git",
       "state" : {
-        "revision" : "636eef0eff00e451de6d5d426e6a6785b90b44e2",
-        "version" : "0.26.5"
+        "revision" : "d3ff56e5dd93a4573ce7950f3428710bbe467f9c",
+        "version" : "0.27.0"
       }
     },
     {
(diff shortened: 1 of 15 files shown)


Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -99,7 +99,27 @@ public actor ContainersService {
...
             do {
-                let config = try Self.getContainerConfiguration(at: dir)
+                let (config, options) = try Self.getContainerConfiguration(at: dir)
+                if options?.autoRemove ?? false {
+                    let label = Self.fullLaunchdServiceLabel(
+                        runtimeName: config.runtimeHandler,
+                        instanceId: config.id)
+
+                    var status: Int32 = -1
+                    try ServiceManager.deregister(fullServiceLabel: label, status: &status)
```
realrajaryan:
```suggestion
                    try? ServiceManager.deregister(fullServiceLabel: label, status: &status)
```

Sources/ContainerPlugin/ServiceManager.swift
```diff
@@ -44,6 +44,10 @@ public struct ServiceManager {
         _ = try runLaunchctlCommand(args: ["bootout", label])
     }
```
saehejkang, line 46: nit: maybe a comment explaining this function and why there needs to be an overload?

JaewonHur: Left a short comment. I made the overload as I didn't want to change the existing function signature (because most of the time they don't need return status). Thanks :)

JaewonHur: Also, it would be good to add test case for it once we have test suite that runs across system lifecycle.

PR #1273, opened 2026-02-26
Fix docker ignore

JaewonHur:
This PR fixes `container build` to use docker specific ignore (#1169).
This PR relies on apple/container-builder-shim#68.

When docker specific ignore file is detected, it creates a hidden docker directory (i.e., `.hidden-docker-dir`) under the build context, and places Dockerfile and ignore file under there. This hidden directory is passed to the builder shim and used by buildkit daemon to correctly figure out the docker specific ignore file.

Docker specific ignore file is detected regardless of whether i) they are under nested directory, ii) outside build context, iii) using custom names. If docker specific ignore is not provided, hidden directory is also not created, and buildkit daemon will use the docker ignore file at build context root as usual.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Docker specific ignore works.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

PR #1300, opened 2026-03-07
Add support for build secrets

percontation:
docker-compatible `--secret id=key,...` arg for `container build`, that works with Dockerfiles with `RUN --mount=type=secret`

Requires https://github.com/apple/container-builder-shim/pull/69

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Adds support for Dockerfiles that use build secrets (e.g. `RUN --mount=type=secret ...`)

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Tests/CLITests/Subcommands/Build/CLIBuilderTest.swift
+++ b/Tests/CLITests/Subcommands/Build/CLIBuilderTest.swift
@@ -189,13 +189,15 @@ extension TestCLIBuildBase {
             let tempFile: URL = try createTempFile(suffix: " _f,i=l.e+ ", contents: testData)
             let tempFile2: URL = try createTempFile(suffix: "file2", contents: Data())
             let imageName: String = "registry.local/secrets:\(UUID().uuidString)"
-            try self.build(tag: imageName, tempDir: tempDir, otherArgs: [
-                "--secret", "id=ENV1",
-                "--secret", "id=env2,env=ENV_VAR",
-                "--secret", "id=env3,env=env3",
-                "--secret", "id=file,src="+tempFile.path,
-                "--secret", "id=empty,src="+tempFile2.path,
-            ])
+            try self.build(
+                tag: imageName, tempDir: tempDir,
+                otherArgs: [
+                    "--secret", "id=ENV1",
+                    "--secret", "id=env2,env=ENV_VAR",
+                    "--secret", "id=env3,env=env3",
+                    "--secret", "id=file,src=" + tempFile.path,
+                    "--secret", "id=empty,src=" + tempFile2.path,
+                ])
             #expect(try self.inspectImage(imageName) == imageName, "expected to have successfully built \(imageName)")
         }
 


percontation:
I haven't done much Swift before, so watch out for any stupid mistakes I may have made.

One design caveat: I passed secrets by copying the build-args code, because they're functionally the same. But, systems are supposed to take care that the secret values aren't logged or stored anywhere, so HTTP headers are a much riskier way to send them than transporting them separately in a new BuildTransfer. I nevertheless used the headers to send the secrets because 1. it seemed easier 2. I didn't notice anything that would log or record these headers 3. it looked like the headers and BuildTransfers alike are just data sent through gRPC, so, not much practical difference right now.

Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -247,6 +250,32 @@ extension Application {
...
+                        throw ContainerizationError(.invalidArgument, message: "secret must start with id=<key> \(secret)")
+                    }
+                    let key = String(parts[0].dropFirst(3))
+                    guard !key.contains("=") else {
+                        throw ContainerizationError(.invalidArgument, message: "secret id cannot contain '=' \(key)")
+                    }
+                    let data: Data
+                    if parts.count == 1 || parts[1].hasPrefix("env=") {
+                        let env = parts.count == 1 ? key : String(parts[1].dropFirst(4))
+                        guard let ptr = getenv(env) else {
```
JaewonHur: Nit comment.
We can avoid using unsafe pointer and strlen by
```
guard let value = ProcessInfo.processInfo.environmentenv] else {
    throw ContainerizationError(.invalidArgument, message: "secret env var doesn't exist \(env)")
}
data = Data(value.utf8)
```

Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -247,6 +250,32 @@ extension Application {
                     buildFileData = try Data(contentsOf: URL(filePath: buildFilePath))
                 }
 
+                var secretsData: [String: Data] = [:]
+                for secret in self.secret {
```
JaewonHur: Could we move this part under `validate` function? #1273 as an example.

Sources/ContainerCommands/BuildCommand.swift
percontation: I couldn't find documentation on how `ProcessInfo.processInfo.environment` would behave on environment variables containing non-UTF8 data, so I went with C getenv. Is there a better Swift way of getting env vars as Data without going through UTF8?

Alternatively, just using `ProcessInfo.processInfo.environment` and accepting whatever limitations that places on non-UTF8 env var values could just be fine, given that env vars needing to be `\x00`-free already makes it iffy to use env vars on non-text data secrets. Would this be preferred here?

Sources/ContainerCommands/BuildCommand.swift
percontation: Moved it. Let me know if I did what you were expecting.

Sources/ContainerCommands/BuildCommand.swift
JaewonHur: Oh It makes sense. I agree it'd better to stay there. Could you leave a comment then?

JaewonHur:
Hi @percontation, could you resolve conflicts one last time? and we can merge it!

JaewonHur (approved):

JaewonHur (approved):

PR #1322, opened 2026-03-17
Pass root directory URL as env variable so that plugins can use it

JaewonHur:
This PR adds a `PluginStateRoot` class so that a plugin can figure out its root URL where the plugin specific data should be stored.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


JaewonHur:
I guess the reason I pass it as env variable is that we get `pluginStateRoot` as an argument in `registerWithLaunchd` (Sources/ContainerPlugin/PluginLoader.swift:L212).

I was not sure it should be always the default path.

jglogan:
I think we only need environment variables for our base directories:

- The installation base for the application.
- The base application data directory.
- A base directory for logging, when writing file based logs (this is really a develop/debug capability, not good for production as the logs grow unbounded)

PR #1326, opened 2026-03-18
volumes: Allow "."

dcantah:
We should allow passing . for the src of a volume.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Parser.swift
+++ b/Sources/Services/ContainerAPIService/Client/Parser.swift
@@ -515,8 +515,8 @@ public struct Parser {
             let src = String(parts[0])
             let dst = String(parts[1])
 
-            // Check if it's a filesystem path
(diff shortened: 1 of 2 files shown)


Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -515,8 +515,8 @@ public struct Parser {
             let src = String(parts[0])
             let dst = String(parts[1])
 
-            // Check if it's a filesystem path
-            guard src.contains("/") else {
+            // Check if it's a filesystem path (absolute, or relative like ".", "..", "./foo", "../foo")
+            guard src.contains("/") || src == "." || src == ".." || src.hasPrefix("./") || src.hasPrefix("../") else {
```
jglogan: Doesn't case 1 short-circuit cases 4 and 5 all the time?

PR #1335, opened 2026-03-20
Add --all-tags flag to image push

### LINKED ISSUE #1334: [Request]: Add `-a/--all-tags` for `container image push`

Add the `-a/--all-tags` flag to `container image push` so that `container image push --all-tags registry/myimage` pushes all locally tagged versions of that image.

### Code of Conduct

realrajaryan:
- Closes #1334.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
Adds `--all-tags` (`-a`) support to `container image push`, which pushes all locally tagged versions of a repository in parallel with bounded concurrency (`--max-concurrent-uploads`, default 3). It continues pushing remaining tags if one fails.

Usage: `container image push --all-tags myregistry.io/myimage`
Example:
```
❯ container image push --all-tags ghcr.io/sometest/alltagstest
The push refers to repository [ghcr.io/sometest/alltagstest]
v1: digest: sha256:28a33e0598da2b341d05e53803c5982377039baedfc8a82a4910d5acc4788c99 size: 5 KB
v2: digest: sha256:28a33e0598da2b341d05e53803c5982377039baedfc8a82a4910d5acc4788c99 size: 5 KB
latest: digest: sha256:28a33e0598da2b341d05e53803c5982377039baedfc8a82a4910d5acc4788c99 size: 5 KB
```

Note: `-a` on `push` now means `--all-tags`, not `--arch`. `--arch` remains available as a long-form flag.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Image/ImagePush.swift
+++ b/Sources/ContainerCommands/Image/ImagePush.swift
@@ -39,7 +39,7 @@ extension Application {
         var imageUploadFlags: Flags.ImageUpload
 
         @Option(
-            name: .long,
+            name: .shortAndLong,
             help: "Limit the push to the specified architecture"
         )
         var arch: String?
@@ -52,7 +52,7 @@ extension Application {
         @Option(help: "Limit the push to the specified platform (format: os/arch[/variant], takes precedence over --os and --arch) [environment: CONTAINER_DEFAULT_PLATFORM]")
         var platform: String?
 
-        @Flag(name: .shortAndLong, help: "Push all tags of an image")
+        @Flag(name: .long, help: "Push all tags of an image")
         var allTags: Bool = false
 
         @OptionGroup
@@ -66,10 +66,10 @@ extension Application {
             if allTags {
                 let ref = try Reference.parse(reference)
                 if ref.tag != nil {
-                    throw ContainerizationError(.invalidArgument, message: "tag can't be used with --all-tags/-a")
+                    throw ContainerizationError(.invalidArgument, message: "tag can't be used with --all-tags")
                 }
                 if ref.digest != nil {
-                    throw ContainerizationError(.invalidArgument, message: "digest can't be used with --all-tags/-a")
+                    throw ContainerizationError(.invalidArgument, message: "digest can't be used with --all-tags")
                 }
             }
         }
@@ -115,29 +115,7 @@ extension Application {
                 log.warning("--platform/--arch/--os with --all-tags filters each tag push to the specified platform; tags without matching manifests may fail")
             }
 
-            // Enumerate matching tags for display before pushing.
-            let allImages = try await ClientImage.list()
             let normalized = try ClientImage.normalizeReference(reference)
-            let parsedRef = try Reference.parse(normalized)
-            let repoName: String
-            if let resolved = parsedRef.resolvedDomain {
-                repoName = "\(resolved)/\(parsedRef.path)"
-            } else {
-                repoName = parsedRef.name
-            }
-
-            let matchingTags = allImages.filter { img in
-                guard !Utility.isInfraImage(name: img.reference) else { return false }
-                guard let ref = try? Reference.parse(img.reference) else { return false }
-                let resolvedName: String
-                if let resolved = ref.resolvedDomain {
-                    resolvedName = "\(resolved)/\(ref.path)"
-                } else {
-                    resolvedName = ref.name
-                }
-                return resolvedName == repoName
-            }
-
             let displayRepo = try ClientImage.denormalizeReference(normalized)
             let displayName = try Reference.parse(displayRepo).name
             print("The push refers to repository [\(displayName)]")
@@ -147,7 +125,7 @@ extension Application {
             case .none: progressConfig = try ProgressConfig(disableProgressUpdates: true)
             case .ansi:
                 progressConfig = try ProgressConfig(
-                    description: "Pushing \(matchingTags.count) tags",
+                    description: "Pushing tags",
                     showPercent: false,
                     showItems: false,
                     showSpeed: false,
@@ -160,13 +138,13 @@ extension Application {
                 progress.finish()
             }
             progress.start()
-            try await ClientImage.pushAllTags(
+            let pushed = try await ClientImage.pushAllTags(
                 reference: reference, platform: platform, scheme: scheme,
                 maxConcurrentUploads: imageUploadFlags.maxConcurrentUploads, progressUpdate: progress.handler)
             progress.finish()
 
             let formatter = ByteCountFormatter()
-            for img in matchingTags {
+            for img in pushed {
                 let tag = (try? Reference.parse(img.reference))?.tag ?? "<none>"
                 let size = formatter.string(fromByteCount: img.descriptor.size)
                 print("\(tag): digest: \(img.descriptor.digest) size: \(size)")
--- a/Sources/ContainerCommands/Image/ImagePush.swift
+++ b/Sources/ContainerCommands/Image/ImagePush.swift
@@ -17,7 +17,9 @@
 import ArgumentParser
 import ContainerAPIClient
 import Containerization
+import ContainerizationError
 import ContainerizationOCI
+import Foundation
 import TerminalProgress
 
 extension Application {
@@ -33,8 +35,11 @@ extension Application {
         @OptionGroup
         var progressFlags: Flags.Progress
 
+        @OptionGroup
+        var imageUploadFlags: Flags.ImageUpload
+
         @Option(
-            name: .shortAndLong,
+            name: .long,
             help: "Limit the push to the specified architecture"
         )
         var arch: String?
@@ -47,6 +52,9 @@ extension Application {
         @Option(help: "Limit the push to the specified platform (format: os/arch[/variant], takes precedence over --os and --arch) [environment: CONTAINER_DEFAULT_PLATFORM]")
         var platform: String?
 
+        @Flag(name: .shortAndLong, help: "Push all tags of an image")
+        var allTags: Bool = false
+
         @OptionGroup
         public var logOptions: Flags.Logging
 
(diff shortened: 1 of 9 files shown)


Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -283,6 +283,48 @@ extension ClientImage {
...
+        // Normalize the reference, then extract the repository name without the tag.
+        let normalized = try Self.normalizeReference(reference)
+        let parsedRef = try Reference.parse(normalized)
+
+        let repositoryName: String
+        if let resolved = parsedRef.resolvedDomain {
+            repositoryName = "\(resolved)/\(parsedRef.path)"
+        } else {
+            repositoryName = parsedRef.name
+        }
```
manojmahapatra: Suggestion: Looks like this is being repeated at couple of places. I _think_ moving this to a package level function inside Utility would make sense.

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -136,6 +136,92 @@ public actor ImagesService {
         }
     }
 
+    public func pushAllTags(repositoryName: String, platform: Platform?, insecure: Bool, maxConcurrentUploads: Int, progressUpdate: ProgressUpdateHandler?) async throws {
```
manojmahapatra: Question: Is this API need to be public? Looks like either package or internal level should suffice.

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -136,6 +136,92 @@ public actor ImagesService {
...
+            guard !Utility.isInfraImage(name: image.reference) else { return false }
+            guard let ref = try? Reference.parse(image.reference) else { return false }
+            let resolvedName: String
+            if let resolved = ref.resolvedDomain {
+                resolvedName = "\(resolved)/\(ref.path)"
+            } else {
+                resolvedName = ref.name
+            }
+            return resolvedName == repositoryName
+        }
```
manojmahapatra: Suggestion: This would be a good candidate which can be moved to Utility.

docs/command-reference.md
```diff
@@ -522,18 +522,20 @@ Pushes an image to a registry. The flags mirror those for `image pull` with the
...
+*   `<reference>`: Image reference to push. When using `--all-tags`, this should be a repository name without a tag.
 
 **Options**
 
 *   `--scheme <scheme>`: Scheme to use when connecting to the container registry. One of (http, https, auto) (default: auto)
 *   `--progress <type>`: Progress type (format: none|ansi) (default: ansi)
-*   `-a, --arch <arch>`: Limit the push to the specified architecture
+*   `-a, --all-tags`: Push all tags of an image
+*   `--max-concurrent-uploads <count>`: Maximum number of concurrent uploads with --all-tags (default: 3)
+*   `--arch <arch>`: Limit the push to the specified architecture
```
katiewasnothere: This would now make the image push command inconsistent with image pull, the kernel commands, and the container commands.

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -78,8 +106,71 @@ extension Application {
...
+            try await image.push(platform: platform, scheme: scheme, progressUpdate: progress.handler)
+            progress.finish()
+        }
+
+        private func pushAllTags(platform: Platform?, scheme: RequestScheme) async throws {
+            if self.platform != nil || arch != nil || os != nil {
+                log.warning("--platform/--arch/--os with --all-tags filters each tag push to the specified platform; tags without matching manifests may fail")
+            }
+
+            // Enumerate matching tags for display before pushing.
```
katiewasnothere: I'd rather `pushAllTags` return the set of images it pushed

Sources/Services/ContainerImagesService/Server/ImagesService.swift
realrajaryan: It's to be consistent with the other methods here, they're all public because ImagesService is consumed across module boundaries by ImagesHelper

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -78,8 +106,49 @@ extension Application {
...
+
+        private func pushAllTags(platform: Platform?, scheme: RequestScheme) async throws {
+            if self.platform != nil || arch != nil || os != nil {
+                log.warning("--platform/--arch/--os with --all-tags filters each tag push to the specified platform; tags without matching manifests may fail")
+            }
+
+            let normalized = try ClientImage.normalizeReference(reference)
+            let displayRepo = try ClientImage.denormalizeReference(normalized)
+            let displayName = try Reference.parse(displayRepo).name
+            print("The push refers to repository [\(displayName)]")
```
katiewasnothere: I don't know if this is super necessary since we get back a list of images we pushed

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -136,6 +136,96 @@ public actor ImagesService {
         }
     }
 
+    public func pushAllTags(repositoryName: String, platform: Platform?, insecure: Bool, maxConcurrentUploads: Int, progressUpdate: ProgressUpdateHandler?) async throws
```
katiewasnothere, line 144: I'm wondering if we should instead move the logic to `ImageStore.ExportOperation` to be similar to how `ImageStore.pull` works. What do you think?

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -136,6 +136,96 @@ public actor ImagesService {
...
+                resolvedName = ref.name
+            }
+            return resolvedName == repositoryName
+        }
+
+        guard !matchingImages.isEmpty else {
+            throw ContainerizationError(.notFound, message: "no tags found for repository \(repositoryName)")
+        }
+
+        let maxConcurrent = maxConcurrentUploads > 0 ? maxConcurrentUploads : 3
```
manojmahapatra: Suggestion: This silent fallback behavior is different than what's in ClientImage (which has a `guard`). I would keep the same behavior to be consistent.

Sources/ContainerCommands/Image/ImagePush.swift
realrajaryan: only downside is the header would print after the push instead of before

Sources/Services/ContainerImagesService/Server/ImagesService.swift
realrajaryan: `ExportOperation` handles single-image blob export, so idt multi-tag orchestration naturally fits there, though I think a shared `RegistryClient` optimization in `ImageStore` would be a good follow-up

realrajaryan: https://github.com/apple/containerization/pull/661

katiewasnothere: @realrajaryan Does this PR need to be updated now that https://github.com/apple/containerization/pull/661 is merged?

realrajaryan: Yes, I'll try update this today. Got caught up

Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
```diff
@@ -328,6 +328,49 @@ class TestCLIImagesCommand: CLITest {
...
+            "--all-tags",
+            "alpine:latest",
+        ])
+
+        #expect(status != 0, "Expected --all-tags with a tag to fail")
+        #expect(
+            error.contains("tag can't be used with --all-tags"),
+            "Expected tag validation error message in output")
+    }
+
```
katiewasnothere, line 344: Maybe doesn't need to be done in this PR, but we should probably have a positive test for this by running a local registry and testing pushing to that.

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -76,9 +104,42 @@ extension Application {
...
+
+        private func pushAllTags(platform: Platform?, scheme: RequestScheme, containerSystemConfig: ContainerSystemConfig) async throws {
+            if self.platform != nil || arch != nil || os != nil {
+                log.warning("--platform/--arch/--os with --all-tags filters each tag push to the specified platform; tags without matching manifests may fail")
+            }
+
+            let normalized = try ClientImage.normalizeReference(reference, containerSystemConfig: containerSystemConfig)
+            let displayRepo = try ClientImage.denormalizeReference(normalized, containerSystemConfig: containerSystemConfig)
+            let displayName = try Reference.parse(displayRepo).name
+            print("The push refers to repository [\(displayName)]")
```
katiewasnothere, line 120: Should this just be a log message? Or maybe the progress description could be `Pushing tags for repository \(displayName)`?

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -76,9 +104,42 @@ extension Application {
...
         }
+
+        private func pushAllTags(platform: Platform?, scheme: RequestScheme, containerSystemConfig: ContainerSystemConfig) async throws {
+            if self.platform != nil || arch != nil || os != nil {
+                log.warning("--platform/--arch/--os with --all-tags filters each tag push to the specified platform; tags without matching manifests may fail")
+            }
+
+            let normalized = try ClientImage.normalizeReference(reference, containerSystemConfig: containerSystemConfig)
+            let displayRepo = try ClientImage.denormalizeReference(normalized, containerSystemConfig: containerSystemConfig)
+            let displayName = try Reference.parse(displayRepo).name
```
katiewasnothere, line 119: Can we use `Utility.repositoryName(for: parsedRef)` instead of denormalizing and calling Reference.parse?

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -284,6 +284,49 @@ extension ClientImage {
         return image
     }
 
+    @discardableResult
+    public static func pushAllTags(
```
katiewasnothere, line 288: Could `pushSingle` also call this? The only difference is the use of maxConcurrentUploads and returning an array of [ClientImage], right?

PR #1343, opened 2026-03-24
Fix: Avoid staging Dockerfiles into read-only build contexts

mazdak:
## Type of Change
- [X] Bug fix

## Motivation and Context

While building our docker-compose-like compose system for container, we noticed an issue in the build phase. The current build path always creates `.com.apple.container.dockerfiles/` inside the context and writes a staged Dockerfile there.

That introduces an issue for:

  - read-only checkouts
  - mounted source trees
  - otherwise immutable build contexts

Those builds can fail before BuildKit starts, even when there is no .dockerignore to rewrite.  Here, we fix container build so it does not require write access to the build context unless it actually needs to synthesize staged Dockerfile artifacts.

### Fix

  - Add a small helper that only produces staged Dockerfile artifacts when an ignore file exists.
  - Fall back to the root .dockerignore if Dockerfile.dockerignore is absent.
  - Skip hidden-directory creation entirely when there is no ignore file to extend.

This preserves the existing staged-ignore behavior where it is needed, while avoiding unnecessary writes for simple read-only contexts.

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -122,6 +122,12 @@ let package = Package(
                 "ContainerBuild"
             ]
         ),
+        .testTarget(
+            name: "ContainerCommandsTests",
+            dependencies: [
+                "ContainerCommands"
+            ]
+        ),
         .executableTarget(
             name: "container-apiserver",
             dependencies: [
--- a/Sources/ContainerCommands/BuildCommand.swift
+++ b/Sources/ContainerCommands/BuildCommand.swift
@@ -30,6 +30,26 @@ extension Application {
     public struct BuildCommand: AsyncLoggableCommand {
         private static let hiddenDockerDir = ".com.apple.container.dockerfiles"
 
+        static func stagedDockerfileArtifacts(
+            buildFileData: Data,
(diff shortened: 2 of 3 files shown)


JaewonHur:
Hi @mazdak Thank you for reporting this issue!

Confirming `container build` doesn't work with read-only build context.
We have to avoid staging Dockerfile (and ignore) in the context.

I guess this solution might not be the best one as it couldn't support docker specific ignore when the build context is read only.

We might need to update fssync to handle Dockerfile and ignore file selectively.
Let me dig more!

Thanks!

JaewonHur:
@mazdak I just made some PRs for this issue.
Sorry to make a duplicate PR for the same issue. Would you mind if I put you as co-author?

mazdak:
> @mazdak I just made some PRs for this issue. Sorry to make a duplicate PR for the same issue. Would you mind if I put you as co-author?

No problem at all. I am happy to just see this resolved.

PR #1349, opened 2026-03-25
Fix using docker specific ignore with read-only build context

JaewonHur:
This PR resolves #1343.
This PR depends on apple/container-builder-shim#72.

Do not create staging directory under build context, but pass dockerignore file bytes to the container-builder-shim.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerCommands/BuildCommand.swift
saehejkang: This can be removed, as it is no longer being used

jglogan (approved):

PR #1355, opened 2026-03-27
stats: restore terminal when streaming is interrupted by a signal

### LINKED ISSUE #1351: [Bug]: container stats does not restore terminal upon exit
### I have done the following

### Steps to reproduce

1. Start `container stats` (streaming mode)
2. <kbd>Ctrl-C</kbd> to exit

### Current behavior

After `container stats` exits, the terminal is still in ["alternate" terminal screen](https://support.apple.com/guide/terminal/display-or-hide-the-alternate-screen-trmld1f46097/mac) state. (I've confirmed this using [Ghostty](https://ghostty.org/) Terminal Inspector feature.) The visible result of this in my terminal emulator is that the mouse scroll feature no longer works to navigate terminal scrollback.

In macOS Terminal.app, the mouse scroll feature does not seem affected, but you can verify that you are still in "alternate" screen by inspecting the "View" menu: “Show alternate screen” will be grayed out.

### Expected behavior

Upon exiting, even as a result of <kbd>Ctrl-C</kbd>, `container stats` should restore terminal properties to their original values, i.e. switch to the "primary" screen, restore cursor visibility, and restore mouse features.

My suspicion is that the `container` process exits before the `defer` block had a chance to run:

https://github.com/apple/container/blob/31c866aa0bb4a9477e27edef8ab33071028cf30d/Sources/ContainerCommands/Container/ContainerStats.swift#L49-L60

### Environment

```markdown
- OS: 26.2
- Xcode: 26.4
- Container: 0.10.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Otoru:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
When `container stats` runs in streaming mode, pressing Ctrl+C sends SIGINT directly to the process before Swift can run the `defer` block responsible for restoring the terminal. This leaves the alternate screen buffer active and the cursor hidden after exit.

Fix this by wrapping the streaming loop in a `withThrowingTaskGroup` that races the work against an `AsyncSignalHandler` for SIGINT and SIGTERM. When a signal arrives, the group is cancelled through normal Swift unwinding, allowing the `defer` to execute and restore the terminal correctly. This follows the same pattern used in `BuildCommand`.

`runStreaming`, `collectStats`, `printStatsTable`, and `clearScreen` are promoted to static methods since none of them read instance state, which is required to safely capture only the `containers` value in the task closure under Swift 6 strict concurrency.

Fixes #1351

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerStats.swift
+++ b/Sources/ContainerCommands/Container/ContainerStats.swift
@@ -19,6 +19,7 @@ import ContainerAPIClient
 import ContainerResource
 import ContainerizationError
 import ContainerizationExtras
+import ContainerizationOS
 import Foundation
 
 extension Application {
@@ -57,7 +58,24 @@ extension Application {
                     fflush(stdout)
                 }
 
-                try await runStreaming()
+                let containerIds = containers
+                try await withThrowingTaskGroup(of: Void.self) { group in
+                    defer { group.cancelAll() }
+                    group.addTask {
+                        let handler = AsyncSignalHandler.create(notify: [SIGINT, SIGTERM])
+                        for await _ in handler.signals {
+                            throw CancellationError()
+                        }
+                    }
+                    group.addTask { [containerIds] in
+                        try await Self.runStreaming(containerIds: containerIds)
+                    }
+                    do {
+                        try await group.next()
+                    } catch is CancellationError {
+                        // Normal exit on signal, defer will restore the terminal
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


Tests/ContainerCommandsTests/ContainerStatsTests.swift
```diff
@@ -0,0 +1,97 @@
+//===----------------------------------------------------------------------===//
```
dcantah: I'd prefer you open a separate change for this

dcantah:
@Otoru We require all commits be signed (gpg/ssh)

Otoru:
@dcantah, thanks for explaining. Fixed.

dcantah:
@Otoru Can you remove the unit tests? We can add separately

Otoru:
@dcantah of course

dcantah (approved):

PR #1358, opened 2026-03-27
build: extend signal handling scope to cover the unpack phase

### LINKED ISSUE #1354: [Bug]: Ctrl+C doesn't interrupt unpacking a built image
### I have done the following

### Steps to reproduce

1. Build an image
2. Press Ctrl+C when unpacking a built image

### Current behavior

```
% container build --no-cache
[+] Building 606.1s (36/36) FINISHED                                                                                          
...
⠴ Unpacking built image for platform linux/arm64 12% (104.7/869.7 MB, Zero KB/s) [42s]
```

### Expected behavior

```
^CError: interrupted: "exiting on signal 2"
```

### Environment

```markdown
- Container: 
../container/bin/container --version
container CLI version 5090093 (build: debug, commit: 5090093)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

Otoru:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #1354.

When `container build` completes the build phase, it enters an unpack phase that runs outside the `withThrowingTaskGroup` containing the signal handler. As a result, pressing Ctrl+C during unpacking has no effect — the SIGINT is never caught and the process keeps running until the unpack finishes.

Fix this by moving the unpack phase inside the existing build task, so both build and unpack run within the same task group that hosts the `AsyncSignalHandler`. The `Task.checkCancellation()` calls already present in the unpack loop will now fire correctly when a signal is received.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/BuildCommand.swift
+++ b/Sources/ContainerCommands/BuildCommand.swift
@@ -350,7 +350,8 @@ extension Application {
                         }
                         return results
                     }()
-                    group.addTask { [terminal, buildArg, secretsData, contextDir, hiddenDockerDir, label, noCache, target, quiet, cacheIn, cacheOut, pull, exports, imageNames, tempURL, log] in
+                    group.addTask {
+                        [terminal, buildArg, secretsData, contextDir, hiddenDockerDir, label, noCache, target, quiet, cacheIn, cacheOut, pull, exports, imageNames, tempURL, log] in
                         let config = Builder.BuildConfig(
                             buildID: buildID,
                             contentStore: RemoteContentStoreClient(),


Otoru:
The delay is expected — `image.unpack()` is an XPC call that doesn't respond to Swift task cancellation mid-stream. The fix ensures the process exits after the current chunk completes, rather than never exiting at all.

To make cancellation immediate there are two paths:

1. Add a cancel endpoint to the XPC unpack protocol and call it from `withTaskCancellationHandler` on the client — server-side changes required
2. Force exit via `Darwin.exit()` from `withTaskCancellationHandler` — no protocol changes, but skips cleanup

Happy to pursue either as a follow-up if that's the preferred direction.

dkovba:
Please run `make fmt` to fix PR checks.

Otoru:
@dkovba done

Issue #514, opened 2025-08-16
[Request]: Implement resource identifier validation consistently.


jglogan:
### Feature or enhancement request details

Resource identifiers are validated inconsistently (some on client, some on server) or not at all (volumes), and allow any length identifier.

We should validate always on the server, limit identifier lengths, and ensure validation patterns are consistent with existing norms.

| Resource | Current validation | Current pattern | Normative (according to Google) |
|-|-|-|-|
| container | client | `[a-zA-Z0-9][a-zA-Z0-9_.-]+` | `[A-Za-z0-9][A-Za-z0-9_-.]{0,62}` |
| image | server | Containerization handles imageref validation properly | lots of rules |
| network | none | N/A | `[a-z0-9][a-z0-9_-]{0,62}` |
| volume | server | `[A-Za-z0-9][A-Za-z0-9_.-]*` | `[a-z0-9][a-z0-9_-]{0,62}` |

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1360, opened 2026-03-27
Create ResourceLabels and use for ManagedResource, NetworkConfiguration.

### LINKED ISSUE #1359: [Request]: TECH DEBT - Custom type for resource labels.

As @Ronitsabhaya75 observed in #1123, our `ManagedResource` uses a `[String: String]` to represent metadata labels, which can admit data that's not valid.

Further, the only label validation that we were performing was in `NetworkConfiguration`.

I'm starting to put together a proposal for what a 1.0 API/SDK should encompass and will be using network resource management as a proving ground. I'll post that as a discussion topic within the next few days and then we can go back and forth between incremental changes and refining the network type until we've established a nice set of patterns that we can apply to the other resource types.

We need to keep individual PRs minimal so that they are readily reviewable. This first bit of work:

- Creates a new `ResourceLabels type`.
- Adds an `AppError` type that supports structured logging and delegates message presentation to the error receiver, and derives a `ResourceLabels.LabelError` subtype to represent label validation errors.
- Extracts the label validation code from `NetworkConfiguration` into this type (relocating unit tests as necessary).
- Revises all affected code paths.

### Code of Conduct

jglogan:
- Closes #1359.
- Part of #514.
- Create a ResourceLabels type and extract the label validation from NetworkConfiguration into the new type.
- Create a base AppError type that is compatible with structured logging and delegates message presentation to the error receiver.
- Define LabelError over AppError for label validation.
- Slightly reworks NetworkConfiguration entity migration code in NetworksService.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
See related issue.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Common/ManagedResource.swift
+++ b/Sources/ContainerResource/Common/ManagedResource.swift
@@ -32,7 +32,9 @@ public protocol ManagedResource: Identifiable, Sendable, Codable {
 
     /// Key-value properties for the resource. The user and system may both
     /// make use of labels to read and write annotations or other metadata.
-    /// A good practice is to use
+    /// A good practice when using labels for automation is to use reverse
+    /// domain name notation (example: `com.example.mytool.role`) for
+    /// label names.
     var labels: ResourceLabels { get }
 
     /// Generates a unique resource ID value.
@@ -52,7 +54,7 @@ extension ManagedResource {
     }
 }
 
-// FIXME: This moves to ManagedResource and/or a ResourceLabels typealias eventually.
 extension ResourceLabels {
+    /// Returns true if for a resource that the system automatically manages.
     public var isBuiltin: Bool { self.contains { $0 == ResourceLabelKeys.role && $1 == ResourceRoleValues.builtin } }
 }
--- a/Sources/ContainerCommands/Container/ContainerStart.swift
+++ b/Sources/ContainerCommands/Container/ContainerStart.swift
@@ -73,7 +73,7 @@ extension Application {
 
             for mount in container.configuration.mounts where mount.isVirtiofs {
                 if !FileManager.default.fileExists(atPath: mount.source) {
-                    throw ContainerizationError(.invalidState, message: "path '\(mount.source)' is not a directory")
+                    throw ContainerizationError(.invalidState, message: "mount source path '\(mount.source)' does not exist")
                 }
             }
 


Sources/ContainerResource/Common/ApplicationError.swift
```diff
@@ -0,0 +1,36 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+/// Protocol for errors with a stable code and structured metadata.
+/// This allows the client to present the error as it chooses.
+
+import Collections
+
+public protocol AppError: Error {
```
saehejkang, line 22: Is this going to be the used for the highest level and for generic type of errors (that all resources will follow)? I kind of worked on creating more lower level, managed resource specific handling approach (that each resource will follow but will vary per resource).

jglogan: I don't know the answer to that for sure yet. I feel like we'll want to try as much as we can in the coming month and evolve to something sensible by the next release.

What I was going for with this error is:

- Supports a caused-by chain.
- Supports structured log output
- Delegates presentation to an error receiver that is responsible for making sense of chained errors, compound errors, and error metadata and mapping it to one or more log messages.
- For easier compatibility, doesn't rely on `enum` for error typing.

saehejkang: Sounds good to me. I am going to take a step back for now and let you lead this. Excited to see what it becomes! Happy to jump back in when it’s ready to be used for managed resources.

Sources/ContainerResource/Common/ManagedResource.swift
```diff
@@ -53,6 +53,6 @@ extension ManagedResource {
 }
 
 // FIXME: This moves to ManagedResource and/or a ResourceLabels typealias eventually.
```
ajemory:
```suggestion
```

Sources/ContainerResource/Common/ManagedResource.swift
```diff
@@ -33,7 +33,7 @@ public protocol ManagedResource: Identifiable, Sendable, Codable {
     /// Key-value properties for the resource. The user and system may both
     /// make use of labels to read and write annotations or other metadata.
     /// A good practice is to use
```
ajemory: Comment cut off?

Issue #1365, opened 2026-03-29
[Request]: Add plain progress mode


nrthakur:
## Description

Add a `plain` progress mode to the `--progress` flag (introduced in #808).

This mode would output one line per progress step using plain text with no ANSI escape codes. This is useful for environments that don't support ANSI, such as CI logs or piped output.

## Related

- #641
- #808

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1367, opened 2026-03-29
Add Plain Progress Output Mode

nrthakur:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

 ## Motivation and Context
Adds a `plain` progress output mode (`--progress plain`) that outputs newline-separated progress lines without ANSI escape sequences. This is useful for CI environments, piped output, and non-TTY contexts where ANSI cursor control is not supported. Unlike `--progress ansi`, plain mode works even when stderr is not a TTY.                                                                                                                
   
Related: #1365, #641, #808 

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Tests/TerminalProgressTests/ProgressBarTests.swift
+++ b/Tests/TerminalProgressTests/ProgressBarTests.swift
@@ -854,6 +854,69 @@ final class ProgressBarTests: XCTestCase {
         XCTAssertFalse(output.contains("\u{001B}"))
     }
 
+    func testPlainModeTerminalOutput() async throws {
+        let pipe = Pipe()
+        let config = try ProgressConfig(
+            terminal: pipe.fileHandleForWriting,
+            description: "Task",
+            showSpinner: false,
+            clearOnFinish: false,
+            outputMode: .plain
+        )
+        let progress = ProgressBar(config: config)
+        progress.render(force: true)
+        progress.finish()
+        try pipe.fileHandleForWriting.close()
+
+        let data = pipe.fileHandleForReading.readDataToEndOfFile()
+        let output = String(data: data, encoding: .utf8) ?? ""
+        let lines = output.components(separatedBy: "\n").filter { !$0.isEmpty }
+        // Expect exactly 2 lines: one from render, one from finish
+        XCTAssertEqual(lines.count, 2)
+        XCTAssertTrue(lines[0].contains("Task"))
+        XCTAssertTrue(lines[1].contains("Task"))
+    }
+
+    func testPlainModeTerminalOutputNoAnsiEscapes() async throws {
+        let pipe = Pipe()
+        let config = try ProgressConfig(
+            terminal: pipe.fileHandleForWriting,
+            description: "Task",
+            showSpinner: false,
+            clearOnFinish: false,
+            outputMode: .plain
+        )
+        let progress = ProgressBar(config: config)
+        progress.render(force: true)
+        progress.finish()
+        try pipe.fileHandleForWriting.close()
+
+        let data = pipe.fileHandleForReading.readDataToEndOfFile()
+        let output = String(data: data, encoding: .utf8) ?? ""
+        XCTAssertFalse(output.contains("\u{001B}"))
+    }
+
+    func testPlainModeTerminalOutputUsesNewlines() async throws {
+        let pipe = Pipe()
+        let config = try ProgressConfig(
+            terminal: pipe.fileHandleForWriting,
+            description: "Task",
+            showSpinner: false,
+            clearOnFinish: false,
+            outputMode: .plain
+        )
+        let progress = ProgressBar(config: config)
+        progress.render(force: true)
+        progress.finish()
+        try pipe.fileHandleForWriting.close()
+
+        let data = pipe.fileHandleForReading.readDataToEndOfFile()
+        let output = String(data: data, encoding: .utf8) ?? ""
+        // Plain mode should use newlines, not carriage returns
+        XCTAssertFalse(output.contains("\r"))
+        XCTAssertTrue(output.contains("\n"))
+    }
+
     func testOutputModeDefaultIsAnsi() async throws {
         let config = try ProgressConfig(description: "Task")
         XCTAssertEqual(config.outputMode, .ansi)
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -89,7 +89,7 @@ container run [<options>] <image> [<arguments> ...]
 
 **Progress Options**
 
-*   `--progress <type>`: Progress type (format: none|ansi) (default: ansi)
+*   `--progress <type>`: Progress type (format: none|ansi|plain) (default: ansi)
 
 **Examples**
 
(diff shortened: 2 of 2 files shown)


Sources/TerminalProgress/ProgressBar.swift
```diff
@@ -157,6 +172,16 @@ extension ProgressBar {
...
         }
+
+        if config.outputMode == .plain && !force {
+            let now = DispatchTime.now()
+            let elapsed = now.uptimeNanoseconds - state.lastPlainRenderTime.uptimeNanoseconds
+            guard elapsed >= 1_000_000_000 else {
+                return
+            }
+            state.lastPlainRenderTime = now
+        }
```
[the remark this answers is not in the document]
nrthakur: By design. Plain mode intentionally throttles to avoid spamming logs. Not an issue.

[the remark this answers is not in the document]
nrthakur: Minor point.  The 40ms timer still fires in plain mode and early-returns. The CPU cost is negligible (one lock + time comparison). An optimization, not a bug. don't think it's worth the complexity right now.

Sources/TerminalProgress/ProgressBar.swift
```diff
@@ -157,6 +172,16 @@ extension ProgressBar {
...
             return
         }
+
+        if config.outputMode == .plain && !force {
+            let now = DispatchTime.now()
+            let elapsed = now.uptimeNanoseconds - state.lastPlainRenderTime.uptimeNanoseconds
+            guard elapsed >= 1_000_000_000 else {
+                return
+            }
+            state.lastPlainRenderTime = now
```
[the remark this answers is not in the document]
nrthakur: Since lastPlainRenderTime defaults to .now(), the first non-forced render() call will always be throttled (elapsed ≈ 0). Users won't see any output for the first second. will fix.

nrthakur:
@jglogan hoping for a review!

Sources/ContainerCommands/Image/ImagePush.swift
```diff
@@ -71,6 +71,17 @@ extension Application {
...
+                progressConfig = try ProgressConfig(
+                    description: "Pushing image \(image.reference)",
+                    itemsName: "blobs",
+                    showSpinner: false,
+                    showItems: true,
+                    showSpeed: false,
+                    ignoreSmallSize: true,
+                    clearOnFinish: false,
+                    outputMode: .plain
+                )
```
dkovba: Can we reduce code duplication in this switch using a helper?

Sources/TerminalProgress/ProgressBar+State.swift
```diff
@@ -66,12 +66,13 @@ extension ProgressBar {
         }
 
         var startTime: DispatchTime
+        var lastPlainRenderTime: DispatchTime
```
dkovba: This property is irrelevant in the ansi mode. Should we make it an optional?

Sources/TerminalProgress/ProgressBar+Terminal.swift
```diff
@@ -80,16 +85,22 @@ extension ProgressBar {
...
-        // Clears previously printed lines.
-        var lines = ""
-        if terminating.hasSuffix("\r") && termWidth > 0 {
-            let lineCount = (text.count - 1) / termWidth
-            for _ in 0..<lineCount {
-                lines += EscapeSequence.moveUp
+        switch config.outputMode {
+        case .plain:
+            guard !text.isEmpty else { return }
+            display("\(text)\n")
```
dkovba: Please add the support for the `terminating` parameter.

dkovba:
Thank you for the contribution!

Sources/TerminalProgress/ProgressBar+Terminal.swift
nrthakur: **Addressed.** The `.plain` case in `displayText` now uses `terminating` instead of a hardcoded `"\n"` (line 91).

The `render()` function explicitly passes `"\n"` for plain mode (line 188), and `finish()` already passes `"\n"`.

See:

* `ProgressBar+Terminal.swift:91`
* `ProgressBar.swift:188`

Sources/TerminalProgress/ProgressBar+State.swift
nrthakur: **Addressed.** `lastPlainRenderTime` is now `DispatchTime?` (line 69), defaults to `nil`, and has been removed from the initializer parameters.

The render throttle now uses `if let` to unwrap it, allowing the first plain render to fire immediately.

See:

* `ProgressBar+State.swift:69`
* `ProgressBar.swift:178`

Sources/ContainerCommands/Image/ImagePush.swift
nrthakur: **Addressed.** The switch has been removed from `ImagePush` (as well as `ImagePull` and `ContainerRun`).

A new `makeConfig` helper on `Flags.Progress` now centralizes this logic. The `.ansi` and `.plain` cases have been consolidated into a single branch, differing only by three computed values:

* `showSpinner: !isPlain`
* `clearOnFinish: !isPlain`
* `outputMode`

See `Flags+ProgressConfig.swift:38–51`.

nrthakur:
@dkovba Apologies, CI was failing, pushed another commit. Would you mind taking another look 👀

dkovba (approved):

Tests/TerminalProgressTests/ProgressBarTests.swift
```diff
@@ -750,4 +750,112 @@ final class ProgressBarTests: XCTestCase {
...
+            description: "Task",
+            showSpinner: false,
+            outputMode: .plain
+        )
+        let progress = ProgressBar(config: config)
+        let output = progress.draw()
+        XCTAssertEqual(output, "Task [0s]")
+    }
+
+    func testPlainModeDrawWithTasks() async throws {
```
manojmahapatra, line 774: Suggestion: The current plain-mode tests only assert `draw()`, so they don't catch any duplicate or missing terminal writes. I would pass `Pipe` as the terminal in the unit tests, read back the bytes, and assert on the emitted output. That would let us catch cases like any duplicate final/extra lines in plain mode.

nrthakur: **Addressed.** Added three pipe-based tests that pass a `Pipe` as the terminal, read back the raw bytes, and assert on the emitted output:

* `testPlainModeTerminalOutput` - verifies exactly 2 lines (one render + one finish), with no duplicates
* `testPlainModeTerminalOutputNoAnsiEscapes` - verifies no ANSI escape codes in the raw output
* `testPlainModeTerminalOutputUsesNewlines` - verifies `\n` terminators and no `\r`

See `ProgressBarTests.swift:857-912`.

jglogan:
@nrthakur Thanks for the contribution!

manojmahapatra:
I see that this is already merged. Minor update https://github.com/apple/container/pull/1379 I think nice-to-have.

Issue #1352, opened 2026-03-26
[Bug]: `container exec -u` does not enforce file permissions


ieee0824:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

```bash
# 1. Start a plain Alpine container
container run -d --name perm-test alpine:3.23 sleep 300

# 2. Verify /etc/shadow permissions (0640 root:shadow)
container exec perm-test ls -la /etc/shadow
# -rw-r-----    1 root     shadow         260 Jan 27 21:19 /etc/shadow

# 3. Exec as "nobody" and try to read the restricted file
container exec -u nobody perm-test cat /etc/shadow

# 4. Confirm uid/gid is switched correctly
container exec -u nobody perm-test id
# uid=65534(nobody) gid=65534(nobody) groups=65534(nobody)

# Cleanup
container stop perm-test && container rm perm-test
```

### Current behavior

`container exec -u nobody perm-test cat /etc/shadow` prints the full contents of `/etc/shadow`, even though:

- The file is `0640 root:shadow`
- The process runs as `nobody` (uid=65534, gid=65534)
- `nobody` is not in the `shadow` group

Docker correctly denies this with `Permission denied` for the same image and commands.

### Expected behavior

```
cat: can't open '/etc/shadow': Permission denied
```

File permission checks should be enforced based on the effective uid/gid set by `-u`, consistent with standard Unix semantics and Docker behavior.

### Environment

```markdown
- OS: macOS 26.3.1 (25D771280a)
- Xcode: 26.2 (17C52)
- Container: CLI version 0.10.0 (build: release, commit: 6bdb647)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

dcantah:
We need capabilities support in here. Our library (https://github.com/apple/containerization) has this, but hasn't been exposed in this project yet.

PR #1381, opened 2026-04-02
fix(exec): enforce permissions for non-root exec user

KevinSailema:
Drop elevated process capabilities when exec is requested with an explicit non-root user.

Add a CLI regression test that validates /etc/shadow access is denied for exec -u nobody.

Refs: #1352

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Issue #1352 reports that container exec with an explicit non-root user could still read privileged files such as /etc/shadow.  
Although uid and gid were switched correctly, elevated process capabilities were still effectively allowing access.  
This change drops elevated capabilities when exec is requested with an explicit non-root user, so file permission checks follow expected Unix and Docker semantics.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Container/ProcessConfiguration.swift
+++ b/Sources/ContainerResource/Container/ProcessConfiguration.swift
@@ -68,6 +68,24 @@ public struct ProcessConfiguration: Sendable, Codable {
                 return name
             }
         }
+
+        /// Returns true when the user specification explicitly resolves to a non-root user.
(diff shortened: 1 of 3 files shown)


dcantah:
This is not the right fix, we just need caps support in this project in general. We're working to get this out shortly.

PR #1383, opened 2026-04-02
Add capabilities support

### LINKED ISSUE #1352: [Bug]: `container exec -u` does not enforce file permissions

dcantah:
Closes #1352

Containerization has had support for a bit, it was just never brought over here. It's exposed on the CLI via the classic `--cap-add` and `--cap-drop` UX.

### DIFF
--- a/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
+++ b/Sources/Services/ContainerSandboxService/Server/SandboxService.swift
@@ -938,6 +938,10 @@ public actor SandboxService {
                 soft: $0.soft
             )
         }
+        czConfig.process.capabilities = try Self.effectiveCapabilities(
+            capAdd: config.capAdd,
+            capDrop: config.capDrop
+        )
         switch process.user {
         case .raw(let name):
             czConfig.process.user = .init(
@@ -984,6 +988,10 @@ public actor SandboxService {
                 soft: $0.soft
             )
         }
+        proc.capabilities = try Self.effectiveCapabilities(
+            capAdd: containerConfig.capAdd,
+            capDrop: containerConfig.capDrop
+        )
         switch config.user {
         case .raw(let name):
             proc.user = .init(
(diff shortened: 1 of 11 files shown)


docs/how-to.md
```diff
@@ -471,6 +471,50 @@ You can also output statistics in JSON format for scripting:
...
+
+By default, containers start with a restricted set of Linux capabilities:
+
+`CAP_AUDIT_WRITE`, `CAP_CHOWN`, `CAP_DAC_OVERRIDE`, `CAP_FOWNER`, `CAP_FSETID`, `CAP_KILL`, `CAP_MKNOD`, `CAP_NET_BIND_SERVICE`, `CAP_NET_RAW`, `CAP_SETFCAP`, `CAP_SETGID`, `CAP_SETPCAP`, `CAP_SETUID`, `CAP_SYS_CHROOT`
+
+You can customize the capability set using `--cap-add` and `--cap-drop`.
+
+Capability names can be specified with or without the `CAP_` prefix, and are case-insensitive:
+
+```console
```
jglogan: Use bash tags for blocks containing just shell commands.

Move the comment outside. Users wouldn't necessarily copy/paste this block but if they were to do it, some shells won't properly deal with comments in an interactive context.

> Capability names are case insensitive, and accept an optional "CAP_" prefix. These commands are all equivalent:

docs/how-to.md
```diff
@@ -471,6 +471,50 @@ You can also output statistics in JSON format for scripting:
...
+```
+
+To drop a single capability from the default set:
+
+```console
+container run --cap-drop CHOWN alpine chown 100 /tmp
+# chown: /tmp: Operation not permitted
+```
+
+The `--cap-add` and `--cap-drop` flags can also be used with `container create`.
```
jglogan: Might want to lead with this, instead of burying it at the bottom.

docs/how-to.md
```diff
@@ -471,6 +471,50 @@ You can also output statistics in JSON format for scripting:
...
+To grant all capabilities except specific ones:
+
+```console
+container run --cap-add ALL --cap-drop NET_ADMIN alpine sh
+```
+
+To drop a single capability from the default set:
+
+```console
+container run --cap-drop CHOWN alpine chown 100 /tmp
```
jglogan: This one could be console since you're showing command output:

```suggestion
% container run --cap-drop CHOWN alpine chown 100 /tmp
chown: /tmp: Operation not permitted
%
```

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -1006,6 +1014,36 @@ public actor SandboxService {
         return proc
     }
 
+    /// Compute effective Linux capabilities from the OCI default set, capAdd, and capDrop:
```
jglogan: Might want to emphasize that the `ALL` processing in steps 1 and 2 is performed in order, such that --cap-add `ALL` overrides everything.

docs/how-to.md
```diff
@@ -471,6 +471,50 @@ You can also output statistics in JSON format for scripting:
...
+# These are equivalent
+container run --cap-add CAP_NET_ADMIN alpine ip link set lo down
+container run --cap-add NET_ADMIN alpine ip link set lo down
+container run --cap-add net_admin alpine ip link set lo down
+```
+
+To grant all capabilities:
+
+```console
+container run --cap-add ALL alpine sh -c "ip link set lo down && echo ok"
```
jglogan, line 494: We should probably explain `--cap-drop ALL --cap-add ALL` behavior here

PR #1385, opened 2026-04-03
Consolidate CLI formatting infrastructure

realrajaryan:
## Motivation and Context
This PR consolidates duplicated list-output formatting across the CLI into shared rendering infrastructure.

Currently, each list command has its own copy of the same json/quiet/table branching. This PR pulls that into shared rendering infrastructure in `ContainerCommands`:
- `ListDisplayable` protocol for table + quiet output                                                                 
- `renderJSON`, `renderTable`, `renderList` as pure functions that return strings                                     
- `emit()` as the single stdout boundary (no-ops on empty strings to avoid blank-line regressions)                    
- `JSONOptions` so all JSON encoding goes through one path, including volume inspect's pretty + ISO 8601 case  

JSON encoding remains separate from display formatting: each command still chooses its own JSON model, while `ListDisplayable` is used only for table and quiet output. `ImageList` remains the intentional exception for quiet mode so it can avoid unnecessary async work.

This change also replaces inline `JSONEncoder` usage with `renderJSON`, removes the old `Codable+JSON.swift` helper, and moves `TableOutput` and `ListFormat` into `ContainerCommands`. It also adds unit tests for the shared rendering helpers and expands integration coverage for image, network, and registry list formatting.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Builder/BuilderStatus.swift
+++ b/Sources/ContainerCommands/Builder/BuilderStatus.swift
@@ -47,7 +47,7 @@ extension Application {
                 let container = try await client.get(id: "buildkit")
 
                 if format == .json {
-                    try emit(renderJSON([PrintableContainer(container)]))
+                    try Output.emit(Output.renderJSON([PrintableContainer(container)]))
                     return
                 }
 
@@ -55,7 +55,7 @@ extension Application {
                     return
                 }
 
-                emit(renderList([PrintableBuilder(container)], quiet: quiet))
+                Output.emit(Output.renderList([PrintableBuilder(container)], quiet: quiet))
             } catch {
                 if let czError = error as? ContainerizationError, czError.code == .notFound {
                     if !quiet {
--- a/Sources/ContainerCommands/OutputRendering.swift
+++ b/Sources/ContainerCommands/OutputRendering.swift
@@ -33,36 +33,56 @@ public struct JSONOptions: Sendable {
     }
 }
 
-/// Renders an `Encodable` value as a JSON string.
-public func renderJSON<T: Encodable>(_ value: T, options: JSONOptions = .compact) throws -> String {
-    let encoder = JSONEncoder()
-    encoder.outputFormatting = options.outputFormatting
-    encoder.dateEncodingStrategy = options.dateEncodingStrategy
-    let data = try encoder.encode(value)
-    return String(decoding: data, as: UTF8.self)
-}
-
-/// Renders a list of displayable items as a table (with header) or quiet-mode identifiers.
-public func renderList<T: ListDisplayable>(_ items: [T], quiet: Bool) -> String {
-    if quiet {
-        return items.map(\.quietValue).joined(separator: "\n")
+/// Shared rendering helpers for CLI output.
+///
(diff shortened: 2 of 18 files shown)


Sources/ContainerCommands/TableOutput.swift
```diff
@@ -16,19 +16,19 @@
 
 import Foundation
 
-public struct TableOutput {
+struct TableOutput {
```
jglogan: One consideration here: what `public` visibility is needed for commands in third-party libraries/plugins that depend on `ContainerCommands` and want to produce consistent collection output?

realrajaryan: Yeah. I don't mind, we can make it public for third party use cases

Sources/ContainerCommands/Builder/BuilderStatus.swift
```diff
@@ -45,60 +45,53 @@ extension Application {
...
-                return
-            }
-
-            if self.quiet {
-                containers
-                    .filter { $0.status == .running }
-                    .forEach { print($0.id) }
-                return
-            }
+private struct PrintableBuilder: ListDisplayable {
```
jglogan, line 72: Is there a reason we can't define a ListDisplayable extension on the bulk of or regular types instead of defining completely separate types that conform to that protocol? It feels like there might be a bit more boilerplate as it currently stands.

jglogan: > Is there a reason we can't define a ListDisplayable extension on the bulk of or regular types instead of defining completely separate types that conform to that protocol?

We'll just continue living with the overhead of these types for now. The problem lies in the need to declare retroactive conformance here on types that are defined in ContainerResource.

Sources/ContainerCommands/OutputRendering.swift
```diff
@@ -0,0 +1,68 @@
...
+        outputFormatting: JSONEncoder.OutputFormatting = [],
+        dateEncodingStrategy: JSONEncoder.DateEncodingStrategy = .deferredToDate
+    ) {
+        self.outputFormatting = outputFormatting
+        self.dateEncodingStrategy = dateEncodingStrategy
+    }
+}
+
+/// Renders an `Encodable` value as a JSON string.
+public func renderJSON<T: Encodable>(_ value: T, options: JSONOptions = .compact) throws -> String {
```
jglogan: Prefer methods and properties to free functions (avoids cluttering the global namespace):

https://www.swift.org/documentation/api-design-guidelines/

Also: a helper function could reduce a little bit of boilerplate everywhere:

```
    public func renderOutput<T: ListDisplayable & Encodable>(                                                                                                                                                                                                   
        _ items: [T], format: ListFormat, quiet: Bool                                                                                                                                                                                                           
    ) throws -> String {                                                                                                                                                                                                                                        
        format == .json ? try renderJSON(items) : renderList(items, quiet: quiet)                                                                                                                                                                               
    }                                                                                                                                                                                                                                                           
```

jglogan (changes requested):

PR #1391, opened 2026-04-05
Archiver: preserve symlinks and explicit build-context entries

mazdak:
## Type of Change
- [X] Bug fix

## Motivation and Context
While building out a Docker Compose-style plugin and validating it against our own real development stack, we ran into two classes of problems in the build-context path:

* Correctness issues in archive generation, especially around symlinks 
* Very high client-side overhead while preparing and streaming build contexts.

This PR fixes build-context archiving in `container` so symlinks are preserved correctly, digest calculation reflects symlink target changes, and fssync archives the exact selected file set in the already-computed order.

Before this change, build-context archiving could mis-handle symlinks, archive a broader tree walk than necessary, and produce a digest that did not change when symlink targets changed.

Part of the fix was also more generally related to the Containerization framework.  This PR intentionally stays independently mergeable against the current released `containerization` dependency, so it keeps the local archiver implementation  needed for these fixes today. The Containerization PR is: https://github.com/apple/containerization/pull/652

## Validation

This is an example of the problem the PR actually fixes.

### Build context:
```
ctx="$(mktemp -d /private/tmp/container-retest-rel.XXXXXX)/ctx"
mkdir -p "$ctx"
printf 'hello\n' > "$ctx/target.txt"
ln -s target.txt "$ctx/link.txt"

cat > "$ctx/Dockerfile" <<'EOF'
FROM alpine:3.20
COPY . /ctx
RUN echo "link target: $(readlink /ctx/link.txt)" && \
      test "$(readlink /ctx/link.txt)" = "target.txt" && \
      test "$(cat /ctx/link.txt)" = "hello"
EOF
```

### Build command:
```
  container build -t rel-patch-retest -f Dockerfile .
```

### Observed results:

  - main branch: failed
      - link target: came back empty
      - exit code 1
  - patched branch: passed
      - link target: target.txt
      - Successfully built `rel-patch-retest:latest`
      - exit code 0
  - Before this fix, container build could lose symlink metadata in the archived build context.
  - A relative symlink like link.txt -> target.txt arrived broken inside the build.
  - After this fix, the same build context preserves the symlink correctly and the build succeeds.

## Testing
- [X] Tested locally
- [X] Added/updated tests

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Archiver.swift
+++ b/Sources/Services/ContainerAPIService/Client/Archiver.swift
@@ -17,9 +17,27 @@
 import ContainerizationArchive
 import ContainerizationOS
 import CryptoKit
+import Darwin
 import Foundation
 
 public final class Archiver: Sendable {
+    private struct FileStatus {
+        enum EntryType {
+            case directory
+            case regular
+            case symbolicLink
+        }
+
+        let entryType: EntryType
+        let permissions: UInt16
+        let size: Int64
+        let owner: UInt32
+        let group: UInt32
+        let creationDate: Date?
+        let modificationDate: Date?
+        let symlinkTarget: String?
+    }
+
     public struct ArchiveEntryInfo: Sendable, Codable {
         public let pathOnHost: URL
         public let pathInArchive: URL
@@ -41,6 +59,19 @@ public final class Archiver: Sendable {
             self.group = group
             self.permissions = permissions
         }
+    }
+
+    private struct ArchiveEntryHashInfo: Encodable {
+        let pathOnHost: String
+        let pathInArchive: String
+        let fileType: String
+        let size: Int64?
+        let permissions: mode_t
+        let owner: uid_t?
+        let group: gid_t?
+        let symlinkTarget: String?
+        let creationDate: Date?
+        let modificationDate: Date?
     }
 
     public static func compress(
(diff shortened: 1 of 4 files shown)


Sources/Services/ContainerAPIService/Client/Archiver.swift
```diff
@@ -108,79 +154,210 @@ public final class Archiver: Sendable {
         return hasher.finalize()
     }
 
+    public static func uncompress(source: URL, destination: URL) throws {
```
jglogan, line 159: I removed this in #1372 because it's insecure and should not be used. See `ArchiveReader.extractContents()` in containerization in this commit: https://github.com/apple/containerization/commit/3e93416b9a6d7b4c25fff7e9dea22a9ca687ee52

jglogan:
Can you provide a simple `container build` command, `Dockerfile` and build context that fails and is addressed by the fix(es) you propose? Thanks!

Sources/Services/ContainerAPIService/Client/Archiver.swift
mazdak: Fixed

mazdak:
> Can you provide a simple `container build` command, `Dockerfile` and build context that fails and is addressed by the fix(es) you propose? Thanks!

Please see PR description

mazdak:
@jglogan are you still considering this?

PR #1395, opened 2026-04-06
Use persistent XPC connection for IP allocation management.

### LINKED ISSUE #1318: [Bug]: Unable to delete a network - `IP allocator cannot be disabled with active containers`
### I have done the following

### Steps to reproduce

There's a corner case where if the sandbox service terminates abruptly, it fails to deallocate the IP addresses for the failed container's interfaces. An orphaned allocation on a user-created network prevents the user from deleting the network with `container network delete`, even if they deleted all other containers attached to the network. 

The workaround is to stop and start the container service, which is hardly ideal.

### Current behavior

```
cannot delete subnet foo because the IP allocator cannot be disabled with active containers
```

### Expected behavior

Network with no referring containers can be deleted.

### Environment

```markdown
- OS: 26.3
- Xcode: 26
- Container: 0.8.0
```

### Relevant log output

```shell
$ container network delete --debug foo
2026-03-15T08:50:46Z failed to delete network: ["error": invalidState: "cannot delete subnet foo because the IP allocator cannot be disabled with active containers", "id": foo]
Error: delete failed for one or more networks: ["foo"]

$ container list
ID  IMAGE  OS  ARCH  STATE  ADDR  CPUS  MEMORY  STARTED

$ container list -a
ID                    IMAGE                                                  OS     ARCH   STATE    ADDR  CPUS  MEMORY   STARTED
buildkit              docker.apple.com/containerkit/builder:0.8.0-10         linux  arm64  stopped        2     2048 MB  2026-03-15T07:25:45Z
```

### Code of Conduct

### LINKED ISSUE #1378: [Bug]: IP address exhaustion with container delete --force
### I have done the following

### Steps to reproduce

If you create containers and then delete them with `--force`, it seems to leak IP addresses. Eventually you will reach a state where you can't create a new container, with the error "no free indices are available for allocation". Looking through the code, I found this comes from the IP address allocation logic.

I wrote a test script that will create a simple container and then `container delete --force` it in a loop. You can find it here:

https://gist.github.com/thomasjm/6a2d60e3233fa55bc7765c2426784b33#file-test_ip_exhaustion-sh

There's also another version that does a graceful shutdown, calling `container stop` and waiting for the container to fully stop before deleting it:

https://gist.github.com/thomasjm/6a2d60e3233fa55bc7765c2426784b33#file-test_ip_exhaustion_graceful-sh

### Current behavior

Running the first script will eventually start generating failures after about ~255 iterations. (Because the default system subnet is a /24, so that's how many IP addresses are available.)

```
...
ip-leak-test-263
  [263] ip-leak-test-263 -> 192.168.64.180/24
ip-leak-test-264
  [264] ip-leak-test-264 -> 192.168.64.191/24
ip-leak-test-265
  [265] ip-leak-test-265 -> 192.168.64.242/24
FAILED at iteration 266: Error: failed to bootstrap container (cause: "unknown: "no free indices are available for allocation"")
```

Running the graceful shutdown version of the script does not result in failures.

### Expected behavior

IP addresses shouldn't leak; creating a container should always succeed if there are <255 currently running containers. (Or whatever the size of the default subnet is.)

### Environment

```markdown
- OS: Version 26.3.1 (a) (25D771280a)
- Xcode: 
- Container: container CLI version 0.10.0 (build: release, commit: 6bdb647)
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

jglogan:
- Closes #1378.
- Closes #1318.
- At present, IP allocation involves four parties: the ContainersService and NetworksService in the API server, the SandboxService running in `container-runtime-linux`, and the NetworkService running in `container-network-vmnet`.
- Each NetworkService `allocate()` requires a corresponding `deallocate()` or an IP address gets leaked.
- This change gets rid of `deallocate()` and instead changes `allocate()` so that the caller received an XPC connection. The caller holds this connection for as long as the network attachment is alive (essentially as long as the container is running). When the connection disconnects, the NetworkService automatically unregisters the hostname:IP association.
- Removed IP allocation from the API server, restoring the old process by which sandboxes directly allocate from networks. To deal with the eventual possibility of more network plugins than `container-network-vmnet`, the bootstrap call requires an additional parameter `networkPluginInfos` that is the same size as the attachments array in the container configuration, and identifies the plugin for each attachment.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

- Current network plumbing is complicated and needed simplification.
- The `deallocate()` call was always a leak hazard.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


vxnkjyffmq-code (approved):

Sources/Services/ContainerNetworkService/Client/NetworkClient.swift
```diff
@@ -53,30 +53,30 @@ extension NetworkClient {
...
+        macAddress: MACAddress? = nil,
+        onDisconnect: @Sendable @escaping () async -> Void = {}
+    ) async throws -> (attachment: Attachment, additionalData: XPCMessage?, connection: XPCClient) {
         let request = XPCMessage(route: NetworkRoutes.allocate.rawValue)
         request.set(key: NetworkKeys.hostname.rawValue, value: hostname)
         if let macAddress = macAddress {
             request.set(key: NetworkKeys.macAddress.rawValue, value: macAddress.description)
         }
 
         let client = createClient()
```
dcantah, line 73: If we changed this client to the model that some of our others have taken (maintain a long lived xpc connection instead of creating a new one for every request), does this screw things up with this persistent routes idea?

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
```diff
@@ -48,6 +49,9 @@ public actor SandboxService {
     private let lock: AsyncLock = AsyncLock()
     private let log: Logging.Logger
     private var state: State = .created
+    /// Persistent XPC connections to each network plugin, one per network attachment.
+    /// Held for the container lifetime; deallocation happens automatically when these are released.
+    private var networkConnections: [XPCClient] = []
```
katiewasnothere, line 54: Maybe this is a naive question, but could this ever cause an issue if we have a lot of running containers with a lot of networks in use?

Sources/Services/ContainerNetworkService/Client/NetworkClient.swift
jglogan: It'd probably be easier to just discuss it. I could see the API server ContainerNetworksService holding long lived XPC connections with each network. The API server uses different calls than the runtime plugin, so I imagine there's a way to make it work.

Sources/Services/ContainerSandboxService/Server/SandboxService.swift
jglogan: Not naive, I can do a bit of research on that.

A napkin calculation is that we can have a max of ~256 VMs, and I think maybe 4 network interfaces per VM (I think I just saw that recently but could be wrong there) so if you worst case the network interfaces per VM number you're looking at 4K XPC connections.

jglogan:
@katiewasnothere @dcantah Thanks for the initial comments!

I pushed a PR to fix a bug in the disconnect handler, it was canceling the server XPC connection even if the container wasn't in the running state, which caused startup errors to get handled improperly.

I still need to think through the sandbox changes more closely myself, so putting this into draft until I do that.

jglogan:
Closing this PR in favor of #1524 and two follow-ons that replace the `persistentRoute` approach with one that adds session awareness to existing routes, and session disconnect handlers, so that we can run long-lived connections that automatically clean up after themselves.

Issue #1375, opened 2026-04-01
[Bug]: builds fail with 'Error: unable to write data to the archive, code 0'


jdmartin:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Reproducing this, for me, is a matter of visiting one of my repos that contains my Dockerfile and Makefile, and running something like `container build --no-cache --arch amd64 --tag mydomain.com/yeti:latest --file Dockerfile .;`  (Sorry, that's very me-specific, which is likely unhelpful...)

### Current behavior

Since upgrading to 0.11.0, my ability to build containers seems to have broken (macOS Sequoia, 15.7.5).  When I run my build command, I see the error below.  The command is this: `container build --no-cache --arch amd64 --tag mydomain.com/yeti:latest --file Dockerfile .;`  (the makefile just polls until container is running, which is that bit above the 'Building AMD64 version...')

I've tried removing each flag one-by-one.  I've tried uninstalling, purging directories, and reinstalling.  The only thing that resolves this is reverting to 0.10.0.  

With full acknowledgment that this might be user error, I thought I'd pass this along.  Happy to share any additional details as needed!

```bash
Neptune ~/D/F/MechanicalYeti (main)> make
Registering API server with launchd...
Verifying apiserver is running...
Waiting for container system to become fully ready...
Container system is fully ready.
Building AMD64 version...
[+] Building 7.3s (2/3)
 => [resolver] fetching image...docker.io/library/node:lts-slim                                                                                                                                    0.0s
 => [internal] load build definition from Dockerfile                                                                                                                                               0.4s
 => => transferring dockerfile: 213B                                                                                                                                                               0.3s
 => [internal] load .dockerignore                                                                                                                                                                  0.3s
 => => transferring context: 30B                                                                                                                                                                   0.3s
Error: unable to write data to the archive, code 0
make[1]: *** [build-container] Error 1
make: *** [default] Error 2
```

### Expected behavior

Historically, with 0.10.0, I'd see a successfully completed build process.

### Environment

```markdown
- OS: macOS 15.7.5
- Container: Container CLI version 0.11.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jdmartin:
Apologies for the noise. Just re-read the Readme, and noticed:

[...]

PR #1403, opened 2026-04-07
Write file data into archive only for regular files

adityaramani:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes https://github.com/apple/container/issues/1375

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Archiver.swift
+++ b/Sources/Services/ContainerAPIService/Client/Archiver.swift
@@ -96,7 +96,7 @@ public final class Archiver: Sendable {
                 guard let entry = try Self._createEntry(entryInfo: info) else {
                     throw Error.failedToCreateEntry
                 }
-                hasher.update(data: try encoder.encode(info))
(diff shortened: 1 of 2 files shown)


wlan0 (changes requested):
Good fix for the archive write path

The digest logic gets broken however. 

`compress()` hashes `ArchiveEntryInfo`, which doesn't include `fileType` or `symlinkTarget`. In case of symlinks, no bytes are hashed either. 

This means, symlink target paths can change without changing the entire digest's hash.

A clean fix would be to either add `fileType`/`symlinkTarget` fields to `ArchiveEntryInfo`, or hash `entry.symlinkTarget`  in `_compressFile()` for symlinks.

adityaramani:
@wlan0  Thanks for catching that. Updated the `compress` method to also account for the metadata of the file / link / directory by hashing on the archive header

PR #1408, opened 2026-04-10
Fix ContainerResource build failure

DePasqualeOrg:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

The `ContainerResource` target uses `OrderedDictionary` (added in #1360) but doesn't declare a dependency on `swift-collections` in `Package.swift`, and the source files import the `Collections` umbrella module instead of `OrderedCollections`. This causes build failures when `OrderedDictionary` is used in public declarations, since the compiler requires the specific submodule to be imported in that context.

- Add `swift-collections` dependency to the `ContainerResource` target in `Package.swift`
- Change `import Collections` to `import OrderedCollections` in `ApplicationError.swift` and `ResourceLabels.swift`

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -350,6 +350,7 @@ let package = Package(
         .target(
             name: "ContainerResource",
             dependencies: [
+                .product(name: "Collections", package: "swift-collections"),
                 .product(name: "Containerization", package: "containerization"),
                 "ContainerXPC",
                 "CAuditToken",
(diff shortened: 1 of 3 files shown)


katiewasnothere (approved):
lgtm thank you!

PR #1412, opened 2026-04-11
Adds journal option to `container volume create`.

### LINKED ISSUE #1411: [Request]: Add journal mode option to `container volume create`.

The PR apple/containerization#672 adds experimental support for journal modes for ext4 filesystems.

We can add this to `container volume create` with an `--option value`, where value is `mode[:size]`, where `mode` is one of `ordered`, `write back`, or `journal`, and `size` is a size value that's handled the same as the `--size` option, suffix-wise.

For now, the default for a new volume will be no journal, just as today. After we test this out a bit, we may want to make something like `ordered` the default; at that point we could add a `none` mode (which rejects the `size` parameter) that creates a filesystem with no journal.

### Code of Conduct

jglogan:
- Closes #1411.
- Default behavior doesn't change today - with no `journal` option, no journal gets created.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
Should make containers more resilient to ungraceful stops.

## Testing
- [ ] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/docs/command-reference.md
+++ b/docs/command-reference.md
@@ -820,7 +820,32 @@ container volume create [--label <label> ...] [--opt <opt> ...] [-s <s>] [--debu
 
 *   `--label <label>`: Set metadata for a volume
 *   `--opt <opt>`: Set driver specific options
-*   `-s <s>`: Size of the volume in bytes, with optional K, M, G, T, or P suffix
+*   `-s <s>`: Size of the volume in bytes, with optional K, M, G, T, or P suffix. Takes precedence over `--opt size=` if both are specified.
+
+**Driver Options**
+
+Driver options are passed with `--opt key=value`. The following options are supported for the default `local` driver:
+
+*   `size=<value>`: Volume size with optional unit suffix (K, M, G, T, P). Minimum 1 MiB. Equivalent to `-s`; if `-s` is also specified, `-s` takes precedence.
+*   `journal=<mode>[:<size>]`: Configure ext4 journaling on the volume. `<mode>` must be one of:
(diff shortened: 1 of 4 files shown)


docs/command-reference.md
```diff
@@ -820,7 +820,32 @@ container volume create [--label <label> ...] [--opt <opt> ...] [-s <s>] [--debu
...
 *   `--label <label>`: Set metadata for a volume
 *   `--opt <opt>`: Set driver specific options
-*   `-s <s>`: Size of the volume in bytes, with optional K, M, G, T, or P suffix
+*   `-s <s>`: Size of the volume in bytes, with optional K, M, G, T, or P suffix. Takes precedence over `--opt size=` if both are specified.
+
+**Driver Options**
+
+Driver options are passed with `--opt key=value`. The following options are supported for the default `local` driver:
+
+*   `size=<value>`: Volume size with optional unit suffix (K, M, G, T, P). Minimum 1 MiB. Equivalent to `-s`; if `-s` is also specified, `-s` takes precedence.
```
JaewonHur, line 829: Why do we have both `-s <s>` and this `size=<value>`?

JaewonHur (approved):

docs/command-reference.md
jglogan: @realrajaryan for any follow-up on this...

I believe it's just because size should be a driver option (not all volumes will allow you to choose a size; see the PR for SMB/NFS volumes, for example), but for the common case of ext4-backed volumes, it's probably the most useful one for the average user so `-s/--size` is there for convenience.

PR #1417, opened 2026-04-14
add -f flag to prune commands for docker compat

### LINKED ISSUE #938: [Request]: add a -f command to image prune for compatability

For compat with docker automation scripts, we have a cleanup step that is `prune -f` that `container` chokes on

### Code of Conduct

knQzx:
## summary
- adds -f/--force to image/container/volume/network prune as a no-op
- prints a warning to stderr when used, so the flag is not silently ignored
- unblocks docker automation scripts calling `prune -f`

## test plan
- added CLI test: `container prune -f` succeeds, prints warning on stderr, normal output on stdout
- did all four prune variants since docker scripts typically call each; happy to narrow to just image prune if preferred

fixes #938

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerPrune.swift
+++ b/Sources/ContainerCommands/Container/ContainerPrune.swift
@@ -31,7 +31,15 @@ extension Application {
         @OptionGroup
         public var logOptions: Flags.Logging
 
+        @Flag(name: .shortAndLong, help: "Accepted for Docker compatibility; has no effect.")
+        var force: Bool = false
+
         public func run() async throws {
+            if force {
+                let warning = "Warning: '-f' has no effect; prune does not prompt for confirmation.\n"
+                FileHandle.standardError.write(Data(warning.utf8))
+            }
+
             let client = ContainerClient()
             let containersToPrune = try await client.list().filter { $0.status == .stopped }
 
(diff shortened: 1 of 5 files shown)


saehejkang:
I don't think this is something we want here. Adding an option for a flag, to only just output it as a no-op is counter intuitive. However, I see the flag actually being useful, if it is properly implemented. This could be broken up into two separate issues:
- Update any `prune` commands with the `--all` flag to ask the user if they are sure they want to continue (like docker)
- Modifying this PR to bypass the prompt if the `--force` flag is used

What do you think?

knQzx:
> I don't think this is something we want here. Adding an option for a flag, to only just output it as a no-op is counter intuitive. However, I see the flag actually being useful, if it is properly implemented. This could be broken up into two separate issues:
> 
> * Update any `prune` commands with the `--all` flag to ask the user if they are sure they want to continue (like docker)
> * Modifying this PR to bypass the prompt if the `--force` flag is used
> 
> What do you think?

i went with option 2 because that's what @ptone asked for in #938 - no-op with a warning. the issue scope is just docker compat so i kept it narrow                                                          
                                                            
no-op feels like a safer first step - it doesn't change behavior for anyone already using prune. prompt + -force is a separate, bigger change, but i can rework it if that's the preferred direction         
                      
@jglogan what do you think?

PR #1418, opened 2026-04-14
improve error message when image save platform is not available

### LINKED ISSUE #874: [Bug]: Image save error output is too low-level.
### I have done the following

### Steps to reproduce

Here's the example from the related #712 

```console
% container image pull --platform linux/arm64 alpine:latest
% container image save --output image.tar alpine:latest
Error: notFound: "Content with digest sha256:85f2b723e106c34644cd5851d7e81ee87da98ac54672b29947c052a45d31dc2f"
/Users/john/Documents/sandboxes % container image save --platform linux/arm64 --output image.tar alpine:latest
alpine:latest  
```

This shows the developer level root cause of the problem. The user-level root cause is "I tried to save platform X but could only find content for platforms (A, B)".

With #712 fixed we could still hit this issue if I were to do:

```bash
container image pull --platform linux/arm64 alpine:latest
container image save --platform linux/amd64 --output image.tar alpine:latest
```

So making the error message more user-friendly is still a must.

### Current behavior

`image save` shows the developer-level cause of the failure.

### Expected behavior

`image save` shows the user-level cause of the failure.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: 0.6.0
```

### Relevant log output

```shell
N/A
```

### Code of Conduct

knQzx:
## summary
- image save now fails fast with a clear error when the requested platform is not available locally
- the message lists the platforms that ARE available so the user knows what they can save

before:
`Error: notFound: "Content with digest sha256:..."`

after:
`Error: image alpine:latest has no content for platform linux/amd64; available platforms: linux/arm64`

## test plan
- verified locally against a multi-arch image with platforms amd64/arm64/ppc64le/s390x/386/arm-v6/arm-v7:
  - `save --platform linux/arm/v5` (not in image) fails with the new error listing all real platforms (attestation `unknown/unknown` entries are filtered out)
  - `save --platform linux/arm64` (in image) succeeds as before
  - `save` without `--platform` still saves the full multi-arch archive (pre-check is skipped)

fixes #874

### DIFF
--- a/Sources/ContainerCommands/Image/ImageSave.swift
+++ b/Sources/ContainerCommands/Image/ImageSave.swift
@@ -84,6 +84,27 @@ extension Application {
                 throw ContainerizationError(.invalidArgument, message: "failed to save image(s)")
             }
 
+            if let p {
+                for (reference, description) in zip(references, images) {
+                    let image = ClientImage(description: description)
+                    do {
+                        _ = try await image.manifest(for: p)
+                    } catch {
+                        var available: [String] = []
+                        if let index = try? await image.index() {
+                            available = index.manifests
+                                .compactMap { $0.platform?.description }
+                                .filter { $0 != "unknown/unknown" }
+                        }
+                        let availableStr = available.isEmpty ? "none" : available.joined(separator: ", ")
+                        throw ContainerizationError(
+                            .invalidArgument,
+                            message: "image \(reference) has no content for platform \(p.description); available platforms: \(availableStr)"
+                        )
+                    }
+                }
+            }
+
             // Write to stdout; otherwise write to the output file
             if output == nil {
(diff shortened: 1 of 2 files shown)


jglogan (approved):

Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -84,6 +84,27 @@ extension Application {
                 throw ContainerizationError(.invalidArgument, message: "failed to save image(s)")
             }
 
+            if let p {
```
jglogan, line 87: @knQzx Could you make a copy of the `testImageSaveAndLoad` integration test and trim it back so it checks the error path here?

jglogan (changes requested):

knQzx:
@jglogan added test

jglogan (approved):

jglogan:
@knQzx Thank you!

PR #1420, opened 2026-04-15
Fix ssh forwarding to use current SSH_AUTH_SOCK value

### LINKED ISSUE #357: [Request]: Mount a unix socket into a container.

As a `container` user, I can mount a unix socket file into a container, so that I can use the ssh agent on my host from a container.

```
container run --volume "${SSH_AUTH_SOCK}:/var/run/ssh-auth.sock" --env SSH_AUTH_SOCK=/var/run/ssh-auth.sock ...
```

### Code of Conduct

JaewonHur:
This PR fixes #357, passing `SSH_AUTH_SOCK` env variable from current terminal to the `SandboxService` so that the container can mount the correct ssh auth socket. For that, it introduces `env` parameters to `bootstrap` RPC of both `ContainersService` and `SandboxService`. This parameter is used only for passing `SSH_AUTH_SOCK` now, but can be extended to pass more runtime env variables.

This PR is a follow up PR of #1214.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Container run `--ssh` was inheriting `SSH_AUTH_SOCK` env variable from launchd, not from current terminal.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -474,7 +474,7 @@ public actor ContainersService {
                     id: id,
                     runtime: runtime
                 )
-                try await sandboxClient.bootstrap(stdio: stdio, allocatedAttachments: allocatedAttachments, env: env ?? [:])
+                try await sandboxClient.bootstrap(stdio: stdio, allocatedAttachments: allocatedAttachments, env: env)
 
                 try await self.exitMonitor.registerProcess(
                     id: id,
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -398,13 +398,13 @@ public actor ContainersService {
     }
 
     /// Bootstrap the init process of the container.
-    public func bootstrap(id: String, stdio: [FileHandle?], env: [String: String] = [:]) async throws {
+    public func bootstrap(id: String, stdio: [FileHandle?], env: [String: String]?) async throws {
         log.debug(
             "ContainersService: enter",
             metadata: [
                 "func": "\(#function)",
                 "id": "\(id)",
-                "env": "\(env)",
+                "env": "\(env ?? [:])",
             ]
         )
         defer {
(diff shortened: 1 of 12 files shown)


JaewonHur:
@chaveMac Hi!

Sorry to make a duplicate PR, but I opened a new one as it's been a while the original one's been updated.
It's mostly based on your PR, so if you don't mind, could I put you as the co-author?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -398,12 +398,13 @@ public actor ContainersService {
     }
 
     /// Bootstrap the init process of the container.
-    public func bootstrap(id: String, stdio: [FileHandle?]) async throws {
+    public func bootstrap(id: String, stdio: [FileHandle?], env: [String: String]) async throws {
```
katiewasnothere: Could we just have `env` be an optional here? Since env will probably be empty most of the time anyways, I feel like this being an optional makes the most sense. What do you think?

JaewonHur: Yeah sure!

hajimohammadsamirazizi-lab: Hello

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -398,12 +398,13 @@ public actor ContainersService {
     }
 
     /// Bootstrap the init process of the container.
-    public func bootstrap(id: String, stdio: [FileHandle?]) async throws {
+    public func bootstrap(id: String, stdio: [FileHandle?], env: [String: String] = [:]) async throws {
```
katiewasnothere: I thought we were making the env variable here an optional?

JaewonHur: Oh I see. sorry!

hajimohammadsamirazizi-lab: Hello

katiewasnothere (approved):

Issue #1404, opened 2026-04-08
[Request]: TECH DEBT: Refactor network resource management for 1.0.


jglogan:
### Feature or enhancement request details

We need to evolve our resource management in the client, API server, and plugins so that we have a maintainable codebase with good compatibility properties for 1.0.

The plan is to tackle this resource by resource, starting with Network. I've put together a checklist which will likely evolve as we work through this first resource:

- [ ] Conformance to ManagedResource
- [ ] Client is an instance with init(servicePrefix:) that performs syntactic validation on all input
- [ ] Client create(config:) -> FooResource
- [ ] Client list() -> [FooResource], sorted by ID
- [ ] Client get(id:) -> FooResource
- [ ] Client delete(id:)
- [ ] Client defines FooResource+XPC.swift with toXPC() -> XPCMessage and fromXPC(xpcMessage:) and round trip unit tests.
- [ ] Client defines FooService.swift protocol
- [ ] Server defines FooServing.swift protocol with FooService.swift implementation and FooTestService.swift implementation for round trip unit tests (for ensuring full XPC compatibility as API evolves).
- [ ] ListableResource conformance for CLI presentation in ContainerCommands
- [ ] No use of URL where FilePath should be used

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1421, opened 2026-04-15
Use NetworkResource for network management in API server.

jglogan:
- Part of #1404.
- Evolves API to use NetworkResource that conforms to ManagedResource.
- BREAKING CHANGE - compile time impact due to changes to Swift client API signatures.  No change to persistent data or API server XPC protocols.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context
Evolve API to 1.0 shape.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Builder/BuilderStart.swift
+++ b/Sources/ContainerCommands/Builder/BuilderStart.swift
@@ -272,7 +272,7 @@ extension Application {
             guard let defaultNetwork = try await networkClient.builtin else {
                 throw ContainerizationError(.invalidState, message: "default network is not present")
             }
-            guard case .running(_, _) = defaultNetwork else {
+            guard defaultNetwork.status.phase == "running" else {
                 throw ContainerizationError(.invalidState, message: "default network is not running")
             }
             config.networks = [
--- a/Sources/ContainerResource/Network/NetworkResource.swift
+++ b/Sources/ContainerResource/Network/NetworkResource.swift
@@ -1,0 +1,116 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerizationExtras
+import Foundation
+
+/// A network resource, representing a configured virtual network and its runtime status.
+///
+/// `NetworkResource` conforms to `ManagedResource` and separates the network's
+/// intrinsic configuration from its ephemeral runtime status — following the same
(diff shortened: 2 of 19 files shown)


Sources/ContainerCommands/Builder/BuilderStart.swift
```diff
@@ -269,7 +269,7 @@ extension Application {
             guard let defaultNetwork = try await networkClient.builtin else {
                 throw ContainerizationError(.invalidState, message: "default network is not present")
             }
-            guard case .running(_, _) = defaultNetwork else {
+            guard defaultNetwork.status.phase == "running" else {
```
ajemory, line 275: nit: "running" gets referenced as a raw string a few times, consider making a computed property or enum for created/running

Sources/ContainerResource/Network/NetworkResource.swift
```diff
@@ -0,0 +1,116 @@
...
+    public var labels: ResourceLabels { config.labels }
+
+    /// Returns `true` for a system-managed network that cannot be deleted by the user.
+    public var isBuiltin: Bool { labels.isBuiltin }
+
+    /// Returns `true` if `name` is a syntactically valid network identifier.
+    ///
+    /// Valid network names are lowercase alphanumeric strings of up to 63
+    /// characters, allowing dots, hyphens, and underscores in interior positions.
+    public static func nameValid(_ name: String) -> Bool {
```
ajemory, line 59: nit: this is the same regex for validating the network id in NetworkConfiguration, could consolidate the functions

Sources/ContainerResource/Network/NetworkResource.swift
```diff
@@ -0,0 +1,116 @@
...
+            self.init(config: config, networkStatus: status)
+        }
+    }
+}
+
+// MARK: - Codable
+
+extension NetworkResource {
+    enum CodingKeys: String, CodingKey {
+        case id
```
ajemory, line 99: `id` is also be in the config, correct? Is it at the top level for convenience (instead of `.config.id`)?

jglogan: Yes. I didn't want to mandate that every resource reports its ID via `config.id`.

For example, let's look at part of a `container image ls --format json | jq` today:

```json
  {
    "reference": "docker.io/library/ubuntu:24.04",
    "descriptor": {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:84e77dee7d1bc93fb029a45e3c6cb9d8aa4831ccfcc7103d36e876938d28895b",
      "size": 6688
    },
    "fullSize": "28.9 MB"
  },
```

If we adapt this to ManagedResource, there's no `config.id`, and we'll need to figure out what `id` and `name` for the resource should be. If we were to use the image digest for `id`, we'd need to relax the language in the protocol definition:

```swift
public protocol ManagedResource: Identifiable, Sendable, Codable {
    /// A 64 byte hexadecimal string, assigned by the system, that uniquely
    /// identifies the resource.
    var id: String { get }
```

Sources/ContainerResource/Network/NetworkResource.swift
jglogan: Good catch! Will fix to use `NetworkResource.nameValid(id)` in `NetworkConfiguration`.

Sources/ContainerCommands/Builder/BuilderStart.swift
jglogan: Let's use this for now, and we can refine this in a subsequent PR.

ajemory (approved):

Issue #608, opened 2025-09-13
[Request]: Use environment variables as system property source instead of UserDefaults.


jglogan:
### Feature or enhancement request details

While UserDefaults are idiomatic for macOS, they apply globally to all sessions, and in some cases session-level control of a property may be desirable.

Environment variables are idiomatic for Unix/Linux developer tools and can apply globally (rc/profile files) and per session.

For this reason we should consider replacing UserDefaults with environment variables. While we could include both as property sources, it results in needless complexity and could create confusing scenarios.

We'd need to look at the variables we have and see which could be problematic if changed per session. Some variables are used only in helpers so the values they use would be established when the helper starts and immutable afterward (whereas a UserDefault could be updated while a helper runs).

It'd be great to get some opinions on this before proceeding. I was thinking that ideally we'd have a system for defining system properties, where each property is enumerable, generic, validated, type-safe, and defines a chain of command for its backing store. 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1422, opened 2026-04-15
Move to TOML configuration for plugin configurations

noah-thor:
This change migrates plugin config format away from using JSON to use TOML.
- Addresses the plugin portion of discussion #1336

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Context within https://github.com/apple/container/discussions/1336

> We're looking to improve the user experience and overall functionality of setting defaults for container to better enable future use cases.
> 
> Today container uses macOS's [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) to configure settings needed at runtime. As mentioned in https://github.com/apple/container/issues/608, UserDefaults may be idiomatic for macOS, but they apply globally to all sessions and do not handle representing complex, hierarchical data well. Additionally, we currently have two ways of setting these defaults, either directly with macOS's defaults command or through container system property.
> 
> https://github.com/apple/container/issues/608 proposes moving to use environment variables in place of UserDefaults. However, we do not believe this is sufficient. Environment variables are not in a consistent location, are not sourced from data, and also do not handle representing complex, hierarchical data well.

As part of this migration to TOML for user settings we want to also move plugin settings to use TOML so that we can have a singular file format for `container` configurations.

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerPlugin/PluginConfig.swift
```diff
@@ -16,19 +16,20 @@
 
 //
 import Foundation
+import TOML
 
 /// PluginConfig details all of the fields to describe and register a plugin.
 /// A plugin is registered by creating a subdirectory  `<application-root>/user-plugins`,
 /// where the name of the subdirectory is the name of the plugin, and then placing a
-/// file named `config.json` inside with the schema below.
+/// file named `config.toml` (or `config.json`) inside with the schema below.
```
katiewasnothere: I think the PR is missing supporting `config.json`. We probably do need to support both for a period of time.

noah-thor: Will add JSON support as a fallback to maintain compatibility until we can fully deprecate json

Sources/ContainerPlugin/PluginFactory.swift
```diff
@@ -28,13 +27,34 @@ public protocol PluginFactory: Sendable {
...
     public init() {}
 
+    /// Returns the URL of the first config file found in `directory`, preferring TOML over JSON.
+    static func findConfigURL(in directory: URL) -> URL? {
+        let fm = FileManager.default
+        for filename in configFilenames {
+            let url = directory.appending(path: filename)
+            if fm.fileExists(atPath: url.path) {
+                if url.pathExtension == "json" {
+                    logger.info(
```
katiewasnothere: Thoughts on this being a warning?

Sources/ContainerPlugin/PluginFactory.swift
```diff
@@ -28,13 +27,34 @@ public protocol PluginFactory: Sendable {
 
 /// Default layout which uses a Unix-like structure.
 public struct DefaultPluginFactory: PluginFactory {
+    // Order matters: earlier entries take priority during config file discovery.
+    private static let configFilenames: [String] = ["config.toml", "config.json"]
+    private static let logger = Logger(label: "DefaultPluginFactory")
```
katiewasnothere: should we just pass the logger in from the APIServer or do we specifically want to make a new logger here?

Sources/ContainerPlugin/PluginFactory.swift
```diff
@@ -28,13 +27,34 @@ public protocol PluginFactory: Sendable {
...
+                if url.pathExtension == "json" {
+                    logger.info(
+                        "Plugin using legacy config.json; please migrate to config.toml",
+                        metadata: ["path": "\(url.path)"]
+                    )
+                }
+                return url
+            }
+        }
+        return nil
```
katiewasnothere, line 53: Maybe throw `notFound` instead?

noah-thor: It more or less conforms to the previous contract, in which `create` returns `Plugin?`.  If we want to use `ContainerizationError.notFound` we will have to add `Containerization` as a dependency of `ContainerPlugin`.

Sources/ContainerPlugin/PluginFactory.swift
noah-thor: Yeah that's a good idea

Sources/ContainerPlugin/PluginFactory.swift
vxnkjyffmq-code: Maybe so who knows

Sources/ContainerPlugin/PluginFactory.swift
noah-thor: Having a logger here should be fine since at all the call sites `Application` and `APIServer+Start` we initialize the logging facilities globally, so any downstream loggers should get created with the correct output.

noah-thor: Ill pass the logger to maintain consistent style, but think we should rethink that approach as perfile/class loggers can be useful

Sources/ContainerPlugin/PluginFactory.swift
katiewasnothere: We'd only need to import `ContainerizationError` as a dependency, which is already used elsewhere in the same package https://github.com/apple/container/blob/e710876ac84e00ae7e5a3f7bcaedb4bb3d8d4993/Sources/ContainerPlugin/ServiceManager.swift#L17

katiewasnothere (approved):
Small request but otherwise LGTM

PR #1424, opened 2026-04-16
Explicitly pass default subnets into network creation

noah-thor:
This changes the semantics around `ReservedVmnetNetwork` and `AllocationOnlyVmnetNetwork` to only create ipv4/6 subnets for networks when the values are explicitly passed in the `NetworkConfiguration`. Previously if the values in the `NetworkConfiguration` were not present it would attempt to source them via `DefaultsStore`.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

The change solves an issue that caused `container network create <name>` to fail when `UserDefaults` was set to a value that is identical to the hard coded default ("192.168.64.1/24"/"fd00::/64"). It would fail because a network with these parameters would already exist, and thus could not be reserved.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
+++ b/Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
@@ -14,7 +14,6 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-import ContainerPersistence
 import ContainerResource
 import ContainerXPC
 import ContainerizationError
@@ -23,6 +22,9 @@ import Foundation
 import Logging
 
 public actor AllocationOnlyVmnetNetwork: Network {
+    // The IPv4 subnet to be used if none explicitly passed in the `NetworkConfiguration`
+    private static let defaultIPv4Subnet = try! CIDRv4("192.168.64.1/24")
+
     private let log: Logger
     private var _state: NetworkState
 
@@ -36,8 +38,8 @@ public actor AllocationOnlyVmnetNetwork: Network {
             throw ContainerizationError(.unsupported, message: "invalid network mode \(configuration.mode)")
         }
 
-        guard configuration.ipv4Subnet == nil else {
-            throw ContainerizationError(.unsupported, message: "IPv4 subnet assignment is not yet implemented")
+        guard configuration.ipv6Subnet == nil else {
+            throw ContainerizationError(.unsupported, message: "IPv6 subnet assignment is not yet implemented")
         }
 
         self.log = log
@@ -65,8 +67,8 @@ public actor AllocationOnlyVmnetNetwork: Network {
             ]
         )
 
-        let defaultIPv4Subnet = try CIDRv4(DefaultsStore.get(key: .defaultSubnet))
-        let ipv4Subnet = configuration.ipv4Subnet ?? defaultIPv4Subnet
+        let ipv4Subnet = configuration.ipv4Subnet ?? Self.defaultIPv4Subnet
+
         let gateway = IPv4Address(ipv4Subnet.lower.value + 1)
         let status = NetworkStatus(
             ipv4Subnet: ipv4Subnet,
--- a/Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
+++ b/Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
@@ -14,7 +14,6 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-import ContainerPersistence
 import ContainerResource
 import ContainerXPC
 import Containerization
@@ -117,11 +116,8 @@ public final class ReservedVmnetNetwork: Network {
 
         vmnet_network_configuration_disable_dhcp(vmnetConfiguration)
 
-        // subnet priority is CLI argument, UserDefault, auto
-        let defaultIpv4Subnet = try DefaultsStore.getOptional(key: .defaultSubnet).map { try CIDRv4($0) }
-        let ipv4Subnet = configuration.ipv4Subnet ?? defaultIpv4Subnet
-        let defaultIpv6Subnet = try DefaultsStore.getOptional(key: .defaultIPv6Subnet).map { try CIDRv6($0) }
-        let ipv6Subnet = configuration.ipv6Subnet ?? defaultIpv6Subnet
+        let ipv4Subnet = configuration.ipv4Subnet
+        let ipv6Subnet = configuration.ipv6Subnet
 
         // set the IPv4 subnet if the caller provided one
         if let ipv4Subnet {
--- a/Sources/APIServer/APIServer+Start.swift
+++ b/Sources/APIServer/APIServer+Start.swift
@@ -19,9 +19,11 @@ import ContainerAPIClient
 import ContainerAPIService
 import ContainerLog
 import ContainerNetworkService
+import ContainerPersistence
 import ContainerPlugin
 import ContainerResource
 import ContainerXPC
+import ContainerizationExtras
 import DNSServer
 import Foundation
 import Logging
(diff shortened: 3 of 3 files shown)


Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
```diff
@@ -65,8 +64,13 @@ public actor AllocationOnlyVmnetNetwork: Network {
             ]
         )
 
-        let defaultIPv4Subnet = try CIDRv4(DefaultsStore.get(key: .defaultSubnet))
-        let ipv4Subnet = configuration.ipv4Subnet ?? defaultIPv4Subnet
+        guard let ipv4Subnet = configuration.ipv4Subnet else {
```
jglogan: This will cause regressions because we don't require macOS 15 users to set `default.subnet`, and now this will fail unless it is set.

I think the easiest would be to define a static constant:

```swift
private static let defaultIPv4Subnet = "192.168.64.1/24"
```

and here, just do

```swift
let ipv4Subnet = configuration.ipv4Subnet.map { try CIDRv4(Self.defaultIPv4Subnet) }
```

jglogan (changes requested):

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
noah-thor: I overlooked the `guard` within the constructor, so with both of these in place it will not work. (also seems like we might want to add a check for not allowing users to provide ipv6)

I do still think we should remove the fallback to `DefaultsStore.get(key: .defaultSubnet)` and do that logic in the `APIServer.Start.initializeNetwork` service as done in this change. That would mean removing both this `guard` and the one within `init`. But that would open the door for macOS 15 users calling `container network create` which an issue. To recreate that error behavior for macOS 15/AllocaitonOnlyVmnet, how do you feel about moving that check into `NetworkCreate` as part of the cli itself?

jglogan:
> But that would open the door for macOS 15 users calling container network create which an issue.

The CLI excludes the `network` subcommand on macOS 15: https://github.com/apple/container/blob/main/Sources/ContainerCommands/Application.swift#L200.

Before we do any other checking in the client, we should probably guard `macOS 26` availability in the API server for the network API calls. Create and delete for sure; we could allow list/inspect but today we don't have any way to provide useful runtime state for an allocation-only network.

jglogan:
> I do still think we should remove the fallback to DefaultsStore.get(key: .defaultSubnet) and do that logic in the APIServer.Start.initializeNetwork service as done in this change.

Yes, but it should be `getOptional` (remember that we don't really support macOS 15 networking, vmnet there doesn't give us everything we need, and on macOS 26 the best usability comes when the user doesn't specify a subnet). We don't want to have some sort of special-casing in the API server like "if it's this network type, use getOptional but for this type, use get". There's no need for the allocation-only network helper to use the defaults store just to supply a fallback for this one case; a private constant in the helper is sufficient.

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
jglogan: > I overlooked the `guard` within the constructor

Yep, you can remove that now such that if the subset is set, it meant that getOptional in the API server found a user override, and if not, we'll use the constant value "192.168.64.1/24" to be defined here.

Tests/ContainerNetworkServiceTests/AllocationOnlyVmnetNetworkTest.swift
```diff
@@ -0,0 +1,39 @@
...
+//===----------------------------------------------------------------------===//
+
+import ContainerResource
+import ContainerizationError
+import ContainerizationExtras
+import Testing
+
+@testable import ContainerNetworkService
+
+struct AllocationOnlyVmnetNetworkTest {
```
jglogan: This test becomes unnecessary.

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
```diff
@@ -23,6 +22,9 @@ import Foundation
 import Logging
 
 public actor AllocationOnlyVmnetNetwork: Network {
+    // The IPv4 subnet to be used if none explicitly passed in the `NetworkConfiguration`
+    private static let defaultIPv4Subnet = try! CIDRv4("192.168.64.1/24")
```
jglogan, line 26: We should avoid force-try, though here it should always work. Fine for now though.

If we ever turn on lint checks we'll either have to add exclusions for these, or rework the code.

Unfortunately `Prefix` in `ContainerizationExtras` is `init?` or we could do something like:

```
    private static let defaultIPv4Subnet = CIDRv4(IPv4Address([192, 168, 64, 1]), prefix: .ipv4(24))
```

Tests/ContainerNetworkServiceTests/AllocationOnlyVmnetNetworkTest.swift
```diff
@@ -0,0 +1,80 @@
...
+    @Test func testStartWithExplicitIPv4Subnet() async throws {
```
jglogan: While we have precedent (image stuff) I believe for unit tests that rely on system side effects, I'd prefer that our unit tests don't do this. I don't think we need these if we have integration test coverage on these paths.

jglogan (commented):
One comment regarding the unit tests.

Tests/ContainerNetworkServiceTests/AllocationOnlyVmnetNetworkTest.swift
noah-thor: Yeah I'll just drop these

noah-thor:
In the latest commit
* dropped the unit tests to avoid the state pollution pointed out

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
noah-thor: am in favor of leaving the forced unwrap for now, the string format reads better in my option and the surface area for errors should be caught in tests. If we do decide to make that a linter rule we should explore a standard way of handling such "known safe" constants and or have a override mechanism

Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
```diff
@@ -117,11 +116,8 @@ public final class ReservedVmnetNetwork: Network {
 
         vmnet_network_configuration_disable_dhcp(vmnetConfiguration)
 
-        // subnet priority is CLI argument, UserDefault, auto
-        let defaultIpv4Subnet = try DefaultsStore.getOptional(key: .defaultSubnet).map { try CIDRv4($0) }
-        let ipv4Subnet = configuration.ipv4Subnet ?? defaultIpv4Subnet
-        let defaultIpv6Subnet = try DefaultsStore.getOptional(key: .defaultIPv6Subnet).map { try CIDRv6($0) }
-        let ipv6Subnet = configuration.ipv6Subnet ?? defaultIpv6Subnet
+        let ipv4Subnet = configuration.ipv4Subnet
+        let ipv6Subnet = configuration.ipv6Subnet
```
katiewasnothere, line 120: Don't we still need to fall back in the reserved case to the internal default if the user didn't set one?

noah-thor: The `if let` blocks below handle the case if a user does not provide a subnet v4/v6. In these cases `vmnet_network_create` will automatically allocate a v4/v6 based on my understanding of the API.

katiewasnothere: Ah okay, I see, sgtm

katiewasnothere (approved):

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
hajimohammadsamirazizi-lab: Hello

jglogan (approved):

Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
hajimohammadsamirazizi-lab: H

PR #1425, opened 2026-04-16
Move to TOML configuration for defaults

noah-thor:
This change migrates away from using `UserDefaults`, instead providing a TOML configuration mechanism for user configurable settings. All exisitng features of user configurability are maintained. However, users will have to migrate any settings they have configured in the `UserDefaults` into TOML for these settings to take effect.

Breaking changes:
* `container system property get` is removed in favor of users directly utilizing `container system property list --format toml | jq<>`. 
* `container system property set` is removed since the TOML configuration is effectively immutable during the lifetime of the `container` daemon. Uses can edit the TOML they have in their home directory, however no changes will take effect until the daemon is restarted via `container system stop && container system start`
* `container system property list --format table` is removed as generating tabular format is non-trivial and the new TOML format is intended to be human readable

Addresses discussion (https://github.com/apple/container/discussions/1336)

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [x] Breaking change
- [x] Documentation update

## Motivation and Context
Context within https://github.com/apple/container/discussions/1336

> We're looking to improve the user experience and overall functionality of setting defaults for container to better enable future use cases.
> 
> Today container uses macOS's [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) to configure settings needed at runtime. As mentioned in https://github.com/apple/container/issues/608, UserDefaults may be idiomatic for macOS, but they apply globally to all sessions and do not handle representing complex, hierarchical data well. Additionally, we currently have two ways of setting these defaults, either directly with macOS's defaults command or through container system property.
> 
> https://github.com/apple/container/issues/608 proposes moving to use environment variables in place of UserDefaults. However, we do not believe this is sufficient. Environment variables are not in a consistent location, are not sourced from data, and also do not handle representing complex, hierarchical data well.

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [X] Added/updated docs


Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
```diff
@@ -117,14 +120,12 @@ public final class ReservedVmnetNetwork: Network {
...
         vmnet_network_configuration_disable_dhcp(vmnetConfiguration)
 
-        // subnet priority is CLI argument, UserDefault, auto
-        let defaultIpv4Subnet = try DefaultsStore.getOptional(key: .defaultSubnet).map { try CIDRv4($0) }
-        let ipv4Subnet = configuration.ipv4Subnet ?? defaultIpv4Subnet
-        let defaultIpv6Subnet = try DefaultsStore.getOptional(key: .defaultIPv6Subnet).map { try CIDRv6($0) }
-        let ipv6Subnet = configuration.ipv6Subnet ?? defaultIpv6Subnet
+        // subnet priority is CLI argument, config file, auto
+        let ipv4Subnet = configuration.ipv4Subnet ?? containerSystemConfig.network.subnet
+        let ipv6Subnet = configuration.ipv6Subnet ?? containerSystemConfig.network.subnetv6
```
noah-thor: This will cause the integration tests to fail as the reservation of ipv4/6 subnets will always occur, even if none are specified via the `configuration`. To address this I opened a [PR](https://github.com/apple/container/pull/1424) that does not have any fallback defaults when `configuration.ipv{4,6}Subnet` is `nil`. Once that PR is merged I can rebase and the tests will pass.

Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -0,0 +1,248 @@
...
+                && !(child.value is Measurement<UnitInformationStorage>)
+            {
+                results += flatten(child.value, prefix: key)
+            } else {
+                results.append((key: key, type: "\(type(of: child.value))", value: "\(child.value)"))
+            }
+        }
+        return results
+    }
+}
```
noah-thor: I don't really like this, but for now I think we can take on this debt. Or if there are more flexible suggestions please let me know.

Sources/APIServer/APIServer+Start.swift
```diff
@@ -37,6 +38,9 @@ extension APIServer {
         static let listenAddress = "127.0.0.1"
         static let localhostDNSPort = 1053
         static let dnsPort = 2053
+        private static let containerSystemConfig: ContainerSystemConfig = try! SystemRuntimeOptions.loadConfig(
```
manojmahapatra: Suggestion: force unwraps are generally unsafe,  I would not use `try!`

Sources/ContainerPersistence/ContainerSystemConfig.swift
manojmahapatra: Suggestion: So `flatten()` uses Mirror with a hardcoded exclusion list (String, CIDRv4, and others) to decide which types are leaves vs containers. I think this is fragile as any new leaf type that isn't added to the list will be silently producing incorrect output.

Since all config types already conform to Codable, I think you could just use a custom Encoder that leverages the same encoding paths as TOMLDecoder and hence no exclusion list needed.

Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -0,0 +1,248 @@
...
+import CVersion
+import ContainerVersion
+import ContainerizationExtras
+import Foundation
+
+/// Top-level configuration decoded from runtime-config.toml.
+///
+/// Each section maps to a nested struct. Missing keys fall back to
+/// hardcoded defaults via custom `init(from:)` implementations.
+final public class ContainerSystemConfig: Codable, Sendable {
```
manojmahapatra, line 26: Suggestion: There are all immutable data models and can be just struct instead of class

Sources/APIServer/APIServer+Start.swift
manojmahapatra: Additionally, containerSystemConfig is being repeated at many places, in addition to I'm seeing every function that needs config accepts it as a parameter, and every function that calls those functions also accepts,  even if it doesn't use the config itself. I think we can do better/different by not passing around into the chain of functions, but hoist it up as a higher level @Environment kinda property.

Sources/ContainerPersistence/ContainerSystemConfig.swift
noah-thor: I had `struct` in the initial prototype, however switched over to `class` to enable passing around by reference instead of copying when necessary to pass through to functions.

Sources/ContainerPersistence/ContainerSystemConfig.swift
noah-thor: The problem is that CIDRv4 and CIDRv6 both encode to formats that are not desirable for TOML output. So this class is really more like `Decodable` and probably not actually conformant to `Codabke` in terms of symmetry.

However I will take another look at adding a custom encoder for these types, because you are right that using the TOMLEncoder would be an improvement.

Sources/APIServer/APIServer+Start.swift
katiewasnothere: @manojmahapatra As an alternative to the current approach in this PR, we also considered having a service on APIServer that managed defaults. We decided against this for a few reasons: we want to avoid services needing to call back to APIServer unless absolutely necessary and we believe configurations values should not change during the lifetime of a service instance (with some rare exceptions). That said, we're still discussing how we want to extend the current approach to allow for user level overrides, such as with environment variables.

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -117,8 +113,8 @@ extension ClientImage {
     private static let dockerRegistryHost = "registry-1.docker.io"
     private static let defaultDockerRegistryRepo = "library"
 
-    public static func normalizeReference(_ ref: String) throws -> String {
-        guard ref != Self.initImageRef else {
+    public static func normalizeReference(_ ref: String, containerSystemConfig: ContainerSystemConfig) throws -> String {
+        guard ref != containerSystemConfig.image.initImage else {
```
katiewasnothere: Thoughts on doing some refactoring here to avoid the client needing the system config?

Sources/ContainerPersistence/ContainerSystemConfig.swift
noah-thor: Actually looking back at this, this is not an issue with the `CIDRv4/6` being encodable. It was a problem with trying to output the table format. There is not a built-in, nor simple way to flatten out a `Codable` class without creating a custom "TableEncoder". That would add a bunch of code for questionable value. I propose just dropping this all together and letting users leverage the `json` and `toml` formats for reference instead.

Sources/Services/ContainerAPIService/Client/ClientImage.swift
noah-thor: We could probably take another pass at that, I was aiming to avoid changing code more than required. In most cases opting for dropping the config directly in place of `DefaultsStore`. However for this PR I would like to avoid adding any more scope

Sources/APIServer/APIServer+Start.swift
manojmahapatra: IIUC are you implying to have some kind of runtime (immutable) context and inject that at module boundaries? (eg SystemStart)

Sources/ContainerPersistence/ContainerSystemConfig.swift
manojmahapatra: hmm.. I took another look. Dropping table support would be a regression, so I don’t think that’s ideal. As a middle ground, could we introduce a PropertyCatalog that defines the supported system properties (for example, build.rosetta, build.cpus) and have all output paths consume that catalog?

A catalog does have downsides: it can get out-of-sync from config fields if not maintained, and every new property requires an update. But at this scale, it seems like a better tradeoff than the current flatten (by reflection) approach.

@katiewasnothere what do you think?

Sources/APIServer/APIServer+Start.swift
katiewasnothere: That's what I would prefer. But we do still want to support some user level overrides, so how to make that experience nice is still an open question. What do you think?

Sources/ContainerPersistence/ContainerSystemConfig.swift
katiewasnothere: In the world where we have a system config, I don't really see the benefit of `property list` in general other than for if/when we're doing merges of multiple sources (like if we supported layering the system config, a user level config, and env variables, for example). But even in that case, my thoughts are that TOML is designed to be human readable, so other than consistency with the other high level list commands like `container ls`, `network ls`, etc, I don't know if I see a great reason to continue to support outputting these values as a table. 

The repository is still only in a minor version and we are certainly trying to avoid breaking changes, but we do also still want to be able to adjust the CLI and general APIs to a shape we'll be happy with longer term. 

That said, I'm open to either option if we think that consistency with other commands should be prioritized. I'm curious to hear @jglogan's thoughts on this.

Sources/APIServer/APIServer+Start.swift
manojmahapatra: That makes sense, thanks for clarifying. I think explicit user overrides are the right direction; probably the other open question is what would be the precedence order.  (we need to define that)

So, a reasonable approach could be then to add a single central config resolver that builds a final config with an explicit order?

Sources/ContainerPersistence/ContainerSystemConfig.swift
noah-thor: I have provided an updated implementation for the table output instead of dropping it. It still relies on a very small amount of reflection, but that is required for `PropertyGet` regardless. The downside is that the key space is hardcoded. For sake of this PR I think that is an acceptable tradeoff, however it may become something we want to change as the configurable options grow/shrink/change.

noah-thor:
I addressed all ongoing discussions as part of this commit. I also noticed that the `Measurement` swift type does not play very nicely with the `TOMLEncoder` in terms of human readability, so added a small wrapper around that type to fix the ux.

Sources/ContainerPersistence/ContainerSystemConfig.swift
jglogan: > I don't really see the benefit of `property list` in general other than for if/when we're doing merges of multiple sources

Agree. I think this has to be a breaking change. The system properties are changing to a tree-valued structure where before we had a collection of properties that happened to have keys that used a domain-name like convention.

I'd think that we could have `system property some-verb` that outputs the serialized property tree. Initially this could be just JSON (convenient for post-processing through `jq`, `python`, et al). Later we could add other format options as needed.

Sources/APIServer/APIServer+Start.swift
katiewasnothere: Assuming we support user overrides with a file and with environment variables, I think the precedence would be environment variables -> user overrides file -> system config. The issue with environment variables is that they can change over the lifetime of `container` so we can't just read them once to create the final config :/

noah-thor:
Pushed new commit. Will resolve conflicts once everything else looks good.

New revision changes a few things:
* Removes `container system property get <>` in favor of recommending users leverage `container system property list --format json | jq <>`
* Removes `container system property list --format table` as maintaining a table format was not simple, and the new toml format is a suitable replacement for human readers.

Sources/APIServer/APIServer+Start.swift
manojmahapatra: yeah, that definitely had crossed my mind, unless we bubble up some kind of hot reloading of config file on demand. I'm not even sure if this is the right direction or hot reloading of config file even makes sense.

noah-thor:
Pushed new commit:
* fixed issue with `DNSConfig` providing a `"test"` default changing behavior of request scheme
* removed static force unwraps in favor of a `try let` in `run()`
* rebased

Sources/APIServer/APIServer+Start.swift
jglogan: > I'm not even sure if this is the right direction or hot reloading of config file even makes sense.

@manojmahapatra Agree...I think it's "we aren't going to need it today". We can start simple and try not to paint ourselves into a corner and evolve from there.

We can move slowly on env variables as well. Here's a rough snapshot of what we have right now:

| NAME | WHERE | NOTES |
|---|---|---|
| CONTAINER_{APP,INSTALL,LOG}_ROOT | services | plumbed from `system start` args into launchd.plist files, no TOML config needed |
| CONTAINER_DEBUG | CLI | commands check `--debug`, then this |
| CONTAINER_DEBUG_LAUNCHD_LABEL | services | plumbed into apiserver environment at `system start`, see #1101, no TOML config needed |
| CONTAINER_DEFAULT_PLATFORM | CLI | similar to `DOCKER_DEFAULT_PLATFORM`, TOML configuration could also make sense here but not immediately required |
| CONTAINER_REGISTRY_{HOST,TOKEN,USER} | services | plumbed into apiserver environment at `system start`, used for limited cases only, no TOML config needed |

None of these require TOML config today, so we can treat them as orthogonal for now.

As far as our existing properties go, none today have environment overrides AFAIK, so we can decide later if say, we want to be able to do something like `export CONTAINER_MEMORY=8g` in one shell session and have `container run` evaluate `--memory`, then `CONTAINER_MEMORY`, then the user TOML, then the system TOML.

docs/how-to.md
```diff
@@ -637,27 +631,43 @@ Check the VM boot logs to confirm your custom init code executed:
...
+rosetta = true
+
+[container]
+cpus = 4
+memory = "1gb"
+
+[dns]
+domain = "test"
+
+[image]
```
jglogan: Wondering that you think about pivoting the image properties into other categories, like:

```toml
[build]
cpus = 2
memory = "2048mb"
rosetta = true
image = "ghcr.io/apple/container-builder-shim/builder:0.11.0"

...

[vminit]
image = "ghcr.io/apple/containerization/vminit:0.30.1"
```

docs/tutorial.md
```diff
@@ -105,16 +105,15 @@ Use the `--help` flag to see which abbreviations exist.
 
 ### Set up a local DNS domain (optional)
 
-`container` includes an embedded DNS service that simplifies access to your containerized applications. If you want to configure a local DNS domain named `test` for this tutorial, run:
+`container` includes an embedded DNS service that simplifies access to your containerized applications. The default DNS domain is `test`. To configure the macOS resolver for this domain, run:
```
jglogan: The "default DNS domain" should be undefined until it is overridden by the user. The original text is correct.

jglogan: The system TOML could define `test` as the backstop instead of leaving it undefined. I think the main effect will be that there will be `{id}.test.` address records for every container even if a domain resource isn't set up, which is a little odd.

Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -0,0 +1,219 @@
...
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.builder = try container.decodeIfPresent(String.self, forKey: .builder) ?? Self.defaultBuilder
+        self.initImage = try container.decodeIfPresent(String.self, forKey: .initImage) ?? Self.defaultInitImage
+    }
+}
+
+final public class KernelConfig: Codable, Sendable {
+    public static let defaultBinaryPath = "opt/kata/share/kata-containers/vmlinux-6.18.15-186"
```
jglogan, line 153: Is there only a user TOML file for this PR, and no system TOML that we include in our distribution?

docs/tutorial.md
jglogan: Looking down in the system config, the domain setting is optional, defaulting to nil, so it seems like we should roll back this one change.

BUILDING.md
```diff
@@ -80,10 +80,11 @@ to prepare your build environment.
     >
     > **Note:** If you have already run `swift package edit`, whether intentionally or by accident, follow the steps in the next section to restore the normal `containerization` dependency. Otherwise, the modified `Package.swift` file will not work, and the project may fail to build.
 
-5. If you want `container` to use any changes you made in the `vminit` subproject of Containerization, update the system property to use the locally built init filesystem image:
+5. If you want `container` to use any changes you made in the `vminit` subproject of Containerization, set the init image in your runtime configuration file at `~/.config/container/runtime-config.toml`:
 
-    ```bash
-    container system property set image.init vminit:latest 
+    ```toml
+    [image]
```
jglogan: Update this if we do restructure.

BUILDING.md
```diff
@@ -133,14 +130,20 @@ To test changes that require the `container-builder-shim` project:
...
 ```bash
 container build -t builder .
-container system property set image.builder builder:latest
 container rm -f buildkit
 ```
 
+Add the following to your `~/.config/container/runtime-config.toml`:
+
+```toml
+[image]
```
jglogan: update if we move images to `build` and `vminitd` categories

Sources/Services/ContainerAPIService/Client/RequestScheme.swift
```diff
@@ -39,22 +40,23 @@ public enum RequestScheme: String, Sendable {
 
     /// Returns the prescribed protocol to use while making a HTTP request to a webserver
     /// - Parameter host: The domain or IP address of the webserver
+    /// - Parameter dnsDomain: The DNS domain used for container name resolution
     /// - Returns: RequestScheme
-    public func schemeFor(host: String) throws -> Self {
+    public func schemeFor(host: String, dnsDomain: String?) throws -> Self {
```
jglogan: `internalDnsDomain` might be clearer

Sources/Services/ContainerAPIService/Client/RequestScheme.swift
```diff
@@ -39,22 +40,23 @@ public enum RequestScheme: String, Sendable {
...
         case .auto:
-            return Self.isInternalHost(host: host) ? .http : .https
+            return Self.isInternalHost(host: host, dnsDomain: dnsDomain) ? .http : .https
         }
     }
 
     /// Checks if the given `host` string is a private IP address
     /// or a domain typically reachable only on the local system.
-    private static func isInternalHost(host: String) -> Bool {
+    private static func isInternalHost(host: String, dnsDomain: String?) -> Bool {
```
jglogan: same here

Sources/ContainerPersistence/MemorySize.swift
```diff
@@ -0,0 +1,56 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// This is a thin wrapper around Measurement<UnitInformationStorage> to enable
+/// better Codable implementations for user provided options. With this wrapper
+/// values will get encoded and decoded from the format "1g" or "10mb".
+public struct MemorySize: Codable, Sendable, Equatable, CustomStringConvertible {
```
jglogan, line 22: Would it make sense to generate some rudimentary unit tests here?

Sources/ContainerPersistence/MemorySize.swift
```diff
@@ -0,0 +1,56 @@
...
+    public var description: String { formatted }
+
+    public let measurement: Measurement<UnitInformationStorage>
+
+    public init(_ string: String) throws {
+        self.measurement = try .parse(parsing: string)
```
jglogan, line 28: Is this case sensitive?

Sources/ContainerPersistence/SystemRuntimeOptions.swift
```diff
@@ -0,0 +1,109 @@
...
+
+import ContainerizationError
+import Foundation
+import Logging
+import SystemPackage
+import TOML
+
+private let log = Logger(label: "SystemRuntimeOptions")
+
+/// TOML-backed configuration loader.
```
jglogan, line 25: Is this how it works?

- `container system start` copies the config to the application root so that the API server and any service plugins is spawns have a consistent system view
- CLI commands use the same application root config, such that the user would need to stop/start the container service if they wanted to make a change to a setting that impacts CLI command behavior

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -117,8 +113,8 @@ extension ClientImage {
     private static let dockerRegistryHost = "registry-1.docker.io"
     private static let defaultDockerRegistryRepo = "library"
 
-    public static func normalizeReference(_ ref: String) throws -> String {
-        guard ref != Self.initImageRef else {
+    public static func normalizeReference(_ ref: String, containerSystemConfig: ContainerSystemConfig) throws -> String {
```
jglogan, line 116: wondering whether these utility functions should couple to system config like this, or should do something like:

```swift
public static func normalizeReference(_ ref: String, passthroughRefs: Set<String>, defaultDomain: String) throws -> String {
```

Sources/Services/ContainerAPIService/Client/ClientImage.swift
```diff
@@ -139,10 +135,10 @@ extension ClientImage {
         return updatedReference.description
     }
 
-    public static func denormalizeReference(_ ref: String) throws -> String {
+    public static func denormalizeReference(_ ref: String, containerSystemConfig: ContainerSystemConfig) throws -> String {
```
jglogan, line 138: same here, and so on

Sources/ContainerBuild/BuildPipelineHandler.swift
```diff
@@ -30,7 +30,9 @@ public actor BuildPipeline {
             [
                 try BuildFSSync(URL(filePath: config.contextDir)),
                 try BuildRemoteContentProxy(config.contentStore),
-                try BuildImageResolver(config.contentStore, quiet: config.quiet, output: config.terminal?.handle ?? FileHandle.standardError, pull: config.pull),
+                try BuildImageResolver(
+                    config.contentStore, quiet: config.quiet, output: config.terminal?.handle ?? FileHandle.standardError, pull: config.pull,
```
jglogan: nit, one parameter per line for readability

Sources/Services/ContainerAPIService/Client/ClientImage.swift
katiewasnothere: I previously suggested we do some refactoring specifically for ClientImage to try to avoid needing to reference the system config in it at all. We agreed that would be done in a follow up PR https://github.com/apple/container/pull/1425#discussion_r3113959737

Sources/ContainerCommands/Builder/BuilderStart.swift
```diff
@@ -191,7 +195,7 @@ extension Application {
                 }
             }
 
-            let useRosetta = DefaultsStore.getBool(key: .buildRosetta) ?? true
+            let useRosetta = containerSystemConfig.build.rosetta
```
jglogan, line 198: I think there's a reason we made `buildRosetta` optional but I don't recall exactly why.

Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
```diff
@@ -44,11 +45,13 @@ public final class ReservedVmnetNetwork: Network {
 
     private let stateMutex: Mutex<State>
     private let log: Logger
+    private let containerSystemConfig: ContainerSystemConfig
```
jglogan: Are we using the system configuration here? Do we need any changes in this file?

Sources/Services/ContainerNetworkService/Server/AllocationOnlyVmnetNetwork.swift
```diff
@@ -26,12 +27,14 @@ public actor AllocationOnlyVmnetNetwork: Network {
     private static let defaultIPv4Subnet = try! CIDRv4("192.168.64.1/24")
 
     private let log: Logger
+    private let containerSystemConfig: ContainerSystemConfig
```
jglogan: Are these changes necessary?

Sources/ContainerCommands/System/SystemStart.swift
```diff
@@ -147,15 +153,11 @@ extension Application {
             }
         }
 
-        private func installDefaultKernel() async throws {
-            let kernelDependency = Dependencies.kernel
-            let defaultKernelURL = kernelDependency.source
-            let defaultKernelBinaryPath = DefaultsStore.get(key: .defaultKernelBinaryPath)
-
+        private func installDefaultKernel(kernelURL: URL, kernelBinaryPath: String) async throws {
```
jglogan, line 159: Is this really `installDefaultKernel()` anymore? It just installs a kernel.

docs/how-to.md
noah-thor: Yeah I think this is a good evolution, I can make that change

docs/tutorial.md
noah-thor: Good catch, yes this one was a bit of a headache as some places were using `get` some were using `getOptional` so previous commits had a hardcoded default of "test" but that resulted in changing behavior

Sources/ContainerPersistence/MemorySize.swift
noah-thor: I can, I think the main new surface area is the `formatted` logic, the rest relies heavily on `Measurement+Parse` with has pretty good coverage via `Tests/ContainerAPIClientTests/Measurement+ParseTests.swift`

Sources/ContainerPersistence/MemorySize.swift
noah-thor: It should be case-insensitive https://github.com/apple/container/blob/main/Sources/Services/ContainerAPIService/Client/Measurement%2BParse.swift#L46

Sources/ContainerCommands/Builder/BuilderStart.swift
noah-thor: I will fix this one as it might change behavior, not having `getOptionalBool` tricked me into thinking this was not an optional.

Sources/Services/ContainerNetworkService/Server/ReservedVmnetNetwork.swift
noah-thor: We were prior to the rebase, I will clean these up. I am surprised the linter doesn't catch unused variables/imports.

Sources/ContainerPersistence/ContainerSystemConfig.swift
noah-thor: That is correct, there is only one TOML that lives in users home

Sources/ContainerCommands/Builder/BuilderStart.swift
noah-thor: On second pass it looks like `DefaultsStore.getBool(key: .buildRosetta) ?? true` is equivalent to defaulting to `true` as far as I can tell

Sources/ContainerBuild/BuildPipelineHandler.swift
noah-thor: surprised the `make fmt` didn't catch this

Sources/ContainerPersistence/MemorySize.swift
noah-thor: Added tests

Sources/Services/ContainerAPIService/Client/RequestScheme.swift
noah-thor: Fixed

jglogan (approved):

katiewasnothere (approved):

noah-thor:
Merged from `main` to update PR

jglogan (approved):

jglogan (approved):

karaday411-sketch (commented):
R

Issue #1413, opened 2026-04-12
[Request]: Add drivers to `container volume` for network NFS or SMB shares


tonycoco:
### Feature or enhancement request details

For example:

- `container volume create --driver smb ...`
- `container volume create --driver nfs ...`

Create the SMB volume once:

```
  container volume create --driver smb \
    --opt share=//server/share \
    --opt username=user \
    --opt password=secret \
    myshare
```

Use it exactly like any other named volume:
  
```
  container run -v myshare:/data alpine ls /data
```

or with container create:

```
container create -v myshare:/data --name mycontainer alpine
```

This would make mounts not hop in the virtiofs.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tonycoco:
#1428 adds this feature

PR #1428, opened 2026-04-18
Add SMB volume driver via direct guest CIFS mount

tonycoco:
Add NFS and SMB volume drivers via direct guest mounts

Adds `--driver nfs` and `--driver smb` to `container volume create`. Both drivers mount network shares directly inside the Linux guest rather than going through `virtiofs`, eliminating the host-side mount hop.

When a volume is created with `--driver smb` or `--driver nfs`, no block image is created. The share path and options are stored as volume metadata. At container start, `Utility.swift` resolves the volume into a `Filesystem.smb` or `Filesystem.nfs` mount, which `SandboxService.swift` translates into a `.any` guest mount with type `cifs` or `nfs` respectively.

Usage:

## SMB

```
container volume create --driver smb \
  --opt share=//192.168.1.1/Media \
  --opt username=user \
  --opt password=secret \
  --opt vers=3.0 \
  myshare
```

## NFS

```
container volume create --driver nfs \
  --opt share=nas.local:/exports/data \
  --opt addr=nas.local \
  --opt vers=3 \
  --opt proto=tcp \
  myexport
```

```
container run -v myshare:/media alpine ls /media
```

Boolean mount flags (e.g. `nolock`, `mfsymlinks`) can be passed as `--opt nolock=` with an empty value and are serialized as bare flags in the mount data string.

## Kernel requirement

This depends on a pending change to apple/containerization. The default guest kernel ships with `CONFIG_CIFS=y` disabled. A PR to that repo is needed to enable it in `kernel/config-arm64` — without it, SMB mounts fail at runtime with `errno 19 (ENODEV)`. NFS (`CONFIG_NFS_FS`) is already enabled in the current kernel config but is not compiled into the shipped kernel binary, so a rebuild is required for both drivers to work end-to-end.

The required containerization change is a single line in `kernel/config-arm64`:

```
# Before
# CONFIG_CIFS is not set
```

```
# After
CONFIG_CIFS=y
# CONFIG_CIFS_STATS2 is not set
CONFIG_CIFS_ALLOW_INSECURE_LEGACY=y
CONFIG_CIFS_UPCALL=y
CONFIG_CIFS_XATTR=y
CONFIG_CIFS_POSIX=y
# CONFIG_CIFS_DEBUG is not set
CONFIG_CIFS_DFS_UPCALL=y
# CONFIG_CIFS_SMB_DIRECT is not set
```

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

#1413 

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs


jglogan:
@tonycoco Could you fix the conflicts? Thanks for the contribution, I need to take a closer look but this seems fine.

One thing to note is that changing the `kernel/config-arm64` file down in containerization won't actually do anything to the kernel in our release distribution, as it uses the Kata kernel defined here: https://github.com/apple/container/blob/main/Sources/ContainerPersistence/DefaultsStore.swift#L222.

Users would need to build and specify their own kernel that includes SMB and/or NFS client capability.

One possible alternative/workaround would be if the public homebrew recipe (which isn't managed by the Containerization project maintainers, incidentally) were capable of building and installing a default kernel from using a more expansive config that it embeds.

tonycoco:
@jglogan Merge conflicts are now fixed. I have commits for the kernel and upstream kata as well in the loop but at least this unlocks the container side of things. I also have some changes to the `container-compose` project that will allow for this feature to get turned on for that too.

Reference:

https://github.com/apple/containerization/pull/681
https://github.com/tonycoco/Container-Compose/tree/feature/smb-volume-driver (not yet a PR but this feature branch works well)

Tracking:
https://github.com/kata-containers/kata-containers/pull/12374

jglogan:
@tonycoco Thanks! I'll let @dcantah handle apple/containerization#681 as he's already working with you on it.

tonycoco:
Looks good and also https://github.com/kata-containers/kata-containers/pull/12374 has been merged!

tonycoco:
@jglogan this change is ready to go

tonycoco:
cleaned up the commit and rebased

tonycoco:
@jglogan @dcantah ping for activity here

tonycoco:
Ping for activity again. Let me know if anything else is needed for the merge.

tonycoco:
Resolved merge conflicts in `VolumesService.swift`. Mostly due to the rename of the `Volume` model to `VolumeConfiguration`. I updated the `switch driver` logic to use the new `VolumeConfiguration` structure. The rebase introduced compilation failures in the newly added test suites `NFSVolumeTests.swift` and `SMBVolumeTests.swift` due to reference of the old `Volume` name so I updated both test files to use the new `VolumeConfiguration` model.

tonycoco:
Any update on this?

tonycoco:
Gentle ping for activity here. @jglogan @dcantah the containerization and kata changes are merged in. This should be ready to roll.

jglogan:
@tonycoco Thanks for the pings, and for your patience. I'll put this into the plan for this sprint.

IIRC I said I wanted to test it locally (which I haven't done yet).

I also need to do a little thinking about the architecture for the volume model. I don't think it means any changes to what your PR does, but what we might want to do is mark these options as experimental so that folks don't get wedded to a particular CLI shape.

jglogan:
@tonycoco Could you create an issue describing the feature and use case(s), and add something to the description above indicating that this PR closes the issue?

tonycoco:
@jglogan https://github.com/apple/container/issues/1911 issue created. Added a bunch of details.

PR #1432, opened 2026-04-20
Fix: memory leak and silent read error in Archiver._compressFile causing build failures under /Users/

### LINKED ISSUE #1423: [Bug]: container build fails under /Users, but succeeds under /tmp
### I have done the following

### Steps to reproduce

 Save the script below as repro.sh and run it. It creates two identical build contexts (a Dockerfile + node_modules from npm install) — one under /tmp/ and one under ~/ — and attempts to
  build both.

  Requires node/npm to be installed.

  #!/usr/bin/env bash
  set -e

  if ! command -v container &>/dev/null; then
      echo "Apple Container CLI not found."
      exit 1
  fi

  setup_context() {
      local dir="$1"
      mkdir -p "$dir"
      cat > "$dir/Dockerfile" <<'EOF'
  FROM node:20-alpine
  WORKDIR /app
  COPY . .
  CMD ["echo", "hello"]
  EOF
      cat > "$dir/package.json" <<'PKGJSON'
  {
    "private": true,
    "dependencies": {
      "preact": "^10.25.4",
      "bootstrap": "^5.3.3",
      "d3": "^7.9.0",
      "ckeditor5": "^44.3.0",
      "typescript": "^5.7.0",
      "@types/d3": "^7.4.0"
    }
  }
  PKGJSON
      echo "Installing node_modules in $dir..."
      (cd "$dir" && npm install --silent 2>&1 | tail -1)
  }

  echo "=== Test 1: build context under /tmp ==="
  DIR1=$(mktemp -d)
  setup_context "$DIR1"
  echo "Context: $(du -sh "$DIR1" | cut -f1), $(find "$DIR1" -type f | wc -l | tr -d ' ') files"
  if container build -t repro-test "$DIR1" 2>&1; then
      echo "RESULT: PASS"
  else
      echo "RESULT: FAIL"
  fi
  rm -rf "$DIR1"

  echo ""
  echo "=== Test 2: same build context under ~/container-bug-repro ==="
  DIR2="$HOME/container-bug-repro"
  rm -rf "$DIR2"
  setup_context "$DIR2"
  echo "Context: $(du -sh "$DIR2" | cut -f1), $(find "$DIR2" -type f | wc -l | tr -d ' ') files"
  if container build -t repro-test "$DIR2" 2>&1; then
      echo "RESULT: PASS"
  else
      echo "RESULT: FAIL"
  fi
  rm -rf "$DIR2"

  Expected behavior

  Both builds should succeed. The build context content is identical.
  Actual behavior
  === Test 1: build context under /tmp ===
  Context: 213M, 21989 files
  Successfully built repro-test:latest
  RESULT: PASS

  === Test 2: same build context under ~/container-bug-repro ===
  Context: 214M, 21989 files
  Error: unable to write data to the archive, code 0
  RESULT: FAIL

  Notes

  - The error occurs during the CLI's tarball creation phase (before anything is sent to the builder VM), specifically during the [internal] load .dockerignore / transferring context step.     - .dockerignore does not help — the error occurs before ignore rules are applied.
  - Builder memory (--memory 8G) does not help — the issue is on the CLI side.                                                                                                                   - --debug flag does not provide additional error detail.
  - Small contexts under /Users/ build fine. The threshold appears to be around ~200MB / ~20k files.                                                                                             - Both paths resolve to the same APFS volume (/dev/disk3s1). The difference may be related to macOS TCC / file-access sandbox behavior, since /private/tmp/ is typically exempt from privacy
  protections that apply to /Users/.

### Problem description

container build fails with Error: unable to write data to the archive, code 0 when the build context directory contains many files (~20k+) and is located under /Users/. The exact same
  context directory builds successfully when located under /tmp/ (/private/tmp/).

### Environment

```markdown
- container CLI version 0.11.0 (build: release, commit: d9b8a8d)
 - macOS 26.4.1 (25E253)
 - Apple Silicon (arm64)
```

### Code of Conduct

### LINKED ISSUE #1439: [Bug]: Archiver `_compressFile` func leaks memory, leaks fd on error, and ignores errors.
### I have done the following

### Steps to reproduce

Should be observable using a build context with a large number of files.

### Problem description

`container build` uses the Archiver type to create an archive of the build context to transfer to the builder VM.

The `_compressFile` function in Archiver writes a single member entry and its data to the archive. To this end it uses a 1 megabyte buffer, and an input stream. The implementation has several bugs though:

- Most seriously, it does not deallocate the buffer on exit from the function.
- The stream is not closed on a throwing error
- If an error occurs on a `read` from the stream (negative return value for error in contrast with zero for end of stream), the function silently exits instead of reporting an error.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

Ronitsabhaya75:
- Closes #1439.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
fixes : #1423 

_compressFile() allocated a 1 MiB read buffer per file but never freed it, leaking ~20 GiB with large contexts (~20k files). It also treated InputStream.read() errors (-1) the same as EOF (0), silently swallowing TCC access denials under /Users/ and corrupting the archive state — which surfaced as the unhelpful "unable to write data to the archive, code 0" error.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Archiver.swift
+++ b/Sources/Services/ContainerAPIService/Client/Archiver.swift
@@ -113,6 +113,7 @@ public final class Archiver: Sendable {
         let writer = archiver.makeTransactionWriter()
         let bufferSize = Int(1.mib())
         let readBuffer = UnsafeMutablePointer<UInt8>.allocate(capacity: bufferSize)
+        defer { readBuffer.deallocate() }
         try writer.writeHeader(entry: entry)
         if entry.fileType == .regular {
             // We need to write the data into the archive only if its a regular file
@@ -121,19 +122,23 @@ public final class Archiver: Sendable {
                 throw Error.failedToCreateInputStream(item)
             }
             stream.open()
+            defer { stream.close() }
             while true {
                 let byteRead = stream.read(readBuffer, maxLength: bufferSize)
-                if byteRead <= 0 {
+                if byteRead < 0 {
+                    // stream.read returns -1 on error (e.g. TCC access denial under /Users/)
+                    let streamError = stream.streamError
+                    throw Error.failedToReadFile(item, streamError)
+                }
+                if byteRead == 0 {
                     break
-                } else {
-                    let data = Data(bytesNoCopy: UnsafeMutableRawPointer(mutating: readBuffer), count: byteRead, deallocator: .none)
-                    hasher.update(data: data)
-                    try data.withUnsafeBytes { pointer in
-                        try writer.writeChunk(data: pointer)
-                    }
+                }
+                let data = Data(bytesNoCopy: UnsafeMutableRawPointer(mutating: readBuffer), count: byteRead, deallocator: .none)
+                hasher.update(data: data)
+                try data.withUnsafeBytes { pointer in
+                    try writer.writeChunk(data: pointer)
                 }
             }
-            stream.close()
         }
         try writer.finish()
     }
@@ -222,6 +227,7 @@ extension Archiver {
         case failedToCreateEntry
         case fileDoesNotExist(_ url: URL)
         case failedToCreateInputStream(_ url: URL)
+        case failedToReadFile(_ url: URL, _ underlying: Swift.Error?)
 
         public var description: String {
             switch self {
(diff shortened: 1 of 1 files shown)


jglogan (approved):
Thanks, @Ronitsabhaya75, good catch!

PR #1448, opened 2026-04-24
Centralize utilities for configuration loading and path parsing

noah-thor:
## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This attempts to create a more centralized utility to find and load configuration files, as a follow up to (pr: https://github.com/apple/container/pull/1425). 

This would make the file apis look something like:
```
let configPath = UserConfiguration.basePath(for: .appRoot).appending("config/config.toml")
```

## Testing
- [ ] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs


noah-thor:
Mainly looking for API feedback, changes will not be mergable until TOML PR is merged

hajimohammadsamirazizi-lab (approved):

Sources/ContainerPersistence/ConfigurationLoader.swift
```diff
@@ -0,0 +1,81 @@
...
+            source = provided
+            guard fm.fileExists(atPath: source.string) else {
+                throw ContainerizationError(
+                    .invalidArgument,
+                    message: "user config file not found at '\(source.string)'"
+                )
+            }
+        } else {
+            guard let resolved = UserConfiguration.resolve(from: .home, relativePath: "config.toml") else {
+                return
```
katiewasnothere: Should we throw here?

Sources/ContainerPersistence/ConfigurationLoader.swift
```diff
@@ -0,0 +1,81 @@
...
+import Foundation
+import SystemPackage
+import TOML
+
+public enum ConfigurationLoader {
+    public static func load<T: Codable & Sendable>(configFile: FilePath) throws -> T {
+        let path = configFile.string
+        guard FileManager.default.fileExists(atPath: path) else {
+            do {
+                return try TOMLDecoder().decode(T.self, from: Data("".utf8))
```
katiewasnothere: Could we make T `Initable` and just return an instance here?

Sources/ContainerPersistence/UserConfiguration.swift
```diff
@@ -0,0 +1,57 @@
...
+            return FilePath(configHome).appending(".container")
+        case .appRoot:
+            if let envPath = ProcessInfo.processInfo.environment["CONTAINER_APP_ROOT"], !envPath.isEmpty {
+                return FilePath(envPath)
+            }
+            let appSupportURL = FileManager.default.urls(
+                for: .applicationSupportDirectory,
+                in: .userDomainMask
+            ).first!.appendingPathComponent("com.apple.container")
+            return FilePath(appSupportURL.path(percentEncoded: false))
```
katiewasnothere: Could we dedup this with `ApplicationRoot`?

noah-thor: You mean completely remove the `ApplicationRoot` and/or just have a single definition? I think that makes sense I was just hesitant to touch that code

Sources/ContainerPersistence/ConfigurationLoader.swift
noah-thor: Yeah, good catch or should I say good throw

Sources/ContainerPersistence/ConfigurationLoader.swift
noah-thor: Is that an existing protocol? I can't seem to find it. I assume its like:
```
protocol Initable {
  init()
}
```
?

Sources/ContainerPersistence/ConfigurationLoader.swift
noah-thor: Wait actually this is the case in which a user has not provided an explicit path to the file, so we just want to  be ok (not throw) if they do not yet have a toml file created at all. I will add a comment

Sources/ContainerCommands/System/SystemStart.swift
```diff
@@ -72,8 +72,10 @@ extension Application {
         public init() {}
 
         public func run() async throws {
-            try ConfigurationLoader.copyConfigToAppRoot()
-            let containerSystemConfig: ContainerSystemConfig = try ConfigurationLoader.load()
+            let appRootPath = FilePath(appRoot.path(percentEncoded: false))
+            try ConfigurationLoader.copyConfigurationToReadOnly(to: appRootPath)
+            let containerSystemConfig: ContainerSystemConfig = try ConfigurationLoader.load(
```
jglogan, line 77: nit for the next commit: we can use type inference here and shorten the line

Sources/ContainerPersistence/PathUtils.swift
```diff
@@ -22,18 +22,18 @@ public enum PathUtils {
         case home
         case appRoot
 
-        public func basePath() -> FilePath {
+        public func basePath(env: [String: String] = ProcessInfo.processInfo.environment) -> FilePath {
             switch self {
             case .home:
                 let configHome: String
-                if let xdg = ProcessInfo.processInfo.environment["XDG_CONFIG_HOME"], !xdg.isEmpty {
+                if let xdg = env["XDG_CONFIG_HOME"], !xdg.isEmpty {
```
jglogan, line 29: Need a follow up PR to scrub ContainerPersistence for string and URL path manipulation. I took care of this for the entity store type.

e.g. we could start here with something like:

```
    let configHome = env["XDG_CONFIG_HOME"]
        .map { FilePath($0) }
        ?? FilePath(NSHomeDirectory()).appending("/.config")
```

jglogan (approved):
lgtm, see follow-up notes

katiewasnothere (approved):

Issue #1329, opened 2026-03-19
[Bug]: `container system start` hangs indefinitely at `Verifying apiserver is running...`


n8henrie:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

I am building from current main and now can't even get container to start. Have rebooted multiple times and cleaned the build directory, the issue persists.

```console
$ git rev-parse HEAD
9fd15f01bf0f7caf757b66e17ac3f8e1be432bde
$ sudo BUILD_CONFIGURATION=release make all install
Building container binaries...
swift-driver version: 1.127.15 Apple Swift version 6.2.4 (swiftlang-6.2.4.1.4 clang-1700.6.4.2)
Target: arm64-apple-macosx26.0
[1/1] Planning build
Building for production...
[10/10] Compiling SwiftProtobuf AnyMessageStorage.swift
Build complete! (23.59s)
make[1]: Entering directory '/Users/n8henrie/git/container'
Installing container binaries from /Users/n8henrie/git/container/.build/arm64-apple-macosx/release into bin/release/staging/...
Install update script
Install uninstaller script
Signing container binaries...
bin/release/staging/bin/container: replacing existing signature
bin/release/staging/bin/container-apiserver: replacing existing signature
bin/release/staging/libexec/container/plugins/container-core-images/bin/container-core-images: replacing existing signature
bin/release/staging/libexec/container/plugins/container-runtime-linux/bin/container-runtime-linux: replacing existing signature
bin/release/staging/libexec/container/plugins/container-network-vmnet/bin/container-network-vmnet: replacing existing signature
Creating application installer
pkgbuild: Inferring bundle components from contents of bin/release/staging/
pkgbuild: Wrote package to bin/release/container-installer-unsigned.pkg
Installing container installer package
make[1]: Leaving directory '/Users/n8henrie/git/container'
Building initfs if containerization is in edit mode
Installing container binaries from /Users/n8henrie/git/container/.build/arm64-apple-macosx/release into bin/release/staging/...
Install update script
Install uninstaller script
Signing container binaries...
bin/release/staging/bin/container: replacing existing signature
bin/release/staging/bin/container-apiserver: replacing existing signature
bin/release/staging/libexec/container/plugins/container-core-images/bin/container-core-images: replacing existing signature
bin/release/staging/libexec/container/plugins/container-runtime-linux/bin/container-runtime-linux: replacing existing signature
[...]

Issue #798, opened 2025-10-22
[Bug]: Can't get container system commands to work


PhantomInTheWire:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. clone apple/container
2. make
4. bin/container system start
5. bin/container system stop(keeps running doesn't exit)
6. ctrl+c then try bin/container system start/status/stop doesn't matter just can't get it to exit

i even tried to use global installation of container to stop it or check status or "start" it again but even that wouldn't exit and was stuck

### Current behavior

i was playing around the codebase a little and when i did git pull this(container system commands stop exiting) happened.

### Expected behavior

it was working properly before git pull everything ran in 3-7 secs

### Environment

```markdown
- OS: macos 26
- Xcode: 
- Container: current main branch commit 2f507fe9d911e9802d0423b4e89bbcf1619cc727
```

### Relevant log output
i mean it just stays stuck but here

```shell
❯ container system stop
2025-10-22T05:38:23+0530 info com.apple.container.cli: [ContainerCommands] checking if APIServer is alive

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

katiewasnothere:
Hi @PhantomInTheWire. I think this may be the same issue discussed in https://github.com/apple/container/discussions/713. If not, please let me know. Unfortunately the only solution we've found for this issue so far is a reboot of the machine. Can you try that and let me know if that resolves the issue?

PhantomInTheWire:
Hey @katiewasnothere,
yeah, when I first ran into this, my initial thought was that it might be a cache invalidation issue, so I did a clean build hoping that would fix it, no luck. Then I rolled back to an older commit that I knew worked, but that didn’t help either. Like you suggested, I also tried rebooting, though I do think it actually stopped the service. After the reboot, I recompiled on the latest commit and tested again, but the issue still persisted.

PhantomInTheWire:
update: after reboot doing uninstall-container.sh -k, then clean build did fix this. should be similar to #713  then.

PhantomInTheWire:
hey @katiewasnothere i think you guys should be able to reproduce this with running `bin/container system start --debug` first(make sure it wasn't running before this). after than the behavior mentioned above happens(container system commands hang). that being said i was able to fix that with `launchctl bootout gui/$(id -u)/com.apple.container.apiserver` without a reboot.

[...]

Issue #621, opened 2025-09-17
[Bug]: "container system start" hangs when app root is placed on an external disk


GarthSnyder:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

The **container system start** command hangs indefinitely for me whenever the app root is on an external disk. It prints "Verifying apiserver is running..." but never progresses beyond that.

Things I've tried that all fail in this same way:

- Starting the 0.4.1 release version installed via package with **container system start --app-root path-to-existing-dir**, where **path-to-existing-dir** is on an external drive.
- Symbolic linking **~/Library/Application Support/com.apple.container** to **path-to-existing dir**.
- Cloning the current repo to an external disk and running **make all test integration** as suggested in the BUILDING document.
- Cloning the current repo to a second external disk and running **make all test integration**.

Things that work fine:

- Using **container** normally with no special configuration.
- Cloning to my home directory (on the root disk) and running **make all test integration**. 

I would imagine it fails similarly on any non-root disk, but in true Apple style, I have only the one internal disk to test with. When failing during **make**, the process hangs at:

```
Running the integration tests...
Warning! Running debug build. Performance may be degraded.
Verifying apiserver is running...
```

Note that the **make test** phase always seems to work fine--even on an external disk--and it runs without errors, even though testing also appears to depend on the apiserver.

Control-C'ing the **container start** command after it hangs leaves the container system in a wedged and unrecoverable state. Containers can't be used until after a reboot.

### Current behavior

As described above.

### Expected behavior

I just want containers stored on an external disk that I specify. In other ways, I don't think there's anything unusual about what I'm trying to do.

### Environment

```markdown
- OS: Tahoe 26.0 release version
- Xcode: 26.0, tool chain 6.2
- Container: Doesn't get far enough to run containers
```

### Relevant log output

```shell
See above.
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
Thanks for the issue...we'll see if we can reproduce this!

jglogan:
@GarthSnyder what sort of external storage are you using, and what kind of volume is on it?

GarthSnyder:
Ah, I should have specified. As mentioned above, I've reproduced this behavior with **--app-root** set to two different devices, both external.

[...]

PR #1455, opened 2026-04-27
fix(help): indefinite hang on container --help / help / no-args

### LINKED ISSUE #1459: [Bug]: help/default commands can hang when apiserver launchd state is wedged
### I have done the following

### Steps to reproduce

No deterministic setup is currently known.

This issue tracks a conditional failure mode: systems where `com.apple.container.apiserver` is registered with launchd but does not answer XPC requests, such as dead, wedged, or stale launchd/XPC states reported in related apiserver hang issues (#1329, #798, #621).

In that state, the pre-fix help/default command paths can block before printing help because they initialize plugin loading, which calls `ClientHealthCheck.ping()` and waits on the apiserver:

```bash
container --help
container help
container
```

A healthy apiserver setup is expected not to reproduce this. For example, the maintainer check in apple/container#1455 ran `make all` and then timed the three commands above; all returned quickly. That result is consistent with this issue being specific to stale/wedged apiserver state rather than normal help rendering.

### Problem description

Help and default command rendering should not require the API server to be reachable. These paths should be able to print local CLI help even when `com.apple.container.apiserver` is unavailable, stale-registered, or wedged.

Current pre-fix behavior: help/default paths can reach `Application.createPluginLoader()`, which calls `ClientHealthCheck.ping()` to fetch roots from the apiserver before printing help. If the apiserver is in a state where the XPC reply never arrives, the command can block rather than showing help.

Expected behavior: `container --help`, `container help`, and `container` should print the base help text without requiring an apiserver XPC round trip.

Observed recovery in related stale-apiserver cases has involved `launchctl bootout ...`, but that is a recovery/cleanup action, not a reliable reproduction step for this issue.

### Environment

- OS: macOS 26
- Xcode: Xcode 26
- Container: main / apple/container#1455 investigation branch

### Additional context

apple/container#1455 proposes a defensive fix: skip API-server/plugin loading for help/default help rendering, and make idempotent XPC sends return on non-firing replies instead of being trapped by structured-concurrency TaskGroup cleanup.

The issue is intentionally scoped to help/default command robustness. It does not claim to fix every apiserver startup/stop hang reported in #1329, #798, or #621.

### Code of Conduct

chrisgeo:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Closes #1459.

`container --help`, `container help`, and `container` (no args) should not require `com.apple.container.apiserver` to be reachable. Today, those paths can enter plugin loading before printing help, which calls `ClientHealthCheck.ping()` and waits on the apiserver.

A healthy build is expected to return quickly; this is not a bug in normal help rendering. The failure mode is conditional on launchd/XPC state where `com.apple.container.apiserver` is registered but unavailable, dead, wedged, or otherwise not replying. That state is the class of issue tracked in #1459 and is similar to stale-apiserver recovery reports in #1329, #798, and #621.

This PR is a defensive robustness fix: help/default-command rendering should remain local, and idempotent XPC requests with explicit timeouts should actually return when the reply callback never fires.

## Summary

This PR fixes two independent defects:

1. **Help/default commands should not ping the API server** — `Application.swift`, `HelpCommand.swift`, and `DefaultCommand.swift` now print base help without constructing a `PluginLoader` first.
2. **Idempotent XPC timeouts should actually unblock** — `XPCClient.send(_:timeoutForIdempotentRequest:)` replaces the old TaskGroup race with a single-resume gate so a non-firing XPC reply cannot trap structured-concurrency cleanup forever.
3. **Mutating XPC sends should not silently drop late replies** — the XPC API now separates mutating/unknown-safety sends from idempotent sends, preventing accidental timeout use on mutating operations.
4. **The investigation is documented** — `docs/internal/help-freeze-analysis.md` captures the call chain, tradeoffs, and remaining follow-ups.

## Why the help-path fix is separate from the XPC fix

There are two independent defects:

- **Defect A (symptom).** Help/default rendering called `createPluginLoader()` → `ClientHealthCheck.ping()` to fetch `appRoot`/`installRoot`/`logRoot` purely so `printModifiedHelpText` could enrich help with plugin commands. The ping is structurally unnecessary for base help output.

- **Defect B (depth).** The old `XPCClient.send(... responseTimeout:)` raced `Task.sleep` against the XPC reply inside a `withThrowingTaskGroup`, but throwing from `group.next()` caused structured-concurrency cleanup to await the still-suspended XPC child task. Cancelling the Swift task does not cancel `xpc_connection_send_message_with_reply`, so a non-firing callback could still block indefinitely.

Defect A removes the apiserver dependency from help/default paths. Defect B makes explicit idempotent XPC timeouts behave as timeouts when the XPC reply never arrives.

## Reproduction scope

No deterministic setup is currently known for creating the stale/wedged launchd state. On a healthy checkout, the following should return quickly, and reviewer testing confirmed that behavior after `make all`:

```bash
make all

time .build/debug/container --help
time .build/debug/container help
time .build/debug/container
```

The bug tracked by #1459 applies when the apiserver is registered but not replying. In that state, pre-fix help/default paths can block before printing help because they ping the apiserver while preparing plugin-enriched help.

`launchctl bootout ...` is a recovery/cleanup action observed in related stale-apiserver cases, not a reliable reproduction step.

## Build/test guidance

For normal source builds and verification, follow `BUILDING.md`:

```bash
make all test integration
```

For installed release-style testing:

```bash
BUILD_CONFIGURATION=release make all test integration
BUILD_CONFIGURATION=release make install
```

The `.build/debug/container` timings above are diagnostic context matching the reviewer command, not the canonical setup for reproducing or validating installed helper/plugin layout.

## Known regression in this PR

Plugin enrichment in `--help` / `help` / no-args output is removed. The output is the original `OVERVIEW: ... USAGE: ... CONTAINER SUBCOMMANDS: ...` block, without a trailing `PLUGINS:` section.

This keeps the freeze fix minimal and avoids entangling it with API design questions on `PluginLoader`. A follow-up can restore plugin enrichment by extracting filesystem-only plugin discovery from `PluginLoader.findPlugins()`; the proposed shape is sketched in `docs/internal/help-freeze-analysis.md`.

## Verification

- `swift build --product container` — clean
- `make test` — 366/366 unit tests pass
- `make swift-fmt-check` — clean
- Added `Tests/ContainerXPCTests/XPCClientTests.swift` coverage for:
  - idempotent timeout returning within the bound
  - reusable clients surviving idempotent timeout
  - late replies after timeout being ignored cleanly
  - plain mutating-safe sends completing when the server replies
  - plain sends ignoring cancellation after dispatch
  - plain sends honoring cancellation before dispatch
- Manual diagnostic smoke test on macOS 26 with `com.apple.container.apiserver` registered with launchd but inactive (`active count = 0`, `state = spawn scheduled`):

  ```bash
  $ time .build/debug/container --help
  ... full help output ...
  real  0m0.11s

  $ time .build/debug/container help
  ... full help output ...
  real  0m0.11s

  $ time .build/debug/container
  ... full help output ...
  real  0m0.11s
  ```

  All exit 0 immediately. Before this PR these paths could block while trying to reach the apiserver before printing help.

## What this PR explicitly does **not** fix

1. A deterministic way to create the stale/wedged apiserver launchd state tracked by #1459.
2. All apiserver startup/stop hangs reported in #1329, #798, or #621.
3. `ClientHealthCheck.ping`'s 60s default `xpcRegistrationTimeout`.
4. Reusable `ContainerClient` / `NetworkClient` mutating calls with no timeout; this is intentional until mutating operations have an idempotency/recovery protocol.
5. `launchctl bootout`-style recovery as a built-in CLI command.

## Workaround for affected users

If the local apiserver launchd state is stale or wedged, related reports have recovered with bootout + restart:

```bash
launchctl bootout gui/$(id -u)/com.apple.container.apiserver 2>/dev/null
launchctl bootout user/$(id -u)/com.apple.container.apiserver 2>/dev/null
container system start
```

---

Full reasoning, the call-chain trace, the audit of affected XPC callers, and the design tradeoffs are in `docs/internal/help-freeze-analysis.md`. **The doc is intentionally written for adversarial review** — please push back on anything that doesn't hold up.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -425,6 +425,12 @@ let package = Package(
                 "CAuditToken",
             ]
         ),
+        .testTarget(
+            name: "ContainerXPCTests",
+            dependencies: [
+                "ContainerXPC"
+            ]
+        ),
         .target(
             name: "ContainerOS",
             dependencies: [
--- a/Sources/APIServer/APIServer+Start.swift
+++ b/Sources/APIServer/APIServer+Start.swift
@@ -324,12 +324,12 @@ extension APIServer {
 
             let harness = NetworksHarness(service: service, log: log)
 
-            // network creation/deletion/list is not supported pre-macOS 26 (refer to AllocationOnlyVmnetNetwork)
+            // network creation is not supported pre-macOS 26 (refer to AllocationOnlyVmnetNetwork)
             if #available(macOS 26, *) {
                 routes[XPCRoute.networkCreate] = harness.create
-                routes[XPCRoute.networkDelete] = harness.delete
-                routes[XPCRoute.networkList] = harness.list
             }
+            routes[XPCRoute.networkList] = harness.list
+            routes[XPCRoute.networkDelete] = harness.delete
 
             return service
         }
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -118,10 +118,13 @@ public struct Application: AsyncLoggableCommand {
         } catch {
             // --help/-h on the root command (e.g. `container --help`) is intercepted
             // by ArgumentParser and lands here.
+            //
+            // Help rendering must not depend on the API server being reachable, otherwise
+            // a wedged or unregistered `com.apple.container.apiserver` causes an
+            // indefinite hang. See docs/internal/help-freeze-analysis.md.
             let containsHelp = fullArgs.contains("-h") || fullArgs.contains("--help")
             if fullArgs.count <= 2 && containsHelp {
-                let pluginLoader = try? await createPluginLoader()
-                await Self.printModifiedHelpText(pluginLoader: pluginLoader)
+                await Self.printModifiedHelpText(pluginLoader: nil, unavailableMessage: nil)
                 return
             }
             let errorAsString: String = String(describing: error)
@@ -234,11 +237,21 @@ public struct Application: AsyncLoggableCommand {
 extension Application {
     // Because we support plugins, we need to modify the help text to display
     // any if we found some.
-    static func printModifiedHelpText(pluginLoader: PluginLoader?) async {
+    //
+    // Pass `unavailableMessage: nil` from contexts that already deliberately skipped
+    // plugin loading (e.g. the `--help` / `help` / no-args paths that must not depend
+    // on the API server). The default preserves prior behavior for paths that *did*
+    // attempt to load plugins but couldn't reach the server.
+    static func printModifiedHelpText(
+        pluginLoader: PluginLoader?,
+        unavailableMessage: String? = "PLUGINS: not available, run `container system start`"
+    ) async {
         let original = Application.helpMessage(for: Application.self)
         guard let pluginLoader else {
             print(original)
-            print("PLUGINS: not available, run `container system start`")
+            if let unavailableMessage {
+                print(unavailableMessage)
+            }
             return
         }
         let altered = pluginLoader.alterCLIHelpText(original: original)
--- a/Sources/ContainerCommands/DefaultCommand.swift
+++ b/Sources/ContainerCommands/DefaultCommand.swift
@@ -33,12 +33,16 @@ struct DefaultCommand: AsyncLoggableCommand {
     var remaining: [String] = []
 
     func run() async throws {
-        // See if we have a possible plugin command.
-        let pluginLoader = try? await Application.createPluginLoader()
+        // No-args invocation prints help and must not depend on the API server.
+        // See docs/internal/help-freeze-analysis.md.
         guard let command = remaining.first else {
-            await Application.printModifiedHelpText(pluginLoader: pluginLoader)
+            await Application.printModifiedHelpText(pluginLoader: nil, unavailableMessage: nil)
             return
         }
+
+        // We have a candidate plugin command; load plugins (which contacts the
+        // API server) only on this path.
+        let pluginLoader = try? await Application.createPluginLoader()
 
         // Check for edge cases and unknown options to match the behavior in the absence of plugins.
         if command.isEmpty {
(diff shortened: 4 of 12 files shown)


jglogan:
Hi @chrisgeo, could you please file a bug issue with clear and reliable reproduction steps, and then in this PR description indicate the the PR "Closes #issue-id."?

Please ensure that you have [Read the Manual](https://github.com/apple/container/blob/main/BUILDING.md) when preparing reproduction steps involving builds, as the ones in this PR seem to indicate that this hasn't been done.

I'm running this and cannot duplicate the issue. It doesn't mean that it's not an issue (there are definitely some XPC cancelation issues but the most pertinent ones relate to the calls from the API server to the container runtime plugin), but we're going to need how to get the system in the state that requires the fix.

I ran:

```bash
$ make all
...build output...

$ time .build/debug/container --help
... full help output ...
.build/debug/container --help  0.03s user 0.02s system 4% cpu 0.945 total

$ time .build/debug/container help
... full help output ...
.build/debug/container help  0.02s user 0.01s system 76% cpu 0.038 total

$ time .build/debug/container
... full help output ...
.build/debug/container  0.03s user 0.01s system 78% cpu 0.054 total
```

chrisgeo:
Thanks — agreed that the original description overstated reproducibility.

I filed #1459 and updated the PR description with `Closes #1459`. I also revised the repro section to distinguish:

1. healthy `make all` behavior, which should return quickly and matches your result;
2. the stale/wedged apiserver launchd/XPC state where this help/default-path bug applies;
3. `launchctl bootout` as recovery/cleanup, not a repro step.

I also added explicit BUILDING.md guidance and clarified that the `.build/debug/container` timings are diagnostic context matching your command, not the canonical installed-helper/plugin-layout setup.

PR #1462, opened 2026-04-28
cli: Add support for --stop-signal

### LINKED ISSUE #1581: [Request]: cli: Add support for `--stop-signal`

The CZ implementation for this is complete so we can add to the CLI now.

### Code of Conduct

dcantah:
Closes #1581.

This adds stop signal support to the cli. The priority is:

1. If an explicit stop signal is passed on the cli use this.
2. If not, check if there is a stop signal in the image config.
3. Finally, use the default (TERM).

### DIFF
--- a/Sources/ContainerResource/Container/ContainerStopOptions.swift
+++ b/Sources/ContainerResource/Container/ContainerStopOptions.swift
@@ -17,15 +17,15 @@
 import Foundation
 
 public struct ContainerStopOptions: Sendable, Codable {
-    public let timeoutInSeconds: Int32
-    public let signal: Int32
+    public var timeoutInSeconds: Int32
+    public var signal: String?
 
     public static let `default` = ContainerStopOptions(
         timeoutInSeconds: 5,
-        signal: SIGTERM
+        signal: nil
     )
 
-    public init(timeoutInSeconds: Int32, signal: Int32) {
+    public init(timeoutInSeconds: Int32, signal: String?) {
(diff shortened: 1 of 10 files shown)


Sources/Services/ContainerSandboxService/Server/SandboxService.swift
saehejkang: should this be updated now as well?

saehejkang: same here

dcantah: I'd prefer them to be strings as well, but I'd rather do in a follow-up.

Sources/ContainerResource/Container/ContainerStopOptions.swift
```diff
@@ -17,15 +17,15 @@
...
+    public var signal: String?
 
     public static let `default` = ContainerStopOptions(
         timeoutInSeconds: 5,
-        signal: SIGTERM
+        signal: nil
     )
 
-    public init(timeoutInSeconds: Int32, signal: Int32) {
+    public init(timeoutInSeconds: Int32, signal: String?) {
```
jglogan, line 28: Can we do validation in this type so that it's correct on construction?

Does this type really need to be mutable (I do see it makes `resolvedOptions` slightly easier below)?

jglogan (approved):

PR #1463, opened 2026-04-28
Add bridge mode networking

### LINKED ISSUE #489: [Request]: Bridge Mode

It would be great if container networking supported bridge mode, allowing each container to obtain an IP address from a DHCP server and to be reachable from the local network.

### Code of Conduct

torarnv:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Bridged networking can be useful for example for mDNS/Bonjour, LAN-visible services without port forwarding, or simply to have the LAN/router manage the IPs of the containers.

Fixes #489

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ PKG_PATH := bin/$(BUILD_CONFIGURATION)/container-installer-unsigned.pkg
 DSYM_DIR := bin/$(BUILD_CONFIGURATION)/bundle/container-dSYM
 DSYM_PATH := bin/$(BUILD_CONFIGURATION)/bundle/container-dSYM.zip
 CODESIGN_OPTS ?= --force --sign - --timestamp=none
+ENABLE_BRIDGE_NETWORKING ?= 0
+ifeq ($(ENABLE_BRIDGE_NETWORKING),1)
+	RUNTIME_LINUX_ENTITLEMENTS := signing/container-runtime-linux.bridge-mode.entitlements
+else
+	RUNTIME_LINUX_ENTITLEMENTS := signing/container-runtime-linux.entitlements
+endif
 
 
 # Conditionally use a temporary data directory for integration tests
@@ -121,7 +127,7 @@ installer-pkg: $(STAGING_DIR)
 	@codesign $(CODESIGN_OPTS) --identifier com.apple.container.cli "$(join $(STAGING_DIR), bin/container)"
 	@codesign $(CODESIGN_OPTS) --identifier com.apple.container.apiserver "$(join $(STAGING_DIR), bin/container-apiserver)"
 	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. "$(join $(STAGING_DIR), libexec/container/plugins/container-core-images/bin/container-core-images)"
-	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. --entitlements=signing/container-runtime-linux.entitlements "$(join $(STAGING_DIR), libexec/container/plugins/container-runtime-linux/bin/container-runtime-linux)"
+	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. --entitlements=$(RUNTIME_LINUX_ENTITLEMENTS) "$(join $(STAGING_DIR), libexec/container/plugins/container-runtime-linux/bin/container-runtime-linux)"
 	@codesign $(CODESIGN_OPTS) --prefix=com.apple.container. --entitlements=signing/container-network-vmnet.entitlements "$(join $(STAGING_DIR), libexec/container/plugins/container-network-vmnet/bin/container-network-vmnet)"
 
 	@echo Creating application installer
--- a/Package.swift
+++ b/Package.swift
@@ -332,6 +332,7 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Containerization", package: "containerization"),
                 "ContainerLog",
+                "ContainerNetworkServiceClient",
                 "ContainerPlugin",
                 "ContainerResource",
                 "ContainerSandboxService",
--- a/Sources/APIServer/ContainerDNSHandler.swift
+++ b/Sources/APIServer/ContainerDNSHandler.swift
@@ -76,10 +76,12 @@ struct ContainerDNSHandler: DNSHandler {
     }
 
     private func answerHost(question: Question) async throws -> ResourceRecord? {
-        guard let ipAllocation = try await networkService.lookup(hostname: question.name) else {
+        guard let ipAllocation = try await networkService.lookup(hostname: question.name),
+            let ipv4Address = ipAllocation.ipv4Address
+        else {
             return nil
         }
-        let ipv4 = ipAllocation.ipv4Address.address.description
+        let ipv4 = ipv4Address.address.description
         guard let ip = try? IPv4Address(ipv4) else {
             throw DNSResolverError.serverError("failed to parse IP address: \(ipv4)")
         }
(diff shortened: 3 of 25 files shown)


torarnv:
This is an early RFC of the feature. It requires the `com.apple.vm.networking` entitlement to use https://developer.apple.com/documentation/virtualization/vzbridgednetworkdeviceattachment, which requires a provisioning profile blessed by Apple, otherwise `container-runtime-linux` crashes on start (unless you locally disable AMFI). So if we land this feature it probably needs some guardrails so it's not enabled for local dev unless explicitly opted in. And the release machinery needs to sign with a blessed profile and signing identity.

dcantah:
@torarnv Sorry for not getting back on the other issues yet, I have a large backlog.. 😆 But, what you point out is essentially why we haven't taken a stab at this yet. We'd have to figure out how we can go about the entitlement deficiencies and we hadn't explored too deeply.

torarnv:
Thanks @dcantah! No rush, just wanted to try this out for myself locally, and for now this solves my immediate need 😄 But would be cool to upstream at some point!

For the entitlement, I assume it won't be a problem to get it blessed for the releases 😉 So the challenge we have to solve is setting the GH release pipeline of this repo up with the right certs/profiles, and ensuring local devs do not pull the feature in accidentally. I don't know how the Homebrew [formula](https://github.com/Homebrew/homebrew-core/blob/f0c124556c25d1d383ae57335c1e892f34da9e9b/Formula/c/container.rb) is going to deal with it, but I guess they'd either need to pull the binary releases from this repo, or live without the bridged network feature.

I'm happy to look further into this if you think there's a path forward for upstreaming this. Let me know if you have other comments/concerns with the code as is. Thanks!

torarnv:
@dcantah 2aa9d21e111feac2c40e5acb0453b47dc6578a6d should take care of the dev experience for people without a blessed cert in their `CODESIGN_OPTS` or run with AMFI off.

jglogan:
@torarnv I'm glad you took a crack at it. As Danny mentioned, we're going to be slammed this month with more basic work, so we won't be in a place where we can merge it soon.

However, this change is an excellent architecture driver. We're in the midst of refactoring the network model to clean up the separation of concerns between client, API server, and network plugin (we're doing this for runtime as well). We need to work out the best way to represent configuration for different network modes, and this needs to work in a scenario where there are additional (builtin or third party) network plugins. Since you have an additional mode here, we can refine this PR and the main together to allow bridged mode to slot in cleanly.

Let's start on that front. Once I have #1421 merged and make a few of the simpler network changes, let's sort out what the config model might like. We can either follow up in here on this, or back in #489.

torarnv:
Totally understandable @jglogan ! 😅 Happy to use this PR as inspiration/discussion for how to do this properly at some point. Some random things that came to mind when working on this:

 * Making `Interface.ipv4Address` optional in https://github.com/apple/containerization/pull/709/changes/23c8a6697810e9729841d99c79274c7a973cddfe spread a bit through the codebase. Dunno if there are better ways to model the lack of an up-front configured IP. OTOH, a machine could strictly speaking have only IPv6, so making IPv4 optional is perhaps in line with that?
 * I opted for extending `container-network-vmnet` with a bridged mode, rather than implementing a new network plugin, as a new plugin didn't seem necessary when the work of creating the final `BridgedNetworkInterface` happens in the Linux runtime plugin anyways, but perhaps is misunderstood the architecture. The `vmnet` in the existing network plugin is a bit of a misnomer already since NAT uses `VZNATNetworkDeviceAttachment` and not `VZVmnetNetworkDeviceAttachment` AFAIU.
 * `container ls` and `container inspect` doesn't show the resulting IP. Should it? We'd need new RPC to feed back the interfaces, etc if so. Since the DHCP happens via the kernel during boot, these should be available very early.
 
Let's discuss further when  #1421  and related changes land 🤝

curdbecker:
Please check out also the adoption of the PR on top of the current state of the main branch with added support for helper-based bridging support. This allows to avoid using the entitlement completely during development while also having the option to work directly with bridged networking (because I actually need it right now 😂).

I imagine also that it is quite okay to have a less optimal performance during testing or even in general, e.g. in my current case where I just would simply bridge a container to a VLAN network interface, but Docker for Mac seems to basically obstruct people from using the features offered by the Virtualization Framework - or basically the entire VM (see also [here](https://github.com/curdbecker/mocker-virtualization), [here](https://github.com/curdbecker/grpc-dnsproxy) and [here](https://github.com/curdbecker/vpnkit) - if someone is interested - had to take their entire virtualization service apart to set simple settings on the VM :-/). And this is exactly where container is amazingly different.

See https://github.com/apple/container/pull/1622

P.S.: @torarnv your idea with having the kernel negotiate the DHCP lease and then also retrieve the nameservers from there without having a dedicated DHCP client in the init container is really amazing. This is quite elegant and simple :)

s3rj1k:
To increase the drama for proper bridge support, I've done another PR that uses `vmnet-helper`, no changes to `github.com/apple/containerization` are required.

https://github.com/apple/container/pull/1661

---

For people that need bridges today :)

curdbecker:
@s3rj1k I completely get you. It's another level of frustration xD 

Hahaha, yes, that was also my initial idea, but I think it's nicer if the support is directly integrated, since then you would need to have `vmnet-helper` installed and in path and... so on. In addition, that only adds another dependency in itself and you would need to modify `container` anyways.

I used it [here](https://github.com/curdbecker/mocker-virtualization/blob/5cfffa6cb2a8f75e812d2862c9d206e60cad83ca/legacy_helpers/vmnet-helper-handoff.c#L202) to get between `Docker Desktop for Mac` and its VM. That's one of the methods to easily intercept the communication for analysis and also add bridging support there. Although, it is easier to approach this directly from the VM runner, i.e. through [injection in the existing client](https://github.com/curdbecker/mocker-virtualization/blob/5cfffa6cb2a8f75e812d2862c9d206e60cad83ca/vzinject/launch_with_console_enabled.ts#L103) or using an [alternative VM runner](https://github.com/curdbecker/mocker-virtualization/blob/5cfffa6cb2a8f75e812d2862c9d206e60cad83ca/mocker-virtualization/main.go#L371) entirely.

The changes in `containerization` are marginal, always quite nice to have around and @torarnv's DHCP improvement makes a lot of sense. You seem to have only support for static IPv4 addresses then? That's okay-ish, but the other method is better.

And btw - you could have already used it today 😂  I am using my branch currently - it also does not require any entitlement, because basically `vmnet-helper` is just built-in :)

s3rj1k:
> I think it's nicer if the support is directly integrated

I fully agree, `vmnet-helper` is a quick hack :)

> You seem to have only support for static IPv4 addresses

yea, only static v4, I wanted to keep this as simple-stupid as possible for maintainability purposes, till something better gets merged.

> you could have already used it today

thing that stopped me is unmerged changes into `containerization`, I'll definitely switch when baseline bridge works lands into `containerization`

curdbecker:
@s3rj1k Ah, yes, that’s what I thought. I also tried to get around changing containerization in the beginning 😂
I explicitly left out the local containerization link from my branch, so that it does not confuse people. It is quite easy to set up using the [documentation](https://github.com/apple/container/blob/main/BUILDING.md#develop-using-a-local-copy-of-containerization) but you must not forget to copy the init container over or you won’t get DHCP for IPv4 but v6 works as it doesn’t need it.

That being said, static IP addresses are still also quite interesting on bridge networks without DHCP :)

At least, people can’t say anymore they don’t have any options to choose from 😂

s3rj1k:
> That being said, static IP addresses are still also quite interesting on bridge networks without DHCP :)

Also works fine with DHCP networks, just need to select IP carefully :)

> At least, people can’t say anymore they don’t have any options to choose from 😂

yea, lol

PR #1468, opened 2026-04-29
[scripts]: add prompt for updating container with unsigned package

### LINKED ISSUE #1467: [Request]: update-container script should not fail if the signed package is not immediately available

After every new version release, the `update-container` script, should not fail if the signed package is not immediately available. Instead, we should prompt users if they would like to continue with installation of the unsigned package.

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #1467 and some minor QOL updates

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/scripts/update-container.sh
+++ b/scripts/update-container.sh
@@ -19,6 +19,7 @@ INSTALL_DIR="/usr/local"
 OPTS=0
 LATEST=false
 VERSION=
+FORCE=false
 TMP_DIR=
 
 # Release Info
@@ -29,8 +30,10 @@ RELEASE_VERSION=
 # Package Info
 PKG_URL=
 PKG_FILE=
-PRIMARY_PKG=
-FALLBACK_PKG=
+SIGNED_PRIMARY_PKG=
+SIGNED_FALLBACK_PKG=
+UNSIGNED_PRIMARY_PKG=
+UNSIGNED_FALLBACK_PKG=
 
 check_installed_version() {
     local target_version="$1"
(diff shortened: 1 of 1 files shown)


scripts/update-container.sh
jglogan: `-f` here?

scripts/update-container.sh
```diff
@@ -50,26 +53,30 @@ usage() {
...
             echo "Invalid option: -${OPTARG}"
             usage
             ;;
     esac
 done
 
-# Default to install the latest release version
-if [ "$OPTS" -eq 0 ]; then
+# Default to upgrade to the latest release version
+if [ "$OPTS" -eq 0 ] && [[ -z "$VERSION" ]]; then
```
jglogan: Why is OPTS == 0 needed?

jglogan (changes requested):
@saehejkang Thanks! Looks like the top line of the usage message needs an update

Also one question about the latest version check.

scripts/update-container.sh
saehejkang: good catch

scripts/update-container.sh
saehejkang: was mainly a redundancy check but can be removed

jglogan (approved):

dcantah (approved):

PR #1470, opened 2026-04-29
ImagePush: print image reference to stdout on success

### LINKED ISSUE #1466: [Bug]: `container image push` should output imageref to stdout on success.
### I have done the following

### Steps to reproduce

Tag an image for a repository (for example, `docker.io/max-mustermann/scratch/test:latest`) and run:

```bash
container image push docker.io/max-mustermann/scratch/test:latest 2> /dev/null
```

### Problem description

Expected: On success, the command writes the reference `docker.io/max-mustermann/scratch/test:latest` to the standard output.

We should discuss here whether to echo back the supplied reference, or print the fully qualified reference.

Actual: The command produces no standard output.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

mvanhorn:
Resolves #1466.

`container image push <ref>` previously produced no standard output on success. Print the fully qualified reference (`image.reference`) after the push completes so callers can pipe the output into subsequent commands.

The issue raised the question: "echo back the supplied reference, or print the fully qualified reference?" — this PR prints `image.reference`, which is the normalized/fully-qualified form (e.g. `docker.io/max-mustermann/scratch/test:latest` from the issue's example), matching the same field that `ImagePull` and the progress bar already use.

This mirrors the existing pattern in `ImageTag.run()`, which prints the resolved target after the tag completes:

```swift
try await existing.tag(new: targetReference)
print(target)
```

Diff: 1 line.

### DIFF
--- a/Sources/ContainerCommands/Image/ImagePush.swift
+++ b/Sources/ContainerCommands/Image/ImagePush.swift
@@ -75,6 +75,7 @@ extension Application {
             progress.start()
             _ = try await image.push(platform: p, scheme: scheme, progressUpdate: progress.handler)
             progress.finish()
+            print(image.reference)
         }
     }
 }


jglogan (approved):

jglogan:
@mvanhorn Thanks for tackling that so quickly!

mvanhorn:
Thanks @jglogan! Glad it landed.

PR #1471, opened 2026-04-29
Tests/CLITests: drain stdout/stderr concurrently in CLITest.run

### LINKED ISSUE #1456: [Bug]: Deadlock risk in CLI tests.
### I have done the following

### Steps to reproduce

We're not seeing this issue in tests today, but a deadlock could happen under certain conditions.

### Problem description

The problem lies in the code here: https://github.com/apple/container/blob/main/Tests/CLITests/Utilities/CLITest.swift#L182

If a process writes more than a pipe buffer's worth of data to stderr, the underlying `write()` sys call will block, which means that the stdout `readDataToEndOfFile()` is unable to make forward progress.

We can use readabilityHandlers, detached Tasks, or redirect to FileHandles to resolve this.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

mvanhorn:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

`CLITest.run(arguments:stdin:currentDirectory:env:)` reads stdout to EOF before stderr begins draining (`Tests/CLITests/Utilities/CLITest.swift:182-183`). If a child process writes more than a pipe-buffer's worth of data to stderr (~64 KB on macOS), the child blocks in `write()` on stderr while we block in `readDataToEndOfFile()` on stdout, and neither side makes progress until `process.waitUntilExit()` returns. This is a latent deadlock that any future test or CLI verbosity bump can trigger.

This PR drains both streams concurrently using `readabilityHandler` closures backed by `Mutex<Data>` buffers. After `process.waitUntilExit()` returns the handlers are cleared and `readDataToEndOfFile()` flushes any bytes the kernel buffered between the last handler invocation and exit. The error path also clears handlers so a failed `process.run()` does not leak callbacks. `Mutex` matches the locking primitive the file already uses for `commandSeq`.

Fixes #1456

## Testing

- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

`swift build --target CLITests` passes against the change. `make swift-fmt-check` is clean. Behavior is unchanged for any test command that previously fit within a single pipe buffer, so the existing CLITests still exercise the helper end-to-end.

I did not add a new test that emits >64 KB to stderr because every existing CLITest invocation goes through `executablePath` (the `container` binary), and reproducing the deadlock requires a child that emits a controllable amount on stderr. Happy to follow up with a small refactor that extracts the drain into a static helper plus a regression test that drives it via `/bin/sh` if that would be useful.

### DIFF
--- a/Tests/CLITests/Utilities/CLITest.swift
+++ b/Tests/CLITests/Utilities/CLITest.swift
@@ -165,31 +165,29 @@ class CLITest {
         }
 
         let inputPipe = Pipe()
-        let outputPipe = Pipe()
-        let errorPipe = Pipe()
         process.standardInput = inputPipe
-        process.standardOutput = outputPipe
-        process.standardError = errorPipe
 
-        // Drain stdout and stderr concurrently via readability handlers so the
-        // child process never blocks on `write()` when one of its streams fills
-        // the kernel pipe buffer. Reading stdout to EOF before stderr deadlocks
-        // any process that emits more than a pipe-buffer's worth of stderr.
-        let outputBuffer = Mutex<Data>(Data())
-        let errorBuffer = Mutex<Data>(Data())
         let outputData: Data
         let errorData: Data
         do {
-            outputPipe.fileHandleForReading.readabilityHandler = { handle in
-                let chunk = handle.availableData
-                guard !chunk.isEmpty else { return }
-                outputBuffer.withLock { $0.append(chunk) }
+            // Redirect stdout/stderr to temp files so the child process never
+            // blocks on `write()` when one stream fills the kernel pipe buffer
+            // before the parent drains it (issue #1456).
+            let tempDir = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
+            try FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)
+            defer {
+                try? FileManager.default.removeItem(at: tempDir)
             }
-            errorPipe.fileHandleForReading.readabilityHandler = { handle in
-                let chunk = handle.availableData
-                guard !chunk.isEmpty else { return }
-                errorBuffer.withLock { $0.append(chunk) }
-            }
+
+            let stdoutURL = tempDir.appendingPathComponent("stdout")
+            let stderrURL = tempDir.appendingPathComponent("stderr")
+            FileManager.default.createFile(atPath: stdoutURL.path, contents: nil)
+            FileManager.default.createFile(atPath: stderrURL.path, contents: nil)
+
+            let stdoutHandle = try FileHandle(forWritingTo: stdoutURL)
+            let stderrHandle = try FileHandle(forWritingTo: stderrURL)
+            process.standardOutput = stdoutHandle
+            process.standardError = stderrHandle
 
             try process.run()
             if let data = stdin {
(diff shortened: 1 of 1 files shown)


katiewasnothere:
Hi @mvanhorn thank you for contributing! What are your thoughts on redirecting the out/err to FileHandles vs using readabilityhandlers?

mvanhorn:
Both are valid - the issue listed each as acceptable. I went with `readabilityHandler` because the diff stays in one file (no temp-file lifecycle or error-path cleanup) and `Mutex` is already used in `CLITest.swift`, so guarding the buffer matches the existing style.

The `FileHandle` route is arguably simpler to reason about (kernel handles buffering, no two-stage drain) at the cost of FS setup and cleanup per test.

If you'd prefer the `FileHandle` version, happy to swap - it's a clean rewrite of the same block. Which way would you like?

mvanhorn:
also @katiewasnothere I live on Mercer Island! PNW represent!

katiewasnothere:
@mvanhorn I would prefer the file based option. Many CLITests and helper functions already follow a pattern of creating and cleaning up temporary directories and files.

mvanhorn:
Done in 20dddd2. Followed the `withTempDir` pattern already in `CLITest.swift` - UUID-named dir under `FileManager.default.temporaryDirectory`, FileHandle for the writes, `defer` cleanup.

Net -13 lines (no Mutex, no handler-clearing, no two-stage drain).

Tests/CLITests/Utilities/CLITest.swift
```diff
@@ -165,23 +165,42 @@ class CLITest {
...
+            // before the parent drains it (issue #1456).
+            let tempDir = FileManager.default.temporaryDirectory.appendingPathComponent(UUID().uuidString)
+            try FileManager.default.createDirectory(at: tempDir, withIntermediateDirectories: true)
+            defer {
+                try? FileManager.default.removeItem(at: tempDir)
+            }
+
+            let stdoutURL = tempDir.appendingPathComponent("stdout")
+            let stderrURL = tempDir.appendingPathComponent("stderr")
+            FileManager.default.createFile(atPath: stdoutURL.path, contents: nil)
```
jglogan, line 184: @mvanhorn How about using `defer` here?

```patch
             let stdoutHandle = try FileHandle(forWritingTo: stdoutURL)
             let stderrHandle = try FileHandle(forWritingTo: stderrURL)
+            defer { try? stdoutHandle.close() }
+            let stderrHandle = try FileHandle(forWritingTo: stderrURL)
+            defer { try? stderrHandle.close() }
             process.standardOutput = stdoutHandle
             process.standardError = stderrHandle
 
             try process.run()
             process.waitUntilExit()
-
-            try? stdoutHandle.close()
-            try? stderrHandle.close()
-
             outputData = try Data(contentsOf: stdoutURL)
             errorData = try Data(contentsOf: stderrURL)
```

jglogan (changes requested):

mvanhorn:
Done in 71c649e. The `defer` approach also covers the `try process.run()` throw path, which the explicit close calls after `waitUntilExit()` didn't.

katiewasnothere (approved):

jglogan (approved):

PR #1474, opened 2026-04-29
Change PR build to generate coverage and comment it

noah-thor:
This extends the github actions to by default run the coverage generation targets. The targets will generate the coverage artifacts, which will then be used by a seprate github action to comment on the PR the coverage numbers. 

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This is a sanity check for testing information for new features and refactors alike. It will allow reviewers to quickly identify if there are major issues with new tests, or large changes to coverage percentage that indicate a problem.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -10,6 +10,14 @@ on:
         type: boolean
         description: "Publish this build for release"
         default: false
+      coverage:
+        type: boolean
+        description: "Run tests with code coverage enabled"
+        default: false
+      pr_number:
+        type: string
+        description: "PR number for coverage artifact naming (required when coverage is true)"
+        default: ""
 
 jobs:
   buildAndTest:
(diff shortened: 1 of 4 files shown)


noah-thor:
Build failing due to `main` having issues, will be fixed when https://github.com/apple/container/pull/1473 is merged

jglogan (approved):

Pedro19875 (changes requested):
96a9d2bb72

FVN222103268[]()

.github/workflows/common.yml
```diff
@@ -82,10 +100,53 @@ jobs:
...
+          echo "Coverage data:"
+          cat pr-coverage/*.txt
+
+      - name: Upload coverage data
+        if: ${{ inputs.coverage }}
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: pr-coverage-${{ inputs.pr_number }}
+          path: pr-coverage/
+          retention-days: 1
```
katiewasnothere, line 135: Should this be 14 as well or what is the reason for only keeping this for a day?

noah-thor: We could make it longer, but this is only used for the comment bot, so it should get read immediately. The html report I am thinking we may link to in a future iteration and people may want to view it, so the 14 days allows time for people to see it.

katiewasnothere (approved):

PR #1451, opened 2026-04-25
Fix "not installing shadowed plugin" startup log noise.

jglogan:
- Closes #1450.
- If we require that bundle-packaged installs must populate a Unix-structured `bin` and `libexec`, we don't need the code that searches the app bundle install root. As it stands, we enumerate both the app bundle and the Linux paths, meaning we're looking at the same plugins (in different paths) twice.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Confusing messages in logs.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

PR #1490, opened 2026-04-30
Add app bundle support to APIServer plugin loader

katiewasnothere:
## Type of Change
- [x] Bug fix

## Motivation and Context
Other locations in the code, including the CLI commands [here](https://github.com/apple/container/blob/1ee2d9d05429008bd98809d7cbfe1e0293248bfc/Sources/ContainerCommands/Application.swift#L160), additionally support loading plugins from a macOS bundle. This PR adds that support to the APIServer's PluginLoader as well for consistency. 

## Testing
- [x] Tested locally


katiewasnothere:
Closing. I missed that we recently removed this explicitly in https://github.com/apple/container/pull/1451.

PR #1518, opened 2026-05-07
Use SystemPath for PacketFilter.

chrisgeo:
- Closes #1517.
- Using URL for filesystem paths is bad practice. FilePath is safer and more ergonomic.
- Sibling of #1480 (HostDNSResolver), same pattern.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
URL conflates web URLs with filesystem paths and carries percent-encoding semantics that are inappropriate for filesystem operations. SystemPackage's FilePath is purpose-built for filesystem paths.

Changes in this PR:
- `configURL` / `anchorsURL` → `configPath` / `anchorsPath` (`FilePath`)
- `defaultConfigPath` / `defaultAnchorsPath` now `FilePath`
- `FileManager` calls switch from `atPath: url.path` to `atPath: path.string`
- Test uses `String(contentsOfFile: path.string, …)` in place of `String(contentsOf: URL, …)`

`Process.executableURL = URL(fileURLWithPath: "/sbin/pfctl")` is preserved as `URL` since `Foundation.Process` requires it.

## Testing
- [x] Tested locally (`swift test --filter PacketFilterTest` passes; full `ContainerAPIClientTests` suite passes — 171 tests, 0 failures)
- [x] Added/updated tests (test fixtures updated to construct `PacketFilter` with `FilePath`)
- [ ] Added/updated docs

PR #1480, opened 2026-04-30
Use SystemPath for HostDNSResolver.

jglogan:
- Closes #1479.
- Using URL for filesystem paths is bad practice. FilePath is safer and more ergonomic.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
See above

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

PR #1522, opened 2026-05-08
Use SystemPath for DirectoryWatcher.

### LINKED ISSUE #1520: [Request]: TECH DEBT - Convert DirectoryWatcher to use FilePath instead of URL.

Continuing the URL → FilePath migration discussed in #1481.

`DirectoryWatcher` (in `Sources/ContainerOS/`) holds a directory URL and emits an array of URLs from its watch handler — both should be `FilePath` per the migration. Same pattern as #1480 (HostDNSResolver) and #1518 (PacketFilter).

### Code of Conduct

chrisgeo:
- Closes #1520.
- Using URL for filesystem paths is bad practice. FilePath is safer and more ergonomic.
- Same pattern as #1480 (HostDNSResolver) and #1518 (PacketFilter).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Continues the URL → FilePath migration discussed in #1481.

Changes in this PR:
- `directoryURL: URL` → `directoryPath: FilePath` (public API)
- `init(directoryURL:, log:)` → `init(directoryPath:, log:)`
- `startWatching` handler signature: `([URL]) throws -> Void` → `([FilePath]) throws -> Void`
- `Sources/APIServer/LocalhostDNSHandler.swift` updated to call the new init directly (the previous `URL(filePath: configPath.string)` wrapping shim is no longer needed)

Behavioral note: `URL.lastPathComponent` returns `String`; `FilePath.lastComponent` returns `FilePath.Component?`. The filename filter in `LocalhostDNSHandler` handles the optional via `$0.lastComponent?.string.starts(with: ...) == true`, preserving the prior "skip malformed entries" behavior.

## Testing
- [x] Tested locally (`swift test --filter DirectoryWatcherTest` and `--filter ContainerOSTests` both pass)
- [x] Added/updated tests (test fixtures updated to construct `DirectoryWatcher` with `FilePath`)
- [ ] Added/updated docs

### DIFF
--- a/Sources/APIServer/LocalhostDNSHandler.swift
+++ b/Sources/APIServer/LocalhostDNSHandler.swift
@@ -51,7 +51,9 @@ actor LocalhostDNSHandler: DNSHandler {
                 if let match = content.firstMatch(of: regex),
                     let ipv4 = (match[1].substring.flatMap { try? IPv4Address(String($0)) })
                 {
-                    let lastName = file.lastComponent?.string ?? ""
+                    guard let lastName = file.lastComponent?.string else {
+                        continue
+                    }
                     let name = String(lastName.dropFirst(HostDNSResolver.containerizationPrefix.count))
                     guard let dnsName = try? DNSName(name) else {
                         continue
--- a/Sources/APIServer/LocalhostDNSHandler.swift
+++ b/Sources/APIServer/LocalhostDNSHandler.swift
@@ -34,22 +34,25 @@ actor LocalhostDNSHandler: DNSHandler {
     public init(configPath: FilePath = HostDNSResolver.defaultConfigPath, ttl: UInt32 = 5, log: Logger) {
         self.ttl = ttl
 
-        self.watcher = DirectoryWatcher(directoryURL: URL(filePath: configPath.string), log: log)
+        self.watcher = DirectoryWatcher(directoryPath: configPath, log: log)
         self.dns = [DNSName: IPv4Address]()
     }
 
     public func monitorResolvers() async {
-        await self.watcher.startWatching { [weak self] fileURLs in
+        await self.watcher.startWatching { [weak self] filePaths in
             var dns: [DNSName: IPv4Address] = [:]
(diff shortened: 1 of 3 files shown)


Sources/APIServer/LocalhostDNSHandler.swift
```diff
@@ -34,22 +34,25 @@ actor LocalhostDNSHandler: DNSHandler {
...
+            for file in filePaths.filter({
+                $0.lastComponent?.string.starts(with: HostDNSResolver.containerizationPrefix) == true
+            }) {
+                let content = try String(contentsOfFile: file.string, encoding: .utf8)
 
                 if let match = content.firstMatch(of: regex),
                     let ipv4 = (match[1].substring.flatMap { try? IPv4Address(String($0)) })
                 {
-                    let name = String(file.lastPathComponent.dropFirst(HostDNSResolver.containerizationPrefix.count))
+                    let lastName = file.lastComponent?.string ?? ""
```
jglogan: nit: instead of nil to empty string here, perhaps this?

```swift
let dnsName = file.lastComponent?.string
    .map { $0.dropFirst(HostDNSResolver.containerizationPrefix.count) }
    .map { try? DNSName($0) }
guard let dnsName else {
    continue
}
```

jglogan (commented):
Building now! One comment, just a nit.

Sources/APIServer/LocalhostDNSHandler.swift
chrisgeo: Great catch, thank you! I'm still learning some idioms in Swift. I switched to using the `guard` from your example and tests pass locally.

```swift
 guard let lastName = file.lastComponent?.string else {
     continue
  }
```

jglogan: That is also idiomatic and easy to read, thanks!

jglogan (approved):

PR #1523, opened 2026-05-08
Use SystemPath for ContainerResource.Filesystem.

### LINKED ISSUE #1521: [Request]: TECH DEBT - Convert ContainerResource.Filesystem to use FilePath instead of URL.

Continuing the URL → FilePath migration discussed in #1481.

`Sources/ContainerResource/Container/Filesystem.swift` has three factory methods (`block`, `volume`, `virtiofs`) that use `URL(fileURLWithPath: source).absolutePath()` to normalize a `String` path. The URL round-trip is unnecessary — every caller passes absolute paths, and `FilePath` provides the equivalent separator normalization.

### Code of Conduct

chrisgeo:
- Closes #1521.
- Using URL for filesystem paths is bad practice. FilePath is safer and more ergonomic.
- Same pattern as #1480 (HostDNSResolver) and #1518 (PacketFilter).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Continues the URL → FilePath migration discussed in #1481. Tiny diff (7 lines, 1 file).

```diff
-source: URL(fileURLWithPath: source).absolutePath(),
+source: FilePath(source).string,
```

Affects the three `Filesystem` factory methods: `block`, `volume`, `virtiofs`.

The original `URL(fileURLWithPath: source).absolutePath()` was a no-op round-trip on absolute inputs (which is what every caller passes today — `Parser.swift`, `Utility.swift`, `SnapshotStore.swift`, and `RuntimeConfigurationTests`). `FilePath(source).string` preserves the same separator normalization without the URL detour.

## Testing
- [x] Tested locally (`swift test --filter ContainerResourceTests` passes — no regressions)
- [ ] Added/updated tests (no behavior change; existing tests cover the contract)
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Container/Filesystem.swift
+++ b/Sources/ContainerResource/Container/Filesystem.swift
@@ -15,6 +15,7 @@
 //===----------------------------------------------------------------------===//
 
 import Foundation
+import SystemPackage
 
 /// Options to pass to a mount call.
 public typealias MountOptions = [String]
@@ -95,7 +96,7 @@ public struct Filesystem: Sendable, Codable {
     ) -> Filesystem {
         .init(
             type: .block(format: format, cache: cache, sync: sync),
-            source: URL(fileURLWithPath: source).absolutePath(),
+            source: FilePath(source).string,
             destination: destination,
             options: options
         )
@@ -108,7 +109,7 @@ public struct Filesystem: Sendable, Codable {
     ) -> Filesystem {
         .init(
             type: .volume(name: name, format: format, cache: cache, sync: sync),
-            source: URL(fileURLWithPath: source).absolutePath(),
+            source: FilePath(source).string,
             destination: destination,
             options: options
         )
@@ -118,7 +119,7 @@ public struct Filesystem: Sendable, Codable {
     public static func virtiofs(source: String, destination: String, options: MountOptions) -> Filesystem {
         .init(
             type: .virtiofs,
-            source: URL(fileURLWithPath: source).absolutePath(),
+            source: FilePath(source).string,
             destination: destination,
             options: options
         )
--- a/Sources/ContainerResource/Container/Filesystem.swift
+++ b/Sources/ContainerResource/Container/Filesystem.swift
@@ -96,7 +96,7 @@ public struct Filesystem: Sendable, Codable {
     ) -> Filesystem {
         .init(
             type: .block(format: format, cache: cache, sync: sync),
-            source: FilePath(source).string,
+            source: absoluteFilePath(for: source).string,
             destination: destination,
             options: options
         )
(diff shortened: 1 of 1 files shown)


Sources/ContainerResource/Container/Filesystem.swift
```diff
@@ -95,7 +96,7 @@ public struct Filesystem: Sendable, Codable {
     ) -> Filesystem {
         .init(
             type: .block(format: format, cache: cache, sync: sync),
-            source: URL(fileURLWithPath: source).absolutePath(),
```
jglogan, line 98: For this if we want to be functionally equivalent, we'd need something like this:

```swift
func absoluteFilePath(for source: String) -> FilePath {
    let path = FilePath(source)
    guard path.isRelative else { return path.lexicallyNormalized() }
    return FilePath(FileManager.default.currentDirectoryPath)
        .pushing(path)
        .lexicallyNormalized()
}
```

jglogan (changes requested):

Sources/ContainerResource/Container/Filesystem.swift
chrisgeo: I went the private function route to keep the functionality centered on this changeset.

Please let me know if you'd prefer that be adjusted. Thanks again for the feedback!

jglogan:
@chrisgeo A little discussion on this one as well. At first I was thinking, why doesn't the PR use FilePath for the source and destination path properties?

For the source it needs to stay string because, possibly among other things, the source could be a network location.

I think that in a follow-on we could consider changing the destination path to a FilePath, but what you have in this PR is a good first step toward that.

Can you think of any possible instance where the destination path would be something else? Pretty sure any `mount(2)` destination on any Unix-like system is a filesystem path.

jglogan (approved):

jglogan:
See discussion above, but the PR is fine as is, merged and thank you!

jglogan:
> Can you think of any possible instance where the destination path would be something else? Pretty sure any `mount(2)` destination on any Unix-like system is a filesystem path.

We shouldn't rush such a change. Leaving it as String allows us to have the flexibility to define other meanings, such as for cases where the receiver of a Filesystem needs to compute the eventual destination path.

chrisgeo:
> We shouldn't rush such a change. Leaving it as String allows us to have the flexibility to define other meanings, such as for cases where the receiver of a Filesystem needs to compute the eventual destination path.

@jglogan Sorry - I missed this before you merged. I took a look couldn't come up with anything we'd miss with with mount(2) options since I recall all unix-based system `fstypes` are paths.

Happy to dive into options for a follow-up when I can

PR #1525, opened 2026-05-11
fix: clamp ProgressBar barLength to non-negative value to avoid crash

### LINKED ISSUE #1547: [Bug]: Progress bar lacks input validation for negative `size` parameter.
### I have done the following

### Steps to reproduce

No known way to reproduce, but makes progress reporting susceptible to logic errors.

### Problem description

If a client calls `set(size: val)` where val is a negative number for any reason, the subsequent `draw` operation can crash when trying to call `String(repeating:count:)`. The set operation should protect against this input by clamping the minimum computed progress to zero.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

ErwanLegrand:
- Closes #1547.
- A negative progress value (e.g. from a race in progress events) would produce a negative barLength and crash inside String(repeating:count:). Wrap the computed length in max(0, ...) and add a regression test that calls set(size: -10) and exercises draw(state:detail:).

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This fixes an Apple container crash

## Testing
- [ ] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/TerminalProgress/ProgressBar.swift
+++ b/Sources/TerminalProgress/ProgressBar.swift
@@ -275,7 +275,7 @@ extension ProgressBar {
             // 45 reserves space for components rendered after the bar (size, speed, time, etc.)
             let usedWidth = (useColor ? joinedComponents.visibleLength : joinedComponents.count) + 45
             let remainingWidth = max(config.width - usedWidth, 1)
-            let barLength = min(remainingWidth, state.finished ? remainingWidth : Int(Int64(remainingWidth) * value / total))
+            let barLength = min(remainingWidth, max(0, state.finished ? remainingWidth : Int(Int64(remainingWidth) * value / total)))
             let barPaddingLength = remainingWidth - barLength
             if useColor {
                 let filledBar = EscapeSequence.colored(String(repeating: config.theme.bar, count: barLength), EscapeSequence.green)
--- a/Tests/TerminalProgressTests/ProgressBarTests.swift
+++ b/Tests/TerminalProgressTests/ProgressBarTests.swift
@@ -749,6 +749,22 @@ final class ProgressBarTests: XCTestCase {
         let _ = progress.draw()
     }
 
+    func testProgressBarNegativeValue() async throws {
+        // Regression test: a negative progress value (e.g. from a race in progress events)
+        // must not cause String(repeating:count:) to be called with a negative count.
+        let config = try ProgressConfig(
+            description: "Task",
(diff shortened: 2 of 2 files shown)


Sources/TerminalProgress/ProgressBar.swift
```diff
@@ -275,7 +275,7 @@ extension ProgressBar {
             // 45 reserves space for components rendered after the bar (size, speed, time, etc.)
             let usedWidth = (useColor ? joinedComponents.visibleLength : joinedComponents.count) + 45
             let remainingWidth = max(config.width - usedWidth, 1)
-            let barLength = min(remainingWidth, state.finished ? remainingWidth : Int(Int64(remainingWidth) * value / total))
+            let barLength = min(remainingWidth, max(0, state.finished ? remainingWidth : Int(Int64(remainingWidth) * value / total)))
```
jglogan, line 278: @dkovba I'd welcome suggestions on how to make this more readable. The clamp makes sense in terms of preventing a negative bar length, but the way it's written, it's pretty complicated to reason about.

There is no need to include the entire thing in the clamp since remainingWidth is already guaranteed to be greater than 1.

The value `Int(Int64(remainingWidth) * value / total))` should be extracted to its own `let` assignment with a constant name that makes sense.

This would make more sense to me if we did something like:

```swift
// reserve spaces between components, plus 45 for components rendered after the bar (size, speed, time, etc.)
let reservedWidth = (useColor ? joinedComponents.visibleLength : joinedComponents.count) + 45
let completedBarWidth = max(config.width - usedWidth, 1)
let progressWidth = max(0, Int(Int64(completedBarWidth) * value / total))
let currentBarWidth: Int
if state.finished {
    currentBarWidth = completedBarWidth
} else {
    currentBarWidth = min(completedBarWidth, progressWidth)
}
```

jglogan (changes requested):

jglogan (changes requested):
@ErwanLegrand Fix looks good but let's make the logic a little more readable. Thanks!

jglogan (approved):

jglogan:
@ErwanLegrand merged, thank you! I'll apply the rework to make the intermediate computations and tests a little more readable.

ErwanLegrand:
@jglogan sorry, I had missed your first message!

PR #1543, opened 2026-05-11
Add support for layered and plugin configurations

noah-thor:
This adds support for reading configurations from a three layer hierarchy:
1. User provided roml
2. Install root toml
3. Code defaults

Additionally we add some code to support plugin configurations via the ConfigurationLoader. Each plugin can provide a struct with an accompanying id that gets used to parse the scoped section of the TOML.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Allows users at install time to configure package widgets defaults without having to recompile `container`

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerCommands/Builder/BuilderStart.swift
```diff
@@ -55,7 +55,7 @@ extension Application {
         public init() {}
 
         public func run() async throws {
-            let containerSystemConfig: ContainerSystemConfig = try ConfigurationLoader.load()
+            let containerSystemConfig: ContainerSystemConfig = try await ConfigurationLoader.load()
```
katiewasnothere, line 58: Wouldn't these not get the right environment variables for when the app root and install root are set by the user in `system start` since `system start` only sets the environment variables on the APIServer's service?

Sources/ContainerPersistence/ConfigurationLoader.swift
```diff
@@ -14,98 +14,204 @@
...
-        let path = configurationFile ?? Self.configurationFile(in: PathUtils.BaseConfigPath.appRoot.basePath())
-        guard FileManager.default.fileExists(atPath: path.string) else {
+    /// - Parameters:
+    ///   - type: The concrete `LoadableConfiguration` type to decode.
+    ///   - configurationFiles: Ordered TOML layers; empty falls back to `defaultConfigFiles()`.
+    ///   - scope: Optional `ConfigKey` to scope the snapshot before decoding.
+    ///   - decodeErrorContext: Prefix used in the `invalidArgument` error thrown on decode failure.
+    private static func loadAndDecode<T: LoadableConfiguration>(
+        _ type: T.Type,
+        configurationFiles: [FilePath],
```
katiewasnothere, line 149: Should this just be 

```
configurationFiles: [FilePath] = defaultConfigFiles()
```

like the other functions?

Sources/ContainerCommands/Builder/BuilderStart.swift
noah-thor: Which env var do you mean? Could you give a concrete example for my understanding?

katiewasnothere: If a user ran `system start --app-root <test app root> --install-root <test install root>`, the configuration would be loaded correctly in the system start code and any services created by APIServer. For the other CLI calls where we fallback to the default config files (aka we just call `ConfigurationLoader.load()`), the base config path for the app root configuration file would first check for the environment variable `CONTAINER_APP_ROOT` which would not be set in this shell instance (but is set in the APIServer's env), then fallback to the default ~/Library/Application\ Support/com.apple.container app root path (see [here](https://github.com/apple/container/blob/940fefb7cd79c0d658e6e9f98ea777aa4738faee/Sources/ContainerPersistence/PathUtils.swift#L36)). Same for the install root.

Sources/ContainerPersistence/ConfigurationLoader.swift
noah-thor: We can do that, but the two call sites of this private are using that same `defaultConfigFiles` logic so the outcome would be unchanged.

Sources/ContainerPersistence/ConfigurationLoader.swift
```diff
@@ -14,98 +14,204 @@
...
+        if paths.allSatisfy({ !fm.fileExists(atPath: $0.string) }) {
             return T()
         }
+
+        var providers: [FileProvider<TOMLSnapshot>] = []
+        for path in paths {
+            do {
+                try providers.append(await FileProvider<TOMLSnapshot>(filePath: path, allowMissing: true))
+            } catch {
+                throw ContainerizationError(
```
jglogan, line 164: Looking at this and L155: if none of the files exist, we return an empty config, but if some exist and some don't we throw?

noah-thor: Right now the logic is:
* If no files exist return `T()`
* else make a provider for every file
  * providers allow for missing files via `allowMissing: true`, any missing files will have no change to the `ConfigReader` view
* an exception is thrown in the case that creating a `FileProvider` causes some unexpected exception like malformed text, etc

jglogan (approved):

PR #1544, opened 2026-05-11
Use allocate with session to automatically clean up IPs.

jglogan:
- Part of #1318.
- Part of #1378.
- Removes network plugin `deallocate()`, and allocate takes an `XPCServerSession` that registers an `onDisconnect` handler that performs deallocation.
- ContainerService now tracks `networkSessions` for allocations.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fix bug and simplify IP allocation.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerNetworkService/Server/NetworkService.swift
+++ b/Sources/Services/ContainerNetworkService/Server/NetworkService.swift
@@ -27,6 +27,7 @@ public actor NetworkService: Sendable {
     private let log: Logger
     private var allocator: AttachmentAllocator
     private var macAddresses: [UInt32: MACAddress]
+    private var allocationsBySession: [XPCServerSession: [(hostname: String, index: UInt32)]] = [:]
 
     /// Set up a network service for the specified network.
     public init(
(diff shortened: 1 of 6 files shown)


Sources/Services/ContainerNetworkService/Server/NetworkService.swift
```diff
@@ -27,6 +27,7 @@ public actor NetworkService: Sendable {
     private let log: Logger
     private var allocator: AttachmentAllocator
     private var macAddresses: [UInt32: MACAddress]
+    private var allocationsBySession: [XPCServerSession: [(hostname: String, index: UInt32)]] = [:]
```
katiewasnothere, line 30: small, small nit but could we set this default value in the init for consistency?

katiewasnothere (approved):

Sources/Services/ContainerNetworkService/Server/NetworkService.swift
jglogan: Sure! I'll fix in the third PR.

PR #1546, opened 2026-05-12
Reduce deployed size by 90% using multicall binary + stripping

thomasjm:
## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

I noticed that the current releases contain 5 large binaries of about 50 MB each: `container`, `container-apiserver`, `container-core-images`, `container-network-vmnet`, and `container-runtime-linux`. These binaries contain substantially similar code.

This PR merges them into a single multicall binary, with symlinks to determine which one to run. This has excellent effects on the size of the `.pkg` file, both before and after extraction:

| Artifact | Master | This PR | Reduction |                                                                                                                                                                                    
| --- | ---: | ---: | ---: |                                                                                                                                                                                                                 
| `.pkg` file | 68.6 MB | 15.1 MB | 78.0% |                                            
| Extracted `bin/` + `libexec/` | 259.1 MB | 57.2 MB | 77.9% |                        
| Installed executables | 5 binaries | 1 binary + 4 symlinks | — |

The downside is that I had to give the combined binary the `com.apple.security.virtualization` entitlement, which was previously only possessed by `container-network-vmnet` and `container-runtime-linux`. This expands the potential attack surface, because now one of the other binaries could be subverted to gain virtualization-related permissions.

FWIW, the way to have the best of both worlds is to extract the common code to a dylib and have all the binaries dynamically link against it.

Anyway, I've tested this locally and it seems to work! Just putting it up to solicit feedback. 

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


jglogan:
@thomasjm That's a cool experiment, thanks! Getting the pkg down to 15MB and the extracted footprint well under 100MB is impressive.

Instead of using symlinks, I suppose we could build an all in one binary with each service as a subcommand, and we'd need to rework the plugin loader / launch types to use slightly different arguments.

Creating a dylib for all the libraries probably means that some apps are getting stuff they don't really strictly need but that doesn't seem like a big deal, but should result in a slightly smaller in-memory footprint in addition to the smaller installer and extracted sizes. It might also yield slightly better startup performance for the first container?

It's worth getting more discussion on it.

jglogan:
FYI see https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests regarding signing commits.

dcantah:
An alternative approach is just `strip`ing the binaries. Saves about 25MB per binary for me.

thomasjm:
> Instead of using symlinks, I suppose we could build an all in one binary with each service as a subcommand

IMO, if you're okay with a single binary approach, then the multicall is nice for modularity. If you think of it just as an optimization applied at the end then the binaries can remain morally separate. Then it's easier to e.g. replace the vmnet helper with a different one as long as the replacement satisfies the helper's (relatively simple) argument format.

> An alternative approach is just striping the binaries.

Right, I was just experimenting with that too. Here's an updated table with stripping included.

| Artifact | Master | Strip only | Multicall only | Multicall + strip |
| --- | ---: | ---: | ---: | ---: |
| `.pkg` file | 68.6 MB | 49.2 MB (−28.3%) | 15.1 MB (−78.0%) | 10.9 MB (−84.1%) |
| Extracted `bin/` + `libexec/` | 259.1 MB | 121.2 MB (−53.2%) | 57.2 MB (−77.9%) | 26.8 MB (−89.7%) |
| Installed executables | 5 binaries | 5 binaries | 1 binary + 4 symlinks | 1 binary + 4 symlinks |

I believe that beyond just disk space, keeping everything in a single binary will save memory at runtime because the OS will only load the code into memory once. Whereas with multiple stripped binaries, it will load each one that is running at any given time. This is the same advantage that would apply to a common dylib + shim binaries.

thomasjm:
Hello, just checking in here. Do you think you'll want to merge this?

PR #1548, opened 2026-05-12
Add `variant` Support

ajemory:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This change adds two types to support opaque runtime data passing through the APIServer:
- RuntimeConfiguration `data`
- RuntimeLinuxData type

LinuxRuntimeData defines runtime specific information. With this change, it will support `variant` only. The optional RuntimeConfiguration `data` field encodes runtime specific data to pass through the APIServer. It is decoded as needed by the runtime. 

The idea is to eventually move all runtime specific data into the `data` field so that the APIServer is only aware of generic container information.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -277,7 +277,7 @@ public actor ContainersService {
     }
 
     /// Create a new container from the provided id and configuration.
-    public func create(configuration: ContainerConfiguration, kernel: Kernel, options: ContainerCreateOptions, initImage: String? = nil) async throws {
+    public func create(configuration: ContainerConfiguration, kernel: Kernel, options: ContainerCreateOptions, initImage: String? = nil, runtimeData: Data? = nil) async throws {
         log.debug(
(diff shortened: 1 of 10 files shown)


Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -281,7 +281,7 @@ public actor ContainersService {
     }
 
     /// Create a new container from the provided id and configuration.
-    public func create(configuration: ContainerConfiguration, kernel: Kernel, options: ContainerCreateOptions, initImage: String? = nil) async throws {
+    public func create(configuration: ContainerConfiguration, kernel: Kernel, options: ContainerCreateOptions, initImage: String? = nil, runtimeData: Data? = nil) async throws {
```
jglogan, line 280: Should we have `bootstrap()` create an empty runtime configuration file if one doesn't exist, at least for a few releases?

jglogan (approved):

PR #1551, opened 2026-05-13
Integration coverage xpc helpers

noah-thor:
## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This fixes the LLVM coverage data not properly being emitted for XPC
services. It requires piping the `LLVM_PROFILE_FILE` environment
variable through to all the services and plugins. The variable itself
also required the "%c" formatter to ensure that it continuously emits
coverage data, otherwise when XPC services are killed via "bootout" they
do not emit coverage.

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerPlugin/PluginLoader.swift
```diff
@@ -205,6 +205,8 @@ extension PluginLoader {
         "http_proxy", "HTTP_PROXY",
         "https_proxy", "HTTPS_PROXY",
         "no_proxy", "NO_PROXY",
+        // Allows LLVM coverage profiling data to be written by launchd-managed helper processes.
+        "LLVM_PROFILE_FILE",
```
jglogan: Would it make sense to conditionally include this using an `#if` defined for coverage builds?

noah-thor: Good idea, will throw an `ifdef` in

jglogan (approved):

Issue #1527, opened 2026-05-11
[Bug]: `container system df` double-counts image storage when the same image has multiple tags


realrajaryan:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Pull one image, tag it, and compare `system df` before and after the tag:

```console
$ container system df --format json
{
  "containers" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  },
  "images" : {
    "active" : 0,
    "reclaimable" : 78729216,
    "sizeInBytes" : 78729216,
    "total" : 1
  },
  "volumes" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  }
}

$ container image tag alpine:latest local/alpine:issue-repro
local/alpine:issue-repro

$ container system df --format json
{
  "containers" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  },
  "images" : {
    "active" : 0,
    "reclaimable" : 157458432,
    "sizeInBytes" : 157458432,
    "total" : 2
  },
  "volumes" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  }
}

$ du -sk "/Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/content/blobs/sha256"
4124    /Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/content/blobs/sha256

$ du -sk "/Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/snapshots"
76884   /Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/snapshots
```

### Problem description

Adding a second tag for the same image doubles the reported image size and reclaimable bytes even though no additional storage is created. `system df` should count unique image storage, not tags/references.

### Environment

```markdown
- OS: macOS 26.4.1 (25E253)
- Xcode: Xcode 26.4.1 (17E202)
- Container: container CLI 0.12.3 (build: release, commit: f989901)
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jglogan:
Closed by #1555.

PR #1555, opened 2026-05-14
Fix system df to count content blobs and deduplicate shared storage

### LINKED ISSUE #1526: [Bug]: `container system df` omits image content blobs from image size and reclaimable bytes
### I have done the following

### Steps to reproduce

Use a clean app root, pull a single-platform image, then compare `system df` with the image content store:

```console
$ container image pull --platform linux/arm64 --progress plain alpine:latest
[1/2] Fetching image [0s]
[1/2] Fetching image [1s]
[1/2] Fetching image 53% (3 of 4 blobs, 2.2/4.0 MB, 8 KB/s) [2s]
[2/2] Unpacking image [2s]
[2/2] Unpacking image for platform linux/arm64/v8 [2s]
[2/2] Unpacking image for platform linux/arm64/v8 100% (514 entries, 8.3 MB) [2s]

$ container system df --format json
{
  "containers" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  },
  "images" : {
    "active" : 0,
    "reclaimable" : 78729216,
    "sizeInBytes" : 78729216,
    "total" : 1
  },
  "volumes" : {
    "active" : 0,
    "reclaimable" : 0,
    "sizeInBytes" : 0,
    "total" : 0
  }
}

$ du -sk "/Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/content/blobs/sha256"
4124    /Users/ras/Downloads/test_container/audit_artifacts/container-issue-repros-20260511-101939/app-root/content/blobs/sha256
```

### Problem description

`system df` reports image size and reclaimable bytes without counting image content blobs. Those blobs are image-managed storage and are reclaimed during image cleanup. `system df` should include blob storage in image size/reclaimable bytes, or explicitly label the current value as unpacked snapshot storage only.

### Environment

```markdown
- OS: macOS 26.4.1 (25E253)
- Xcode: Xcode 26.4.1 (17E202)
- Container: container CLI 0.12.3 (build: release, commit: f989901)
```

### Code of Conduct

realrajaryan:
- Closes #1526 and #1527.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This PR fixes `system df` to report actual on-disk allocated bytes (content blobs + snapshots) instead of summing per-image snapshot sizes. Orphaned blobs are now included as reclaimable, and storage shared across tags is no longer double counted. Also consolidates three identical `calculateDirectorySize` implementations into a shared `FileManager.allocatedSize(of:)` extension.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
+++ b/Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
@@ -18,6 +18,10 @@ import Foundation
 
 extension FileManager {
     /// Total bytes allocated on disk for all files in a directory (recursive).
+    ///
+    /// Caveats: hidden files are skipped, symlinks to directories are not followed, but
+    /// symlinks-to-files and hard links each contribute their target's full allocation
+    /// so shared inodes are counted multiple times.
     public func allocatedSize(of directory: URL) -> UInt64 {
         guard
             let enumerator = self.enumerator(
--- a/Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
+++ b/Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
@@ -1,0 +1,43 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension FileManager {
+    /// Total bytes allocated on disk for all files in a directory (recursive).
+    public func allocatedSize(of directory: URL) -> UInt64 {
+        guard
+            let enumerator = self.enumerator(
+                at: directory,
+                includingPropertiesForKeys: [.totalFileAllocatedSizeKey],
+                options: [.skipsHiddenFiles]
+            )
+        else {
+            return 0
+        }
+
+        var size: UInt64 = 0
+        for case let fileURL as URL in enumerator {
+            guard let resourceValues = try? fileURL.resourceValues(forKeys: [.totalFileAllocatedSizeKey]),
+                let fileSize = resourceValues.totalFileAllocatedSize
+            else {
+                continue
+            }
+            size += UInt64(fileSize)
+        }
+        return size
+    }
+}
--- a/Sources/Plugins/CoreImages/ImagesHelper.swift
+++ b/Sources/Plugins/CoreImages/ImagesHelper.swift
@@ -95,10 +95,11 @@ extension ImagesHelper {
             // TODO: remove as part of ImageStore URL removal PR
             let rootURL = URL(fileURLWithPath: root.string)
             let contentStore = RemoteContentStoreClient()
+            let contentBlobsPath = root.appendingPathComponent("content/blobs/sha256")
             let imageStore = try ImageStore(path: rootURL, contentStore: contentStore)
             let unpackStrategy = SnapshotStore.defaultUnpackStrategy(initImage: containerSystemConfig.vminit.image)
             let snapshotStore = try SnapshotStore(path: rootURL, unpackStrategy: unpackStrategy, log: log)
-            let service = try ImagesService(contentStore: contentStore, imageStore: imageStore, snapshotStore: snapshotStore, log: log)
+            let service = try ImagesService(contentStore: contentStore, contentBlobsPath: contentBlobsPath, imageStore: imageStore, snapshotStore: snapshotStore, log: log)
             let harness = ImagesServiceHarness(service: service, log: log)
 
             routes[ImagesServiceXPCRoute.imagePull.rawValue] = XPCServer.route(harness.pull)
--- a/Sources/Plugins/CoreImages/ImagesHelper.swift
+++ b/Sources/Plugins/CoreImages/ImagesHelper.swift
@@ -100,7 +100,6 @@ extension ImagesHelper {
             let snapshotStore = try SnapshotStore(path: rootURL, unpackStrategy: unpackStrategy, log: log)
             let service = try ImagesService(
                 contentStore: contentStore,
-                contentStoreTotalSize: { try await contentStore.totalAllocatedSize() },
                 imageStore: imageStore,
                 snapshotStore: snapshotStore,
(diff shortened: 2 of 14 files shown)


Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
```diff
@@ -0,0 +1,43 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension FileManager {
+    /// Total bytes allocated on disk for all files in a directory (recursive).
```
jglogan, line 20: The comments should include more details on what is and isn't counted. As far as I can tell:

- Symlinks to files outside `directory` will include those files' sizes in the total.
- Symlinks to files inside `directory` will include those files' sizes in the total, resulting in inaccuracy due to multiple counting.
- The same issues apply for hard links.
- Symlinks to directories aren't followed (which is good).

Do we need to address the double counting issues, or will the directories we enumerate never have these sorts of things?

Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
```diff
@@ -0,0 +1,43 @@
...
+    public func allocatedSize(of directory: URL) -> UInt64 {
+        guard
+            let enumerator = self.enumerator(
```
jglogan, line 27: I added `FileDescriptorOps.enumerate()` to ContainerizationOS for TOCTOU-safe traversal of directory trees; it might make sense to use here.

Sources/Plugins/CoreImages/ImagesHelper.swift
```diff
@@ -92,10 +92,11 @@ extension ImagesHelper {
 
         private func initializeImagesService(root: URL, containerSystemConfig: ContainerSystemConfig, log: Logger, routes: inout [String: XPCServer.RouteHandler]) throws {
             let contentStore = RemoteContentStoreClient()
+            let contentBlobsPath = root.appendingPathComponent("content/blobs/sha256")
             let imageStore = try ImageStore(path: root, contentStore: contentStore)
             let unpackStrategy = SnapshotStore.defaultUnpackStrategy(initImage: containerSystemConfig.vminit.image)
             let snapshotStore = try SnapshotStore(path: root, unpackStrategy: unpackStrategy, log: log)
-            let service = try ImagesService(contentStore: contentStore, imageStore: imageStore, snapshotStore: snapshotStore, log: log)
+            let service = try ImagesService(contentStore: contentStore, contentBlobsPath: contentBlobsPath, imageStore: imageStore, snapshotStore: snapshotStore, log: log)
```
jglogan: @adityaramani should the ImagesService know about the contentBlobsPath? That feels like an encapsulation break.

It's only used for getting the total size of the content store - could we just add a content client call that returns that value?

Sources/ContainerResource/Common/FileManager+AllocatedSize.swift
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+extension FileManager {
+    /// Total bytes allocated on disk for all files in a directory (recursive).
+    public func allocatedSize(of directory: URL) -> UInt64 {
```
jglogan, line 25: We shouldn't write any more code that uses URL for fs paths.

jglogan: Also noting that this would be a good candidate for extracting to a FilePathOps utility type in ContainerizationOS: https://github.com/apple/containerization/issues/744.

jglogan (changes requested):
Think everything looks functionally fine but we shouldn't be accessing the content store directly. The current approach would be broken for any other content store implementation.

Sources/Plugins/CoreImages/ImagesHelper.swift
adityaramani: Yes +1!

Between container and containerization there are two types are implement the `ContentStore` protocol

1) The `LocalContentStore` from Containerization
2) The `RemoteContentStoreClient` from this repo

They exist because there should be only one entity that is actually touching the files on disk - if we had multiple instances of `LocalContentStore` going about they could trample over each other and mess with the state of the content store.

The `RemoteContentStoreClient` type talk to the `ContentService` (also defined in this file) which houses the one singular instance of `LocalContentStore`

When you ask the `RemoteContentStoreClient` to ingest something into the content store it proxies that request through to the `LocalContentStore` which ensures that all operations happen atomically.


With this change - it slightly breaks that abstraction and makes ImageService directly check a path on disk where the files the `LocalContentStore` store may be.

To solve the "get me the total size of the content store" problem - we can probably do two things.
1) Add that API to the `ContentStore` protocol and implement it in both the types
2) Add a list api to the `ContentStore` which returns a list of `Content` https://github.com/apple/containerization/blob/main/Sources/ContainerizationOCI/Content/Content.swift#L23 objects  (this is probably a little more involved and needs some deliberation)

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -290,49 +297,41 @@ public actor ImagesService {
...
-
-        for descriptor in index.manifests {
-            // Skip attestation manifests
-            if let refType = descriptor.annotations?["vnd.docker.reference.type"],
-                refType == "attestation-manifest"
-            {
-                continue
-            }
+        let snapshotDiskSize = await self.snapshotStore.totalAllocatedSize()
+        let contentDiskSize = try await self.contentStoreTotalSize()
```
jglogan: What prevents us from invoking `self.contentStore.totalAllocatedSize()` here?

Sources/Services/ContainerImagesService/Server/ImagesService.swift
```diff
@@ -290,49 +294,41 @@ public actor ImagesService {
...
-            if isActive {
-                activeCount += 1
-            } else {
-                reclaimableSize += imageSize
+            guard activeReferences.contains(image.reference) else { continue }
+            activeCount += 1
+            let imageDigest = image.digest.trimmingDigestPrefix
+            guard processedDigests.insert(imageDigest).inserted else { continue }
+
+            for digest in try await image.referencedDigests() where activeContentSizes[digest] == nil {
```
jglogan, line 308: It might be possible to see inaccuracy in `activeSnapshotSizes` if other operations are in progress, since the state may change while this call is at any of the suspension points in this loop. It's probably not anything to lose sleep over right now, just wanted to note the issue.

jglogan (approved):

PR #1562, opened 2026-05-15
Expose container exit code and exit time via inspect

### LINKED ISSUE #1501: [Request]: Surface lastExitCode on ContainerSnapshot
> _Suggested triage: type=`Feature`, label=`enhancement`. (External contributor — cannot self-apply.)_

#### Summary

Add an optional `lastExitCode: Int32?` field to `ContainerSnapshot`, populated from the existing `ExitMonitor` callback when a container transitions to `.stopped`.

A draft PR with the implementation is open at #1503.

#### Motivation

External orchestrators that drive the API server (the canonical use case is a Compose-spec orchestrator implementing `depends_on: condition: service_completed_successfully`) need to distinguish between *clean* exit (exit code 0) and *failed* exit (non-zero exit code) of a one-shot container.

Today `ContainerSnapshot` exposes `RuntimeStatus.stopped`, but the underlying exit code is not surfaced. The orchestrator can observe \"the container stopped\" but not \"the container stopped successfully\". This forces consumers to treat `.stopped` as a success, silently misinterpreting non-zero exits.

The same data already exists in the daemon's path: `ContainersService.handleContainerExit(id:code:context:)` receives `code: ExitStatus?` from the existing `ExitMonitor` callback wiring. The proposed change just stamps that value onto the snapshot.

#### Proposed change

```swift
// Sources/ContainerResource/Container/ContainerSnapshot.swift
public var lastExitCode: Int32?

public init(
    configuration: ContainerConfiguration,
    status: RuntimeStatus,
    networks: [Attachment],
    startedDate: Date? = nil,
    lastExitCode: Int32? = nil
) { ... }
```

```swift
// Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
// In handleContainerExit, terminal-state block:
state.snapshot.status = .stopped
state.snapshot.lastExitCode = code?.exitCode   // NEW
state.snapshot.networks = []
```

#### Wire compatibility

`ContainerSnapshot` is marshaled as `Codable` JSON over XPC. Adding an optional field is forward-compatible:

- Older clients reading from a newer server: ignore the new key.
- Newer clients reading from an older server: decode `lastExitCode` as `nil` (the documented \"never exited or exit not captured\" case).

#### Scope

Intentionally minimal — *in-memory only*. The exit code lives in the in-memory `ContainerState` snapshot for the duration of API server uptime. A daemon restart resets all snapshots to `.stopped` without exit codes (existing behavior). Bundle persistence for survival across daemon restarts (e.g. an `exit_status.json`) is a deliberate out-of-scope follow-up.

#### Use case context

This is a straightforward upstream-ing of a small change we have been carrying in a downstream fork to power Compose-style orchestration on top of `apple/container` ([container-compose](https://github.com/full-chaos/container-compose)). We would much prefer to consume the field directly from `apple/container`'s `ContainerSnapshot` rather than maintain a fork. We have existing internal coverage of this change against a real `apple/container` daemon; happy to share notes.

### Code of Conduct

I agree to follow this project's Code of Conduct.

bilby91:
## Summary

`ContainerSnapshot` currently drops the exit status of the init process, so `container inspect` and `container list` cannot tell callers whether a stopped container exited cleanly or with an error. The daemon already has the value (it observes `ExitStatus` in `handleContainerExit`); this PR just plumbs it through to the snapshot type and the CLI projection.

Fixes #1501.

## Changes

- `ContainerSnapshot` gains two optional fields:
  - `exitCode: Int32?` — exit code of the init process
  - `exitedDate: Date?` — when the container transitioned to stopped
- `ContainersService.handleContainerExit` populates both fields before calling `setContainerState`.
- `PrintableContainer` (the CLI's `Codable` projection used by `inspect` / `list`) carries the new fields through.

Both fields default to `nil`, so:
- Snapshots produced by an older daemon decode unchanged (covered by a back-compat test).
- A newer CLI talking to an unmodified daemon sees `nil` and `JSONEncoder` strips the keys from output — no breaking change for downstream consumers parsing today's JSON.

## Motivation

We're building [crunchloop/devcontainer](https://github.com/crunchloop/devcontainer) — an open-source Go runtime for [Dev Containers](https://containers.dev/) — and are adding `apple/container` as a first-class backend alongside Docker. Part of that work is a native implementation of the compose subset Dev Containers relies on for multi-service workspaces.

Two concrete things break today without exit-code visibility:

1. **`depends_on: condition: service_completed_successfully`** — a one-shot service (DB seeder, migration, fixture loader) has to finish *and* exit cleanly before dependents start. With no exit code in `inspect`, we can only observe "stopped" and have to guess at success.
2. **CI / automation flows** driving one-shot containers (lint, test, build steps) need to react to the exit code without scraping container logs.

This is a small plumbing change — the daemon already captures `ExitStatus`, it just isn't propagated. It's also the kind of low-level primitive that helps anyone building an orchestrator or higher-level tooling on top of `apple/container`, not only us.

## Test plan

- [x] `swift build` (full project, 258 modules) — clean
- [x] `swift test --filter ContainerSnapshotTests` — 3/3 pass
  - Legacy snapshot JSON decodes with `nil` exit fields (back-compat)
  - Snapshot round-trips through JSON with the new fields populated
  - Non-zero exit codes preserved across `Int32` range (incl. negative, `Int32.min`, `Int32.max`)
- [x] New CLI binary against an **unmodified** running daemon (0.12.3): runs cleanly, new fields omitted from output (Swift `JSONEncoder` strips `nil` optionals)
- [x] `make fmt` — no changes

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerList.swift
+++ b/Sources/ContainerCommands/Container/ContainerList.swift
@@ -81,11 +81,15 @@ struct PrintableContainer: Codable, Sendable {
     let configuration: ContainerConfiguration
     let networks: [Attachment]
     let startedDate: Date?
+    let exitCode: Int32?
+    let exitedDate: Date?
 
     init(_ container: ContainerSnapshot) {
         self.status = container.status
         self.configuration = container.configuration
         self.networks = container.networks
         self.startedDate = container.startedDate
+        self.exitCode = container.exitCode
+        self.exitedDate = container.exitedDate
     }
 }
--- a/Sources/ContainerResource/Container/ContainerSnapshot.swift
+++ b/Sources/ContainerResource/Container/ContainerSnapshot.swift
@@ -39,16 +39,28 @@ public struct ContainerSnapshot: Codable, Sendable {
     public var networks: [Attachment]
     /// When the container was started.
     public var startedDate: Date?
+    /// The exit code of the container's init process. Nil while the
+    /// container is running or has never been started. Populated when
+    /// the container transitions to `.stopped`.
+    public var exitCode: Int32?
+    /// When the container's init process exited. Nil while the
+    /// container is running or has never been started. Populated
+    /// alongside `exitCode` on transition to `.stopped`.
+    public var exitedDate: Date?
 
     public init(
         configuration: ContainerConfiguration,
         status: RuntimeStatus,
         networks: [Attachment],
-        startedDate: Date? = nil
+        startedDate: Date? = nil,
+        exitCode: Int32? = nil,
+        exitedDate: Date? = nil
     ) {
         self.configuration = configuration
         self.status = status
         self.networks = networks
         self.startedDate = startedDate
+        self.exitCode = exitCode
+        self.exitedDate = exitedDate
     }
 }
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -961,6 +961,10 @@ public actor ContainersService {
 
         state.snapshot.status = .stopped
         state.snapshot.networks = []
+        if let code {
+            state.snapshot.exitCode = code.exitCode
+            state.snapshot.exitedDate = code.exitedAt
+        }
         state.client = nil
         await self.setContainerState(id, state, context: context)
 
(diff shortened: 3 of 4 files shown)


jglogan:
Hi @bilby91, could you have a look at this and #1503 which was submitted shortly after #1501 was filed?

PR #1570, opened 2026-05-17
[images]: fix image load error to write to stderr and use FilePath

### LINKED ISSUE #1535: [Bug]: `container image load -i <missing-file>` writes the error to stdout
### I have done the following

### Steps to reproduce

Run `image load` with a missing input file and redirect stdout/stderr:

```console
$ container image load -i /path/that/does/not/exist >stdout.txt 2>stderr.txt
$ echo $?
1

$ cat stdout.txt
File does not exist /path/that/does/not/exist

$ cat stderr.txt
```

### Problem description

Related to https://github.com/apple/container/issues/642. The command fails but writes the error to stdout and leaves stderr empty. User-facing errors should go to stderr so scripts can separate command output from failures.

### Environment

```markdown
- OS: macOS 26.4.1 (25E253)
- Xcode: Xcode 26.4.1 (17E202)
- Container: container CLI 0.12.3 (build: release, commit: f989901)
```

### Code of Conduct

### LINKED ISSUE #1573: [Request]: TECH DEBT: Convert image load/save to use FilePath

Replace `string` with `FilePath` for the input file used for both commands.

### Code of Conduct

saehejkang:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #1535
Closes #1573

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Image/ImageLoad.swift
+++ b/Sources/ContainerCommands/Image/ImageLoad.swift
@@ -19,6 +19,7 @@ import ContainerAPIClient
 import Containerization
 import ContainerizationError
 import Foundation
+import SystemPackage
 import TerminalProgress
 
 extension Application {
@@ -32,9 +33,13 @@ extension Application {
         @Option(
             name: .shortAndLong, help: "Path to the image tar archive", completion: .file(),
             transform: { str in
-                URL(fileURLWithPath: str, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false)
+                let path = FilePath(str)
+                guard path.isRelative else { return path.lexicallyNormalized() }
+                return FilePath(FileManager.default.currentDirectoryPath)
+                    .pushing(path)
+                    .lexicallyNormalized()
             })
-        var input: String?
+        var input: FilePath?
 
         @Flag(name: .shortAndLong, help: "Load images even if the archive contains invalid files")
         public var force = false
@@ -49,7 +54,14 @@ extension Application {
             }
 
             // Read from stdin; otherwise read from the input file
-            guard let input else {
+            let resolvedPath: FilePath
+            if let input {
+                guard FileManager.default.fileExists(atPath: input.string) else {
+                    log.error("file does not exist", metadata: ["path": "\(input)"])
+                    Application.exit(withError: ArgumentParser.ExitCode(1))
+                }
+                resolvedPath = input
+            } else {
                 guard FileManager.default.createFile(atPath: tempFile.path(), contents: nil) else {
                     throw ContainerizationError(.internalError, message: "unable to create temporary file")
                 }
@@ -65,12 +77,7 @@ extension Application {
                     fileHandle.write(chunk)
                 }
                 try fileHandle.close()
-                return
-            }
-
-            guard FileManager.default.fileExists(atPath: input) else {
-                log.error("file does not exist", metadata: ["path": "\(input)"])
-                Application.exit(withError: ArgumentParser.ExitCode(1))
+                resolvedPath = FilePath(tempFile.path())
             }
 
             let progressConfig = try ProgressConfig(
@@ -86,7 +93,7 @@ extension Application {
 
             progress.set(description: "Loading tar archive")
             let result = try await ClientImage.load(
-                from: input,
+                from: resolvedPath.string,
                 force: force)
             if !result.rejectedMembers.isEmpty {
                 log.warning("archive contains invalid members", metadata: ["paths": "\(result.rejectedMembers)"])
--- a/Sources/ContainerCommands/Image/ImageLoad.swift
+++ b/Sources/ContainerCommands/Image/ImageLoad.swift
@@ -49,7 +49,7 @@ extension Application {
             }
 
             // Read from stdin; otherwise read from the input file
-            if input == nil {
+            guard let input else {
                 guard FileManager.default.createFile(atPath: tempFile.path(), contents: nil) else {
                     throw ContainerizationError(.internalError, message: "unable to create temporary file")
                 }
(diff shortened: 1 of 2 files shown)


Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -67,7 +67,10 @@ extension Application {
                 try fileHandle.close()
             } else {
                 guard FileManager.default.fileExists(atPath: input!) else {
-                    print("File does not exist \(input!)")
+                    let errorMessage = "File does not exist \(input!)\n"
+                    if let data = errorMessage.data(using: .utf8) {
+                        FileHandle.standardError.write(data)
```
jglogan: @saehejkang You can just use `log.error()` (with structured logging) for this.

See https://github.com/apple/container/blob/main/Sources/ContainerCommands/Image/ImageDelete.swift#L76 for an example (the message on L85 of that file should use `log.info()` actually).

jglogan (changes requested):

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: Forgot about the structured logging updates that were made!

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -67,7 +67,7 @@ extension Application {
                 try fileHandle.close()
             } else {
                 guard FileManager.default.fileExists(atPath: input!) else {
-                    print("File does not exist \(input!)")
+                    log.error("file does not exist", metadata: ["path": "\(input!)"])
```
jglogan: Dang, I should have noticed the force try.

OK, one more change, sorry: can you change L52 to be `guard let input else {` with a return at the end of the block, and then we won't need the else at L68 as well as the force try.

jglogan (approved):

jglogan (changes requested):
See note about force try on L70

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -49,7 +49,7 @@ extension Application {
             }
 
             // Read from stdin; otherwise read from the input file
-            if input == nil {
+            guard let input else {
```
jglogan: @saehejkang OK, I remember now why the original work here settled for force-try.

For this to pass integration tests we have to modify this slightly. We can't use guard unless we duplicate the image load code here using a temp file. Since we're in the process of changing from URL and string manipulation to FilePath, let's creep the scope a bit.

Here, transform the input string to a SystemPackage.FilePath?, converting a relative FilePath to an absolute one:

```
        @Option(
            name: .shortAndLong, help: "Path to the image tar archive", completion: .file(),
            transform: { str in
                URL(fileURLWithPath: str, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false)
            })
        var input: String?
```


```suggestion
            let resolvedPath: FilePath
            if let input {
                guard FileManager.default.fileExists(atPath: input!) else {
                    log.error("file does not exist", metadata: ["path": "\(input)"])
                    Application.exit(withError: ArgumentParser.ExitCode(1))
                    resolvedPath = input
                }
            } else {
                // same as L53-67 but resolvedPath gets the temp file
            }
            
            // the remaining image load code, using tempPath for the input   
```

jglogan (changes requested):
See suggestion to cut this over to FilePath. We can do ImageSave in a follow-up issue

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: It all makes a bit sense now, why we went with the first way. This was one of the first issues I ever worked on 😄 .

Sources/ContainerCommands/Image/ImageLoad.swift
```diff
@@ -32,9 +33,13 @@ extension Application {
         @Option(
             name: .shortAndLong, help: "Path to the image tar archive", completion: .file(),
             transform: { str in
-                URL(fileURLWithPath: str, relativeTo: .currentDirectory()).absoluteURL.path(percentEncoded: false)
+                let path = FilePath(str)
```
saehejkang, line 36: I noticed that this logic is pretty much 1:1 from the `Filesystem.absoluteFilePath` function [here](https://github.com/apple/container/blob/bd1916f9b314b9b9e335362bbcfc8ee11f7d860b/Sources/ContainerResource/Container/Filesystem.swift#L189). Thoughts on making it public and useable elswhere?

jglogan: I'm thinking the same. I'm working on doing FilePath conversion for ContainerBuild and there's a pallet load of FilePath extensions that it needs. No PR for it yet but it has an `.absolutePath` implementation.

I also created some path resolution extensions in ContainerPlugin for the FilePath changes for the data/install/log root paths: https://github.com/apple/container/pull/1558.

If you feel like proposing it, what do you think about creating an issue to add FilePath extensions down in containerization, under ContainerizationOS? I think that would make it easier for us to make much-needed changes through the whole stack. We could use the issue discussion to build up the list of what we think we'd need (I can summarize what ContainerBuild wants there.)

Sources/ContainerCommands/Image/ImageLoad.swift
jglogan: LOL I forgot that it was you! Well it worked, this just gets rid of the awkwardness of the force-try (if we ever turned on linting we'd have to put linter exceptions on them).

Thanks for your patience on this one - it's a PR I should have reviewed by looking at the whole file rather than just the changes and we could have saved an iteration or two.

Sources/ContainerCommands/Image/ImageLoad.swift
saehejkang: https://github.com/apple/containerization/issues/744

jglogan (approved):

Issue #1534, opened 2026-05-11
[Bug]: Command reference and how-to docs are stale relative to 0.12.3


realrajaryan:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Compare the docs with the installed CLI help:

```console
$ container image pull --help
OVERVIEW: Pull an image

USAGE: container image pull [--scheme <scheme>] [--progress <type>] [--max-concurrent-downloads <max-concurrent-downloads>] [--arch <arch>] [--os <os>] [--platform <platform>] [--debug] <reference>

ARGUMENTS:
  <reference>

OPTIONS:
  --scheme <scheme>       Scheme to use when connecting to the container
                          registry. One of (http, https, auto) (default: auto)
  --progress <type>       Progress type (format: auto|none|ansi|plain|color)
                          (default: auto)
  --max-concurrent-downloads <max-concurrent-downloads>
                          Maximum number of concurrent downloads (default: 3)
                          (default: 3)
  -a, --arch <arch>       Limit the pull to the specified architecture
  --os <os>               Limit the pull to the specified OS
  --platform <platform>   Limit the pull to the specified platform (format:
                          os/arch[/variant], takes precedence over --os and
                          --arch) [environment: CONTAINER_DEFAULT_PLATFORM]
  --debug                 Enable debug output [environment: CONTAINER_DEBUG]
  --version               Show the version.
  -h, --help              Show help information.

$ container system version --help
OVERVIEW: Show version information

USAGE: container system version [--format <format>] [--debug]

OPTIONS:
  --format <format>       Format of the output (values: json, table, yaml;
                          default: table)
  --debug                 Enable debug output [environment: CONTAINER_DEBUG]
  --version               Show the version.
  -h, --help              Show help information.

$ container system dns list --help
OVERVIEW: List local DNS domains

USAGE: container system dns list [--format <format>] [--quiet] [--debug]

OPTIONS:
  --format <format>       Format of the output (values: json, table, yaml;
                          default: table)
  -q, --quiet             Only output the domain
  --debug                 Enable debug output [environment: CONTAINER_DEBUG]
  --version               Show the version.
  -h, --help              Show help information.

$ container system dns create --help
OVERVIEW: Create a local DNS domain for containers (must run as an
administrator)

USAGE: container system dns create [--debug] [--localhost <localhost>] <domain-name>

ARGUMENTS:
  <domain-name>           The local domain name

[...]

PR #1571, opened 2026-05-17
[command-reference]: add diff script, CI check, and update reference docs

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Related to #1534 

Thought it would be worthwhile to add a sort of mandatory check to the CI to fail, if the `command-reference` docs are not updated appropriately, whenever changes are made to the CLI. When the lint command is ran, there is an output that shows the diff between the current CLI build and what is written in the docs. Hopefully this will help keep the docs up to date and reduce tech debt over time.

NOTE: The `scripts/generate-command-reference.py` file was generated with AI but reviewed by me.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/.github/workflows/common.yml
+++ b/.github/workflows/common.yml
@@ -56,6 +56,11 @@ jobs:
         env:
           DEVELOPER_DIR: "/Applications/Xcode-latest.app/Contents/Developer"
 
+      - name: Check command reference usage drift
+        run: make lint-command-reference
+        env:
+          DEVELOPER_DIR: "/Applications/Xcode-latest.app/Contents/Developer"
+
       - name: Set build configuration
         env:
           RELEASE: ${{ inputs.release }}
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@ test_results/
 *.ext4
 *.pkg
 *.swp
+__pycache__/
+*.pyc
 
 # Coverage artifacts
 *.profraw
--- a/Makefile
+++ b/Makefile
@@ -337,6 +337,11 @@ docs:
 	@rm -rf _site
 	@scripts/make-docs.sh _site container
 
+.PHONY: lint-command-reference
+lint-command-reference: cli
+	@echo Linting command reference usage against container CLI...
+	@python3 scripts/generate-command-reference.py --cli ./bin/container --lint-usage-against docs/command-reference.md
+
 .PHONY: cleancontent
 cleancontent:
 	@bin/container system stop || true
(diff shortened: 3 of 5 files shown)


katiewasnothere:
@saehejkang I wonder if we can use ArgumentParser's [experimental help dump](https://swiftpackageindex.com/apple/swift-argument-parser/1.7.0/documentation/argumentparser/experimentalfeatures) feature here instead of needing to manually parse the help with python. What do you think?

jglogan:
I think I'm more inclined toward a two-layer approach for this:

- (As developers and reviewers) we all need to be vigilant about CLI-related changes and ensure that docs get updated.
- (As maintainers) we have a recurring task to burn some API tokens on correcting the drift (and an agent can use the experimental dump help metadata to establish a baseline).

Having a script do things can get the job done for the command reference but isn't as adaptable as using a natural language based approach to solving a natural language related problem. Neither approach will be perfect but I feel the result will be better if we just lean into LLMs for this.

saehejkang:
@jglogan @katiewasnothere

> I wonder if we can use ArgumentParser's [experimental help dump](https://swiftpackageindex.com/apple/swift-argument-parser/1.7.0/documentation/argumentparser/experimentalfeatures) feature here instead of needing to manually parse the help with python. What do you think?

LOL, I completely missed this feature. I can see it being something that simplifies a lot of the script and then we don't need to add any python stuff.

> Having a script do things can get the job done for the command reference but isn't as adaptable as using a natural language based approach to solving a natural language related problem. 

FWIW, I did use an LLM to actually go through and make the necessary updates to the command reference docs. The script made it a lot quicker for the LLM to see the drift and make the updates. Saves a lot more tokens and we only have to prompt it to:

```run the diff script and make updates to the docs```

Furthermore, if there is ever a time when a developer wants to see the drift (and has no AI), they can read the output of the script and make updates promptly. Also, the CI, would now have a way to check for the drift automatically (and fail if updates were missed), because I assume we can't use an LLM in the pipeline (yet)?

jglogan:
@saehejkang Let's look at this more closely after the dust settles on this release. My concern with integrating anything that's complicated into the build that isn't directly about building or testing the code, it becomes a showstopper issue should it stop working properly.

saehejkang:
> Let's look at this more closely after the dust settles on this release.

Sounds good to me. On second thought, there could even be a bot or something that opens automatic PRs to update the command docs. But again, a discussion we can table for the near future.

saehejkang:
Closing for now, but will hopefully revisit this in the near future.

PR #1592, opened 2026-05-23
Add container log retrieval options

### LINKED ISSUE #1591: [Request]: Add container log retrieval options

External orchestrators that drive `container` through the API server need to implement log UX that matches common container tooling, including retrieving only recent log lines and optionally requesting timestamped output.

Today `ContainerClient.logs(id:)` returns the raw container log and boot log file handles unconditionally. Consumers that need `--since` semantics have to replay the entire log stream and filter client-side, even though the API server is closest to the files and can apply the filter before returning handles.

Requested enhancement:

- Add a source-compatible `ContainerClient.logs(id:options:)` overload.
- Add a `ContainerLogOptions` API surface with:
  - `since: Date?` to filter ISO-8601-prefixed log lines older than the requested timestamp.
  - `timestamps: Bool` to reserve/forward timestamp preference through the API boundary.
- Preserve existing `logs(id:)` behavior as the default zero-option path.

This would unblock external compose-style orchestrators from implementing `logs --since <timestamp>` and `logs --timestamps` against the `container` API without duplicating log filtering logic downstream.

### Code of Conduct

chrisgeo:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Add a source-compatible log options API for `ContainerClient.logs` so API clients can request recent log output and forward timestamp preferences through the existing XPC log route.

External orchestrators that implement compose-style log commands need `--since` and `--timestamps` behavior without replaying and filtering the full log stream downstream. This change introduces `ContainerLogOptions`, preserves the existing `logs(id:)` default behavior, and plumbs the options through the client, XPC harness, and container service.

The `since` option filters ISO-8601-prefixed lines before file handles are returned. Lines without a parseable timestamp are preserved. The `timestamps` option is forwarded through the API boundary as a hint; decorating raw log lines that lack timestamps remains future work.

Resolves #1591.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Local verification:
- `swift build`
- `swift test --filter ContainerLogsTests`
- `swift test --filter ContainerLogOptionsTests`
- LSP diagnostics clean on all changed Swift files
- External Swift package driver imported `ContainerResource`, encoded/decoded `ContainerLogOptions`, and verified `.default` values

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -203,9 +203,12 @@ let package = Package(
             name: "ContainerAPIServiceTests",
             dependencies: [
                 .product(name: "Containerization", package: "containerization"),
+                "ContainerAPIClient",
+                "ContainerAPIService",
                 "ContainerResource",
                 "ContainerRuntimeLinuxClient",
                 "ContainerRuntimeClient",
+                "ContainerXPC",
             ]
         ),
         .target(
--- a/Sources/ContainerResource/Container/ContainerLogOptions.swift
+++ b/Sources/ContainerResource/Container/ContainerLogOptions.swift
@@ -1,0 +1,42 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// Options that refine how `ContainerClient.logs(id:options:)` returns
+/// container log file handles.
+///
+/// Both fields are optional / additive; ``default`` is the zero-value
+/// equivalent to the original `logs(id:)` behavior.
+public struct ContainerLogOptions: Sendable, Codable {
+    /// If non-nil, log lines whose ISO-8601 timestamp prefix is older than
+    /// this date are filtered out before the file handle is returned to the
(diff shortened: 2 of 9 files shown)


stephenlclarke:
Thanks again for the base API work here. I opened two draft stacked follow-ups so the remaining log-filtering pieces can be reviewed without turning this into one large Compose-shaped change:

- #1764 adds `tail` / `until` retrieval filters on top of this `ContainerLogOptions` direction and wires the `container logs` CLI/docs/tests.
- #1765 stacks on #1764 and moves Docker-compatible timestamp parsing into a reusable `ContainerResource` parser.

Both are deliberately draft. Because GitHub can only target `apple/container:main`, their diffs include this PR's branch; the intended review deltas are called out in each PR body.

If you or the maintainers would rather fold the retrieval-only option shape from #1764 into this PR before it lands, I’m happy to reshape the stack around that. My goal is still to keep this aligned with your #1592 direction and keep Compose-specific policy out of `apple/container`.

stephenlclarke:
@chrisgeo 

Hi Chris, thanks again for pushing #1592 forward.

I’ve been working through the log API shape from the perspective of Compose parity and did a stricter local pass against this branch. I do not want to take over your PR, but I found a few edge cases where the current raw-log timestamp filtering can look successful even when the log data cannot actually be proven to satisfy the filter.

The main local changes I tried are:

- fail closed when `since` is applied to non-UTF8 logs
- fail closed when a raw log record has no parseable timestamp prefix
- treat blank application log records as unfilterable under timestamp filters
- preserve the underlying `ContainerizationError` instead of wrapping it as a generic log-open failure
- make temporary filtered-handle creation report write/seek/unlink failures instead of silently returning unfiltered output

My thinking is that this keeps #1592’s API direction intact while making the current raw-log boundary explicit. Full Docker/Compose parity still looks like it belongs in later structured log record work, but this avoids giving callers a false sense that raw stdio logs were accurately filtered.

I have the changes locally as a small patch series and I’m happy to either share the branch, turn it into suggested changes, or open a small stacked follow-up if that would be easier. I’d appreciate your preference on what fits best with your intended direction for #1592.

Issue #1593, opened 2026-05-23
[Request]: Migrate PublishSocket.containerPath and hostPath to FilePath


chrisgeo:
### Summary

Migrate `PublishSocket.containerPath` and `PublishSocket.hostPath` from `Foundation.URL` to `SystemPackage.FilePath`, consistent with the ongoing project-wide effort to replace `URL`/`String` paths with strongly-typed `FilePath` / `SystemPath` at API boundaries.

### Motivation

`PublishSocket` is the only remaining `URL`-typed path in `ContainerResource` after the recent migrations:

- #1493 NetworksService, VolumesService, EntityStore → FilePath
- #1518 PacketFilter → SystemPath
- #1522 DirectoryWatcher → SystemPath
- #1523 `ContainerResource.Filesystem` → SystemPath
- #1570, #1574, #1580 image load/save/copy → FilePath
- #1480 HostDNSResolver → SystemPath

`URL` is the wrong type for filesystem paths — it implies a navigable resource, allows non-file schemes, and forces consumers through percent-encoding/decoding when interoperating with POSIX APIs. `FilePath` is purpose-built, `Sendable`, and already used by callers on both sides of this struct (e.g. `Parser`, `RuntimeService` after the recent runtime reorg).

A `Parser`-side companion change (CHAOS-1463 in our tracker) is blocked on this migration because it needs to consume `FilePath` from `PublishSocket` directly without a `URL ↔ FilePath` roundtrip.

### Proposed change

- `PublishSocket.containerPath: URL → FilePath`
- `PublishSocket.hostPath: URL → FilePath`
- Provide a custom `Codable` implementation that preserves the existing JSON wire format byte-for-byte (encoding the `FilePath` as `URL.absoluteString`, i.e. `"file:///..."`), so persisted container bundles and any older readers continue to decode correctly. The decoder accepts both the legacy `file://` form and plain absolute paths, and rejects malformed / empty / relative inputs.
- Update the two call sites: `Parser` constructs `FilePath` directly; `RuntimeService` converts at the `UnixSocketConfiguration` (Containerization) boundary.

### Compatibility

- **On-disk container bundles**: encoder emits the same `"file:///..."` strings as before; existing bundles round-trip cleanly. Tests pin this byte-compatibility.
- **XPC**: same encoder output, plus decoder tolerates either form, so a new client/old service or new service/old client combination both decode correctly.
[...]

PR #1580, opened 2026-05-20
Update container copy to use FilePath

JaewonHur:
Replace `URL` with `FilePath` in `container copy` (#1557).
In addition, make `copyIn`/`copyOut` API to use `String` for path as we need to preserve the trailing slash to the `LinuxContainer.copy`---i.e., this trailing slash is used in `LinuxContainer.copy` to determine copy behavior.

PR #1577, opened 2026-05-20
Reorganize runtime targets for scalable plugin support

ajemory:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #1569

Separates generic runtime infrastructure from Linux-specific implementation into distinct SPM targets:

- ContainerRuntimeClient — generic runtime plugin client API
- ContainerRuntimeLinuxClient — Linux-specific types
- ContainerRuntimeLinuxServer — Linux runtime service + interface strategy implementations

Renames types to reflect the runtime/sandbox distinction:
- SandboxClient → RuntimeClient
- SandboxService → RuntimeService
- SandboxRoutes → RuntimeRoutes
- SandboxKeys → RuntimeKeys

Breaking for external consumers:
- ContainerSandboxService removed, replaced by ContainerRuntimeClient + ContainerRuntimeLinuxServer
- Public types renamed

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

PR #1493, opened 2026-04-30
Switch NetworksService, VolumesService, EntityStore to FilePath.

jglogan:
- Closes #1485.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
URL for filesystem paths is tech debt.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

PR #1594, opened 2026-05-23
Use FilePath for PublishSocket

chrisgeo:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context

Migrates `PublishSocket.containerPath` and `PublishSocket.hostPath` from `Foundation.URL` to `SystemPackage.FilePath`, consistent with the project-wide effort to replace `URL`/`String` paths with strongly-typed `FilePath` / `SystemPath` at API boundaries (#1480, #1493, #1518, #1522, #1523, #1570, #1574, #1580). `PublishSocket` was the last `URL`-typed path in `ContainerResource`.

`URL` is the wrong type for filesystem paths: it implies a navigable resource, permits non-file schemes, and forces consumers through percent-encoding when interoperating with POSIX APIs. `FilePath` is purpose-built, `Sendable`, and already used by both callers of this struct after the recent runtime reorg (#1577).

Tracking issue: #1593

### Wire-format compatibility

`PublishSocket` is serialized to disk as part of container bundles and over XPC. To keep this migration non-breaking on the wire:

- The encoder emits each path as `URL(filePath: path.string).absoluteString` — byte-identical to the prior `URL`-typed `JSONEncoder` output (e.g. `"file:///var/run/docker.sock"`). Existing bundles and any reader still decoding these fields as `URL` continue to work.
- The decoder accepts the legacy `file://` form, plain absolute paths (forward-compat), and resolves percent-encoding. Empty, relative, malformed, or non-`localhost`-host inputs throw `DecodingError.dataCorrupted` so corrupt state fails loudly rather than silently producing an invalid socket path.

### Source compatibility

Source-breaking for any external consumer constructing `PublishSocket` directly with `URL`. Acceptable given the pre-1.0 project status and the consistent direction of prior migrations.

## Testing
- [x] Tested locally (`swift test --filter ContainerResourceTests` — 50/50 passing)
- [x] Added/updated tests — `Tests/ContainerResourceTests/PublishSocketTests.swift`, 14 cases covering:
  - Encoder produces byte-identical output to the prior `URL` encoding (pinned via a sibling `LegacyPublishSocket` mirror)
  - Encoder percent-encodes spaces and special characters
  - Decoder accepts legacy `file://` form, percent-encoded paths, and `file://localhost/...`
  - Decoder accepts plain absolute paths (forward-compat)
  - Encode → decode round-trip preserves all fields including `permissions`
  - Decoder rejects empty strings, `"file:"`, `"file://"`, relative paths, and non-local-host file URLs
- [ ] Added/updated docs — public API change only; no doc surface impacted

Build verified clean: `swift build --target ContainerRuntimeLinuxServer`.


Sources/ContainerResource/Container/PublishSocket.swift
```diff
@@ -20,21 +20,100 @@ import SystemPackage
...
+    public var hostPath: FilePath
 
     /// File permissions for the socket on the host.
     public var permissions: FilePermissions?
 
     public init(
-        containerPath: URL,
-        hostPath: URL,
+        containerPath: FilePath,
+        hostPath: FilePath,
```
jglogan, line 48: What constraints/invariants exist on these paths today? If any do exist, should we enforce them in init() so that PublishSocket objects are correct by construction?

We should docc the parameters here and if there are constraints, describe them as well.

Sources/ContainerResource/Container/PublishSocket.swift
```diff
@@ -20,21 +20,100 @@ import SystemPackage
...
+    /// These fields were previously typed `URL`; `JSONEncoder` special-cases
+    /// `URL` to emit `absoluteString`. `FilePath`'s synthesized `Codable`
+    /// would instead emit a keyed container (`{"_storage": "..."}`), changing
+    /// the on-disk and XPC wire format. We therefore encode each path as the
+    /// equivalent `URL.absoluteString` so the byte form remains compatible
+    /// with persisted bundles and any readers (e.g. an older service binary)
+    /// that still decode these fields as `URL`.
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(Self.encodePath(containerPath), forKey: .containerPath)
```
jglogan: If we can get the change in before the end of the week or very early next week (1.0 release), we should do a breaking change on the XPC wire format and encode the absolute path but as the string representation of the path instead of as a URL.

The decoder should preserve compatibility (which it looks like you have with `decodePath`) for a few versions so that persisted config values from older releases continue to work.

We could in a separate PR add a persistent config migration at containers service startup in the API server.

Sources/ContainerResource/Container/PublishSocket.swift
```diff
@@ -20,21 +20,100 @@ import SystemPackage
 /// Represents a socket that should be published from container to host.
 public struct PublishSocket: Sendable, Codable {
```
jglogan, line 26: Add a comment here describing the change, something like:

```swift
/// - Deprecated: New for 1.0.0, path types changed from `URL` to `FilePath`.
/// - Note: Decoder handles `FilePath` and `URL` for persistent data compatibility;
///    this compatibility will be removed in a later release.
```

Sources/Services/ContainerAPIService/Client/Parser.swift
jglogan: If we move validation into PublishSocket this can go away.

jglogan: We can update the containerization dependency to 0.33.2 and use `FilePathOps.absolutePath(path:)`.

Sources/ContainerResource/Container/PublishSocket.swift
```diff
@@ -20,21 +20,100 @@ import SystemPackage
...
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.containerPath = try Self.decodePath(from: container, forKey: .containerPath)
+        self.hostPath = try Self.decodePath(from: container, forKey: .hostPath)
+        self.permissions = try container.decodeIfPresent(FilePermissions.self, forKey: .permissions)
+    }
+
+    /// Encode a `FilePath` as a file-URL `absoluteString` to match the prior
+    /// `URL`-typed wire format byte-for-byte.
+    private static func encodePath(_ path: FilePath) -> String {
```
jglogan: This can go away if we enforce absolute paths on init().

jglogan (changes requested):
Looks good, but we've added an absolutePath helper in containerization that we can now use, and I think things will be more maintainable if we add validation to the `PublishSocket.init()` calls so that a `PublishSocket` is at lexically correct by construction (we don't need to do semantic checks, such as file/directory existence in the initializers).

chrisgeo:
@jglogan I believe I addressed all concerns, please let me know if there's any I missed.

Sources/ContainerResource/Container/PublishSocket.swift
```diff
@@ -14,27 +14,137 @@
...
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(containerPath.string, forKey: .containerPath)
+        try container.encode(hostPath.string, forKey: .hostPath)
+        try container.encodeIfPresent(permissions, forKey: .permissions)
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let containerPath = try Self.decodePath(from: container, forKey: .containerPath)
```
jglogan, line 90: What do you think about enforcing absolute path here (or in decodePath, since it's a private func dedicated to just these paths) as well? Legacy should always be absolute, but we should protect against bad inputs should someone muck about with a config manually.

jglogan (changes requested):
@chrisgeo Spot on, one last change suggestion for validating decoded PublishSocket object paths and then we should be good to go.

Tests/ContainerResourceTests/PublishSocketTests.swift
```diff
@@ -0,0 +1,229 @@
...
+        let json = try JSONEncoder().encode(socket)
+        let decoded = try #require(try JSONSerialization.jsonObject(with: json) as? [String: Any])
+        #expect(decoded["containerPath"] as? String == "/tmp/a b.sock")
+        #expect(decoded["hostPath"] as? String == "/tmp/dir with spaces/sock")
+    }
+
+    // MARK: - Decoding (canonical plain-path form)
+
+    @Test
+    func testDecodePlainAbsolutePath() throws {
```
jglogan, line 99: Add a test for decoding a PublishSocket that has relative paths.

jglogan (changes requested):

Sources/ContainerResource/Container/PublishSocket.swift
chrisgeo: I think that's a valid path to cover until it's phased out.

chrisgeo: Done in 01ec891. `decodePath` now validates absoluteness directly and throws `DecodingError.dataCorrupted` ("decoded socket path must be absolute") for non-absolute decoded paths, covering both the plain-path and legacy `file:` branches. The by-construction check in `init()` stays as defense in depth, but the private helper is now self-validating so a manually edited/corrupt config is rejected at the decode layer.

Tests/ContainerResourceTests/PublishSocketTests.swift
chrisgeo: Added `testDecodeRelativeHostPathThrows` (relative `hostPath`) in 01ec891 and updated the existing `testDecodeRelativePathThrows` (relative `containerPath`) comment to reflect decode-layer enforcement. Both assert `DecodingError`. `swift test --filter ContainerResourceTests` → 53/53 passing.

chrisgeo: @jglogan To verify, you’d like relative paths to fail?

jglogan: Exactly. The configuration need to contain absolute paths, as it's not clear what a relative path should resolve to if a relative path were valid.

jglogan (approved):

jglogan:
@chrisgeo Thank you for these fixes!

jglogan:
Added @katiewasnothere for approval since I merged a conflict fix for Package.resolved.

katiewasnothere (approved):

PR #1595, opened 2026-05-24
feat(api): Add blkio cgroup tuning via --blkio flag

### LINKED ISSUE #1512: [Request] Add blkio/cgroup BFQ tuning flags to container run/create
### Feature request

Please add block I/O tuning flags to `container run` and `container create` for cgroup/BFQ-style workload shaping.

### Compose compatibility

Docker Compose exposes this through `service.blkio_config`. Container-Compose can parse the field, but it must currently warn-skip it because apple/container has no equivalent runtime flag surface.

### Requested CLI shape

- `--blkio-weight <weight>`
- `--blkio-weight-device <path>:<weight>`
- `--device-read-bps <path>:<rate>`
- `--device-write-bps <path>:<rate>`
- `--device-read-iops <path>:<rate>`
- `--device-write-iops <path>:<rate>`
- matching support on `container create`

The exact implementation can be constrained to what the Linux guest and host virtualization stack can support, but a first-class CLI/API surface would let compose translators pass through supported block-I/O constraints instead of silently dropping them.

chrisgeo:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Closes #1512.

### Use case

Docker Compose exposes block-I/O tuning through [`service.blkio_config`][compose-blkio]
(weights, weight-per-device, and read/write bps/iops throttles). Compose
translators such as container-compose can parse the field today, but they
must warn-skip it on `apple/container` because there is no runtime flag
surface for it — `blkio_config:` silently degrades into an unconfigured
container. This PR adds the missing surface so:

1. **container-compose (and other compose translators) can pass
   `service.blkio_config` straight through** to `container run` /
   `container create` without dropping fields.
2. Operators tuning **noisy-neighbor / IO-bound workloads** (databases,
   build steps, log shippers sharing a host disk) can apply BFQ-style
   weights and per-device throttles directly from the CLI, matching the
   `docker run --blkio-weight*` / `--device-{read,write}-{bps,iops}`
   ergonomics they already know.
3. The cgroup `blockIO` config flows through the runtime data path so a
   future `podman`-style translator or higher-level orchestrator gets the
   same plumbing for free.

### CLI shape

Per the grouped, repeatable `key=value` form proposed in
[#1512 (comment)][cli-shape] (lives under `MANAGEMENT OPTIONS`, not
`RESOURCE OPTIONS`, to keep the resource help wall short):

```bash
container run --blkio weight=500 ...
container run --blkio device=/dev/sda,weight=700,leaf-weight=300 ...
container run --blkio device=/dev/sda,read-bps=1048576,write-bps=1048576 ...
container run --blkio device=/dev/sda,read-iops=1000,write-iops=1000 ...
```

`device=` accepts an absolute host path (resolved via `stat(2)` for
major/minor) or a literal `<major>:<minor>`. Per-device keys: `weight`,
`leaf-weight`, `read-bps`, `write-bps`, `read-iops`, `write-iops`. Global
keys (no `device=`): `weight`, `leaf-weight`. The parser rejects unknown
keys, conflicting global weights, and global-only keys mixed onto
device-less specs.

### Runtime plumbing

To keep OS-specific options out of the cross-platform container config,
`blockIO` is carried as **opaque runtime data**, not on
`ContainerConfiguration.Resources`:

- CLI (`ContainerRun` / `ContainerCreate`) calls `Parser.blockIO(specs:)`,
  wraps the result in `LinuxRuntimeData(blockIO:)`, JSON-encodes it, and
  passes it as `runtimeData` to `ContainerClient.create`.
- `RuntimeService.bootstrap` reads `RuntimeConfiguration.runtimeData`
  from disk and threads it into `configureContainer`, which decodes
  `LinuxRuntimeData` and applies `czConfig.blockIO` (a
  `Containerization.LinuxBlockIO`, the `Sendable` wrapper around the OCI
  type).

## Dependency

Depends on [apple/containerization#739][cz-pr], which introduces
`Containerization.LinuxBlockIO` and wires it into
`LinuxContainer.Configuration`. While #739 is unmerged this branch pins
`containerization` to its branch
(`full-chaos/containerization@feat/chaos-1380-blkio-runtime`,
revision `3d009df`). Before merge, the pin moves back to
`apple/containerization` at whatever revision lands #739.

[compose-blkio]: https://docs.docker.com/reference/compose-file/services/#blkio_config
[cli-shape]: https://github.com/apple/container/issues/1512#issuecomment-4526251917
[cz-pr]: https://github.com/apple/containerization/pull/739

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

`swift build` clean, `swift test --filter ParserTest` passes
(105 tests, 5 new) — tests call `Parser.blockIO` directly now that the
field is no longer on `Resources`:
- `testResourcesBlockIOFlags` — combined global + per-device + throttle
- `testResourcesBlockIOAcceptsMajorMinorLiteral` — `device=8:0` literal
- `testResourcesRejectsInvalidBlockIOWeight` — weight outside 10..1000
- `testResourcesRejectsUnknownBlockIOKey` — typo detection
- `testResourcesRejectsGlobalKeyOnDeviceSpec` — `read-bps` without `device=`

`RuntimeConfigurationTests` pass (runtime-data round-trip).
`container run --help` confirmed renders the new `--blkio` flag under
`MANAGEMENT OPTIONS`.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -13,10 +13,11 @@
     {
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/containerization.git",
+      "location" : "https://github.com/full-chaos/containerization.git",
       "state" : {
         "revision" : "9275f365dd555c8f072e7d250d809f5eb7bdd746",
         "version" : "0.33.4"
+
       }
     },
     {
--- a/Package.resolved
+++ b/Package.resolved
@@ -13,11 +13,10 @@
     {
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/full-chaos/containerization.git",
+      "location" : "https://github.com/apple/containerization.git",
       "state" : {
         "revision" : "9275f365dd555c8f072e7d250d809f5eb7bdd746",
         "version" : "0.33.4"
-
       }
     },
     {
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "35d1f07a3595a3ebabac1430f46ebb8b4e54d4531874f4ff7df33b64092df5f3",
+  "originHash" : "72e23775adf1d2847eb1bb074e288ba36b7691127416439ad99d1bbff4edadb7",
   "pins" : [
     {
       "identity" : "async-http-client",
@@ -13,10 +13,9 @@
     {
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/containerization.git",
+      "location" : "https://github.com/full-chaos/containerization.git",
       "state" : {
-        "revision" : "5427fd21ded4b84034126caef5b3182900b4776d",
-        "version" : "0.41.0"
+        "revision" : "63527275f9ad41e13984038e155f13f056a75def"
       }
     },
     {
--- a/Package.resolved
+++ b/Package.resolved
@@ -13,7 +13,7 @@
     {
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/full-chaos/containerization.git",
+      "location" : "https://github.com/apple/containerization.git",
       "state" : {
         "revision" : "5427fd21ded4b84034126caef5b3182900b4776d",
         "version" : "0.41.0"
--- a/Package.resolved
+++ b/Package.resolved
@@ -13,7 +13,7 @@
     {
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/containerization.git",
+      "location" : "https://github.com/full-chaos/containerization.git",
       "state" : {
         "revision" : "5427fd21ded4b84034126caef5b3182900b4776d",
         "version" : "0.41.0"
--- a/Package.swift
+++ b/Package.swift
@@ -350,7 +350,9 @@ let package = Package(
         ),
         .target(
             name: "ContainerRuntimeLinuxClient",
-            dependencies: [],
+            dependencies: [
+                .product(name: "ContainerizationOCI", package: "containerization"),
+            ],
             path: "Sources/Services/RuntimeLinux/Client"
         ),
         .executableTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -53,7 +53,7 @@ let package = Package(
         .library(name: "MachineAPIService", targets: ["MachineAPIService"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/containerization.git", exact: Version(stringLiteral: scVersion)),
+        .package(url: "https://github.com/full-chaos/containerization.git", branch: scVersion),
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-configuration", from: "1.0.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -53,7 +53,7 @@ let package = Package(
         .library(name: "MachineAPIService", targets: ["MachineAPIService"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/full-chaos/containerization.git", branch: scVersion),
+        .package(url: "https://github.com/apple/containerization.git", exact: Version(stringLiteral: scVersion)),
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.3.0"),
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-configuration", from: "1.0.0"),
--- a/Package.swift
+++ b/Package.swift
@@ -401,7 +401,9 @@ let package = Package(
         ),
         .target(
             name: "ContainerRuntimeLinuxClient",
-            dependencies: [],
+            dependencies: [
+                .product(name: "ContainerizationOCI", package: "containerization"),
+            ],
             path: "Sources/Services/RuntimeLinux/Client"
         ),
         .executableTarget(
--- a/Package.swift
+++ b/Package.swift
@@ -55,7 +55,10 @@ let package = Package(
         .library(name: "ContainerK8s", targets: ["ContainerK8s"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/containerization.git", exact: Version(stringLiteral: scVersion)),
+        .package(
+            url: "https://github.com/full-chaos/containerization.git",
+            revision: "63527275f9ad41e13984038e155f13f056a75def"
+        ),
         .package(url: "https://github.com/apple/swift-argument-parser.git", from: "1.7.0"),
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.2.0"),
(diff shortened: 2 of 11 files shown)


Package.swift
```diff
@@ -49,7 +49,7 @@ let package = Package(
         .library(name: "TerminalProgress", targets: ["TerminalProgress"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/containerization.git", exact: Version(stringLiteral: scVersion)),
+        .package(url: "https://github.com/full-chaos/containerization.git", branch: scVersion),
```
chrisgeo: Will remove before merge.

.gitignore
```diff
@@ -31,3 +31,6 @@ coverage-reports/
 # API docs for local preview only.
 _site/
 _serve/
+.vscode/**
+.worktrees/**
+.worktress/**
```
chrisgeo: Will remove before merge.

Package.resolved
```diff
@@ -13,10 +13,10 @@
...
       "identity" : "containerization",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/apple/containerization.git",
+      "location" : "https://github.com/full-chaos/containerization.git",
       "state" : {
-        "revision" : "db5b5b98405d53543f69105087130ffd623a5b9a",
-        "version" : "0.32.2"
+        "branch" : "feat/chaos-1380-blkio-runtime",
+        "revision" : "3d009dfc2724fc924dd18ca2014b5e839d554163"
       }
```
chrisgeo: Will revert before merge.

Sources/ContainerResource/Container/ContainerConfiguration.swift
```diff
@@ -149,6 +149,8 @@ public struct ContainerConfiguration: Sendable, Codable {
         public var cpus: Int = 4
         /// Memory in bytes allocated.
         public var memoryInBytes: UInt64 = 1024.mib()
+        /// Block I/O resource limits.
+        public var blockIO: LinuxBlockIO?
```
jglogan: This definitely moves it in the desired direction. Since this is a Linux-specific option, it'd be better that this become a field in `LinuxRuntimeData` and then if `LinuxBlockIO` needs to be present, you populate it there, encode the `LinuxRuntimeData` as `runtimeData` for the [`RuntimeConfiguration`](https://github.com/apple/container/blob/main/Sources/Services/Runtime/RuntimeClient/RuntimeConfiguration.swift#L33) type, and then the RuntimeService can decode the runtime config, check for the presence of `runtimeData`, decode out `LinuxRuntimeData`, and do as it pleases with the result.

We don't want to have things that are overtly OS-specific in the `ContainerConfiguration`.

Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -114,6 +119,24 @@ public struct Flags {
             help: "Amount of memory (1MiByte granularity), with optional K, M, G, T, or P suffix"
         )
         public var memory: String?
+
+        @Option(
+            name: .customLong("blkio"),
```
jglogan: We don't really have a good mechanism for providing OS-specific options on the CLI (it's going to require some non-trivial rework on our command line processing), so I think the best thing here is just to have this be `options` like we do for non-generic volume options (I'm aiming to do this networks as well).

I wouldn't document the option list here as it'll turn the help output into a text wall. Better to update the command reference under docs where we can have a huge list, table, whatever.

jglogan: Since it would be `options` this should go under `Flags.Management` instead of `Flags.Resource`.

jglogan (changes requested):

chrisgeo:
Pushed [`5d11a29`](https://github.com/apple/container/pull/1595/commits/5d11a29) addressing review feedback.

### 1. `blockIO` no longer lives on `ContainerConfiguration`

Per the suggestion on `ContainerConfiguration.swift`, the field is now on `LinuxRuntimeData` and is carried opaquely through `RuntimeConfiguration.runtimeData`:

- CLI (`ContainerRun` / `ContainerCreate`) calls `Parser.blockIO(specs:)`, wraps the result in `LinuxRuntimeData(blockIO:)`, JSON-encodes it, and passes it as `runtimeData` to `ContainerClient.create`.
- `RuntimeService.bootstrap` reads `RuntimeConfiguration.runtimeData` from disk and threads it into `configureContainer`, which decodes `LinuxRuntimeData` and applies `czConfig.blockIO`.
- `ContainerConfiguration.Resources.blockIO` is removed; `Parser.resources` no longer takes `blkio`.

This keeps OS-specific options out of the generic container config type, as you described.

### 2. `--blkio` reshaped as `options` under `Flags.Management`

- The flag moved from `Flags.Resource` to `Flags.Management` (it shows up under `MANAGEMENT OPTIONS` in `container run --help` now).
- Help text collapsed to a single line: `Block I/O cgroup tuning options (Linux only; see command reference for the supported keys)`. The detailed key/value schema will live in the command reference doc rather than the help wall.
- The structured key=value parsing in `Parser.blockIO` is unchanged, so existing behavior is preserved.

### 3. Tests

`swift test --filter ParserTest` — 105 tests pass, including the 5 blkio cases (now rewired to call `Parser.blockIO` directly since the field is gone from `Resources`). `RuntimeConfigurationTests` also still pass.

### Deferred cleanup

The three self-review notes on `Package.swift`, `Package.resolved`, and `.gitignore` are intentionally left for now:

- `Package.swift` / `Package.resolved` still pin to the [containerization#739](https://github.com/apple/containerization/pull/739) branch because that upstream PR is still open. They will revert to `apple/containerization` at whatever revision lands #739, as the PR description notes.
- `.gitignore` shows no diff against `main` on the current branch, so there is nothing to revert there — flagging in case I'm missing something on your end.

Sources/Services/ContainerAPIService/Client/Flags.swift
```diff
@@ -334,6 +339,15 @@ public struct Flags {
         @Option(name: .customLong("shm-size"), help: "Size of /dev/shm (e.g. 64M, 1G)")
         public var shmSize: String?
 
+        @Option(
+            name: .customLong("blkio"),
+            help: .init(
+                "Block I/O cgroup tuning options (Linux only; see command reference for the supported keys)",
```
jglogan:
```suggestion
                "Block I/O cgroup tuning options (experimental: see command reference for the supported keys)",
```

Let's explain the platform-specific nature in the documentation.

I'm not sure yet whether we want to have a dedicated option for this yet, or just a generic `--runtime-options` option list that can carry any amount of runtime options (among which are `blkio.*`). Marking it experimental and explaining the nature of that in the reference allows you to move forward while alerting users that this is unstable from a UX perspective.

jglogan (changes requested):
Thanks for reworking things to make the data specific to the Linux runtime type.

One last pair of changes and AFAICT it looks fine, and then we should be able to proceed once @dcantah weighs in on the containerization change. See the comment regarding the CLI help, and update the command reference document with a really brief discussion of the option, and mention that the CLI option is experimental and may change in the future.

jglogan:
@chrisgeo Haven't forgotten about this but we've been busy preparing for the release next week. We'll be able to start addressing the PR backlog once that's out.

chrisgeo:
@jglogan not a problem and no rush at all

PR #1603, opened 2026-05-26
Add shmSize to management flag's init

katiewasnothere:
## Type of Change
- [x] Bug fix

## Motivation and Context
Ensure all fields are initialized in the management flags' init function. Without this change, if someone calls init() on this set of flags, they will get an error like 

```
Can't read a value from a parsable argument definition.

This error indicates that a property declared with an `@Argument`,
`@Option`, `@Flag`, or `@OptionGroup` property wrapper was neither
initialized to a value nor decoded from command-line arguments.

To get a valid value, either call one of the static parsing methods
(`parse`, `parseAsRoot`, or `main`) or define an initializer that
initializes _every_ property of your parsable type.
```

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -189,6 +189,7 @@ public struct Flags {
             rosetta: Bool,
             runtime: String?,
             ssh: Bool,
+            shmSize: String?,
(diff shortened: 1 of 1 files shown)


jglogan (approved):

hernandezadonai587 (commented):
Q

PR #1605, opened 2026-05-27
Print warning message when page cache thrashing is detected

JaewonHur:
Container build process can (seemingly) hang when a lots of threads contend for the disk IO, thrashing each other's page cache. User had no means to tell the hang is due to this issue or the issue in Dockerfile itself. This PR makes build procedure print warning message on thrashing is detected.

## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Help users root causing the hang while `container build`

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/BuildCommand.swift
+++ b/Sources/ContainerCommands/BuildCommand.swift
@@ -25,6 +25,7 @@ import ContainerizationError
 import ContainerizationOCI
 import ContainerizationOS
 import Foundation
+import Logging
 import NIO
 import TerminalProgress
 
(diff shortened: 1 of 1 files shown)


Sources/ContainerCommands/BuildCommand.swift
```diff
@@ -333,6 +334,39 @@ extension Application {
...
+                        processConfig.executable = "/bin/cat"
+                        processConfig.arguments = ["/sys/fs/cgroup/memory.stat"]
+                        processConfig.terminal = false
+
+                        var previousStats: [String: UInt64] = [:]
+
+                        while true {
+                            let stdout = Pipe()
+
+                            let process = try await client.createProcess(
```
jglogan, line 349: Any throwing non-cancel  error in here will kill the build, I think. Is this ok?

jglogan (approved):
Approved but see comment.

dcantah:
We have `container.stats()` today which already tallies up this info, but just doesn't surface these two specific stats. Can we just make a change over in CZ to actually expose them?

dcantah:
https://github.com/apple/containerization/pull/758

PR #1607, opened 2026-05-27
Use ManagedResource for volumes in CLI commands

katiewasnothere:
## Type of Change
- [x] New feature  

## Motivation and Context
Related to https://github.com/apple/container/issues/1404.

This PR adds the initial work to have volume resources conform to ManagedResource, in alignment with other resources such as networks (see [here](https://github.com/apple/container/pull/1421)). Further work is necessary to move the use of `VolumeResource` down to the APIServer (in the VolumesService) and in the volume client. Volumes do not currently have any plugin or runtime state, so that information is not included in the `VolumeResource`, but could be added later if needed. 

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/ContainerCommands/Volume/VolumeResource+ListDisplayable.swift
+++ b/Sources/ContainerCommands/Volume/VolumeResource+ListDisplayable.swift
@@ -1,0 +1,36 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerResource
+
+extension VolumeResource: ListDisplayable {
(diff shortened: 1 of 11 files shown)


Sources/ContainerCommands/Volume/VolumeResource+ListDisplayable.swift
```diff
@@ -0,0 +1,36 @@
...
+import ContainerResource
+
+extension VolumeResource: ListDisplayable {
+    public static var tableHeader: [String] {
+        ["NAME", "TYPE", "DRIVER", "OPTIONS"]
+    }
+
+    public var tableRow: [String] {
+        [
+            id,
```
jglogan: We don't have to do it today, but just wanted to get the thought out there.

If we switch ID over to a system-assigned identifier (for container, volume, network it'd be a 32 digit hex value like Docker), we'd probably want to report `name` and not `id` here. What do you think?

Sources/ContainerResource/Volume/VolumeResource.swift
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// A volume resource, representing a configured volume.
+public struct VolumeResource: ManagedResource {
+    /// The volume's configuration — its persistent, intrinsic properties.
+    public let config: Volume
```
jglogan: Again not necessarily for this PR `Volume` become `VolumeConfiguration`?

Sources/ContainerCommands/Volume/VolumeResource+ListDisplayable.swift
katiewasnothere: Seems very reasonable to me. I'll change it to that.

jglogan (approved):

PR #1611, opened 2026-05-28
Align JSON output with shape for container resource.

jglogan:
- Part of #1404.
- Use `configuration` for configuration properties, `state` for current state label, `status` for status properties.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [x] Breaking change
- [ ] Documentation update

## Motivation and Context
The hope here is to break CLI JSON and XPC compatibility with previous versions all at once at the 1.0.0 release so that output is uniform and cleanly extensible.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs


ajemory (approved):
Having `state`, `status`, and `phase` is pretty confusing and unintuitive. We should address that in a future PR. For the purpose of aligning the JSON output, looks good

PR #1614, opened 2026-05-28
Add TOML configuration for default DNS nameservers, search, options

### LINKED ISSUE #1449: [Request]: system property for default `--dns`.

NOTE: Depends on #1425; there's no point tackling this until the TOML configuration is in place.

We've a number of issues relating to DNS resolution from containers not working properly when a third party DNS server conflicts with the server that macOS mDNSResponder wants to start.

Until an OS-level solution can be provided, it would be useful to have a system property (and potentially environment variables if session level overrides seem sensible) that allows the user to configure default values for `--dns`.

We can also consider properties for the other DNS CLI options here as well.

### Code of Conduct

0xMH:
## Type of Change
- [x] New feature

## Motivation and Context

Closes #1449. Adds default values for `--dns`, `--dns-search`, `--dns-option`, and `--dns-domain` to the `[dns]` section of `~/.config/container/config.toml`, so users hitting macOS `mDNSResponder` conflicts can set the workaround once instead of repeating it on every invocation. Depends on the merged TOML configuration introduced by #1425.

Defaults are read by `container run`, `container build`, and `container builder start` via a shared `Utility.dnsConfiguration(from:defaults:)` helper. CLI flags take precedence; `--no-dns` still disables DNS. Two pre-existing bugs in the build path were fixed to make the feature work end-to-end: `BuildCommand` was forwarding only `dnsNameservers` to the builder (now forwards all four DNS fields), and `BuilderStart`'s `dnsChanged` check only compared the first non-empty field (now compares all four).

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerPersistence/ContainerSystemConfig.swift
+++ b/Sources/ContainerPersistence/ContainerSystemConfig.swift
@@ -113,16 +113,23 @@ final public class ContainerConfig: Codable, Sendable {
 
     public let cpus: Int
     public let memory: MemorySize
+    public let dns: ContainerDNSConfig
 
-    public init(cpus: Int = defaultCPUs, memory: MemorySize = defaultMemory) {
+    public init(
+        cpus: Int = defaultCPUs,
+        memory: MemorySize = defaultMemory,
+        dns: ContainerDNSConfig = .init()
+    ) {
         self.cpus = cpus
         self.memory = memory
+        self.dns = dns
     }
 
     public init(from decoder: any Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.cpus = try container.decodeIfPresent(Int.self, forKey: .cpus) ?? Self.defaultCPUs
         self.memory = try container.decodeIfPresent(MemorySize.self, forKey: .memory) ?? Self.defaultMemory
+        self.dns = try container.decodeIfPresent(ContainerDNSConfig.self, forKey: .dns) ?? .init()
     }
 }
 
@@ -136,6 +143,37 @@ final public class DNSConfig: Codable, Sendable {
     public init(from decoder: any Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         self.domain = try container.decodeIfPresent(String.self, forKey: .domain)
+    }
+}
+
+final public class ContainerDNSConfig: Codable, Sendable {
+    public static let defaultNameservers: [String] = []
+    public static let defaultSearchDomains: [String] = []
+    public static let defaultOptions: [String] = []
+
+    public let domain: String?
+    public let nameservers: [String]
+    public let searchDomains: [String]
+    public let options: [String]
+
+    public init(
+        domain: String? = nil,
+        nameservers: [String] = defaultNameservers,
+        searchDomains: [String] = defaultSearchDomains,
+        options: [String] = defaultOptions
+    ) {
+        self.domain = domain
+        self.nameservers = nameservers
+        self.searchDomains = searchDomains
+        self.options = options
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.domain = try container.decodeIfPresent(String.self, forKey: .domain)
+        self.nameservers = try container.decodeIfPresent([String].self, forKey: .nameservers) ?? Self.defaultNameservers
+        self.searchDomains = try container.decodeIfPresent([String].self, forKey: .searchDomains) ?? Self.defaultSearchDomains
+        self.options = try container.decodeIfPresent([String].self, forKey: .options) ?? Self.defaultOptions
     }
 }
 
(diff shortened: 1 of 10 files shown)


Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -127,15 +127,33 @@ final public class ContainerConfig: Codable, Sendable {
 }
 
 final public class DNSConfig: Codable, Sendable {
+    public static let defaultNameservers: [String] = []
```
katiewasnothere, line 150: Today there is a subtle difference between what the default DNS settings are for `container` aka the apiserver itself, which would determine the hostname resolution (aka the A record) for a given container on the _host_, and the DNS settings for setting up the resolv.conf in a container. I think we should maintain this distinction. There could be scenarios where a user does not want the default DNS domain on the host to necessarily match the default DNS domain that the container application uses in the container itself. 

I think we should add a new field on the `ContainerConfig` type that has the default DNS settings for running a container. 

Essentially we'd end up with something like this in the TOML:
```
[dns] <-------- default DNS settings for the APIServer
domain = "test"

[container.dns] <----------- default DNS settings for containers 
server = "8.8.8.8"
domain = "foo"
search = ["foo", "test"]
options = ["haha"]
```

0xMH: Thanks, good insight. I removed that extra builder startup call.

0xMH:
Thanks for the review, @katiewasnothere. Happy to iterate further if anything else needs addressing.

katiewasnothere:
Hi @0xMH thank you for all the work you've done on this change! We are planning to make a new release of the container package some time in the next week and we want to hold off on merging this change until then.

am-saksham:
Hi @0xMH and @katiewasnothere! First, thanks for the great work figuring out the core logic for this. 

I noticed this PR has been stalled since the June release and has picked up some merge conflicts. I also saw that a couple of newer PRs for #1449 were opened recently, but they seem to have missed the `[dns]` vs `[container.dns]` architectural requirement that Katie pointed out above.

@0xMH, if you are currently swamped with other things, would you mind if I pull your branch, resolve the conflicts against `main`, and open a fresh PR to get this over the finish line? I will carry over your commits and make sure to include a `Co-authored-by` tag so you keep full credit for your work. 

Let me know if that sounds good!

0xMH:
Thanks for the offer! but no need since I'm waiting for the green flag to continue and rebase and prepare for the merge.

am-saksham:
Awesome, sounds good! Looking forward to seeing it merged. Let me know if you end up needing a hand later on!

katiewasnothere:
@0xMH Really sorry for the delay here, we've been swamped. I'd love to get this in for this sprint (which ends at the end of August). Could you rebase your PR? I will take another pass through the code. Thank you for your work here!

docs/command-reference.md
```diff
@@ -85,7 +85,7 @@ container run [<options>] <image> [<arguments> ...]
             - `10.*.*.*`
             - `192.168.*.*`
             - `172.16.*.*` through `172.31.*.*`
-        - The host ends with the machine's default container DNS domain (as defined in `DNSConfig.defaultDomain`, located [here](../Sources/ContainerPersistence/ContainerSystemConfig.swift))
+        - The host ends with the machine's configured internal DNS domain from `[dns].domain`
```
katiewasnothere, line 88: Could we add a link to the reference doc here instead?

Sources/Services/ContainerAPIService/Client/Utility.swift
```diff
@@ -330,6 +328,32 @@ public struct Utility {
         return [AttachmentConfiguration(network: builtinNetworkId, options: AttachmentOptions(hostname: fqdn ?? containerId, macAddress: nil, mtu: 1280))]
     }
 
+    public static func dnsConfiguration(
+        from flags: Flags.DNS,
+        defaults: ContainerDNSConfig,
+        hostDomainFallback: String? = nil
```
katiewasnothere, line 334: What is this for?

Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -139,6 +146,37 @@ final public class DNSConfig: Codable, Sendable {
...
 
+final public class ContainerDNSConfig: Codable, Sendable {
+    public static let defaultNameservers: [String] = []
+    public static let defaultSearchDomains: [String] = []
+    public static let defaultOptions: [String] = []
+
+    public let domain: String?
+    public let nameservers: [String]
+    public let searchDomains: [String]
+    public let options: [String]
```
katiewasnothere, line 157: What do you think about making these optionals instead of empty arrays?

Sources/ContainerPersistence/ContainerSystemConfig.swift
```diff
@@ -113,16 +113,23 @@ final public class ContainerConfig: Codable, Sendable {
 
     public let cpus: Int
     public let memory: MemorySize
+    public let dns: ContainerDNSConfig
```
katiewasnothere, line 116: What do you think about making this an optional as well?

PR #1618, opened 2026-05-29
feat(plugin): add user-level plugin directory that survives upgrades

### LINKED ISSUE #1617: [Request]: User-installed CLI plugins are lost on every container upgrade
CLI plugins installed into `{install-root}/libexec/container-plugins/` survive inside the current keg but get wiped when `container` is upgraded (Homebrew moves the symlink from `Cellar/container/0.12.3/` → `0.12.4/`, old keg gets cleaned).

Steps to reproduce (Homebrew, but applies to any package manager):

```bash
# Install a CLI plugin
mkdir -p /opt/homebrew/Cellar/container/0.12.3/libexec/container-plugins/myplugin/bin
cp my-plugin /opt/homebrew/Cellar/container/0.12.3/libexec/container-plugins/myplugin/bin/myplugin
chmod +x /opt/homebrew/Cellar/container/0.12.3/libexec/container-plugins/myplugin/bin/myplugin

cat > /opt/homebrew/Cellar/container/0.12.3/libexec/container-plugins/myplugin/config.toml << 'EOF'
abstract = "my plugin"
author = "me"
EOF

container myplugin   # works

brew upgrade container  # keg rotates

container myplugin   # Plugin 'container-myplugin' not found.
```

Docker solves this with `~/.docker/cli-plugins/` — a stable, user-writable directory that survives Docker upgrades. Could `container` search a similar path?

Proposal: add a fourth search directory to `PluginLoader` that's outside the versioned keg, e.g.:

- `~/.config/container/plugins/`
- `~/.local/lib/container/plugins/`
- `$XDG_DATA_HOME/container/plugins/`

Priority would be: user plugins dir (new, highest) → install-root plugins → app-bundle plugins. Same `config.toml` + `bin/` layout as existing plugins. No changes to the plugin format or daemon plugin registration.

This is the main blocker for shipping a Homebrew-distributed compose plugin — right now every `brew upgrade container` silently breaks `container compose`.

container 0.12.3, macOS 26.3, Homebrew.

Cyb3rDudu:
CLI plugins installed into `{install-root}/libexec/container-plugins/` are lost on every `container` upgrade because the path lives inside the versioned keg (e.g. Homebrew rotates `Cellar/container/0.12.3/` → `0.12.4/`).

Add a fourth search directory to `PluginLoader` that is stable across upgrades: `XDG_DATA_HOME/container/plugins/` (falls back to `~/.local/share/container/plugins/`). The directory is opt-in — it must exist before `container` will scan it.

User-home plugins take priority over install-root plugins, so a package-manager upgrade cannot silently shadow plugins the user installed themselves. Search order is now: user-home → install-root → app-bundle.

Closes #1617

### DIFF
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -158,7 +158,11 @@ public struct Application: AsyncLoggableCommand {
             .appending(FilePath.Component("container"))
             .appending(FilePath.Component("plugins"))
         let installRootPluginsURL = URL(fileURLWithPath: installRootPluginsPath.string)
+        // user-level plugins directory (stable across upgrades)
+        let userHomePluginsURL = PluginLoader.userHomePluginsDir()
+
         let pluginDirectories = [
+            userHomePluginsURL,
             userPluginsURL,
             appBundlePluginsURL,
             installRootPluginsURL,
--- a/Sources/ContainerCommands/DefaultCommand.swift
+++ b/Sources/ContainerCommands/DefaultCommand.swift
@@ -62,7 +62,8 @@ struct DefaultCommand: AsyncLoggableCommand {
             .appending(FilePath.Component("container"))
             .appending(FilePath.Component("plugins"))
         let installRootPluginsURL = URL(fileURLWithPath: installRootPluginsPath.string)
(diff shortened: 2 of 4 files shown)


Cyb3rDudu:
@jglogan any updates on this. Sorry for bumping :)

PR #1622, opened 2026-06-01
Enhance bridge networking with helper-based bridging support

### LINKED ISSUE #489: [Request]: Bridge Mode

curdbecker:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Based on @torarnv's [existing pull request](https://github.com/apple/container/pull/1463), this pull request introduces support via a `BridgeNetworkService` in the vmnet plugin helper to enable socket-based bridge forwarding. The `BridgeNetworkService` is heavily inspired by [nirs/vmnet-helper](https://github.com/nirs/vmnet-helper) and draws on the same concept - although simplified only towards bridge networking with a vmnet interface and lacking basically all performance optimization so far. 

This allow bridging to a real host network interface without having the binary signed with the `com.apple.vm.networking` entitlement, e.g. during container development.

In addition, this pull request rebases the changes of the existing pull request on the refactoring done in the current main branch which seems to working towards a more general network plugin interface and also introduces several, likely temporary enhancements to bridge the current gap of getting the network configuration options actually to the network helper during creation.

Specifically, the bridged networking understands two additional variants: 
`.bridged` network attachments to a BridgedNetworkInterface for the given host interface from the original PR.
`.bridgedViaHelper` network attachments to a FileHandleNetworkInterface for the FileHandle endpoint where the plugin helper will provide bridging services to the runtime.

Simple example:
Screenshot 2026-06-01 at 15 58 32

Requires https://github.com/apple/containerization/pull/759 (also based on [@torarnv's work](https://github.com/apple/containerization/pull/759))
Extends https://github.com/apple/container/pull/1463
Fixes https://github.com/apple/container/issues/489

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/Network/Server/BridgeNetworkService.swift
+++ b/Sources/Services/Network/Server/BridgeNetworkService.swift
@@ -1,0 +1,488 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerNetworkClient
+import ContainerResource
+import ContainerXPC
+import ContainerizationError
+import ContainerizationExtras
+import Foundation
+import Logging
+import vmnet
+
+internal final class Bridge {
+
+    internal final class PacketBuffer {
+        public let buffers: UnsafeMutableRawPointer
+        public let iovs: UnsafeMutablePointer<iovec>
+        public let packets: UnsafeMutablePointer<vmpktdesc>
+
+        public let maxPacketSize: Int
+        public let maxPktCount: Int
+
+        public init(maxPktCount: Int, maxPktSize: Int) {
+            self.buffers = .allocate(byteCount: maxPktCount * maxPktSize, alignment: 16)
+            self.iovs = .allocate(capacity: maxPktCount)
+            self.packets = .allocate(capacity: maxPktCount)
+            for i in 0..<maxPktCount {
+                iovs[i] = iovec(
+                    iov_base: buffers.advanced(by: i * maxPktSize),
+                    iov_len: maxPktSize)
+                packets[i] = vmpktdesc(
+                    vm_pkt_size: maxPktSize,
+                    vm_pkt_iov: iovs.advanced(by: i),
+                    vm_pkt_iovcnt: 1,
+                    vm_flags: 0)
+            }
+            self.maxPacketSize = maxPktSize
+            self.maxPktCount = maxPktCount
+        }
+
+        public func reset() {
+            // Reset fields - vmnet_read mutates them
+            for i in 0..<maxPktCount {
+                packets[i].vm_pkt_size = maxPacketSize
+                packets[i].vm_flags = 0
+                iovs[i].iov_len = maxPacketSize
+            }
+        }
+
+        deinit {
+            buffers.deallocate()
+            iovs.deallocate()
+            packets.deallocate()
+        }
+    }
+
+    // an unique identifier that serves as also as an identifier of the vmnet packet queue
+    public let identifier: String
+
+    // "our" network-side endpoint of the UNIX socket pipe to the sandbox service
+    public let networkEndpoint: FileHandle
+
+    // the remote-side endpoint of the UNIX socket pipe that the sandbox service can
+    // use to communicate with us
+    //
+    // NOTE: shall be closed as soon as we do not need it anymore, since it will be
+    // dup-ed when we send it to the sandbox service via XPC automatically
+    public let sandboxEndpoint: FileHandle
+
+    // MAC assigned to the interface by automatically vmnet
+    // Note: unfortunately currently it does not seem possible to explicitly get vmnet to assign a custom MAC address
+    public let macAddress: MACAddress
+
+    // MTU reported by vmnet (usually 1500).
+    public let mtu: UInt32
+
+    private let log: Logger
+
+    private var stopped: Bool = false
+
+    private let interfaceRef: interface_ref
+    private let queue: DispatchQueue
+    private var readSource: DispatchSourceRead?
+
+    // host->sandbox
+    private let hostPacketBuffer: PacketBuffer
+    // sandbox->host
+    private let sandboxPacketBuffer: PacketBuffer
+
+    public init(
+        identifier: String,
+        hostInterface: String,
+        logger: Logger,
+        enableTso: Bool,
+        enableChecksumOffload: Bool,
+        bufferedPacketCount: Int
+    ) throws {
+        var fds: [Int32] = [-1, -1]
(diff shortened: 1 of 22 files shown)


curdbecker:
Rebased on `main` including version `1.0.0` - also updated `containerization` in related PR :)

Sources/Services/Network/Server/BridgeNetworkService.swift
```diff
@@ -0,0 +1,488 @@
...
+    // the remote-side endpoint of the UNIX socket pipe that the sandbox service can
+    // use to communicate with us
+    //
+    // NOTE: shall be closed as soon as we do not need it anymore, since it will be
+    // dup-ed when we send it to the sandbox service via XPC automatically
+    public let sandboxEndpoint: FileHandle
+
+    // MAC assigned to the interface by automatically vmnet
+    // Note: unfortunately currently it does not seem possible to explicitly get vmnet to assign a custom MAC address
+    public let macAddress: MACAddress
```
nirs, line 85: The MAC address of the interface does not matter - the vm using the interface can use any mac address.

Sources/Services/Network/Server/BridgeNetworkService.swift
```diff
@@ -0,0 +1,488 @@
...
+        )
+        xpc_dictionary_set_string(desc, vmnet_shared_interface_name_key, hostInterface)
+        xpc_dictionary_set_bool(desc, vmnet_enable_tso_key, enableTso)
+        xpc_dictionary_set_bool(
+            desc, vmnet_enable_checksum_offload_key,
+            enableChecksumOffload)
+
+        // Unfortunately, apparently not understood currently - otherwise this would maybe provide us the opportunity to
+        // set a custom MAC address
+        // xpc_dictionary_set_bool(desc, vmnet_allocate_mac_address_key, false)
```
nirs, line 140: Reading vmnet.h this should disable the allocation of the MAC address by vmnet but it seems to be ignored and you always get mac address from vmnet.

Sources/Services/Network/Server/BridgeNetworkService.swift
```diff
@@ -0,0 +1,488 @@
...
+        }
+        return count
+    }
+
+    private func writeToSandbox(count: Int) -> Int {
+        var written = 0
+        for i in 0..<count {
+            let pktSize = hostPacketBuffer.packets[i].vm_pkt_size
+            guard write(networkEndpoint.fileDescriptor, hostPacketBuffer.iovs[i].iov_base, pktSize) == pktSize else {
+                continue
```
nirs, line 252: Under load this will fail with ENOBUFS - With this code you drop the packet which will cause retransmission.

vmnet-helper handles this by retrying after a short delay. Typically the first retry succeeds.

Sources/Services/Network/Server/BridgeNetworkService.swift
```diff
@@ -0,0 +1,488 @@
...
+            }
+            written += 1
+        }
```
nirs, line 255: Logging how many packets were dropped can help debug performance issues under load.

Sources/Services/Network/Server/BridgeNetworkService.swift
```diff
@@ -0,0 +1,488 @@
...
+        guard pktsOut == pktsIn else {
+            log.warning("could not forward all packets to sandbox: \(pktsIn) != \(pktsOut)")
+            return
+        }
+    }
+
+    private func relayFromSandboxToHost() {
+        // vmnet-helper uses Apple-private system calls sendmsg_x/recmsg_x to forward multiple packets
+        // in a single system call on the bridge... that sounds intuitively as it will likely achieve quite a
+        // performance improvement, but it also might break without further notice
```
nirs, line 276: This doubles the throughput in iperf3 and reduce cpu usage.

nirs (commented):
Adding network helper that handles the data path can work but the network helper is typically the control path - reserving the network.

curdbecker:
> Adding network helper that handles the data path can work but the network helper is typically the control path - reserving the network.

Yes, my first version actually had that in the runtime - I think then it was still called sandbox service. However, that is actually a bit counter-intuitive, since the network server then does not serve a real purpose. And that's also actually what I would expect from a network component that handles the bridging. And it came with a bit more subtle issues... Sorry for the vagueness. If you would like to have more specific details, I would have to dig through the reflog again :)

The current setup actually feels to me more correct - especially after the networking refactoring. I would exactly keep it like that - unless somebody seriously objects... Although it would be certainly nicer to work out the entitlement issues than working around them :)

In general my point is that having services that don't provide actual value in a certain context always feels like a waste of resources to me :) I also wouldn't run a DHCP server on a network that doesn't use dynamic address allocation. Especially if I never plan to use it.

Sources/Services/Network/Server/BridgeNetworkService.swift
curdbecker: Ah - interesting. Then I have to try that out. Rather strange then, right? I would have expected that there is some deeper sense here. Maybe also connected to the reason why the entitlement is so privileged? Preventing arbitrary spoofing?

Sources/Services/Network/Server/BridgeNetworkService.swift
curdbecker: Yep, I also read `vmnet.h`, but if I remember correctly, it actually makes then `vmnet_start_interface` actually fail. That's why I left in there in a comment in case I was doing something wrong - or there was something else needed.

Sources/Services/Network/Server/BridgeNetworkService.swift
curdbecker: Yes, right. I saw your implementation, but for me this is just fine. I actually have here a version that does that, but I then really had no need, since also in real networks sometimes packets fail... and then the protocols are supposed to handle this. Right now, the bridging network does RTP streaming and it works well enough for my current activities.

Sources/Services/Network/Server/BridgeNetworkService.swift
curdbecker: Yes, right, I think I also have that somewhere here when I suspected last week that I had issues with dropped packets in the bridge... Turned out that I built a network loop during testing with the Linux Wifi emulator in a containeri-zed environment when trying to bridge it also to the real VLAN. And then my Unifi switch at the desktop did not like me very much anymore... 😂

Sources/Services/Network/Server/BridgeNetworkService.swift
curdbecker: Yes, I believe that in a heartbeat - that's why I commented. I think I also saw that you did quite the performance testing. That was pretty cool :)

However, in the spirit of at least getting the PR in a shape that is upstream-able, I opted to stay away from undocumented APIs, since the API could change without further notice. And especially since I imagine that the native bridging support will again likely outperform it quite a bit.

nirs:
How per-container network helper lifetime is managed?

I hope that creating a "bridged" network is needed only once before you use it, and for every container a new helper connected to the same bridged network is started, and stopped when the container is stopped.

curdbecker:
Yep, exactly like that - starting a new container instance is then like plugging yet another cable into a dumb L2 switch :)

Sources/Services/Network/Server/BridgeNetworkService.swift
nirs: You need to push hard to see the failures - lightweight usage can work fine for weeks before it fails. It also depends on the socket receive buffer size - smaller buffer leads to quicker failures.

Try `iperf3 -c addr -P 8 --bidir`

PR #1630, opened 2026-06-03
[container]: add `container export` for live containers

### LINKED ISSUE #1400: [Request]: add --live flag to container export

Use this flag to export an OCI layout tar file from a running container

### Relates To

#1399

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Closes #1400

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -901,14 +901,22 @@ public actor ContainersService {
         self.log.debug("\(#function)")
 
         let state = try self._getContainerState(id: id)
-        guard state.snapshot.status == .stopped else {
-            throw ContainerizationError(.invalidState, message: "container is not stopped")
-        }
-
         let path = self.containerRoot.appendingPathComponent(id)
         let bundle = ContainerResource.Bundle(path: path)
         let rootfs = bundle.containerRootfsBlock
-        try EXT4.EXT4Reader(blockDevice: FilePath(rootfs)).export(archive: FilePath(archive))
+
+        switch state.snapshot.status {
+        case .running:
+            let client = try state.getClient()
+            let snapshot = rootfs.appendingPathExtension("snapshot")
+            defer { try? FileManager.default.removeItem(at: snapshot) }
+            try await client.snapshotDisk(imagePath: rootfs.path, destinationPath: snapshot.path)
+            try EXT4.EXT4Reader(blockDevice: FilePath(snapshot)).export(archive: FilePath(archive))
+        case .stopped:
+            try EXT4.EXT4Reader(blockDevice: FilePath(rootfs)).export(archive: FilePath(archive))
+        default:
+            throw ContainerizationError(.invalidState, message: "container must be running or stopped")
+        }
     }
 
     private func handleContainerExit(id: String, code: ExitStatus? = nil) async throws {
--- a/Sources/ContainerCommands/Container/ContainerExport.swift
+++ b/Sources/ContainerCommands/Container/ContainerExport.swift
@@ -67,7 +67,11 @@ extension Application {
                 }
                 try fileHandle.close()
             } else {
-                try FileManager.default.moveItem(at: archive, to: URL(fileURLWithPath: output!))
+                let outputURL = URL(fileURLWithPath: output!)
+                if FileManager.default.fileExists(atPath: outputURL.path(percentEncoded: false)) {
+                    try FileManager.default.removeItem(at: outputURL)
+                }
+                try FileManager.default.moveItem(at: archive, to: outputURL)
             }
         }
     }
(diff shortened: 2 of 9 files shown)


jglogan:
@saehejkang Let's tackle these two PRs right after next week's release.

Sources/APIServer/APIServer+Start.swift
```diff
@@ -297,6 +297,8 @@ extension APIServer {
             routes[XPCRoute.containerBootstrap] = XPCServer.route(harness.bootstrap)
             routes[XPCRoute.containerDial] = XPCServer.route(harness.dial)
             routes[XPCRoute.containerStop] = XPCServer.route(harness.stop)
+            routes[XPCRoute.containerFreeze] = XPCServer.route(harness.freeze)
```
jglogan: Do we need to expose freeze/thaw on the API server today?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -897,17 +959,40 @@ public actor ContainersService {
...
             throw ContainerizationError(.invalidState, message: "container is not stopped")
         }
 
         let path = self.containerRoot.appendingPathComponent(id)
         let bundle = ContainerResource.Bundle(path: path)
         let rootfs = bundle.containerRootfsBlock
+
+        if live {
+            let client = try state.getClient()
+            try await client.filesystemOperation(operation: .freeze, path: "/")
```
jglogan: It might be possible to make the PR quite a bit simpler and at the same time, make the amount of time that the container filesystem is frozen as short as possible.

Instead of exposing any freeze/thaw APIs, we could have something like this on the runtime API:

```swift
  // RuntimeRoutes
  static let snapshotDisk = "com.apple.container.runtime/snapshotDisk"

  // RuntimeKeys (reuses existing destinationPath key; adds imagePath)
  static let imagePath = "imagePath"

  // RuntimeClient
  public func snapshotDisk(imagePath: String, destinationPath: String) async throws
```

where `snapshotDisk` in RuntimeService does freeze/clonefile/thaw in a single call.

jglogan (commented):
@saehejkang see the suggestion for simplifying this change. What do you think?

Sources/APIServer/APIServer+Start.swift
saehejkang: I was just following current patterns but looks like we don't want this now?

saehejkang: Ahh I see the difference now between the `APIServer` and `RuntimeAPI`. Looks like I made the changes in the correct place and you mentioned that below.

jglogan:
Local test configuration, run `./test.sh` to test:

entrypoint.sh

```bash
#!/bin/sh
#
# Continuously writes random 256-byte files under /tmp/output.
#
set -eu

mkdir -p /tmp/output

while true; do
    tmpfile=$(mktemp /tmp/output/.tmp.XXXXXX)
    dd if=/dev/urandom of="$tmpfile" bs=1 count=256 2>/dev/null
    hash=$(sha256sum "$tmpfile" | awk '{print $1}')
    mv "$tmpfile" "/tmp/output/$hash"
done
```

Dockerfile:

```Dockerfile
FROM alpine:3.20

COPY entrypoint.sh /entrypoint.sh
RUN chmod +x /entrypoint.sh

ENTRYPOINT ["/entrypoint.sh"]
```

test.sh:

```
#!/bin/sh
#
# Build and run a container that performs continuous I/O.
# Use live export to obtain two different tar archives.
# Import these in Colima or other Docker-like and run
# to measure how many files were created.
#
set -euox pipefail

CONTAINER="${CONTAINER:-container}"
EXPORT_OPTS=${EXPORT_OPTS:-}
IMAGE=liveexport-test
NAME=liveexport-test-run

${CONTAINER} build -t "${IMAGE}" .
${CONTAINER} run -d --rm --name "${NAME}" "${IMAGE}"

trap '${CONTAINER} rm -f "${NAME}"' EXIT

sleep 2
${CONTAINER} export ${EXPORT_OPTS} -o "${NAME}-1.tar" "${NAME}"

sleep 2
${CONTAINER} export ${EXPORT_OPTS} -o "${NAME}-2.tar" "${NAME}"

docker import "${NAME}-1.tar" "${IMAGE}-1"
docker import "${NAME}-2.tar" "${IMAGE}-2"

docker run -it --rm "${IMAGE}-1" ls /tmp/output | wc -l
docker run -it --rm "${IMAGE}-2" ls /tmp/output | wc -l
```

jglogan (changes requested):
Test feedback, from exporting with `container` and `docker` (via Colima):

```bash
EXPORT_OPTS=--live ./test.sh
CONTAINER=docker ./test.sh
```

- Basic functionality looks good. With the `--live` flag, and for the first run, it works fine.
- Please remove the `--live` flag, and instead, in the RuntimeService, just use the is-running detection result to determine whether freeze/thaw is necessary. The user shouldn't have to concern themselves with the state of the container. The test should pass using `./test.sh` without `EXPORT_OPTS`.
- `container export` fails on back-to-back runs because it refuses to overwrite the output file (`Error: “archive.tar” couldn’t be moved to “20260801-container-liveexport” because an item with the same name already exists.`). The command should overwrite the previous file without consulting the user.

jglogan (approved):

PR #1636, opened 2026-06-03
Add docs on using container system configurations

### LINKED ISSUE #1635: [Request]: Docs for ContainerSystemConfig

Container recently added a system for setting default configurations for `container` CLI and services in https://github.com/apple/container/pull/1425 and https://github.com/apple/container/pull/1543. We should add documentation for how users are expected to use this system. 

### Code of Conduct

katiewasnothere:
## Type of Change
- [x] Documentation update

## Motivation and Context
Closes https://github.com/apple/container/issues/1635. Provide tutorial docs for users to customize default configuration settings.


README.md
```diff
@@ -76,7 +76,7 @@ To retain your user data so that it is available should you reinstall later, run
 
 ## Next steps
 
-- Take [a guided tour of `container`](./docs/tutorial.md) by building, running, and publishing a simple web server image.
+- Take [a guided tour of `container`](./docs/tutorials/tutorial.md) by building, running, and publishing a simple web server image.
```
jglogan: `tutorials/tutorial.md` isn't super descriptive.

perhaps `tutorials/start-here.md`?

docs/container-system-config.md
```diff
@@ -0,0 +1,114 @@
+# `config.toml` reference
+
+> [!IMPORTANT]
```
jglogan, line 3: It would be nicer if our configuration schema had enough metadata that we could get rid of the schema and table instead just say: run `container system property describe` or something like that

jglogan (changes requested):
Looks fine, just need to change `tutorial.md` to something that stands out a little more.

README.md
katiewasnothere: Updated!

jglogan (approved):

docs/container-system-config.md
```diff
@@ -0,0 +1,114 @@
...
+
+### MemorySize format
+
+Quoted string with a numeric prefix and a binary unit suffix. Parsing is case-insensitive; the suffix may be one of:
+
+| Suffix family       | Unit                     | Example values         |
+|---------------------|--------------------------|------------------------|
+| `b`                 | bytes                    | `"1024b"`              |
+| `k`, `kb`, `kib`    | kibibytes (1024 bytes)   | `"512k"`, `"512kb"`    |
+| `m`, `mb`, `mib`    | mebibytes (1024 KiB)     | `"2048mb"`, `"2g"`     |
```
0xMH, line 101: nit: `"2g"` is in the mebibytes row, but `g` is the gibibytes suffix. Probably belongs in the `g`/`gb`/`gib` row.

docs/container-system-config.md
```diff
@@ -0,0 +1,114 @@
...
+
+| Key      | Type     | Default      | Description                                                                                                  |
+|----------|----------|--------------|--------------------------------------------------------------------------------------------------------------|
+| `domain` | `String` | `"docker.io"` | Registry assumed when an image reference omits the registry host (e.g. `alpine` → `docker.io/library/alpine`). |
+
+## `[vminit]`
+
+| Key     | Type     | Default                                                | Description                                                                                              |
+|---------|----------|--------------------------------------------------------|----------------------------------------------------------------------------------------------------------|
+| `image` | `String` | `ghcr.io/apple/containerization/vminit:<tag>`     | Reference for the `vminitd` image used to boot container VMs. The tag segment is taken from the project's bundled `containerization` version.  |
```
0xMH, line 81: question: I might be missing something but looking at [`VminitConfig.defaultImage`](https://github.com/apple/container/blob/bf13a7caca6b370b94337aa65476a22cdf1df403/Sources/ContainerPersistence/ContainerSystemConfig.swift#L143-L148), it seems like when the version tag resolves to `latest` the default becomes the bare `vminit:latest` rather than `ghcr.io/apple/containerization/vminit:latest`. Is that actually the case and is it intended?

katiewasnothere: Hi, that is correct. The intention behind doing that is that `containerization` is still unstable so we depend on a specific version of it for the container project. As a result, we also want to use a specific version of the vminitd image so that we're matching the version of the containerization code that is used to build the container project and everything behaves as expected. 

HOWEVER, when a developer is developing changes in `containerization` that they want to test using `container`, they can edit the swift package specified for `containerization` (see [here](https://github.com/apple/container/blob/main/BUILDING.md#develop-using-a-local-copy-of-containerization)). When they do this, the version of `containerization` will fallback to "latest" (see [here](https://github.com/apple/container/blob/89fa98c55ad4ad015687eda8f8b2f660c5863602/Sources/CVersion/include/Version.h#L18)). Then the code you linked above will select the locally build vminitd image with potential changes you've made in your local `containerization` instead of an official one.

jglogan (approved):

docs/container-system-config.md
0xMH: Thanks for the explanation, that clears it up.

PR #1660, opened 2026-06-07
Exclude application data directory from backups

### LINKED ISSUE #1554: [Request]: Exclude application data from backups via `isExcludedFromBackupKey`

The application data at `~/Library/Application Support/com.apple.container/` holds large sparse files whose size exceeds actual allocation, causing backup tools to fail. Consider setting `URLResourceKey.isExcludedFromBackupKey = true` on the application data folder to resolve this.

### Code of Conduct

somiljain2006:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
[Why is this change needed?]

Fixes #1554 

The application data directory (~/Library/Application Support/com.apple.container) can contain large, sparse files whose apparent size is significantly larger than their allocated size. This may cause backup tools to process unnecessary data or fail when backing up the container application data.

This change marks the application root directory with URLResourceKey.isExcludedFromBackupKey, preventing it from being included in backups while preserving normal container functionality.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Automated Testing:
- Added test coverage for backup exclusion behavior.
- Verified all existing tests continue to pass.

Manual Verification:
1. Built locally with make.
2. Started the API server using ./bin/container system start.
3. Verified ~/Library/Application Support/com.apple.container was created.
4. Verified isExcludedFromBackupKey == true.
5. Confirmed normal startup completed successfully.


somiljain2006:
@jglogan @katiewasnothere @realrajaryan  Can you review this pr?

Tests/ContainerAPIServiceTests/RuntimeConfigurationTests.swift
```diff
@@ -133,3 +133,24 @@ struct RuntimeConfigurationTests {
         #expect(decodedData.variant == "test-variant", "Variant should round-trip through RuntimeConfiguration")
     }
 }
+
+struct BackupExclusionTests {
+    @Test
+    func testExcludeFromBackupMarksDirectory() throws {
```
dkovba: This test verifies the `Foundation` framework. Consider testing the actual code in container where we set `resourceValues.isExcludedFromBackup` to `true`.

somiljain2006:
@dkovba CI failure appears unrelated to this change.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -72,6 +72,19 @@ public actor ContainersService {
         log: Logger,
         debugHelpers: Bool = false
     ) throws {
+        try FileManager.default.createDirectory(at: appRoot, withIntermediateDirectories: true)
```
dkovba: This `FileManager.default.createDirectory()` call seems to be redundant.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -72,6 +72,19 @@ public actor ContainersService {
...
+        do {
+            var mutableAppRoot = appRoot
+            var resourceValues = URLResourceValues()
+            resourceValues.isExcludedFromBackup = true
+            try mutableAppRoot.setResourceValues(resourceValues)
```
dkovba: Consider doing this where we actually create the app root.

somiljain2006:
@dkovba, I have addressed the changes. Can you review them

Package.swift
```diff
@@ -210,6 +210,11 @@ let package = Package(
                 "ContainerResource",
                 "ContainerRuntimeLinuxClient",
                 "ContainerRuntimeClient",
+                "ContainerAPIService",
+                "ContainerPlugin",
+                "ContainerPersistence",
+                "DNSServer",
```
dkovba: Is `DNSServer` actually used?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -72,8 +72,12 @@ public actor ContainersService {
...
         debugHelpers: Bool = false
     ) throws {
+        try Self.createAppRoot(at: appRoot, log: log)
         let containerRoot = appRoot.appendingPathComponent("containers")
-        try FileManager.default.createDirectory(at: containerRoot, withIntermediateDirectories: true)
+        try FileManager.default.createDirectory(
+            at: containerRoot,
+            withIntermediateDirectories: true
+        )
         self.exitMonitor = ExitMonitor(log: log)
```
dkovba, line 77: Not an issue, but this change wasn't needed.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -1147,6 +1151,31 @@ public actor ContainersService {
             return (config, runtimeConfig.options)
         }
     }
+
+    private static func createAppRoot(at appRoot: URL, log: Logger) throws {
+        try FileManager.default.createDirectory(
+            at: appRoot,
+            withIntermediateDirectories: true
+        )
```
dkovba: It seems when this code is executed, we already have the app root directory. Would `Sources/ContainerPlugin/ApplicationRoot.swift` be a better place for setting `isExcludedFromBackup = true`?

Tests/ContainerAPIServiceTests/ContainersServiceTests.swift
```diff
@@ -0,0 +1,68 @@
...
+import ContainerPersistence
+import ContainerPlugin
+import Foundation
+import Logging
+import Testing
+
+struct ContainersServiceTests {
+
+    @Test
+    func testAppRootIsExcludedFromBackups() async throws {
```
dkovba: If we move `isExcludedFromBackup = true` to `Sources/ContainerPlugin/ApplicationRoot.swift`, we can simplify the test.

Tests/ContainerAPIServiceTests/ContainersServiceTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        let pluginLoader = try PluginLoader(
+            appRoot: appRoot,
+            installRoot: installRoot,
+            logRoot: nil,
+            pluginDirectories: [],
+            pluginFactories: [],
+            log: logger
+        )
+
+        let systemConfig = try await ConfigurationLoader.load()
```
dkovba: Should we avoid using `ConfigurationLoader.load()` in tests?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
somiljain2006: My concern with moving this into ApplicationRoot is that it currently seems to be a path-resolution utility used throughout the codebase. Adding filesystem writes there may introduce regression. I put the logic in ContainersService because that's where the application data directory is used and created.

dkovba: This is a valid concern. We could create the app root in a new method in `ApplicationRoot` (e.g., `ensureCreated()`) or in any different place that executes first.

Package.swift
```diff
@@ -210,6 +210,10 @@ let package = Package(
                 "ContainerResource",
                 "ContainerRuntimeLinuxClient",
                 "ContainerRuntimeClient",
+                "ContainerAPIService",
+                "ContainerPlugin",
+                "ContainerPersistence",
+                .product(name: "Logging", package: "swift-log"),
```
dkovba: It seems none of these dependencies are used in `ContainerAPIServiceTests`. Should it be `ContainerPluginTests`?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -72,6 +72,7 @@ public actor ContainersService {
         log: Logger,
         debugHelpers: Bool = false
     ) throws {
+        try ApplicationRoot.ensureCreated(at: appRoot, log: log)
```
dkovba: Would `APIServer.run()` be a better place for this call?

dkovba (commented):
Thank you for the contribution 👍

dkovba (approved):

Tests/ContainerPluginTests/RootPathTests.swift
```diff
@@ -27,6 +29,31 @@ struct ApplicationRootTests {
...
+        try ApplicationRoot.ensureCreated(at: appRoot, log: logger)
+
+        var isDirectory: ObjCBool = false
+        let exists = FileManager.default.fileExists(atPath: appRoot.path, isDirectory: &isDirectory)
+        #expect(exists && isDirectory.boolValue, "appRoot should be created")
+
+        let readBack = try appRoot.resourceValues(forKeys: [.isExcludedFromBackupKey])
+        #expect(
+            readBack.isExcludedFromBackup == true,
+            "ApplicationRoot.ensureCreated should explicitly exclude appRoot from Time Machine backups."
```
dkovba: It might be reasonable to remove "Time Machine" since the same flag applies to other backup products.

dkovba (approved):

somiljain2006:
@dkovba Can this be merged?

Sources/ContainerPlugin/ApplicationRoot.swift
```diff
@@ -45,4 +46,29 @@ public struct ApplicationRoot {
 
     /// The pathname to the root directory
     public static let pathname = path.string
+
+    /// Explicitly creates the application data root directory and excludes it from backups.
+    public static func ensureCreated(at appRoot: URL, log: Logger) throws {
```
dkovba: The function name doesn't mention that any folder that is passed to it will be excluded from backup. The external parameter name doesn't make it clear that it accepts only the app root.

Sources/ContainerPlugin/ApplicationRoot.swift
```diff
@@ -45,4 +46,29 @@ public struct ApplicationRoot {
...
+        try FileManager.default.createDirectory(
+            at: appRoot,
+            withIntermediateDirectories: true
+        )
+
+        do {
```
dkovba: Can we exclude the folder from backup only when it's first created, so the user could change that?

Sources/ContainerPlugin/ApplicationRoot.swift
```diff
@@ -45,4 +46,29 @@ public struct ApplicationRoot {
 
     /// The pathname to the root directory
     public static let pathname = path.string
+
+    /// Explicitly creates the application data root directory and excludes it from backups.
```
dkovba: Would it be better not to exclude the whole application folder from backup? For example, we likely don't want to exclude volumes.

dkovba (changes requested):
Sorry for the delay. Please consider additional feedback.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -73,7 +73,28 @@ public actor ContainersService {
         debugHelpers: Bool = false
     ) throws {
         let containerRoot = appRoot.appendingPathComponent("containers")
+        let isFirstCreation = !FileManager.default.fileExists(atPath: containerRoot.path)
```
dkovba: If I remember this correctly, we already discussed this. When this code is executed, the application folder may have already been created, and therefore, we can't exclude it from backup here. Is this true?

somiljain2006: Sorry, It’s been quite a while since then, so I had forgotten about that.

Sources/ContainerPlugin/ApplicationRoot.swift
```diff
@@ -45,4 +46,12 @@ public struct ApplicationRoot {
 
     /// The pathname to the root directory
     public static let pathname = path.string
+
+    /// Creates the application data root directory.
+    public static func ensureCreated(at appRoot: URL) throws {
```
dkovba: There is no need for this simple wrapper.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -73,7 +73,26 @@ public actor ContainersService {
         debugHelpers: Bool = false
     ) throws {
         let containerRoot = appRoot.appendingPathComponent("containers")
```
dkovba, line 75: Consider excluding from backup:
- `snapshots`
- `containers`
- `Virtual Machines.vzvmlibrary`

dkovba (changes requested):

dkovba (changes requested):
Users should be able to change which folders are excluded from backup, so we shouldn't overwrite that choice on every start. Set the exclusion only when a folder is first created.

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
somiljain2006: I couldn't find any code in this repository that creates Virtual Machines.vzvmlibrary, so I'll leave that to the underlying virtualization layer.

Sources/APIServer/APIServer+Start.swift
```diff
@@ -59,6 +59,8 @@ extension APIServer {
             }
 
             do {
+                let appRootURL = URL(fileURLWithPath: appRoot.string)
+                try FileManager.default.createDirectory(at: appRootURL, withIntermediateDirectories: true)
```
dkovba, line 63: Is this still needed?

Tests/ContainerPluginTests/RootPathTests.swift
```diff
@@ -14,6 +14,8 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
+import Foundation
+import Logging
```
dkovba, line 18: Are we supposed to have a test here?

Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
```diff
@@ -73,7 +73,26 @@ public actor ContainersService {
...
+
         try FileManager.default.createDirectory(at: containerRoot, withIntermediateDirectories: true)
+
+        do {
```
dkovba, line 79: It seems this hasn't been addressed: https://github.com/apple/container/pull/1660#pullrequestreview-5096385095

dkovba (changes requested):

PR #1661, opened 2026-06-08
Add bridged physical networking via `vmnet-helper`

s3rj1k:
Add bridged physical networking via vmnet-helper

A `container-network-vmnet-helper` plugin bridges container machines
onto a physical interface with static IPv4 and an address pool.
Machines attach via `container machine set network=<spec>` and set
extra kernel command-line args with `kernel-arg=`. Bridged networks
are machine-only; a regular container requesting one is rejected.

## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
I, as a user, would like to have bridge support in machine containers, this change adds bridge support via external [vmnet-helper](https://github.com/nirs/vmnet-helper) and ability to set kernel args for custom kernels

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
+++ b/Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
@@ -1,0 +1,317 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerResource
+import ContainerRuntimeClient
+import ContainerXPC
+import Containerization
+import ContainerizationError
+import ContainerizationExtras
+import CryptoKit
+import Foundation
+import Logging
+
+/// Creates network interfaces for bridged networks by launching a
+/// vmnet-helper process per interface.
+///
+/// vmnet-helper (https://github.com/nirs/vmnet-helper) attaches to the
+/// physical host interface in vmnet bridged mode and relays Ethernet
+/// frames over a `SOCK_DGRAM` socketpair. The strategy spawns the helper,
+/// reads the interface parameters it reports on stdout, and hands the VM
+/// end of the socketpair to the virtual machine. The MAC address comes
+/// from vmnet, not from the caller.
+public struct VmnetHelperInterfaceStrategy: InterfaceStrategy {
+    /// Default helper locations for Homebrew on macOS 26 and newer,
+    /// then the install script location for macOS 15 and earlier.
+    private static let defaultHelperPaths = [
+        "/opt/homebrew/opt/vmnet-helper/libexec/vmnet-helper",
+        "/usr/local/opt/vmnet-helper/libexec/vmnet-helper",
+        "/opt/vmnet-helper/bin/vmnet-helper",
+    ]
+
+    private static let startupTimeout = Duration.seconds(15)
+    // The Virtualization framework requires SO_RCVBUF of at least twice
+    // SO_SNDBUF on the attachment socket and recommends four times.
+    private static let socketSendBufferSize: Int32 = 1024 * 1024
+    private static let socketReceiveBufferSize: Int32 = 4 * 1024 * 1024
+
+    private let log: Logger
+
+    public init(log: Logger) {
+        self.log = log
+    }
+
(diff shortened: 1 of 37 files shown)


Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
```diff
@@ -0,0 +1,290 @@
...
+        var fds: [Int32] = [-1, -1]
+        guard socketpair(AF_UNIX, SOCK_DGRAM, 0, &fds) == 0 else {
+            throw ContainerizationError(.internalError, message: "socketpair failed with errno \(errno)")
+        }
+        let vmFd = fds[0]
+        let helperFd = fds[1]
+        for fd in fds {
+            var sendSize = Self.socketSendBufferSize
+            var receiveSize = Self.socketReceiveBufferSize
+            _ = setsockopt(fd, SOL_SOCKET, SO_SNDBUF, &sendSize, socklen_t(MemoryLayout<Int32>.size))
```
servusdei2018: We're ignoring the return values on these `setsockopt` calls (`_ = setsockopt(...)`). If a system has a restrictive buffer ceiling set via sysctl (`kern.ipc.maxsockbuf`), the OS will truncate or reject these 1MB/4MB sizes. If that happens, the Virtualization framework might start throwing silent packet drops under high throughput. We should probably log a warning or check the return code here instead of discarding it completely.

Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
```diff
@@ -0,0 +1,290 @@
...
+            }
+            scanned = buffer.count
+
+            let remaining = deadline - ContinuousClock.now
+            guard remaining > .zero else {
+                throw ContainerizationError(.internalError, message: "timed out waiting for vmnet-helper to start")
+            }
+
+            var pollFd = pollfd(fd: fd, events: Int16(POLLIN), revents: 0)
+            let timeoutMs = Int32(min(remaining / .milliseconds(1), 1000))
```
servusdei2018: Although this loop runs on a background GCD thread, `poll` can still lock up the thread for up to a full second per iteration. If the runtime tries to fire a cancel or shutdown signal, it won't be seen until the poll timeout drops, which will lag our teardown. Dropping the timeout value or passing a cancellation token down would make this a lot more responsive.

servusdei2018 (changes requested):

Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
s3rj1k: Hi, thanks for the review, I've rebased on top of recent changes and hopefully solved both highlighted issues.

Sources/Services/RuntimeLinux/Server/VmnetHelperInterfaceStrategy.swift
s3rj1k: ditto

s3rj1k:
Does not look like there is an upstream interest in this, closing PR

nirs:
@jglogan this looks like an interesting integration. Can we someone review this?

s3rj1k:
@nirs Thanks for the interest in this and for the network helper 😃 

I am trying to follow the latest releases of `container` in main branch of forked repo.

https://github.com/s3rj1k/container/releases/tag/v1.1.0%2B5973b9c
https://s3rj1k.github.io/container/

---

Regarding this feature in general, for my use-case I need to have a VM that I can run easily with bridge and shard networks and where I can run wireguard in kernel mode and nested virtualization, essentially replacing UTM.

For `container` specifically this means, using `machine` mode, building custom OS image, rebuilding Linux kernel (for wireguard and nested virtualization), and supporting this feature across releases.

So I am actually starting to think at this point that taking `libkrun` and conjuring some simple CLI wrapper with exactly limited amount of features that I need from VMM will be more usable in terms of support in the long run.

nirs:
> @nirs Thanks for the interest in this and for the network helper 😃
> 
> I am trying to follow the latest releases of `container` in main branch of forked repo.
> 
> https://github.com/s3rj1k/container/releases/tag/v1.1.0%2B5973b9c https://s3rj1k.github.io/container/
> 
> Regarding this feature in general, for my use-case I need to have a VM that I can run easily with bridge and shard networks and where I can run wireguard in kernel mode and nested virtualization, essentially replacing UTM.
> 
> For `container` specifically this means, using `machine` mode, building custom OS image, rebuilding Linux kernel (for wireguard and nested virtualization), and supporting this feature across releases.
> 
> So I am actually starting to think at this point that taking `libkrun` and conjuring some simple CLI wrapper with exactly limited amount of features that I need from VMM will be more usable in terms of support in the long run.

You can use krunkit (using libkrun) or vfkit (using virtualization framework). Both support nested virtualization and have good integration with vmnet-helper. You can check examples in [vmnet-helper testing harness](https://github.com/nirs/vmnet-helper/blob/6076be89236965b0751605585062c09c7d622d09/testing/vm.py#L159) and minikube [vfkit](https://github.com/kubernetes/minikube/blob/master/pkg/drivers/vfkit/vfkit.go) and [krunkit](https://github.com/kubernetes/minikube/blob/master/pkg/drivers/krunkit/krunkit.go) drivers. krunkit supports also GPU hardware acceleration.

Integrating with Apple container looks interesting and useful for the community.

PR #1694, opened 2026-06-11
docs: clarify home directory paths in container-machine.md

### LINKED ISSUE #1678: [Bug]: Container machine docs disagree on the home directory path
The new container machine docs describe the home directory in two different ways.

### What I found

In `docs/container-machine.md`, the first section says:

```text
Your repo lives in $HOME on macOS and is mounted at /Users/<username> inside the container machine.
```

But the quickstart right below it says:

```bash
container machine run -n dev pwd          # /home/<you> — your Mac home dir, mounted in
```

Those two paths point users in different directions:

- `/Users/<username>` sounds like the macOS path is preserved inside the Linux machine
- `/home/<username>` sounds like a normal Linux home directory

### What the code seems to do

The implementation and tests point to `/home/<username>` as the machine user's home.

`Sources/Services/MachineAPIService/Client/MachineConfiguration.swift`:

```swift
public var home: String {
    "/home/\(username)"
}
```

The CLI test also checks for that exact path:

```swift
#expect(home == "/home/\(hostUsername)", "HOME should be set to /home/<username>")
```

### Expected

The docs should use one clear story for this.

Something like:

```text
Your macOS home is shared into the container machine and exposed as the Linux user's home at /home/<username>.
```

That keeps the docs aligned with the code and avoids users looking for their files under `/Users/<username>` inside the machine.

### Why this matters

Container machines are meant to feel like a persistent Linux dev environment. The first thing users will check is `pwd`, `$HOME`, and where their repo is mounted. If the docs mention both `/Users/<username>` and `/home/<username>`, it is easy to think the mount is broken when it is actually just the doc wording.

Otoru:
## Summary

- Fixes contradictory descriptions of the home directory paths in `docs/container-machine.md`
- The doc said the macOS home is mounted at `/Users/<username>` in one place, then referred to `/home/<you>` as "your Mac home dir, mounted in" — which contradicts itself
- `/home/<username>` is the persistent Linux home; the macOS home is what is mounted at `/Users/<username>` (preserving the host path)

Fixes #1678

## Test plan

- [ ] Read through the updated doc and verify the two paths are no longer contradictory

### DIFF
--- a/docs/container-machine.md
+++ b/docs/container-machine.md
@@ -5,19 +5,19 @@ Container machine provides a highly integrated Linux environment that works seam
 ## Why container machines
 
 Containers are typically modeled after an application. A container machine is modeled after a Linux environment. It runs the image's init system allowing you to register long running services or test your application under a process supervisor.
-A container machine automatically maps your username and home directory into the Linux environment. Your repositories and dotfiles are available on both platforms. Use editors and tools directly on macOS simultaneously building and running your application inside of the Linux environment.
+A container machine automatically maps your username into the Linux environment and mounts your macOS home directory at `/Users/<username>` inside the machine. Your repositories are accessible from both platforms at the same path. Use editors and tools directly on macOS simultaneously building and running your application inside of the Linux environment.
 
-- **Edit on the Mac, build inside.** Your repo lives in `$HOME` on macOS and is mounted at `/Users/<username>` inside the container machine. Use your macOS editor or IDE; compile and run inside your container machine.
+- **Edit on the Mac, build inside.** Your macOS home directory is mounted at `/Users/<username>` inside the container machine (preserving the host path). Use your macOS editor or IDE; compile and run inside your container machine.
 - **Use macOS-native tooling against Linux artifacts.** Profilers, screenshot tools, browsers, and GUI debuggers on your Mac all see the same files the container machine sees — there is no copy step between "I built it" and "I am inspecting it".
 - **Real Linux services for testing.** Run a database or whatever your stack needs as a system service — `systemctl start postgresql` works on images with `systemd` installed.
-- **One environment per target distro.** Create as many container machines as you have target distros — `alpine`, `ubuntu`, `debian`. Each has the same `$HOME` and the same dotfiles from your Mac. Quickly test your application in various distributions.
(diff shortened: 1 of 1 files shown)


docs/container-machine.md
```diff
@@ -5,19 +5,19 @@ Container machine provides a highly integrated Linux environment that works seam
...
-- **One environment per target distro.** Create as many container machines as you have target distros — `alpine`, `ubuntu`, `debian`. Each has the same `$HOME` and the same dotfiles from your Mac. Quickly test your application in various distributions.
+- **One environment per target distro.** Create as many container machines as you have target distros — `alpine`, `ubuntu`, `debian`. Each shares the same macOS home mount at `/Users/<username>`. Quickly test your application in various distributions.
 
 ## Quickstart
 
 ```bash
 container machine create alpine:latest --name dev
 container machine run -n dev whoami       # your host username, not root
-container machine run -n dev pwd          # /home/<you> — your Mac home dir, mounted in
+container machine run -n dev pwd          # /home/<you> — your persistent Linux home directory
```
afbjorklund: This `pwd` is unlikely to _ever_ print /home, since by default it will use the $PWD **from the host.**

```
$ coma pwd
/Users/afb
```

However, if you do a `cd` and `pwd` or look at $HOME inside the guest - then you will see /home.

```
$ coma echo '$HOME'
/home/afb
```

docs/container-machine.md
```diff
@@ -5,19 +5,19 @@ Container machine provides a highly integrated Linux environment that works seam
...
 container machine run -n dev              # interactive shell; cd into your repos in $HOME
```
afbjorklund: Note that you have to use $HOME in the initial command, once inside the guest - $HOME changes.

docs/container-machine.md
Otoru: Good catch, thanks! Updated the example — `pwd` now shows `/Users/<you>` (inherited from host `$PWD`) and added an `echo '$HOME'` line to show the Linux home at `/home/<you>`.

docs/container-machine.md
Otoru: Noted — updated the interactive shell comment to make that explicit.

PR #1699, opened 2026-06-11
Fix: race conditions in DefaultNetworkService.lookup by removing async from AttachmentAllocator.lookup

GarySkywalker-droid:
## Type of Change
- [*] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Removed the async from AttachmentAllocator.lookup, causing the race conditions because the deallocator being called while lookup is running as it overrides the actor's reentrancy protection. Leading to at times mac address being returned as nil and fixes the #1693 

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/Network/Server/AttachmentAllocator.swift
+++ b/Sources/Services/Network/Server/AttachmentAllocator.swift
@@ -1,4 +1,4 @@
-/===----------------------------------------------------------------------===//
+//===----------------------------------------------------------------------===//
 // Copyright © 2026 Apple Inc. and the container project authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


klood21 (approved):

katiewasnothere:
Hi @GarySkywalker-droid could you describe the race condition this change fixes a bit more? I'm not clear where the issue is

kiwigitops (commented):
I think this still needs an await at the call site. Even if lookup(hostname:) is no longer declared async, AttachmentAllocator is an actor, and this call is still crossing from DefaultNetworkService to a different actor instance. Swift treats that as an asynchronous actor-isolated access, so try allocator.lookup(hostname:) should not compile as written.

That also means this change may not remove the suspension point the PR is trying to avoid. The race probably needs a different shape, such as performing the hostname lookup inside the allocator operation that owns the related state transition, or otherwise avoiding a separate cross-actor read.

PR #1708, opened 2026-06-12
docs: add [machine] section to config.toml reference

Jonathan2567:
## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
The `[machine]` section was already implemented in `ContainerSystemConfig.swift` and consumed by `MachineCreate.swift` as the baseline configuration for `container machine create`, defining default CPU count, memory allocation, and home directory mount mode for new machines. However, it was missing from the `config.toml` reference documentation. This change brings the documentation in line with the source code and provides users with a complete reference of available configuration sections.

Source: `Sources/ContainerPersistence/MachineConfig.swift`

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/docs/container-system-config.md
+++ b/docs/container-system-config.md
@@ -16,6 +16,7 @@ Source of truth: [`Sources/ContainerPersistence/ContainerSystemConfig.swift`](..
 [container]  # default per-container resources
 [dns]        # default DNS domain for DNS resolution on host
 [kernel]     # guest kernel binary path and download URL
+[machine]    # default per-machine resources and home mount
 [network]    # default subnets for new networks
 [registry]   # default registry domain
 [vminit]     # default vminitd image to use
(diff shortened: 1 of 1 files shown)


docs/container-system-config.md
```diff
@@ -59,6 +60,17 @@ Guest kernel used when launching container VMs. Defaults change per release as k
 | `binaryPath` | `String` | `"opt/kata/share/kata-containers/vmlinux-6.18.15-186"`                                                 | Path **inside** the downloaded kernel archive that points to the kernel binary. |
 | `url`        | `URL`    | `"https://github.com/kata-containers/kata-containers/releases/download/3.28.0/kata-static-3.28.0-arm64.tar.zst"` | Archive to download when no kernel is installed. Encoded and decoded as a plain string in TOML. |
 
+## `[machine]`
+
+Defaults applied when `container machine create` is invoked without `--cpus`, `--memory`, or `--home-mount`.
+Does not affect existing machines — use `container machine set` to change running machines.
```
katiewasnothere: I think this statement is a bit confusing. If a user changed the defaults in the MachineConfig, they would need to restart the container service, which would bring down all running `machines`. Then when those `machines` are started again, the new defaults would take effect. 

If a user changes a specific `machine`'s config with `container machine set`, the `machine` needs to be stopped and restarted.

Jonathan05643: Thanks for the review!

You're right that the previous wording was confusing. I've updated it to clarify that the machine must be stopped and restarted for container machine set changes to take effect.

Regarding changes to config.toml defaults requiring a daemon restart — that behavior applies to all config sections and is already covered in the [Container system config tutorial](https://github.com/apple/container/blob/main/docs/tutorials/container-system-config-tutorial.md#restart-the-container-service), so I kept the section focused on the per-machine behavior.

katiewasnothere (approved):

PR #1715, opened 2026-06-13
docs: clarify container machine vs container semantics

### LINKED ISSUE #1700: [Request]: strange doc comments about machines versus containers

The `container machine` documentations talks about filesystems and init:

"each machine keeps its filesystem and runs the image's own init system"

But a container _also_ keeps its filesystem, and it can use the `--init` flag?

So there is no major difference between the two, it's more of a convenience...

The real difference is in the user setup and in the default volume mounts:

"Host integrations such as automatic user and home directory sharing"

----

Talking about `container machine` versus `container` here, not about containers.

Otherwise it would be more about reusing the old kernel, vs starting a new one...

A container machine (with user and home) is usually referred to as a "toolbox".

And you can run those on top of `container machine` (or `container`), if needed:

* Toolbx <https://containertoolbx.org/>

* Distrobox <https://distrobox.it/>

### Code of Conduct

Bortlesboat:
## Summary
- clarify that container machines use standard OCI images to provide long-lived Linux environments
- explain that both regular containers and container machines persist changes until deletion
- frame the distinction around reusable environment workflows and host user/home integration

## Verification
- `git diff --check`
- reviewed the updated `docs/container-machine.md` wording against issue #1700 and maintainer comments

Closes #1700

### DIFF
--- a/docs/container-machine.md
+++ b/docs/container-machine.md
@@ -4,8 +4,8 @@ Container machine provides a highly integrated Linux environment that works seam
 
 ## Why container machines
 
-Containers are typically modeled after an application. A container machine is modeled after a Linux environment. It runs the image's init system allowing you to register long running services or test your application under a process supervisor.
-A container machine automatically maps your username and home directory into the Linux environment. Your repositories and dotfiles are available on both platforms. Use editors and tools directly on macOS simultaneously building and running your application inside of the Linux environment.
+Containers are typically modeled after an application or task. A container machine is a convenience wrapper around a Linux environment: it pairs a standard OCI image with dedicated persistent root storage, boots the image's init system, and adds host integrations such as automatic user and home directory sharing.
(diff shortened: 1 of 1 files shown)


afbjorklund:
> emphasize the real differentiators: dedicated persistent root storage, init/system-service support,

I don't think the first two are real differentiators, but the new wording makes the intent clear enough.

afbjorklund:
> a container machine is a convenience wrapper around a Linux environment, not a wholly separate runtime model

Not necessarily true, currently `container ls` and `container machine ls` _are_ totally separated runtime models...

afbjorklund (approved):

Bortlesboat:
Rechecked against current main: this remains mergeable, the diff is still only two clarified documentation lines, and the prior approval is on the current head commit. Re-surfacing for maintainer review when convenient.

Issue #1721, opened 2026-06-14
[Bug]:XPC connection invalid when the user home directory is on an external volume; internal --app-root works


jincaiw:
### I have done the following

- [x] I have searched the existing issues
- [ ] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

1. Use a macOS account whose home directory is stored on an external volume:

echo "$HOME"

Example output:

/Volumes/ex-data/<user>

2. Start Apple Container normally:

container system start

3. The command reports that it is launching container-apiserver, but the health check fails.
4. Check the system status:

container system status

Current behavior

Starting the service produces:

Launching container-apiserver...
Testing access to container-apiserver...
Error: internalError: "failed to get a response from apiserver: interrupted: "XPC connection error: Connection invalid""
Ensure container system service has been started with `container system start`.

The status command then reports:

apiserver is not running and not registered with launchd

Stopping the service and terminating possible stale processes did not resolve the problem:

container system stop 2>/dev/null || true
pkill -f container-apiserver 2>/dev/null || true
pkill -f container-network-vmnet 2>/dev/null || true
pkill -f container-core-images 2>/dev/null || true
sleep 2
container system start --timeout 60
container system status

The same XPC error occurred.

Workaround that resolved the issue

I created an application data directory on the internal system disk and passed it through --app-root:

container system stop 2>/dev/null || true
pkill -f container-apiserver 2>/dev/null || true
pkill -f container-network-vmnet 2>/dev/null || true
pkill -f container-core-images 2>/dev/null || true
APP_ROOT="/Users/Shared/apple-container-${USER}"
mkdir -p "$APP_ROOT"
chmod 700 "$APP_ROOT"
CONTAINER_DEBUG=1 container system start \
  --app-root "$APP_ROOT" \
  --timeout 60

The API server then started successfully:

Launching container-apiserver...
Testing access to container-apiserver...
Verifying machine API server is running...
No default kernel configured.
Install the recommended default kernel from [https://github.com/kata-containers/kata-containers/releases/download/3.28.0/kata-static-3.28.0-arm64.tar.zst]? [Y/n]: y
Installing kernel...

After the kernel installation, container system status returned:

FIELD              VALUE
status             running
appRoot            /Users/Shared/apple-container-<user>/
installRoot        /usr/local/
logRoot
apiserver.version  container-apiserver version 1.0.0 (build: release, commit: ee848e3)
[...]

PR #1724, opened 2026-06-14
Fix XPC connection failure when home directory is on external volume

majidfeiz:
## Problem

When a user's home directory resides on an external or removable volume
(e.g. `/Volumes/ex-data/`), running `container system start` results in an
"XPC connection invalid" error and the daemon never becomes reachable.

Reported in: #1721

**Root cause:** `launchctl bootstrap <domain> <plist-path>` silently fails
when the plist file is located on an external or removable volume. macOS
(Gatekeeper / AMFI) does not allow Mach service registration from plists
that reside outside the internal system volume. Because `ApplicationRoot`
defaults to `~/.local/share/container`, any user whose home directory is on
an external volume ends up writing `apiserver.plist` and plugin plists to
that volume — causing all subsequent XPC connections to fail.

## Solution

Introduce `ServiceRoot`: a new root type whose sole responsibility is to
provide a location for launchd plist files. Its default path is:

```
<NSTemporaryDirectory()>/com.apple.container/
```

`NSTemporaryDirectory()` (i.e. `FileManager.default.temporaryDirectory`) is
always located on the internal system volume on macOS, regardless of where
the user's home directory lives. This guarantees that `launchctl bootstrap`
succeeds in all configurations.

`ServiceRoot` follows the exact same pattern as the existing `ApplicationRoot`,
`InstallRoot`, and `LogRoot` types:

| Type | Env var | Default |
|---|---|---|
| `ApplicationRoot` | `CONTAINER_APP_ROOT` | `~/.local/share/container` |
| `InstallRoot` | `CONTAINER_INSTALL_ROOT` | `<binary>/../../` |
| `LogRoot` | `CONTAINER_LOG_ROOT` | *(macOS log facility)* |
| `ServiceRoot` | `CONTAINER_SERVICE_ROOT` | `$TMPDIR/com.apple.container` |

`CONTAINER_SERVICE_ROOT` is propagated from `SystemStart` into the
`container-apiserver` plist environment, so the daemon's `PluginLoader`
automatically picks it up for plugin plist registration as well — no
changes are required at any other call site.

## Changes

- **`Sources/ContainerPlugin/ServiceRoot.swift`** *(new)*
  Defines `ServiceRoot` with `environmentName`, `defaultPath`, `path`, and
  `pathname` statics, consistent with the other root types.

- **`Sources/ContainerCommands/System/SystemStart.swift`**
  - Adds `--service-root` CLI option (overrides `CONTAINER_SERVICE_ROOT`).
  - Propagates `CONTAINER_SERVICE_ROOT` into the plist environment.
  - Writes `apiserver.plist` under `serviceRoot` instead of `appRoot`.

- **`Sources/ContainerPlugin/PluginLoader.swift`**
  - Separates `plistResourceRoot` (system volume, for plists) from
    `pluginResourceRoot` (appRoot, for runtime state).
  - Adds `serviceRoot: URL? = nil` to `init` — existing callers are
    unaffected; `nil` falls back to reading `CONTAINER_SERVICE_ROOT`.

- **`Tests/ContainerPluginTests/RootPathTests.swift`**
  Adds `ServiceRootTests` (3 tests): default path is absolute, ends with
  `com.apple.container`, and is located under the temporary directory.

- **`Tests/ContainerPluginTests/PluginLoaderTest.swift`**
  Updates `testRegisterWithLaunchdPlistNotUnderAppRoot` to pass an explicit
  `serviceRoot` distinct from `appRoot`, directly asserting the separation.

## Testing

All existing tests pass:

```
swift test --filter ContainerPluginTests
# Test run with 48 tests in 9 suites passed
```

Manual verification path (for reviewers with an external-volume home):
1. `container system start` — previously failed with XPC connection invalid.
2. After this fix, the daemon registers successfully and all XPC calls work.

To override the service root manually:

```bash
container system start --service-root /private/tmp/my-container-service
# or via env var:
CONTAINER_SERVICE_ROOT=/private/tmp/my-container-service container system start
```

### DIFF
--- a/Sources/ContainerCommands/System/SystemStart.swift
+++ b/Sources/ContainerCommands/System/SystemStart.swift
@@ -54,7 +54,7 @@ extension Application {
             name: .long,
             help: "Path to the root directory for launchd service registration files; must reside on an internal volume",
             transform: { FilePath(FileManager.default.currentDirectoryPath).resolve($0, defaultPath: FilePath($0)) })
-        var serviceRoot = ServiceRoot.defaultPath
+        var serviceRoot = ServiceRoot.path
 
         @Flag(
             name: .long,
@@ -136,7 +136,8 @@ extension Application {
             // CONTAINER_SERVICE_ROOT is forwarded into the plist environment so the
             // daemon's PluginLoader writes plugin plists to the same root without
             // requiring any additional wiring at other call sites.
-            let plistDir = serviceRoot
+            let plistDir =
+                serviceRoot
                 .appending(FilePath.Component("apiserver"))
             let plistDirURL = URL(fileURLWithPath: plistDir.string)
             try FileManager.default.createDirectory(at: plistDirURL, withIntermediateDirectories: true)
--- a/Sources/ContainerCommands/System/SystemStart.swift
+++ b/Sources/ContainerCommands/System/SystemStart.swift
@@ -50,6 +50,12 @@ extension Application {
             transform: { FilePath(FileManager.default.currentDirectoryPath).resolve($0, defaultPath: FilePath($0)) })
         var logRoot: FilePath? = nil
 
+        @Option(
+            name: .long,
+            help: "Path to the root directory for launchd service registration files; must reside on an internal volume",
+            transform: { FilePath(FileManager.default.currentDirectoryPath).resolve($0, defaultPath: FilePath($0)) })
+        var serviceRoot = ServiceRoot.defaultPath
+
         @Flag(
             name: .long,
             inversion: .prefixedEnableDisable,
@@ -109,6 +115,7 @@ extension Application {
             var env = PluginLoader.filterEnvironment()
             env[ApplicationRoot.environmentName] = appRoot.string
             env[InstallRoot.environmentName] = installRoot.string
+            env[ServiceRoot.environmentName] = serviceRoot.string
             if let logRoot {
                 env[LogRoot.environmentName] = logRoot.string
             }
@@ -121,8 +128,16 @@ extension Application {
                 machServices: ["com.apple.container.apiserver"]
             )
 
-            let plistPath = apiServerDataPath.appending(FilePath.Component("apiserver.plist"))
-            let plistURL = URL(fileURLWithPath: plistPath.string)
+            // Write the plist under serviceRoot, which must be on an internal system
+            // volume. launchd cannot register Mach services from plists on external or
+            // removable volumes, so serviceRoot is kept separate from appRoot.
+            // CONTAINER_SERVICE_ROOT is propagated to the daemon via the plist
+            // environment so that plugin registration also uses this root.
+            let plistDir = serviceRoot
+                .appending(FilePath.Component("apiserver"))
+            let plistDirURL = URL(fileURLWithPath: plistDir.string)
+            try FileManager.default.createDirectory(at: plistDirURL, withIntermediateDirectories: true)
+            let plistURL = plistDirURL.appendingPathComponent("apiserver.plist")
             let data = try plist.encode()
             try data.write(to: plistURL)
 
--- a/Sources/ContainerCommands/System/SystemStart.swift
+++ b/Sources/ContainerCommands/System/SystemStart.swift
@@ -128,11 +128,14 @@ extension Application {
                 machServices: ["com.apple.container.apiserver"]
             )
 
-            // Write the plist under serviceRoot, which must be on an internal system
-            // volume. launchd cannot register Mach services from plists on external or
-            // removable volumes, so serviceRoot is kept separate from appRoot.
-            // CONTAINER_SERVICE_ROOT is propagated to the daemon via the plist
-            // environment so that plugin registration also uses this root.
+            // Write the plist under serviceRoot, intentionally separate from appRoot.
+            // launchd cannot register Mach services from plists on external or removable
+            // volumes; keeping service files on the internal system volume (serviceRoot)
+            // while application data lives in appRoot preserves that guarantee even when
+            // the user's home directory is on an external volume.
+            // CONTAINER_SERVICE_ROOT is forwarded into the plist environment so the
+            // daemon's PluginLoader writes plugin plists to the same root without
+            // requiring any additional wiring at other call sites.
             let plistDir = serviceRoot
                 .appending(FilePath.Component("apiserver"))
             let plistDirURL = URL(fileURLWithPath: plistDir.string)
--- a/Sources/ContainerPlugin/PluginLoader.swift
+++ b/Sources/ContainerPlugin/PluginLoader.swift
@@ -34,16 +34,16 @@ public struct PluginLoader: Sendable {
 
     public typealias PluginQualifier = ((Plugin) -> Bool)
 
-    // A path on disk managed by the PluginLoader, where it stores
-    // runtime data for loaded plugins. This includes log files.
-    private let pluginResourceRoot: URL
-
     // A path on the system (internal) volume where launchd plists are written.
     // Derived from serviceRoot, which is always on an internal volume, so that
     // plist registration succeeds even when appRoot is on an external or
     // removable volume — launchd cannot bootstrap Mach services from plists on
     // such volumes.
     private let plistResourceRoot: URL
+
+    // The resolved service root propagated to child services so nested plugin
+    // loaders use the same internal-volume location for their launchd plists.
+    private let serviceRoot: URL
 
     public init(
         appRoot: URL,
@@ -54,10 +54,8 @@ public struct PluginLoader: Sendable {
         log: Logger? = nil,
         serviceRoot: URL? = nil
     ) throws {
-        let pluginResourceRoot = appRoot.appendingPathComponent("plugin-state")
-        try FileManager.default.createDirectory(at: pluginResourceRoot, withIntermediateDirectories: true)
-        self.pluginResourceRoot = pluginResourceRoot
         let resolvedServiceRoot = serviceRoot ?? URL(fileURLWithPath: ServiceRoot.pathname, isDirectory: true)
+        self.serviceRoot = resolvedServiceRoot
         let plistResourceRoot = resolvedServiceRoot.appendingPathComponent("plugin-state", isDirectory: true)
         try FileManager.default.createDirectory(at: plistResourceRoot, withIntermediateDirectories: true)
         self.plistResourceRoot = plistResourceRoot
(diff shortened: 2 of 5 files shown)


kasc0206 (approved):
## 审查意见 / Review

This is an excellent contribution. The root cause analysis is thorough, and the solution is clean and well-isolated.

### 优点 / Strengths

1. **Root cause analysis**: Correctly identifies that `launchctl bootstrap` silently fails on external volumes due to Gatekeeper/AMFI restrictions
2. **Solution design**: Introducing `ServiceRoot` as a new root type follows the existing `ApplicationRoot`/`InstallRoot`/`LogRoot` pattern perfectly — consistency with the codebase is excellent
3. **Minimal changes**: Only 5 files changed, with clear separation of concerns
4. **Test coverage**: Both unit tests (`RootPathTests`, `PluginLoaderTest`) properly validate the behavior
5. **Backward compatibility**: `serviceRoot: URL? = nil` defaults mean existing callers are unaffected
6. **Documentation**: The doc comments on `ServiceRoot` clearly explain *why* the separation exists

### 小建议 / Nitpicks

- Consider adding a fallback or warning when `NSTemporaryDirectory()` itself returns nil or an unexpected path, though this is extremely rare on macOS
- The `plistDir` construction `serviceRoot.appending("apiserver")` vs the `PluginLoader` using `plugin-state` subdirectory is intentional for audit separation — consider noting this in a comment

### 已验证 / Verified
- Code compiles with `swift build` ✅
- All existing `ContainerPluginTests` pass ✅
- No breaking API changes to `PluginLoader.init` (new parameter is optional with default nil)

Great work!

kasc0206:
## Code Review Summary / 代码审查摘要

### Strengths / 优点

1. **Accurate root cause**: Correctly identifies that `launchctl bootstrap` silently fails when the plist resides on external/removable volumes due to macOS Gatekeeper/AMFI restrictions
   **准确的根因分析**：正确识别出 launchctl bootstrap 在外部卷上静默失败的问题
2. **Elegant design**: Introducing `ServiceRoot` as a new root type follows the existing `ApplicationRoot`/`InstallRoot`/`LogRoot` pattern — consistent with codebase conventions
   **优雅的设计**：新增 ServiceRoot 类型遵循了已有的 root 类型模式
3. **Minimal footprint**: Only 5 files changed, clear separation of concerns
   **最小改动**：仅修改 5 个文件，职责清晰
4. **Proper test coverage**: Unit tests in `RootPathTests` and `PluginLoaderTest` validate the behavior
   **充分的测试覆盖**：单元测试验证了行为正确性
5. **Backward compatible**: `serviceRoot: URL? = nil` default means all existing callers work unchanged
   **向后兼容**：可选参数默认为 nil，不影响现有调用方

### Nitpicks / 小建议

- Consider adding a fallback or warning when `NSTemporaryDirectory()` returns nil or an unexpected path (rare on macOS but would cause cryptic failures)
  建议考虑 NSTemporaryDirectory() 返回空值时的降级或警告
- The directory structure (`serviceRoot/apiserver/` for the API server plist vs `serviceRoot/plugin-state/` for plugin plists) is intentional — consider adding a brief comment explaining this audit separation
  建议添加注释说明 apiserver/ 和 plugin-state/ 子目录的职责区分

### Verification / 验证
- Code compiles with `swift build` ✅
- All existing `ContainerPluginTests` pass ✅
- No breaking API changes to `PluginLoader.init`

Great contribution! / 优秀的贡献！

majidfeiz:
Hi @kasc0206,

Thank you for the thorough review and the kind words — really appreciate you taking the time to go through the changes carefully.

I've addressed both of your suggestions:

**1. Defensive check for `NSTemporaryDirectory()`**
Converted `ServiceRoot.defaultPath` from a plain `static let` to a closure-based initializer that includes a `precondition` guard. If `NSTemporaryDirectory()` ever returns an empty path, the process fails at startup with a clear message rather than silently constructing a broken path.

**2. Clearer comment around `plistDir` construction**
Expanded the comment in `SystemStart` to make the intentional `appRoot`/`serviceRoot` separation explicit — noting both why the split exists (external-volume launchd restriction) and how `CONTAINER_SERVICE_ROOT` propagates to the daemon without requiring changes at other call sites.

Both changes are in the latest commit. Let me know if anything else needs adjustment.

kasc0206:
Thanks for the quick follow-up @majidfeiz! The changes look great:

1. ✅ **Defensive precondition** for `NSTemporaryDirectory()` — clean and explicit
2. ✅ **Improved comment** in `SystemStart` — clearly explains the separation rationale
3. ✅ **Test enhancement** — verifying plist is under `serviceRoot` AND NOT under `appRoot` is exactly the right assertion

LGTM! The PR is ready to merge. 🚀

Lucky2755:
Thank you for working on this. I tested this PR with `ApplicationRoot` on an external SSD and found one remaining case where launchd plists are still written under the external application root.

The `ServiceRoot` approach correctly fixes registration for `container-apiserver` and for plugins that do not provide a custom state directory. However, `PluginLoader.registerWithLaunchd` currently contains:

```swift
let rootURL =
    pluginStateRoot ??
    self.plistResourceRoot.appending(path: plugin.name)
```

This means that `plistResourceRoot` is only used when `pluginStateRoot` is `nil`.

Some important callers do provide `pluginStateRoot`:

- `NetworksService.registerService` passes the network entity path and an `instanceId`, such as `default`.
- `ContainersService.registerService` passes the container bundle path and the container ID.

When `ApplicationRoot` is on an external volume, those paths are also on the external volume. As a result, the plist for the default network is still written outside `ServiceRoot`.

In my test, the following plists were created correctly under the internal service root:

```text
$CONTAINER_SERVICE_ROOT/apiserver/apiserver.plist
$CONTAINER_SERVICE_ROOT/plugin-state/machine-apiserver/service.plist
$CONTAINER_SERVICE_ROOT/plugin-state/container-core-images/service.plist
```

However, the plist for the instanced network service was missing from the service root. The API server then attempted to connect to:

```text
com.apple.container.network.container-network-vmnet.default
```

and the XPC lookup failed because that launchd service had not been registered.

The runtime state and the launchd registration file need to be treated as two independent locations:

```text
pluginStateRoot
    Runtime state, network data and container bundles.
    May remain under ApplicationRoot on the external volume.

plistResourceRoot
    launchd plist files only.
    Must always remain under ServiceRoot on the internal volume.
```

A possible fix is to always derive the plist directory from `plistResourceRoot`, while using `instanceId` to prevent different network or container instances from overwriting one another:

```swift
var rootURL = self.plistResourceRoot
    .appendingPathComponent(plugin.name, isDirectory: true)

if let instanceId {
    rootURL = rootURL.appendingPathComponent(
        instanceId,
        isDirectory: true
    )
}
```

This produces paths such as:

```text
$CONTAINER_SERVICE_ROOT/plugin-state/
    container-network-vmnet/default/service.plist

$CONTAINER_SERVICE_ROOT/plugin-state/
    container-runtime-linux/<container-id>/service.plist
```

`pluginStateRoot` can remain in the public method signature for compatibility, but it should no longer determine where `service.plist` is written. It could eventually be deprecated or removed, since the affected callers already pass their runtime state paths to the plugin through command-line arguments.

I also made two related adjustments locally:

1. Store the resolved `serviceRoot` in `PluginLoader` and explicitly propagate it to every plugin plist:

```swift
env[ServiceRoot.environmentName] =
    serviceRoot.path(percentEncoded: false)
```

This keeps the child service environment consistent even when `PluginLoader` is initialized with an explicit `serviceRoot`.

2. Initialize the `--service-root` option with `ServiceRoot.path` rather than `ServiceRoot.defaultPath`:

```swift
var serviceRoot = ServiceRoot.path
```

Otherwise, a user-provided `CONTAINER_SERVICE_ROOT` can be replaced by the default before the API server plist is generated.

For regression coverage, the important missing test case is a registration that supplies all three of these:

```swift
pluginStateRoot: externalStateRoot
instanceId: "default"
serviceRoot: internalServiceRoot
```

The test should verify that:

```text
internalServiceRoot/plugin-state/<plugin>/default/service.plist
```

exists, while:

```text
externalStateRoot/service.plist
```

does not exist.

A second instance ID should also be registered to verify that the two plist files do not overwrite each other.

I applied these changes locally on top of the PR branch, built the installer package, installed it, and tested the following scenario:

```text
ApplicationRoot: external SSD
ServiceRoot:     internal per-user temporary directory
Architecture:    arm64
```

Manual verification succeeded:

- `container system start` completed successfully.
- `container system status` reported the service as running.
- The default `container-network-vmnet` plist was created under the internal service root.
- Container networking worked.
- An OCI image for a systemd-based Kali Linux environment was built.
- A persistent `container machine` was created from that image and started successfully.
- Network access and package installation inside the Kali machine worked.
- The machine remained usable after stopping and starting it again.
- Images, machine data and runtime state remained under the external application root.

So the overall `ServiceRoot` design appears correct; the remaining issue is the precedence currently given to `pluginStateRoot` for instanced plugin plist placement.

I can provide the complete patch and regression test if that would be useful.

majidfeiz:
Hi @Lucky2755,

Thank you for the detailed investigation, the external-volume testing, and the clear explanation of the remaining instanced-plugin case. You were correct that `pluginStateRoot` was still taking precedence over `plistResourceRoot`, which allowed network and container plugin plists to be written under `ApplicationRoot`.

I addressed the feedback in commit `7952989`:

1. Plugin launchd plists are now always written under `plistResourceRoot`.
2. When an `instanceId` is provided, it is included in the plist path to prevent registrations for different instances from overwriting each other:

   ```text
   $CONTAINER_SERVICE_ROOT/plugin-state/<plugin-name>/<instance-id>/service.plist
   ```

3. `pluginStateRoot` remains in the public method signature for source compatibility, but it no longer controls launchd plist placement. Runtime state can therefore remain under `ApplicationRoot`, including when it is located on an external volume.
4. `PluginLoader` now stores the resolved service root and propagates `CONTAINER_SERVICE_ROOT` to every child plugin service.
5. The `--service-root` option now initializes from `ServiceRoot.path`, preserving a user-provided `CONTAINER_SERVICE_ROOT` value.
6. Regression coverage now registers two instances while supplying a separate runtime state root. The test verifies that both plists are created under the service root, no plist is written under the runtime state root, and the resolved service root is present in the child environment.

Validation completed successfully:

```text
swift test --filter PluginLoaderTest.testRegisterWithLaunchdInstancedPlistsStayUnderServiceRoot
# 1 test passed

swift test --filter ContainerPluginTests
# 48 tests in 9 suites passed
```

The project builds successfully, and the modified files pass the repository's strict Swift formatting checks.

Thanks again for reproducing this on an external SSD and for providing such actionable feedback.

Lucky2755 (approved):
Thank you for the quick turnaround and for addressing all the points so thoroughly. The updated plist placement, service-root propagation, and regression coverage look great.

I really appreciate the clear explanation and the work you put into resolving this.

PR #1731, opened 2026-06-14
Remove lost+found from named volume EXT4 images

### LINKED ISSUE #1730: [Bug]: Named volumes contain lost+found, breaking database image initialization (Postgres, MySQL, etc.)
## I have done the following

## Steps to reproduce

```bash
# 1. Create a named volume and confirm lost+found is present
container run --rm -v pgdata:/mnt alpine ls /mnt
# Output: lost+found

# 2. Run Postgres — initdb finds the non-empty directory and fails
container run --rm \
  --name pg-test \
  -e POSTGRES_PASSWORD=secret \
  -v pgdata:/var/lib/postgresql/data \
  postgres:17
```

## Problem description

Named volumes are formatted as EXT4, which always creates a `lost+found` directory at the root. Postgres (and other database images) treat a non-empty volume root as a fatal error at `initdb` time:

```
initdb: error: directory "/var/lib/postgresql/data" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
initdb: hint: Using a mount point directly as the data directory is not recommended.
Create a subdirectory under the mount point.
```

The same failure affects MySQL, Elasticsearch, MongoDB, and any image whose entrypoint checks for an empty data directory on first use.

Docker Desktop, Colima, and OrbStack all present fresh named volumes as empty directories. `lost+found` is not required for EXT4 to function — `e2fsck` recreates it automatically if a consistency check is ever run. The fix is a single `formatter.unlink(path: FilePath("/lost+found"))` call in `VolumesService.createVolumeImage`, before `formatter.close()`.

## Environment

- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)

## Code of Conduct

sylvaincombes:
## What

Remove `lost+found` from freshly created named volume EXT4 images by calling
`formatter.unlink(path: FilePath("/lost+found"))` before `formatter.close()` in
`VolumesService.createVolumeImage`.

## Why

The `EXT4.Formatter` in `containerization` correctly creates `lost+found` for
general-purpose e2fsck compatibility. Named container volumes, however, are never
checked with e2fsck — and many widely-used database images treat a non-empty
volume root as a fatal error at initialization time:

```
initdb: error: directory "/var/lib/postgresql/data" exists but is not empty
initdb: detail: It contains a lost+found directory, perhaps due to it being a mount point.
```

Affected images include Postgres, MySQL, Elasticsearch, MongoDB, and any image
whose entrypoint runs `initdb`-style initialization against the raw volume mount point.

Docker Desktop, Colima, and OrbStack all present fresh named volumes as empty
directories. This fix restores that parity and unblocks a broad class of
real-world use cases.

## Safety

`lost+found` is not required for EXT4 to function. If e2fsck is ever run on a
volume, it recreates `lost+found` automatically when it finds it missing. No data
loss, no corruption risk. Only newly created volumes are affected; existing
volume images are unchanged.

The fix is at the right semantic layer: `VolumesService` is where the decision
that "this EXT4 image is for a container named volume" belongs. The
`containerization` library comment ("required for e2fsck to pass") remains
correct for its own scope.

## Tests

Two new tests in `Tests/ContainerResourceTests/VolumeImageTests.swift`:

- `formatterCreatesLostAndFoundByDefault` — baseline confirming the default
  formatter behaviour that this fix deliberately overrides
- `volumeImageHasNoLostAndFound` — confirms the root directory is empty after
  the unlink, using `EXT4.EXT4Reader.listDirectory`

Fixes #1730

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -414,6 +414,7 @@ let package = Package(
             name: "ContainerResourceTests",
             dependencies: [
                 .product(name: "Containerization", package: "containerization"),
+                .product(name: "ContainerizationEXT4", package: "containerization"),
                 .product(name: "ContainerizationExtras", package: "containerization"),
                 "ContainerAPIService",
                 "ContainerResource",
--- a/Sources/Services/ContainerAPIService/Server/Volumes/VolumesService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Volumes/VolumesService.swift
@@ -288,16 +288,17 @@ public actor VolumesService {
     }
 
     private func createVolumeImage(for name: String, sizeInBytes: UInt64 = VolumeStorage.defaultVolumeSizeBytes, journal: EXT4.JournalConfig? = nil) throws {
-        let blockPath = blockPath(for: name)
+        try Self.formatVolumeImage(at: FilePath(blockPath(for: name)), size: sizeInBytes, journal: journal)
+    }
 
-        // Use the containerization library's EXT4 formatter
-        let formatter = try EXT4.Formatter(
-            FilePath(blockPath),
-            blockSize: 4096,
-            minDiskSize: sizeInBytes,
-            journal: journal
-        )
-
+    /// Formats a new EXT4 image at `path` and removes `lost+found`.
+    ///
+    /// Named container volumes are never checked with e2fsck, and many database
+    /// images (Postgres, MySQL, etc.) refuse to initialize in a non-empty directory.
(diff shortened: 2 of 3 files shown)


jglogan:
Thank you for the contribution!

We don't want to apply blanket destructive behavior like this on all volumes; see comments on #1730.

PR #1735, opened 2026-06-15
add ID validation to daemon service

0xf00sec:
The CLI validates container IDs via Utility.validEntityName before sending XPC requests, but the daemon never checked the ID itself. Any process talking directly to com.apple.container.apiserver could pass a path-traversal string (../../tmp/foo) as a container ID, causing the daemon to create, measure, or export bundles outside of containerRoot.

I've Added Utility.validEntityName checks in ContainersService at the entry point of create, containerDiskUsage, and exportRootfs -- the three functions that build filesystem paths from the container ID. The daemon now enforces the same naming rules as the CLI regardless of how the XPC service is reached.


realrajaryan:
Hey @0xf00sec, thank you for the contribution! Could you please add some tests as well

0xf00sec:
Hey @realrajaryan  I've added two tests to UtilityTests.swift one for valid names, one for traversal sequences. All 14 tests pass.

PR #1740, opened 2026-06-17
Add sticky container IPs across stop/start

vsarunas:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Currently upon a container stop/start, the IP address will always be incremented, even if there is free IP addresses available. This can sometimes cause issues with other software running inside the container where the IPs are always changing and require reconfiguration/handling.

The PR adds simple best attempt to restore the previous IP addresses allocated if the network still has available IPs, essentially giving static IPs under normal use. Upon network IP exhaustion, a starting container takes an IP from a stopped container.

### Sticky

```
container run -d --name stick alpine sleep infinity
```

Will show the same IP persisted:
```
container list
ID     IMAGE                            OS     ARCH   STATE    IP               CPUS  MEMORY   STARTED
stick  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.3/24  4     1024 MB  2026-06-17T11:24:31Z

container stop stick && container start stick

container list
ID     IMAGE                            OS     ARCH   STATE    IP               CPUS  MEMORY   STARTED
stick  docker.io/library/alpine:latest  linux  arm64  running  192.168.64.3/24  4     1024 MB  2026-06-17T11:24:51Z
```

Same can be done with start order, IP will be preserved:
```bash
for n in s1 s2 s3; do container run -d --name $n alpine sleep infinity; done
for n in s1 s2 s3; do container stop $n; done
for n in s3 s1 s2; do container start $n; done
```

### Pool exhaustion and re-claim

Add a small network:
```
container network create --subnet 192.168.210.0/29 tiny
```

4 containers will be created:
```bash
for i in 1 2 3 4 5 6 7; do
  container run -d --name t$i --network tiny -m 256m alpine sleep infinity
done
```

Cannot allocate more once exhausted:
```
Error: failed to bootstrap container (cause: "internalError: "failed to bootstrap container t5 (cause: "invalidState: "network tiny (192.168.210.0/29) has no free IP addresses"")"")
```

The IP can be re-allocated from a stopped container on exhaustion and a new container can be now started:
```
container stop t1
container run -d --name t5 --network tiny -m 256m alpine sleep infinity
```

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs


vsarunas:
@jglogan, could you have a look?

I originally started with a mix of this and static allocation similar to https://github.com/apple/container/pull/751, but this has various edge cases on verifying the IP requested against stopped and running containers, but this gives majority of the benefit.

vsarunas:
@jglogan - what do you think ?

florescercadojulio:
Since this is a best-effort strategy, would it make sense to notify the user when the previous IP cannot be reused and a new one has been assigned? Otherwise, users may assume the IP is stable and inadvertently build a fragile setup around that assumption.

Also, is there a common workflow in this project where users are expected to rely on container IPs instead of service discovery? If DNS-based service discovery is the recommended approach, I’m curious what use cases motivated adding sticky IP behavior.

vsarunas:
There is a debug log of when the IP address changes, I can change it from debug to a warning level.

There are certain places that you need to use an IP address only instead of DNS name (Kafka broker IP addresses list for example).

Yes, using DNS as well, it hasn't been without problems it self and sometimes have to flush cache straight after creation/restart:
```
sudo dscacheutil -flushcache; sudo killall -HUP mDNSResponder
```

vsarunas:
MAC address is now also preserved.

vsarunas:
@jglogan could you have a look at this PR please?

PR #1742, opened 2026-06-17
add container machine nested virt

crosbymichael:
This also includes custom kernels for container machine. Its required with nested virt as CONFIG_KVM needs to be enabled.

### DIFF
--- a/Sources/ContainerCommands/Machine/MachineCapabilities.swift
+++ b/Sources/ContainerCommands/Machine/MachineCapabilities.swift
@@ -1,0 +1,70 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerizationError
+import Foundation
+import Virtualization
+
+/// Pre-flight checks for container machine settings that depend on host capabilities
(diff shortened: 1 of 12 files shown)


Sources/ContainerCommands/Machine/MachineCapabilities.swift
```diff
@@ -0,0 +1,70 @@
...
+            throw ContainerizationError(
+                .unsupported,
+                message: "nested virtualization is not supported on this host (requires Apple Silicon M3+ and macOS 15+)"
+            )
+        }
+    }
+
+    /// Resolves the user-supplied kernel path to absolute and confirms it points to a
+    /// readable, non-empty regular file. Returns the absolute path.
+    static func validateKernelPath(_ path: String) throws -> String {
```
jglogan: Use SystemPackage.FilePath for paths, except in cases where path really can be a network URL.

jglogan (changes requested):

Sources/Services/MachineAPIService/Server/MachinesService.swift
```diff
@@ -363,13 +363,28 @@ public actor MachinesService {
...
             )
 
             config.resources.cpus = bootConfig.cpus
             config.resources.cpuOverhead = 0
             config.resources.memoryInBytes = bootConfig.memory.toUInt64(unit: .bytes)
 
-            let kernel = try await ClientKernel.getDefaultKernel(for: .current)
+            let kernel: Kernel
+            if let kernelPath = bootConfig.kernelPath {
+                guard FileManager.default.isReadableFile(atPath: kernelPath.string) else {
```
yibozhuang: should we use `MachineCapabilities.validateKernelPath`  to perform the validation here?

Sources/ContainerPersistence/MachineConfig.swift
```diff
@@ -146,4 +199,17 @@ extension MachineConfig {
         }
         return opt
     }
+
+    /// Parse a boolean setting accepting only "true" or "false".
+    private static func parseBool(_ value: String, for key: String) throws -> Bool {
```
yibozhuang, line 204: we already have a `parseBool` defined here: https://github.com/apple/container/blob/main/Sources/Services/ContainerAPIService/Client/Parser.swift#L1043 

could we reuse this?

jglogan (approved):

yibozhuang (approved):

PR #1744, opened 2026-06-17
Allow custom kernel boot args via --kernel-arg

arirubinstein:
## Type of Change
- [ ] Bug fix
- [X] New feature  
- [ ] Breaking change
- [ ] Documentation update

## What

Adds a repeatable `--kernel-arg` flag to `container run`/`container create` for appending arbitrary boot arguments to the kernel command line.

```
container run --kernel /path/to/custom-bzImage \
  --kernel-arg lsm=lockdown,capability,landlock,yama,apparmor,bpf \
  ...
```

## Why

The runtime hardcodes `lsm=lockdown,capability,landlock,yama,apparmor` (and `oops=panic`) onto every kernel command line in `RuntimeService.bootstrap`. With a custom kernel there is no way to adjust this — e.g. to enable **BPF LSM** you need `lsm=...,bpf`, which is currently impossible. More generally there is no escape hatch for any boot-time kernel argument.

## How

- New `--kernel-arg <arg>` option (repeatable) on `Flags.Management`.
- `Utility.getKernel` appends the user args onto `kernel.commandLine.kernelArgs`, which is persisted into the container bundle.
- `RuntimeService.bootstrap` now applies its built-in defaults **per-key**, skipping any default whose key the user already supplied. Defaults are expressed as a small keyed table, so this also lets `oops=` be overridden and makes future defaults easy to add.

Default behavior is unchanged for anyone who does not pass `--kernel-arg` — the same `oops=panic` and `lsm=...` args are applied.

## Testing

- `swift build` clean.
- `container run --help` renders the new flag.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -176,6 +176,7 @@ public struct Flags {
             entrypoint: String?,
             initImage: String?,
             kernel: String?,
+            kernelArgs: [String],
             labels: [String],
             mounts: [String],
             name: String?,
@@ -205,6 +206,7 @@ public struct Flags {
             self.entrypoint = entrypoint
             self.initImage = initImage
             self.kernel = kernel
+            self.kernelArgs = kernelArgs
             self.labels = labels
             self.mounts = mounts
             self.name = name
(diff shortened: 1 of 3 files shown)


noah-thor:
Can you please sign the commit: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Actual code change looks good to me

arirubinstein:
Signed and rebased

noah-thor (approved):

crosbymichael (approved):

PR #1741, opened 2026-06-17
fix(cp): resolve relative host paths against current directory

adityabagchi24:
Fixes #1738

`container cp` fails when the host source path is relative (e.g. `container cp file foo:/root/`), because `NSString.standardizingPath` only canonicalizes paths but does not make them absolute. The unchanged relative path is then interpreted as `/file` (root-absolute) by `URL(fileURLWithPath:)` on the runtime side.

Fixed by resolving relative paths against the current working directory before use, matching the pattern already used by `container export`, `container image save`, and `container image load`.

The same fix was also applied to the copy-out destination path (line 68), which had the same issue.

## Type of Change

- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

`container cp file foo:/root/` fails with `"copyIn: source not found '/file'"` because the relative path `file` is never expanded to an absolute path. Using `$PWD/file` works, but relative paths should work too — every other command in the codebase handles this correctly.

## Testing

- [x] Tested locally — builds and all existing tests pass
- [ ] Added/updated tests
- [ ] Added/updated docs

PR #1749, opened 2026-06-18
Resolve relative host paths in `container cp`

### LINKED ISSUE #1738: [Bug]: Container copy doesn't expand host path correctly.
### I have done the following

### Steps to reproduce

```
container run -d --rm --name foo alpine:latest sleep 1d
touch file
container cp file foo:/root/
```

This fails with
```
Error: failed to copy into container foo (cause: "internalError: "failed to copy into container foo (cause: "notFound: "copyIn: source not found '/file'"")"")
```

Following succeeds
```
container cp $PWD/file foo:/root/
```

### Problem description

Container cp fails to expand host path.

### Environment

```markdown
ProductName:		macOS
ProductVersion:		26.2
BuildVersion:		25C56
ReleaseType:		Internal
```

### Code of Conduct

uiharu-kazari:
## Summary

`container cp <relative-path> <id>:<dest>` failed with "source not found" because the local path was passed to the API server unmodified. The server runs with a working directory of `/`, so a relative path such as `file` was resolved to `/file` instead of the user's intended file. Only absolute host paths worked.

This resolves local paths to absolute, standardized paths before handing them to the API server: it expands a leading tilde and joins relative paths to the caller's current working directory. Applied to both the copy-in source and the copy-out destination so the two directions behave consistently.

## Fixes
Fixes #1738

## Testing
- Full `make all` build passes; `swift format lint --strict` clean.
- Verified the resolution logic: `file`, `./sub/x`, `../x` now resolve to absolute paths against the cwd, while absolute paths (`/etc/hosts`) and tilde paths (`~/z`) are preserved; the previous code left relative paths relative.

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerCopy.swift
+++ b/Sources/ContainerCommands/Container/ContainerCopy.swift
@@ -29,6 +29,21 @@ extension Application {
             case container(id: String, path: String)
         }
 
+        /// Resolve a local filesystem path to an absolute, standardized path.
+        ///
+        /// Expands a leading tilde and resolves relative paths against the current
+        /// working directory before standardizing. Without this, a relative path such
+        /// as `file` is forwarded verbatim to the API server, which resolves it
+        /// against its own working directory (`/`) and fails to find the source.
+        static func resolveLocalPath(_ path: String) -> String {
+            let expanded = (path as NSString).expandingTildeInPath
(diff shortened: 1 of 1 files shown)


JaewonHur:
Hi! Thanks for the PR, but we already have a working for this issue: #1741. Sorry! will close now!

PR #1762, opened 2026-06-19
[container]: add commit command

### LINKED ISSUE #1399: [Request]: container commit command

Add support for the `container commit` command to create a new container image from the current state of a container. 

### Relates To

#1400

### Code of Conduct

saehejkang:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Closes #1399

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerCommands/Image/OCIImageArchive.swift
+++ b/Sources/ContainerCommands/Image/OCIImageArchive.swift
@@ -1,0 +1,103 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ContainerResource
+import ContainerizationArchive
+import ContainerizationOCI
+import CryptoKit
+import Foundation
+
+enum OCIImageArchive {
+    static func create(
+        from rootfsArchive: URL,
+        in temporaryDirectory: URL,
+        reference: String,
+        container: ContainerSnapshot,
+        sourceImage: Image
+    ) throws -> URL {
+        let platform = container.configuration.platform
+        let process = container.configuration.initProcess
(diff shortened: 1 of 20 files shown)


jglogan:
@saehejkang Build's failing for this one, looks like a simple fix:

```console
/Users/runner/actions-runner/_work/container/container/Sources/ContainerCommands/Image/ImageLoad.swift:83:73: error: cannot find type 'Logger' in scope
 81 |         }
 82 | 
 83 |         static func load(from path: FilePath, force: Bool = false, log: Logger) async throws {
    |                                                                         `- error: cannot find type 'Logger' in scope
 84 |             let progressConfig = try ProgressConfig(
 85 |                 showTasks: true,
[3208/3318] Compiling ContainerCommands ContainerStop.swift
```

saehejkang:
> Build's failing for this one, looks like a simple fix:

Fixed. Also this PR is almost ready for a full review, but I need to go through some things before I am confident in the changes I made.

Sources/ContainerCommands/Image/OCIImageArchive.swift
```diff
@@ -0,0 +1,103 @@
...
+            workingDir: process.workingDirectory,
+            labels: labels,
+            stopSignal: container.configuration.stopSignal
+        )
+
+        let layoutDirectory = temporaryDirectory.appendingPathComponent("oci-layout")
+        let blobsDirectory = layoutDirectory.appendingPathComponent("blobs/sha256")
+        try FileManager.default.createDirectory(at: blobsDirectory, withIntermediateDirectories: true)
+
+        let layerDescriptor = try writeBlob(fromFile: rootfsArchive, mediaType: MediaTypes.imageLayer, into: blobsDirectory)
```
katiewasnothere, line 48: Is there a way we can reuse some other code instead of recreating an image layout here?

katiewasnothere: Would a commit API call on apiserver be a good option? Thoughts?

saehejkang: > Is there a way we can reuse some other code instead of recreating an image layout here?

Hmm I am not sure if I even saw some code that can be reused. Maybe I missed it 🤔 

> Would a commit API call on apiserver be a good option? Thoughts?

I had initially added a commit API call but thought just keeping it in a separate helper file/functions. Could go  either way here.

katiewasnothere: I suppose we could take this change and decide to do a commit API later. If you already have a branch that has the work using a commit API, I could take look as well. Otherwise, I'd ask just to do another pass on this to see if there's some code we can reuse and if not, then we can merge it as it is now.

Issue #1752, opened 2026-06-18
Support Docker Compose-compatible container log semantics


stephenlclarke:
## Summary

I am working on a Docker Compose v2 compatibility plugin for apple/container:

https://github.com/stephenlclarke/container-compose

While implementing `container compose logs`, I found that some Compose-level behavior can live cleanly in the plugin, but several lower-level log capabilities need support from apple/container itself. This issue tracks the remaining runtime/API behavior needed for Compose-compatible log commands.

This is related to:

- #1591
- #1592

#1591/#1592 cover an important first step by adding log retrieval options such as `since` and `timestamps`. Compose-compatible logging also needs a few additional capabilities so external orchestrators can avoid replaying/filtering complete log files in user space.

## Why this matters

Docker Compose-style orchestrators need to support commands such as:

```sh
container compose logs
container compose logs --follow
container compose logs --tail 100
container compose logs --since 10m
container compose logs --until 2026-06-18T10:00:00Z
container compose logs --timestamps
```

The plugin can own Compose presentation behavior such as:

- project/service/container selection
- service prefixes
- colorized output
- interleaving logs from multiple services
- replica labels and Compose metadata
- command-line compatibility mapping

However, the plugin needs apple/container to provide efficient and stable access to the underlying log records.

## Requested apple/container capabilities

Building on #1591/#1592, it would be useful for `ContainerClient.logs` or an equivalent direct API to support:

- `tail`: return only the last N log lines/records without requiring the caller to replay the full log
- `since`: return records after a timestamp
- `until`: return records before a timestamp
- `timestamps`: stable timestamp behavior, either by preserving stored timestamps or decorating emitted records consistently
- stream identity where available, for example stdout vs stderr
- follow/stream behavior that works reliably for orchestrators reading multiple containers concurrently
- enough returned metadata for callers to associate log records with a container and preserve ordering where possible

An illustrative API shape might be:

```swift
public struct ContainerLogOptions: Sendable {
    public var since: Date?
    public var until: Date?
    public var tail: Int?
    public var timestamps: Bool
    public var follow: Bool
}
```

If apple/container already intends #1592 to become the extension point for this, these could potentially be added there incrementally.

## Compose compatibility impact

Without runtime/API support, `container-compose` can still approximate some behavior, but it has to do inefficient or incomplete work in the plugin:

- `--tail` requires reading and buffering full logs
- `--since` / `--until` require parsing and filtering full logs
[...]

PR #1764, opened 2026-06-20
feat(logs): add tail and until retrieval filters

stephenlclarke:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

This is a draft stacked follow-up to #1592. Chris's #1592 introduces the initial `ContainerLogOptions` API shape for `since` and `timestamps`; this draft extends that direction with the remaining line-retrieval filters needed by `container logs` and external API clients: `tail` and `until`.

The motivation is to keep Docker-style log retrieval semantics close to the service that owns the log files, rather than requiring every caller to replay and filter the complete stream locally. This is useful for the `container logs` CLI and for external orchestrators, while keeping Compose-specific formatting and policy out of `apple/container`.

Draft/stacking note: this branch is based on `full-chaos/container:feat/chaos-1322-log-options` from #1592. Because GitHub PRs to `apple/container` target `main`, this diff includes #1592 plus the follow-up commits below. The intended review delta for this draft is:

- `ece5d5d feat(logs): add tail and until retrieval filters`
- `bd178e8 fix(logs): harden tail replay filters`
- `acab75a test(logs): cover byte-preserving log replay`
- `881f9e9 fix(logs): fail closed when filtering logs`
- `b8d7675 test(logs): close filtered log handles`
- `f4e2723 test(logs): close command log handles`
- `03c2594 docs(logs): clarify followed time filter limitation`
- `4884d25 fix(logs): create filtered logs with private permissions`
- `897665a fix(logs): require unlinking filtered log files`

## What Changed

- Extends `ContainerLogOptions` with `tail` and `until` while preserving the existing zero-option behavior.
- Plumbs `tail`, `since`, and `until` through `ContainerClient`, XPC, the harness, and `ContainersService`.
- Applies `tail` after service-side line filtering, with Docker-compatible negative-tail behavior treated as "all".
- Adds a bounded service-side tail scan for static tail-only requests instead of reading the whole log into memory.
- Adds byte-preserving local output for followed raw logs so negative tail and non-UTF-8 log bytes do not crash or decode incorrectly.
- Adds `container logs --tail` long-form support and `--since` / `--until` CLI arguments.
- Preserves unparseable and non-UTF-8 raw log bytes while filtering what can be safely filtered.
- Fails closed when filtered log replay cannot create the temporary filtered handle, so callers do not silently receive unfiltered logs.
- Creates filtered replay files with exclusive owner-only permissions and requires successful unlink before returning the replay handle.
- Closes filtered and command log handles explicitly in focused tests.
- Clarifies that followed time filters are not supported for this static-replay draft.
- Updates the command reference for the new CLI options and the current followed time-filter limitation.

## Relationship to Existing Work

- Builds on #1592 rather than replacing it.
- Related to #1591 and #1752.
- If maintainers prefer, the retrieval-only shape from this draft can be folded into #1592 before that PR lands.

## Non-Goals

- This does not add structured log records.
- This does not add timestamp rendering for raw logs.
- This does not add Compose-specific behavior.
- This does not support followed `--since` / `--until` yet; those filters are static replay only in this draft.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

Local verification:

```bash
swift format lint --strict --configuration .swift-format-nolint Sources/ContainerCommands/LogFileOutput.swift Sources/ContainerCommands/Container/ContainerLogs.swift Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift Tests/ContainerCommandsTests/ContainerLogsCommandTests.swift Tests/ContainerAPIServiceTests/ContainerLogsTests.swift Tests/ContainerResourceTests/ContainerLogOptionsTests.swift
git diff --check
swift build
swift test --filter ContainerLogsCommandTests
swift test --filter ContainerLogsTests
swift test --filter ContainerLogOptionsTests
```

Result: focused validation passed locally. The focused tests ran 25 tests across `ContainerLogsCommandTests`, `ContainerLogsTests`, and `ContainerLogOptionsTests`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -203,9 +203,12 @@ let package = Package(
             name: "ContainerAPIServiceTests",
             dependencies: [
                 .product(name: "Containerization", package: "containerization"),
+                "ContainerAPIClient",
+                "ContainerAPIService",
                 "ContainerResource",
                 "ContainerRuntimeLinuxClient",
                 "ContainerRuntimeClient",
+                "ContainerXPC",
             ]
         ),
         .target(
--- a/Sources/ContainerCommands/Container/ContainerLogs.swift
+++ b/Sources/ContainerCommands/Container/ContainerLogs.swift
@@ -16,6 +16,7 @@
 
 import ArgumentParser
 import ContainerAPIClient
+import ContainerResource
 import ContainerizationError
 import Darwin
 import Dispatch
@@ -36,8 +37,14 @@ extension Application {
         @Flag(name: .shortAndLong, help: "Follow log output")
         var follow: Bool = false
 
-        @Option(name: .short, help: "Number of lines to show from the end of the logs. If not provided this will print all of the logs")
+        @Option(name: [.short, .customLong("tail")], help: "Number of lines to show from the end of the logs. If not provided this will print all of the logs")
         var numLines: Int?
+
+        @Option(name: .long, help: "Show logs after the specified RFC 3339 timestamp")
+        var since: ContainerLogTimestamp?
+
+        @Option(name: .long, help: "Show logs before the specified RFC 3339 timestamp")
+        var until: ContainerLogTimestamp?
 
         @OptionGroup
         public var logOptions: Flags.Logging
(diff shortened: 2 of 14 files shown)


stephenlclarke:
@chrisgeo, I tried to add you as a reviewer but GitHub would not let me request review from this fork. These draft follow-ups are stacked on your log retrieval options work in #1592. No pressure for a full review yet; early guidance on API shape and whether this fits your intended direction would be really useful before I polish them for maintainer review.

stephenlclarke:
I’ve done a stricter local review pass on this branch against the log API direction in #1592, the structured-log work tracked in #1752, and Docker/Compose log option behavior.

I have not pushed the branch update yet, but the local branch has moved on substantially from the current PR head. The main local changes are:

- fail closed for timestamp filters when raw logs cannot prove a timestamped record
- reject followed `--since` / `--until` rather than following a filtered temporary snapshot
- handle negative and zero tail values using Docker-shaped semantics
- avoid eager/full-log replay for filtered static logs by scanning in bounded chunks
- retain filtered tail results with a fixed-size ring buffer
- add `ContainerLogOptions.stream` so CLI/direct API callers can filter only the selected stdio or boot stream
- preserve `ContainerizationError` failures from the filtering path
- expand command/API tests around non-UTF8 logs, blank records, stream selection, zero tail, negative tail, and filtered tail ordering

Validation passing locally:

- `swift format lint --strict --configuration .swift-format-nolint ...`
- `markdownlint docs/upstream/PR-1764.md`
- `git diff --check`
- `swift test --filter ContainerLogsTests --filter ContainerLogsCommandTests --filter ContainerLogOptionsTests`

The local branch is currently at `3df80a3`. I’m calling this out before pushing because the update is materially larger than the current visible PR diff and I want the review context to be clear.

PR #1778, opened 2026-06-21
Fix forwarded process signal encoding

### LINKED ISSUE #1747: [Bug]: Terminal resize SIGWINCH forwarding prints missing signal in xpc message
## Description

When running a long-lived foreground command with `container run`, terminal resize events can produce repeated XPC signal forwarding errors on stderr:

```text
failed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 28]
```

Signal 28 is `SIGWINCH`. The workload continues compiling, so this appears to be a CLI/terminal resize forwarding issue rather than an application failure, but it pollutes stderr during long-running builds and makes it harder to detect real build errors.

## Environment

```text
container CLI version 1.0.0 (build: release, commit: unspeci)
macOS 26.5.1 (25F80)
Darwin 25.5.0 arm64
Host hardware: Apple Silicon
Shell: zsh
```

`container system status` reports the apiserver running:

```text
status             running
apiserver.version  container-apiserver version 1.0.0 (build: release, commit: unspeci)
```

## Observed output

This happened while running a Rust build in a foreground Apple container session. The build was still making progress, then stderr repeatedly printed:

```text
   Compiling tokio v1.52.3
   Compiling futures-macro v0.3.32
   Compiling futures-task v0.3.32
   Compiling serde_json v1.0.149
   Compiling futures-util v0.3.32
   Compiling pkg-config v0.3.32
failed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 28]
failed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 28]
failed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 28]
failed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 28]
```

## Expected behavior

Terminal resize events should either:

- be forwarded successfully as `SIGWINCH`, or
- be ignored without printing an internal XPC error to stderr.

The CLI should not emit repeated internal `missing signal in xpc message` errors during normal terminal resize handling.

## Notes

A direct signal send did not reproduce the same error:

```shell
CID="$(container run --detach quay.io/libpod/alpine:latest sleep 300)"
container kill --signal WINCH "$CID"
container kill "$CID"
```

That suggests the issue is likely specific to the foreground CLI's terminal resize/signal forwarding path rather than generic signal delivery.

vincent-peng:
## Summary
- Encode forwarded process signals as signal names instead of integer XPC values.
- Preserve Darwin-to-Linux signal semantics for foreground signal forwarding.
- Add coverage for the kill request payload and forwarded INT, TERM, WINCH, USR1, and USR2 mappings.

## Root cause
Foreground non-tty signal forwarding sent `.signal` as an integer, but the API server reads that field as a string before forwarding it to the runtime. That made forwarded resize signals fail with `missing signal in xpc message`. A numeric-only fix would also risk mis-mapping Darwin USR1/USR2 values to Linux.

Fixes #1747

## Validation
- `swift test --filter ContainerAPIClientTests.ClientProcessTests`
- `swift test --filter ContainerAPIClientTests`
- `swift build --target ContainerAPIClient`


JeronimoColon:
I hit this independently while building on ContainerAPIClient, in a form broader than the resize noise in #1747: `ClientProcess.kill` fails for every signal, so there is no working way to signal an exec'd process at all. A timeout built on `createProcess` can stop waiting for a command, but cannot terminate it. I filed #1941 with a standalone reproducer and the root-cause trace.

Two notes from checking this change against the server path:

- The encoding checks out. The server-side parse (`Signal.init(_:)` in containerization) uppercases, trims an optional `SIG` prefix, and resolves the name against the Linux table, so the prefix-less names this PR sends parse correctly. That parser and its tables are unchanged from containerization 0.33.3 through current main, so a client carrying this fix also works against already-installed 1.0.0 and 1.1.0 API servers.
- Consider adding a `KILL` case to the tests. Timeout and cancel implementations mostly send SIGKILL, so it is the value most callers will hit first. It passes the guard today (`Signal.platformName(SIGKILL)` resolves to "KILL", which is in the Linux table); a test would pin that.

Sending names instead of numbers also avoids the Darwin/Linux mismatch on USR1 and USR2, which a numeric fix would have gotten wrong. Nice work.

vincent-peng:
Thanks, this was helpful. Added explicit `SIGKILL` coverage in b08d725 to pin the timeout/cancel path.

I also added a Darwin-only `SIGINFO` rejection case so the client-side guard keeps host-only signal names from reaching the runtime. Keeping the name-based encoding unchanged so USR1/USR2 continue mapping by signal name rather than Darwin numeric value.

PR #1779, opened 2026-06-21
Add Apple container sentinel file

### LINKED ISSUE #140: Detecting Apple’s Container Environment
I am exploring methods to determine if an application is running within Apple’s container environment. In platforms like `Docker`, the presence of the `/.dockerenv` file or specific environment variables (e.g., container=docker) can serve as indicators. Similarly, for Podman or OCI containers, environment variables like container=podman are commonly used.

Could you advise on reliable techniques or best practices to detect if an application is executing inside Apple’s container environment? Are there specific environment variables, files, or system attributes that can be checked for this purpose?

Any guidance or references to documentation would be greatly appreciated.

culpen90:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #140.

This adds a read-only `/run/.containerenv` sentinel file for Linux containers so applications have a reliable Apple container detection signal without changing the process environment.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

```console
swift build --target ContainerRuntimeLinuxServer
```


jglogan:
@culpen90 Thank you for the enhancement!

Have a look at the comment from Anders at the bottom of the issue: https://github.com/apple/container/issues/140#issuecomment-3015180849.

Can you implement the container sentinel using an env file instead? Ideally this change could be limited to just the RuntimeService.

culpen90:
@jglogan Thanks, updated this to use an env file sentinel instead.

The runtime now writes `.containerenv` with `container=apple` under the RuntimeService root and mounts it read-only at `/run/.containerenv`. The process environment is no longer modified, and the PR diff is limited to `RuntimeService.swift`.

Validated locally with:

```console
swift build --target ContainerRuntimeLinuxServer
```

PR #1781, opened 2026-06-21
Remove duplicate release workflow that double-builds and races on every tag

### LINKED ISSUE #1755: Duplicate tag-triggered release workflows (release.yml + release-build.yml) double-build every tag and can race to publish the same release
## Summary

`.github/workflows/release.yml` and `.github/workflows/release-build.yml` are currently byte-for-byte identical (1527 bytes each, same MD5; `diff` exits clean) and both subscribe to the **same** trigger — a `push` of a tag matching `[0-9]+\.[0-9]+\.[0-9]+`. When a release tag is pushed, GitHub starts **both** workflows independently and simultaneously. Neither file declares a `concurrency:` group, so the two runs are not serialized.

The result is (1) the full release build runs **twice** per tag, and (2) both runs concurrently call `softprops/action-gh-release` to create/update the same draft prerelease and upload the same assets, which can race.

(For history: `release-build.yml` was added later as a copy of `release.yml`, and the two have since been kept in lockstep — they are not two intentionally distinct pipelines.)

## Affected files

- `.github/workflows/release.yml` — full file (lines 1–58)
- `.github/workflows/release-build.yml` — full file (lines 1–58, identical content)
- Trigger: lines 6–9 · `build` job (calls `common.yml` with `release: true`): lines 11–21 · `release` job (`softprops/action-gh-release`): lines 23–57
- Neither file contains a `concurrency:` key.

Shared trigger and publish step (identical in both):

```yaml
on:
  push:
    tags:
      - "[0-9]+\.[0-9]+\.[0-9]+"
```

```yaml
- name: Create release
  uses: softprops/action-gh-release@b4309332981a82ec1c5618f44dd2e27cc8bfbfda # v3
  with:
    token: ${{ github.token }}
    name: ${{ github.ref_name }}-prerelease
    draft: true
    make_latest: false
    prerelease: true
    fail_on_unmatched_files: true
    files: |
      outputs/container-package/*.zip
      outputs/container-package/*.pkg
```

## Steps to reproduce

1. Push a release tag, e.g. `git tag 0.4.2 && git push origin 0.4.2`.
2. GitHub matches the tag against the trigger in **both** files and starts two independent workflow runs (GitHub does not deduplicate separate workflow files even when their `name:` is identical).
3. Each run executes the full `common.yml` build (`release: true`, `timeout-minutes: 75`), then a `release` job calling `softprops/action-gh-release` for the tag, creating a draft prerelease `<tag>-prerelease` and uploading the same `*.zip` / `*.pkg` assets.
4. With no `concurrency:` group and no path/branch distinction between the files, the two runs are not serialized.

## Expected behavior

A single workflow run per release tag — one build, one publish.

## Actual behavior

Two identical runs per tag:
- The full build + packaging pipeline (`timeout-minutes: 75`) runs **twice**, unconditionally doubling CI time/cost for every tag.
- Both runs target the same GitHub Release with the same name and assets. `softprops/action-gh-release` upserts by tag, so concurrent runs can collide — duplicate/overwritten draft releases, asset-upload conflicts, or one run intermittently failing on asset upload (`fail_on_unmatched_files: true`) depending on interleaving. The outcome is non-deterministic per tag push.

## Suggested fix

Either:
- Delete the duplicate (`release.yml`), keeping a single release workflow; **or**
- If two workflows are intentional, add a shared tag-keyed concurrency group to serialize/deduplicate runs:

```yaml
concurrency:
  group: release-${{ github.ref }}
  cancel-in-progress: false
```

Removing the duplicate is the cleaner option since the two files are identical.

## Severity & scope

Confined to CI/release tooling — it does not affect the shipped `container` CLI, runtime, or security. The publish race is partly self-mitigated (the pinned `v3` action handles concurrent-run races; `draft: true` keeps a human in the loop before final publish). The **unconditional** cost is the doubled ~75-minute build on every release tag.

radheradhe01:
Closes #1755.

`.github/workflows/release.yml` and `.github/workflows/release-build.yml` are byte-for-byte identical (same blob SHA) and both trigger on the same release-tag push (`[0-9]+\.[0-9]+\.[0-9]+`), with no `concurrency:` group. So every release tag starts **two** identical workflow runs:

- the full build + package pipeline (`common.yml`, `timeout-minutes: 75`) runs twice; and
- both runs call `softprops/action-gh-release` for the same tag and assets, which can race.

This is observable in the Actions history: both workflows ran at the identical timestamp for 1.0.0, 0.12.3, 0.12.1, 0.12.0 and 0.11.0 — and for 0.12.2 one of the two runs was `cancelled` while the other succeeded.

This change removes `release-build.yml`, which was added later as a copy of `release.yml` (#550), restoring a single release workflow.

If you'd prefer to keep `release-build.yml` and drop `release.yml` instead, or keep both behind a `concurrency: { group: release-${{ github.ref }} }` guard, I'm happy to switch — they're identical, so any of those resolves the double-run.


jglogan:
@radheradhe01 let's keep `release-build.yaml`. That was an age ago, but it looks like we went with the convention `-build.yaml` and I forgot to delete the older `release.yaml`.

Thank you for cleaning this up!

jglogan (approved):
@radheradhe01 thanks! I'll rename the file for consistency with the others in a follow-on PR, since I'm working on the integration tests and workflows atm.

PR #1782, opened 2026-06-21
Log the graceful-stop error instead of silently discarding it in gracefulStopContainer

### LINKED ISSUE #1756: gracefulStopContainer silently swallows the graceful-stop error in an empty catch {} (no log)
## Summary

`RuntimeService.gracefulStopContainer(_:signal:timeout:)` wraps its graceful-stop attempt in a `do`/`catch`, but the `catch` block is empty (`} catch {}`). Any error thrown during the graceful-stop attempt is discarded — not logged, not rethrown. This is the only error-handling site in the file that does not log; every other `catch` in `RuntimeService.swift` logs the error via `self.log.error(..., metadata: ["error": "\(error)"])`. When a graceful stop fails and the VM is force-stopped instead, there is no log line recording that it happened or why.

## Affected file

`Sources/Services/RuntimeLinux/Server/RuntimeService.swift`

- Function `gracefulStopContainer` at lines **1214–1246**
- Empty catch at line **1240** (the `do`/`catch` spans **1218–1240**)
- For contrast, other catches in the same file log the error — e.g. lines 300, 538, 974, 1255. The logger is held at line 49 (`private let log: Logging.Logger`).

```swift
private func gracefulStopContainer(_ lc: LinuxContainer, signal: Signal, timeout: Duration) async throws -> ExitStatus {
    // Try and gracefully shut down the process. Even if this succeeds we need to power off
    // the vm, but we should try this first always.
    var code = ExitStatus(exitCode: 255)
    do {
        code = try await withThrowingTaskGroup(of: ExitStatus.self) { group in
            group.addTask { try await lc.wait() }
            group.addTask {
                try await lc.kill(signal)
                try await Task.sleep(for: timeout)
                try await lc.kill(.kill)
                return ExitStatus(exitCode: 137)
            }
            guard let code = try await group.next() else {
                throw ContainerizationError(
                    .internalError,
                    message: "failed to get exit code from gracefully stopping container"
                )
            }
            group.cancelAll()
            return code
        }
    } catch {}                 // line 1240: error discarded, nothing logged

    // Now actually bring down the vm.
    try await lc.stop()
    return code
}
```

Any of these throwing operations can fail and the error is silently swallowed: `lc.wait()`, `lc.kill(signal)` / `lc.kill(.kill)` (e.g. guest agent / VM unresponsive, process already gone), `Task.sleep(for:)` being cancelled, or the explicit `ContainerizationError` when `group.next()` returns `nil`. For the case where the graceful attempt throws but the subsequent unconditional `lc.stop()` succeeds, the error is logged at no layer at all.

## Expected behavior

When the graceful-stop attempt throws, the error should be logged (at `error` or `warning` level), matching the file's existing convention, before falling through to `lc.stop()`.

## Actual behavior

The error is silently discarded. Execution falls through to `try await lc.stop()` with no record that the graceful attempt failed or why.

## Suggested fix

A single log line, preserving the intentional fall-through to `lc.stop()`:

```swift
} catch {
    self.log.error("graceful stop failed; forcing vm shutdown", metadata: ["error": "\(error)"])
}
```

## Severity & scope

Low — a diagnostic/observability gap, not a functional defect. `lc.stop()` runs unconditionally after the catch, so the VM is always brought down (no leak/hang); `255` is the codebase's deliberate "unknown exit" sentinel (same use at line 1263); and the total-failure case where `lc.stop()` itself throws is still surfaced (it propagates to the caller, logged as "failed to clean up container"). The practical impact is harder debugging of failed/forced container shutdowns. The fix is a one-line log statement matching the existing pattern.

radheradhe01:
Closes #1756.

`RuntimeService.gracefulStopContainer(_:signal:timeout:)` wraps the graceful-stop attempt in `do { … } catch {}` — an empty catch that silently discards any thrown error before falling through to the unconditional `lc.stop()`. It is the only catch in this file that does not log; every other one uses `self.log.error(…, metadata: ["error": "\(error)"])`.

This adds a single log line matching that convention, so a failed graceful stop (and the resulting fall-through to a forced VM shutdown) is diagnosable. The intentional fall-through to `lc.stop()` is unchanged.

### DIFF
--- a/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
+++ b/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
@@ -1242,7 +1242,9 @@ public actor RuntimeService {
 
                 return code
             }
-        } catch {}
+        } catch {
+            self.log.error("graceful stop failed; forcing vm shutdown", metadata: ["error": "\(error)"])
+        }
 
         // Now actually bring down the vm.
         try await lc.stop()


jglogan (approved):
@radheradhe01 This looks good to me - thank you for adding that bit of observability.

PR #1783, opened 2026-06-22
Fix/exec empty arguments crash

SEPURI-SAI-KRISHNA:
## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Motivation and Context
  `swift-argument-parser` enforces that `arguments` is non-empty before
  `run()` is invoked, so the force-unwrap of `arguments.first!` is not
  reachable in practice.

  However, the guard makes the invariant explicit in the code itself,
  removes reliance on ArgumentParser's implicit enforcement, and would
  satisfy force-unwrap lint rules if enabled in the future

  ## Testing
  - [x] Tested locally
  - [ ] Added/updated tests
  - [ ] Added/updated docs

### DIFF
--- a/.idea/inspectionProfiles/profiles_settings.xml
+++ b/.idea/inspectionProfiles/profiles_settings.xml
@@ -1,6 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <settings>
-    <option name="USE_PROJECT_PROFILE" value="false" />
-    <version value="1.0" />
-  </settings>
-</component>
--- a/.idea/inspectionProfiles/profiles_settings.xml
+++ b/.idea/inspectionProfiles/profiles_settings.xml
@@ -1,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <settings>
+    <option name="USE_PROJECT_PROFILE" value="false" />
+    <version value="1.0" />
+  </settings>
+</component>
--- a/.idea/modules.xml
+++ b/.idea/modules.xml
@@ -1,8 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="ProjectModuleManager">
-    <modules>
-      <module fileurl="file://$PROJECT_DIR$/.idea/container.iml" filepath="$PROJECT_DIR$/.idea/container.iml" />
-    </modules>
-  </component>
-</project>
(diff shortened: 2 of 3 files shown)


jglogan:
@SEPURI-SAI-KRISHNA I don't see it crash at runtime.

```console
% container ls
ID                 IMAGE                                         OS     ARCH   STATE    IP               CPUS  MEMORY   STARTED
6f6ca580-registry  docker-upstream.apple.com/library/registry:3  linux  arm64  running  192.168.64.3/24  4     1024 MB  2026-06-23T02:38:21Z

% container exec 6f6ca580-registry
Error: Missing expected argument '<arguments> ...'
Help:  <arguments>  New process arguments
Usage: container exec [<options>] <container-id> <arguments> ...
  See 'container exec --help' for more information.
```

That's because swift-argument-parser enforces the invariant `arguments.count > 0` before `run()` is even invoked.

The fix you provide is good defensive programming though, and if we were to turn on lint checks at some point we would need to create exceptions for the few force-try expressions scattered through the code.

SEPURI-SAI-KRISHNA:
@jglogan Thanks for the context. You're right that ArgumentParser enforces this before run() is invoked I missed that invariant. Happy to close this if the team prefers keeping the codebase as is. Alternatively, since you mentioned lint checks would flag force-unwraps eventually, I can keep it as a preemptive cleanup but I'll defer to your preference.

jglogan:
@SEPURI-SAI-KRISHNA I think the change is fine - you just need to configure commit signing, and perhaps update your **Motivation and Context** description.

SEPURI-SAI-KRISHNA:
@jglogan I have added signature. please check. I have also updated the Motivation and Context.

JaewonHur:
@SEPURI-SAI-KRISHNA Oops sorry! The signatures have gone while I rebase them on latest main. Could you force push the signed commit again..?

jglogan:
@SEPURI-SAI-KRISHNA Could you rebase this commit to the latest main and force-push the changes? We tried the rebase from the GitHub UI but that dropped the commit signatures.

The PR needs to be rebased to pick up an integration test workaround unrelated to your change. Thank you!

SEPURI-SAI-KRISHNA:
@jglogan @JaewonHur I have rebase and force-pushed.

JaewonHur (approved):

jglogan:
@SEPURI-SAI-KRISHNA Merged, thanks again for your contribution!

PR #1784, opened 2026-06-22
fix: replace try! with try? for stdout/stderr writes in ProcessIO

SEPURI-SAI-KRISHNA:
## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Motivation and Context
  In `ProcessIO.swift`, the readability handlers for stdout and stderr used
  `try!` when writing data to the output file handles. If the pipe is
  broken (e.g. the terminal is closed mid-execution), this crashes the
  entire process instead of gracefully continuing.

  Changed both occurrences to `try?` so a failed write is silently dropped
  the process and its exit code are still correctly handled.

  ## Testing
  - [x] Tested locally
  - [ ] Added/updated tests
  - [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/ProcessIO.swift
+++ b/Sources/Services/ContainerAPIService/Client/ProcessIO.swift
@@ -105,7 +105,12 @@ public struct ProcessIO: Sendable {
                     cc.yield()
                     return
                 }
-                try! pout.write(contentsOf: data)
+                do {
+                    try pout.write(contentsOf: data)
+                } catch {
+                    rout.readabilityHandler = nil
+                    cc.yield()
+                }
             }
         }
 
@@ -126,7 +131,12 @@ public struct ProcessIO: Sendable {
                     cc.yield()
                     return
                 }
-                try! perr.write(contentsOf: data)
+                do {
+                    try perr.write(contentsOf: data)
+                } catch {
+                    rerr.readabilityHandler = nil
+                    cc.yield()
+                }
             }
             stdio[2] = stderr.fileHandleForWriting
         }


Sources/Services/ContainerAPIService/Client/ProcessIO.swift
```diff
@@ -105,7 +105,7 @@ public struct ProcessIO: Sendable {
                     cc.yield()
                     return
                 }
-                try! pout.write(contentsOf: data)
+                try? pout.write(contentsOf: data)
```
jglogan: Thanks for catching these force-try cases.

I think we'd be better served in this case if we treat these failed writse like an EOF:

```suggestion
                do {
                    try pout.write(contentsOf: data)
                } catch {
                    rout.readabilityHandler = nil
                    cc.yield()
                }
```

Sources/Services/ContainerAPIService/Client/ProcessIO.swift
```diff
@@ -126,7 +126,7 @@ public struct ProcessIO: Sendable {
                     cc.yield()
                     return
                 }
-                try! perr.write(contentsOf: data)
+                try? perr.write(contentsOf: data)
```
jglogan:
```suggestion
                do {
                    try perr.write(contentsOf: data)
                } catch {
                    rerr.readabilityHandler = nil
                    cc.yield()
                }
```

jglogan (changes requested):
@SEPURI-SAI-KRISHNA Thank you for the contribution! Please have a look at the comments and let me know what you think.

SEPURI-SAI-KRISHNA:
@jglogan try? silently drops the error, but treating a failed write as EOF is the right behavior (stop reading, signal completion). Let me apply it.

SEPURI-SAI-KRISHNA:
@jglogan please check now, i have made the changes.

jglogan (approved):
@SEPURI-SAI-KRISHNA Thank you for the fix! It's nice to chip away at this technical debt.

And thanks for submitting these as individual PRs - it's easier to review and approve them individually (I'm still thinking about the app-root one, so having it as a separate change means it doesn't hold up these).

SEPURI-SAI-KRISHNA:
@jglogan Appreciate your support and patience. Thanks for sharing your thoughts on separate PR. I follow this across GIT which I consider a good practice rather than bulky PR with huge code changes. Would definitely love to  contribute further.

Sources/Services/ContainerAPIService/Client/ProcessIO.swift
heptalianyaftali-boop: @

PR #1787, opened 2026-06-22
fix: remove force-unwrap on session dictionary in DefaultNetworkService

SEPURI-SAI-KRISHNA:
## Type of Change
  - [x] Bug fix
  - [ ] New feature
  - [ ] Breaking change
  - [ ] Documentation update

  ## Motivation and Context
  In `DefaultNetworkService.allocate`, the dictionary entry for `session`
  is initialized at line 98 if nil, then immediately force-unwrapped at
  line 104 with `allocationsBySession[session]!.append(...)`.

  While the actor serializes access and the entry should always exist at
  that point, the force-unwrap is an unnecessary crash risk if the
  invariant is ever broken. Replaced with optional chaining (`?`) which
  silently no-ops rather than crashing.

  ## Testing
  - [x] Tested locally
  - [ ] Added/updated tests
  - [ ] Added/updated docs

### DIFF
--- a/Sources/Services/Network/Server/DefaultNetworkService.swift
+++ b/Sources/Services/Network/Server/DefaultNetworkService.swift
@@ -95,13 +95,13 @@ public actor DefaultNetworkService: NetworkService {
         }
         macAddresses[index] = macAddress
 
-        if allocationsBySession[session] == nil {
-            allocationsBySession[session] = []
+        let isNewSession = allocationsBySession[session] == nil
+        allocationsBySession[session, default: []].append((hostname: hostname, index: index))
+        if isNewSession {
             await session.onDisconnect { [weak self] in
                 await self?.releaseSession(session)
             }
         }
-        allocationsBySession[session]!.append((hostname: hostname, index: index))
 
         return (attachment: attachment, additionalData: additionalData)
     }


Sources/Services/Network/Server/DefaultNetworkService.swift
```diff
@@ -101,7 +101,7 @@ public actor DefaultNetworkService: NetworkService {
                 await self?.releaseSession(session)
             }
         }
-        allocationsBySession[session]!.append((hostname: hostname, index: index))
+        allocationsBySession[session]?.append((hostname: hostname, index: index))
```
jglogan: This approach to L98-104 would be better:

```swift
        let isNewSession = allocationsBySession[session] == nil
        allocationsBySession[session, default: []].append((hostname: hostname, index: index))
        if isNewSession {
            await session.onDisconnect { [weak self] in
                await self?.releaseSession(session)
            }
        }
```

jglogan (commented):
@SEPURI-SAI-KRISHNA See the suggested fix for this force-try.

jglogan (changes requested):

Sources/Services/Network/Server/DefaultNetworkService.swift
SEPURI-SAI-KRISHNA: I have made the changes. please check. Thank you !

jglogan (approved):
@SEPURI-SAI-KRISHNA approved, thank you!

SEPURI-SAI-KRISHNA:
@jglogan Thanks for your support and patience. Would love to contribute further.

PR #1791, opened 2026-06-23
perf(parser): add collection capacity hints to known-size loops

hluaguo:
## Type of Change
- [x] Bug fix

## Motivation and Context
Methods in Parser.swift and Utility.swift build arrays and dictionaries in loops where the final size is known from the input parameter count, but none pre-allocate storage. Without a capacity hint, Array.append grows the backing store exponentially (2x), causing O(log n) reallocation+copies per call. These methods execute on every `container run/create` command, so the savings compound at scale.

## Testing
- [x] Tested locally
- [x] Added/updated tests

Added 7 new tests with large inputs (50-100 entries) that exercise the capacity-allocated code paths. All 183 existing tests pass with zero regressions.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Parser.swift
+++ b/Sources/Services/ContainerAPIService/Client/Parser.swift
@@ -24,6 +24,14 @@ import ContainerizationOS
 import Foundation
 import SystemPackage
 
+// MARK: - Collection capacity hints
+// Methods in this file build arrays and dictionaries in loops where the final
+// size is known from the input parameter count. reserveCapacity() and
+// Dictionary(minimumCapacity:) avoid O(log n) reallocation copies as the
+// collection grows incrementally. While this is a micro-optimization for each
+// individual call, these methods execute on every `container run/create` and
+// the savings compound at scale.
+
 /// A parsed volume specification from user input
 public struct ParsedVolume {
     public let name: String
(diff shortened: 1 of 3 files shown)


Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -331,6 +339,7 @@ public struct Parser {
 
     public static func tmpfsMounts(_ mounts: [String]) throws -> [Filesystem] {
         var result: [Filesystem] = []
+        result.reserveCapacity(mounts.count)
```
jglogan, line 343: Would it make more sense to reserve capacity for the deduped list?

```swift
        let mounts = mounts.dedupe()
        var result: [Filesystem] = []
        result.reserveCapacity(mounts.count)
```

Sources/Services/ContainerAPIService/Client/Parser.swift
```diff
@@ -342,6 +351,7 @@ public struct Parser {
 
     public static func mounts(_ rawMounts: [String], relativeTo basePath: URL? = nil) throws -> [VolumeOrFilesystem] {
         var mounts: [VolumeOrFilesystem] = []
+        mounts.reserveCapacity(rawMounts.count)
```
jglogan, line 355: Same comment applies here.

jglogan (commented):
@hluaguo thank you for these changes. Can you have a look at the comments on the collections that are deduped prior to processing? I think we can specify the actual capacity more precisely for those.

hluaguo:
Oh, the dedupe can shrink the input, so reserving before it over-allocates. I'll shift reserveCapacity after the dedupe in both tmpfsMounts and mounts.

jglogan (approved):

jglogan:
@hluaguo Merged, thank you for the optimization!

PR #1804, opened 2026-06-24
Route `container image save` reference list to stderr in stdout mode

### LINKED ISSUE #1801: [Bug]: `container image save` appends image references to stdout archive
### I have done the following

### Steps to reproduce

1. Build or use `container` from current `main`.
2. Save an image without `--output`, redirecting stdout to a tar file:

   ```bash
   container image save alpine:latest > alpine.tar
   ```

3. Inspect the resulting archive with a tar/OCI consumer that expects the stream to end at the tar EOF markers.

The current implementation writes the archive bytes to stdout, then prints each saved image reference to stdout afterward:

- `Sources/ContainerCommands/Image/ImageSave.swift`, lines 132-137: writes tar chunks to `FileHandle.standardOutput`
- `Sources/ContainerCommands/Image/ImageSave.swift`, lines 141-144: calls `print(reference)` after the stdout archive path completes

### Problem description

When `container image save` is used without `--output`, stdout is the archive stream. The command should emit only archive bytes on stdout so shell redirection and pipelines produce a valid OCI-compatible tar archive.

Currently, after writing the tar data to stdout, the command prints image references to stdout. That appends text bytes after the archive stream, so a command such as:

```bash
container image save alpine:latest > alpine.tar
```

creates an archive with trailing non-archive output. `container image load` may tolerate that trailing data, so the existing stdin/stdout integration test can pass while other tar or OCI tooling rejects or warns about the archive. Status/reference output for the stdout mode should go to stderr/logging, or be suppressed when stdout is carrying the archive.

### Environment

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: Full Xcode is not selected locally; active developer directory is /Library/Developer/CommandLineTools. Swift reports Apple Swift 6.3.2 targeting arm64-apple-macosx26.0.
- Container: container CLI version 1.0.0-15-g1d70dd6 (build: debug, commit: 1d70dd6)
```

### Code of Conduct

costajohnt:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Fixes #1801.

When `container image save` runs without `--output`, stdout carries the OCI tar archive. The command writes the archive bytes to stdout and then `print(reference)`s each saved image reference to stdout afterward, appending non-archive text after the tar EOF marker. That corrupts the stream for strict tar/OCI consumers, so a command like:

```bash
container image save alpine:latest > alpine.tar
```

produces an archive with trailing non-archive bytes. The existing round-trip test passes only because `container image load` tolerates the trailing data, while other tooling rejects or warns about it.

This routes the saved-reference list to stderr in the no-`--output` branch, so stdout contains only archive bytes. When saving to a file via `--output`, stdout is free, so the references continue to print to stdout exactly as before.

I used a raw `FileHandle.standardError.write(...)` to match the existing stderr writes in this codebase (e.g. `Application.swift`) and to keep the reference list as plain text rather than formatted log lines. Happy to route it through the logger instead if you'd prefer.

## Testing
- [ ] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Added `testImageSaveToStdoutProducesCleanArchive`: it saves an image to stdout and asserts that the captured stdout ends at the tar EOF marker (1024 zero bytes), with the saved reference surfaced on stderr. With the bug, the trailing bytes are reference text and the assertion fails.

`swift build` and `swift format` are clean. I was not able to run the CLI integration suite locally (it needs the running container runtime), so I left "Tested locally" unchecked. The new test sits alongside the existing `image save`/`load` integration tests and is written to run in CI.

### DIFF
--- a/Sources/ContainerCommands/Image/ImageSave.swift
+++ b/Sources/ContainerCommands/Image/ImageSave.swift
@@ -140,7 +140,15 @@ extension Application {
 
             progress.finish()
             for reference in references {
-                print(reference)
+                if output == nil {
+                    // stdout is carrying the OCI archive in this branch, so the
+                    // saved-reference list goes to stderr via the logger. Printing
+                    // it to stdout appends non-archive bytes after the tar EOF and
+                    // corrupts the stream for redirection and pipelines (#1801).
+                    log.info("\(reference)")
+                } else {
+                    print(reference)
+                }
             }
         }
     }
--- a/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
+++ b/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
@@ -266,6 +266,44 @@ class TestCLIImagesCommand: CLITest {
         }
     }
 
+    @Test func testImageSaveToStdoutProducesCleanArchive() throws {
+        do {
+            // 1. pull and tag an image to save
+            try doPull(imageName: alpine)
+            let alpineRef: Reference = try Reference.parse(alpine)
+            let alpineTagged = "\(alpineRef.name):testImageSaveToStdout"
+            try doImageTag(image: alpine, newName: alpineTagged)
+            defer {
+                try? doRemoveImages(images: [alpineTagged])
+            }
+
+            // 2. save to stdout (no --output): stdout is the archive stream
+            let saveArgs = [
+                "image",
+                "save",
+                alpineTagged,
+            ]
+            let (outputData, _, error, status) = try run(arguments: saveArgs)
+            if status != 0 {
+                throw CLIError.executionFailed("save to stdout failed: \(error)")
+            }
+
+            // 3. The archive on stdout must end at the tar EOF marker (two
+            //    512-byte zero blocks). With the bug, the saved-reference list
+            //    is printed to stdout after the archive, so the trailing bytes
+            //    are reference text rather than the tar EOF zeros (#1801).
+            #expect(outputData.count >= 1024, "stdout archive is too small to contain a tar EOF marker")
+            let trailer = outputData.suffix(1024)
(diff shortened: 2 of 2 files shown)


Sources/ContainerCommands/Image/ImageSave.swift
```diff
@@ -140,7 +140,15 @@ extension Application {
 
             progress.finish()
             for reference in references {
-                print(reference)
+                if output == nil {
+                    // stdout is carrying the OCI archive in this branch, so the
+                    // saved-reference list goes to stderr. Printing it to stdout
+                    // appends non-archive bytes after the tar EOF and corrupts
+                    // the stream for redirection and pipelines (#1801).
+                    FileHandle.standardError.write(Data("\(reference)\n".utf8))
```
jglogan: Use `log.info` for this.

jglogan (changes requested):
@costajohnt Thank you for the fix! Please see the one suggested change.

Sources/ContainerCommands/Image/ImageSave.swift
jglogan: Also, could you rebase your change onto the latest main and push it again? We needed to disable a CI test that was flaky, so rebase and push will allow us to build this change without fear of an unrelated test failing.

jglogan:
@costajohnt The fix looks good now, but one more thing - we need you to submit signed commits to merge the change. See https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests for details.

jglogan (approved):
@costajohnt Thank you for the fix and the test!

jglogan:
@costajohnt The point of using the logger is that by default it's like a `print(stderr, msg)` - no special log formatting, just the message. But if you use `--debug` you'll get log-formatted messages with timestamps.

Issue #1809, opened 2026-06-25
[Request]: First-class container-to-container DNS discovery on custom networks


thromel:
### Feature or enhancement request details

Please add first-class container-to-container DNS discovery for containers attached to the same `container` network, including bare container names and network aliases.

This came up while testing Compose-compatible workflows with `container-compose`. Compose users expect service names such as `web`, `db`, and network aliases to resolve from peer containers on the same network without a global host DNS setup step.

Related tracking:

- Existing DNS bug/discussion: https://github.com/apple/container/issues/856
- Compose compatibility tracker: https://github.com/Mcrich23/Container-Compose/issues/118
- Container-Compose service discovery issue: https://github.com/Mcrich23/Container-Compose/issues/114
- Container-Compose network aliases issue: https://github.com/Mcrich23/Container-Compose/issues/116

Current behavior on `container` 1.0.0:

```sh
container system start
container network create appnet
container run -d --name network-web --network appnet nginx:alpine
container run --rm --network appnet busybox:latest nslookup network-web
```

The peer container receives the network gateway as its resolver:

```sh
container exec network-web cat /etc/resolv.conf
# nameserver 192.168.65.1
```

The container is reachable by IP from the same network, but name lookup returns NXDOMAIN:

```text
Server:         192.168.65.1
Address:        192.168.65.1:53

** server can't find network-web: NXDOMAIN
```

The currently documented workaround appears to be configuring a DNS domain with `container system dns create <domain>` plus setting `dns.domain`. That can make FQDN/search-domain flows work, but it is not the same as network-scoped, Compose-style service discovery. It also requires global host DNS configuration for what is usually expected to be a per-network container runtime feature.

A useful end state would be one of these:

1. Containers on the same network can resolve each other by container name, for example `web` -> that container's network IP.
2. The network layer supports additional network-scoped aliases so higher-level tools can map Compose service names and aliases onto Apple Container networks.
3. If bare names are intentionally out of scope, expose a supported CLI/API path that tools can use to create network-scoped DNS domains/aliases without requiring users to manually toggle global DNS settings.

Environment used for the repro:

```markdown
- OS: macOS 26.5.1 (25F80)
- Xcode: 26.5 (17F42)
- Container: container CLI version 1.0.0 (build: release, commit: ee848e3)
- Architecture: arm64
```

I agree to follow this project's Code of Conduct.

thromel:
Update: I opened a draft implementation PR for the container-facing DNS part:

- #1813

What it covers:

[...]

Issue #856, opened 2025-11-06
[Bug]: internal DNS does not work (clean install)


skinny:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

Last week I had a setup where internal DNS worked pretty well. I wiped my device and did a clean reinstallation of container 0.6.0 (and also tested a debug build on the latest github code)

I run :
```
container run -d --rm --name web  nginx
container run -it --rm alpine sh

/ # ping web
ping: bad address 'web'
/ # ping -4 web
ping: bad address 'web'
/ # nslookup web
Server:         192.168.2.1
Address:        192.168.2.1:53

** server can't find web: NXDOMAIN

** server can't find web: NXDOMAIN

/ # nslookup google.com
Server:         192.168.2.1
Address:        192.168.2.1:53

Non-authoritative answer:
Name:   google.com
Address: 142.250.64.174

Non-authoritative answer:
Name:   google.com
Address: 2607:f8b0:4008:815::200e
```

### Current behavior

container name resolution does not work at all

### Expected behavior

container names should resolve correctly

### Environment

```markdown
- OS: macOS 26.0.1
- Xcode: latest
- Container: Container CLI 0.6.0 (and latest from `main`)
```

Maybe I'm  doing something wrong, but I expect this to work actually because it did. Please let me know if you need more information from my system

### Relevant log output

```shell
N/A
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

seven1240:
Could be alpine related, can you try debian? I have run into this before: https://github.com/apple/container/issues/480

skinny:
Yes I also tried debian , same behavior unfortunately. I am looking for a way to debug this further as it remains a mystery why this was working properly last week and as this is such a basic scenario, I must be doing something wrong on my machine

skinny:
```
sudo container system dns create mycorp
container system property set dns.domain mycorp
```

After a couple of hours trial and error today the lines above is what makes it work. I must have had that done during my initial testing (before wiping the device). Not sure it's in the docs yet but it definately should be in there as users might expect inter-container dns working out of the box

AT-main:
I ran into the same thing. Setting those properties got the containers talking to each other, but I later had to remove them when building images. With the custom DNS domain set, connections from inside the containers couldn’t reach the outside world.

Clearing those settings fixed it for me:

```bash
container system property clear dns.domain
[...]

PR #1810, opened 2026-06-25
Normalize network hostname lookup

thromel:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Network attachments are allocated under the hostname supplied by the runtime, while DNS queries are canonical DNS names and may include a trailing dot. DNS hostnames are also case-insensitive. Because `AttachmentAllocator` keyed hostnames exactly as supplied, a container allocated as `web` could fail lookup when the embedded DNS path queried `web.`.

This normalizes allocator keys by lowercasing hostnames and treating a single trailing dot as optional across allocate, lookup, and deallocate. The original hostname remains preserved in the returned attachment.

This is a narrow fix for the embedded DNS lookup path. First-class Compose-style bare service discovery from containers through the network gateway is a separate feature request: https://github.com/apple/container/issues/1809. Related DNS discussion: https://github.com/apple/container/issues/856.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Checks run:

```sh
swift test --filter AttachmentAllocatorTest
swift test -c debug -Xswiftc -warnings-as-errors --filter AttachmentAllocatorTest
```

Additional manual validation:

- On released `container` 1.0.0, created a custom network and container; direct `dig @127.0.0.1 -p 2053 <container-name> A` returned no A record.
- With this patched debug build installed under a temporary `/tmp` install root, the same direct queries returned the allocated container IP for both `<container-name>` and `<container-name>.`.

### DIFF
--- a/Sources/Services/Network/Server/AttachmentAllocator.swift
+++ b/Sources/Services/Network/Server/AttachmentAllocator.swift
@@ -30,6 +30,7 @@ actor AttachmentAllocator {
 
     /// Allocate a network address for a host.
     func allocate(hostname: String) async throws -> UInt32 {
+        let hostname = Self.normalized(hostname: hostname)
         // Client is responsible for ensuring two containers don't use same hostname, so provide existing IP if hostname exists
         if let index = hostnames[hostname] {
             return index
@@ -44,6 +45,7 @@ actor AttachmentAllocator {
     /// Free an allocated network address by hostname.
     @discardableResult
     func deallocate(hostname: String) async throws -> UInt32? {
+        let hostname = Self.normalized(hostname: hostname)
         guard let index = hostnames.removeValue(forKey: hostname) else {
             return nil
         }
@@ -54,6 +56,12 @@ actor AttachmentAllocator {
 
     /// Retrieve the allocator index for a hostname.
     func lookup(hostname: String) async throws -> UInt32? {
-        hostnames[hostname]
+        let hostname = Self.normalized(hostname: hostname)
+        return hostnames[hostname]
+    }
+
+    private static func normalized(hostname: String) -> String {
+        let hostname = hostname.hasSuffix(".") ? String(hostname.dropLast()) : hostname
+        return hostname.lowercased()
     }
 }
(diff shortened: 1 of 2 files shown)


jglogan:
@thromel I'll need to go back to the DNS refactor PR and look at our design decisions there, and then look at what the change is doing here.

devops-thiago:
Supporting data for this fix, plus one gap it leaves open.

We hit the same bug on container 1.2.0. With a container named `dnsweb` running, the hostnames resolver on 2053 answered NXDOMAIN for `dnsweb.`:

```
$ dig @127.0.0.1 -p 2053 dnsweb. A +short
(no answer, status NXDOMAIN)
```

After normalizing the lookup the same query answers with the container's address, so the trailing-dot handling here fixes a real failure, not a theoretical one.

The gap: queries that arrive domain-qualified still miss after this change. A resolver file written by `container system dns create test` routes queries to this server as `dnsweb.test.`, and after dot-stripping the allocator is asked for `dnsweb.test` while the key is `dnsweb`. So the documented opt-in path stays broken for exactly the queries it is designed to route.

Two ways to close it: strip the registered local domain suffix before lookup, or fall back to the first label after an exact miss. The first-label version is only sound if attachment hostnames can never contain a dot; if they can, suffix-stripping against the known domains is the safe one. In a downstream patch we went with first-label fallback (the names there are always single labels) and `dnsweb.`, `dnsweb.test.` and `dnsweb.container.internal.` all resolve to the same address, with unknown names still NXDOMAIN.

Happy to send a follow-up PR for the suffix handling if that is easier than folding it in here.

thromel:
Thanks for the concrete reproduction and for confirming the
trailing-dot fix on 1.2.0. You’re right that this patch only
canonicalizes case and a single trailing dot; it won’t map
dnsweb.test. back to the stored dnsweb key.

I also agree that stripping a known registered suffix is safer than a
blanket first-label fallback if dotted attachment names are valid. A
focused follow-up with bare-name, registered-suffix, and unknown-name
tests seems like the clearest way to evaluate that behavior separately
from this narrow fix.

PR #1811, opened 2026-06-25
Add container run hostname flag

thromel:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context
Apple `container run` and `container create` currently use the container name as the guest hostname through the first network attachment. Docker Compose also has a separate `hostname:` field, where the requested guest hostname can differ from the container management name.

This PR adds `--hostname <hostname>` to the shared management flags and uses it as the hostname for the first network attachment. The Linux runtime already derives the guest hostname from the first attachment hostname, so the change uses the existing persisted network attachment model rather than introducing a separate container configuration field.

## Design status

This is separate from the alias work in #1815/#1839. Aliases may cover most Compose `hostname:` translation cases while keeping the container name and guest hostname coupled. This PR keeps the explicit hostname-override path available for review, but it is not required for the alias PR.

Related to Compose compatibility work in Mcrich23/Container-Compose#119 and service/alias work in #1815/#1839.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

Checks run:

```sh
git diff --check origin/main..HEAD
swift test -c debug -Xswiftc -warnings-as-errors --filter 'ParserTest|UtilityTests'
```

The focused test run passed 122 tests after rebasing onto current `origin/main` and linked the CLI stack as part of the package test build.

Earlier manual smoke test against the normal running service:

```sh
.build/arm64-apple-macosx/debug/container run --rm --name codex-hostname-flag --hostname custom-host busybox hostname
```

The command printed:

```text
custom-host
```

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -174,6 +174,7 @@ public struct Flags {
             dns: Flags.DNS,
             dnsDisabled: Bool,
             entrypoint: String?,
+            hostname: String?,
             initImage: String?,
             kernel: String?,
             labels: [String],
@@ -203,6 +204,7 @@ public struct Flags {
             self.dns = dns
             self.dnsDisabled = dnsDisabled
             self.entrypoint = entrypoint
+            self.hostname = hostname
             self.initImage = initImage
             self.kernel = kernel
             self.labels = labels
@@ -257,6 +259,9 @@ public struct Flags {
             )
         )
         public var entrypoint: String?
+
+        @Option(name: .long, help: .init("Set the container hostname", valueName: "hostname"))
+        public var hostname: String?
 
         @Flag(name: .customLong("init"), help: "Run an init process inside the container that forwards signals and reaps processes")
         public var useInit = false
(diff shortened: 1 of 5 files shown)


jglogan:
@thromel The container `--name` is the hostname in `container`.

thromel:
Thanks, that makes sense. I verified locally that `container run --rm --name codex-name-hostname-check busybox hostname` prints `codex-name-hostname-check`.

The gap I was trying to cover is Compose `hostname:`, where the requested guest hostname can be distinct from the container management name/service-derived ID. If `container` intentionally wants those to stay the same, then this PR is probably not the right direction and I can close it. If preserving that separation is acceptable, I can update the PR description/tests to make the distinction explicit.

jglogan:
Let me think about this one. I'm not in love (right at this moment) with the idea of a hostname flag.

We could use the alias feature you propose to get Compose hostname working for all but seemingly perverse cases where someone puts `hostname: existing-container-name` in their YAML.

thromel:
That matches the current split.

I updated this PR to make the hostname flag an optional/separate path rather than a prerequisite for Compose support. #1815 carries aliases as the primary path; aliases can represent most Compose `hostname:` cases while keeping the container management name and guest hostname coupled.

If keeping name and guest hostname coupled is the preferred API, this PR can be closed without blocking #1815/#1839.

PR #1816, opened 2026-06-25
fix: avoid writing image references to stdout during image save

### LINKED ISSUE #1801: [Bug]: `container image save` appends image references to stdout archive

SayanDey322:
## Summary
Fixes corruption of tar archives when `container image save` writes to stdout.

## Changes
- Moved status messages from stdout to stderr.
- Ensured stdout contains only archive data.
- Added a regression test covering stdout output.

Fixes #1801

### DIFF
--- a/Sources/ContainerCommands/Image/ImageSave.swift
+++ b/Sources/ContainerCommands/Image/ImageSave.swift
@@ -140,7 +140,7 @@ extension Application {
 
             progress.finish()
             for reference in references {
-                print(reference)
+                log.info("saved image \(reference)")
             }
         }
     }
--- a/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
+++ b/Tests/CLITests/Subcommands/Images/TestCLIImagesCommand.swift
@@ -492,6 +492,11 @@ class TestCLIImagesCommand: CLITest {
                 throw CLIError.executionFailed("command failed: \(error)")
             }
 
+            // Tar archives must end with two 512-byte zero blocks.
+            #expect(stdoutData.count >= 1024, "expected save output to contain tar EOF blocks")
+            let trailingBlocks = stdoutData.suffix(1024)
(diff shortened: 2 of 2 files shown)


jglogan:
@SayanDey322 thank you for the submission, but please check to see whether someone is already assigned to the issue, or whether a PR already exists (#1804 in this case) before submitting. Closing as a duplicate.

Issue #1833, opened 2026-06-26
[Request]: Integration tests are fail frequently and are difficult to maintain.


jglogan:
### Feature or enhancement request details

Many causes for this, all stemming from our current test support infrastructure being set up early in the project and slowly becoming inadequate for all we've thrown at it:

- Inadequate identification and isolation of tests that need to run serially.
- Patchwork approach to test concurrency management.
- Patchwork approach to test isolation and resource management.

The tests also run painfully slowly. There are several fixes we can apply here but the one that we should address immediately is part of improving test resource management. We should identify images we use commonly and preload these into the local image store, so that we don't have many tests all trying to fetch them from the remote repository at once.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1834, opened 2026-06-26
Enhanced test fixtures for integration tests.

jglogan:
- Part of #1833.
- Adds `ContainerFixture` with scoped resource lifecycle and cleanup in place of implementation inheritance for test support functions. The fixture also handles resource prefixing and uses a more ergonomic `CommandResult` in place of a tuple for return values.
- `ImageWarmup` suite pre-pulls well-known images, and `copyWarmupImage()` tags test-local refs, keeping the canonical image store untouched.
- Three-phase `integration-new`: warmup, followed by concurrent tests (managed by the swift test `--experimental-maximum-parallelization-width` flag), followed by serialized tests.
- `coverage-new` merges unit + integration-new profraw, replacing `coverage` in CI as a migration progress indicator.
- Updates GH workflow so non-coverage invokes both the `integration` and `integration-new` Makefile targets, while coverage runs invoke the `coverage-new` target.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Recent CI integration test issues and test time increase, especially after adding `container machine` tests.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -151,7 +151,7 @@ dsym:
 
 .PHONY: test
 test:
-	@$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --skip TestCLI
+	@$(SWIFT) test -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --skip TestCLI --skip IntegrationTests
 
 .PHONY: install-kernel
 install-kernel:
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -194,6 +194,56 @@ define GENERATE_COV_REPORTS
...
+INTEGRATION_SWIFT_EXTRA ?=
+INTEGRATION_POST_TEST ?=
+
+define RUN_INTEGRATION
+	@echo Ensuring apiserver stopped before the CLI integration tests...
+	@bin/container system stop && sleep 3 && scripts/ensure-container-stopped.sh
+	@if [ -n "$(APP_ROOT)" ]; then \
+		echo "Clearing application data under $(APP_ROOT) (preserving kernels)..." ; \
+		mkdir -p $(APP_ROOT) ; \
+		find "$(APP_ROOT)" -mindepth 1 -maxdepth 1 ! -name kernels -exec rm -rf {} + ; \
```
katiewasnothere, line 215: I would prefer we also remove the kernel just so we're guaranteed to have a clean setup.

katiewasnothere (approved):

Makefile
jglogan: We could add some sort of preserve-kernel env in a follow-up and make the default the slower setup option.

PR #1838, opened 2026-06-26
fix(plugins): bound discovery when apiserver stalls

stephenlclarke:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Plugin discovery can hang before root help or plugin dispatch completes when `com.apple.container.apiserver` is registered but not answering XPC requests. This affects paths such as `container --help`, `container help`, bare `container`, and `container compose ...`.

Addresses #1459.

Related to #995, but this PR intentionally keeps a narrower scope:
- #995 also changes apiserver startup/default-network behavior and adds broader NetworkClient timeout policy.
- This PR only bounds plugin discovery/help and fixes the shared `XPCClient.send(... responseTimeout:)` timeout primitive.
- This PR keeps XPC clients reusable after a timed-out request; late replies are ignored by a single-resume gate instead of closing the connection.
- This PR adds deterministic command-health and anonymous-XPC regression tests.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Validation:
- `swift test --disable-automatic-resolution --enable-code-coverage --filter 'ApplicationHealthTests|XPCClientTests|HelpCommandTests'`
- `make swift-fmt-check`
- `git diff --check`

The focused coverage suite covers the new timeout/race behavior, root/default/help command paths, XPC timeout return, nil-route timeout messaging, normal XPC replies, and client reuse after timeout.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -150,6 +150,7 @@ let package = Package(
         .testTarget(
             name: "ContainerCommandsTests",
             dependencies: [
+                "ContainerAPIClient",
                 "ContainerCommands",
                 "ContainerResource",
             ]
@@ -472,6 +473,12 @@ let package = Package(
                 .product(name: "ContainerizationExtras", package: "containerization"),
                 .product(name: "Logging", package: "swift-log"),
                 "CAuditToken",
+            ]
+        ),
+        .testTarget(
+            name: "ContainerXPCTests",
+            dependencies: [
+                "ContainerXPC"
             ]
         ),
         .target(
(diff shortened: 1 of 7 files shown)


stephenlclarke:
Superseded by the current open, ready-for-review replacement #1935. The replacement imports the preferred XPC cancellation work from #1862 in its own commit, adds deterministic coverage, and keeps the remaining root-help policy delta focused.

PR #1848, opened 2026-06-27
Adds build fixture and migrates build CLI tests.

jglogan:
- Part of #1833.
- Also changes test directory names to include test ID so that they are easily identifiable.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Improve integration test reliability.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs


Tests/IntegrationTests/Build/BuildFixture.swift
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// MARK: - Build context helpers
+
+extension ContainerFixture {
+
+    /// Creates a new scratch directory under ``testDir`` and returns its path.
+    ///
+    /// The directory is removed automatically when the fixture scope exits.
+    func createTempDir() throws -> FilePath {
```
jglogan, line 155: I think we can get rid of this in a final cleanup pass. Since the tests themselves run in test-specific temp directories, creating a temp directory underneath is just unneeded boilerplate.

katiewasnothere (approved):
There's a couple places where we could be using helper functions (such as builderDelete) and a couple places we could refactor some helper functions to make them more versatile (like the helper for outputting to local dir). Since this is a big move, I think those things can be cleaned up later, so this PR LGTM.

PR #1868, opened 2026-06-29
Migrate image, volume and miscellaneous system tests.

jglogan:
- Part of #1833.
- Distributes fixture extensions for images and containers added as part of previous builder PR to more sensible locations.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Improve integration test reliability.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs


Tests/IntegrationTests/Utilities/ContainerFixture+SystemHelpers.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+// MARK: - System helpers
+
+extension ContainerFixture {
+    /// Creates a temporary directory under ``testDir``, passes its `URL` to `body`,
+    /// then removes it when `body` exits (cleanup handled by the fixture scope).
+    func withTempDir<T>(_ body: (URL) async throws -> T) async throws -> T {
+        let dir = URL(filePath: testDir.appending(UUID().uuidString).string)
+        try FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
```
JaewonHur: Should delete the directory on exit?

Tests/IntegrationTests/System/TestCLIKernelSetSerial.swift
```diff
@@ -0,0 +1,88 @@
...
+            .appending(path: "vmlinux.container")
+            .relativePath
+
+        try await ContainerFixture.with { f in
+            f.addCleanup { _ = try? f.run(["system", "kernel", "set", "--recommended", "--force"]) }
+            try await f.withTempDir { tempDir in
+                let localTarPath = tempDir.appending(path: remoteTar.lastPathComponent)
+                try await ContainerAPIClient.FileDownloader.downloadFile(url: remoteTar, to: localTarPath)
+                try f.run(["system", "kernel", "set", "--force", "--tar", localTarPath.path, "--binary", symlinkBinaryPath]).check()
+                try await validateContainerRun(f)
```
katiewasnothere: This isn't actually validating that the container is running with the kernel that was set above though right?

Tests/IntegrationTests/Volumes/TestCLIVolumesSerial.swift
```diff
@@ -0,0 +1,110 @@
...
+                f.doVolumeDeleteIfExists(v2)
+            }
+
+            try f.doVolumeCreate(v1)
+            try f.doVolumeCreate(v2)
+            let list = try f.run(["volume", "list", "--quiet"]).check().output
+            #expect(list.contains(v1) && list.contains(v2))
+
+            let result = try f.run(["volume", "prune"]).check()
+            #expect(result.output.contains(v1) || !result.output.contains("No volumes to prune"))
```
JaewonHur: Shouldn't it check `result.output.contains(v1)` only?

Tests/IntegrationTests/Volumes/TestCLIAnonymousVolumes.swift
```diff
@@ -0,0 +1,251 @@
...
+
+            try f.doStop(c)
+            try f.doRemove(c)
+            let result = try f.run(["volume", "rm", volumeID])
+            #expect(result.status == 0, "manual deletion of unmounted anonymous volume should succeed")
+            #expect(!(try f.volumeExists(volumeID)))
+        }
+    }
+
+    @Test func testAnonymousVolumeDetachedMode() async throws {
```
JaewonHur, line 234: Not critical, but the original test was setting `autoRemote: true` so that container is deleted on stop. New version slightly changes the test semantics.

Tests/IntegrationTests/System/TestCLIKernelSetSerial.swift
```diff
@@ -0,0 +1,88 @@
...
+    // MARK: - Tests
+
+    @Test func fromLocalTar() async throws {
+        let symlinkBinaryPath = URL(filePath: defaultBinaryPath)
+            .deletingLastPathComponent()
+            .appending(path: "vmlinux.container")
+            .relativePath
+
+        try await ContainerFixture.with { f in
+            f.addCleanup { _ = try? f.run(["system", "kernel", "set", "--recommended", "--force"]) }
```
JaewonHur: Should catch and log if this fails? because it affects following tests.

Tests/IntegrationTests/Volumes/TestCLIVolumes.swift
```diff
@@ -0,0 +1,269 @@
...
+struct TestCLIVolumes {
+    private let alpine = ContainerFixture.warmupImages[0]
+
+    @Test func testVolumeDataPersistenceAcrossContainers() async throws {
+        try await ContainerFixture.with { f in
+            let vol = "\(f.testID)-vol"
+            let c1 = "\(f.testID)-c1"
+            let c2 = "\(f.testID)-c2"
+            try f.doPull(alpine)
+            let image = try f.copyWarmupImage(alpine)
```
katiewasnothere: Why do we need to make copies of the alpine image for these tests?

Tests/IntegrationTests/Utilities/ContainerFixture+SystemHelpers.swift
jglogan: According to the docc, yes.

It exists in the fixture scratch directory which gets deleted on exit from the fixture scope though (unless you specify the environment variable for preserving the scratch directories).

Let me review the use cases for this helper and figure out what makes the most sense.

jglogan: This pattern resulted from literally translating the old tests to the new fixtures.

The new fixtures automatically run in a temp directory, so we can just get rid of the helper and have the tests operate directly in the scratch directory.

I noticed the same thing on the build tests but decided to leave those as is until we merged everything. Then, I've got a whole list of cleanups that remove unnecessary code like this.

Tests/IntegrationTests/System/TestCLIKernelSetSerial.swift
jglogan: This is a weakness in the original test. I'll document the test better and change the name so that it doesn't mislead, and we can also run `uname -r` as a rough check.

Tests/IntegrationTests/Volumes/TestCLIVolumesSerial.swift
jglogan: Prune should delete both `v1` and `v2`. Checking for those should suffice.

The "no volumes" thing is weird. I'm not sure why the absence of "No volumes" passes the test if the output doesn't contain `v1. Let's remove those.

Tests/IntegrationTests/Volumes/TestCLIAnonymousVolumes.swift
jglogan: Good catch...updated to match the old test.

Tests/IntegrationTests/System/TestCLIKernelSetSerial.swift
jglogan: Agree. Changed to record test issues on non-zero status or throw.

Tests/IntegrationTests/Volumes/TestCLIVolumes.swift
jglogan: We don't. Claude got a little overexuberant I think. We only need to make a copy if we intend to delete the image as part of the test.

katiewasnothere (approved):

PR #1922, opened 2026-07-08
Fix build context symlink resolution and XPC concurrency

### LINKED ISSUE #1899: [Bug]: `container build` ignores default/relative build context in `mktemp` directory, but works with absolute path
### I have done the following

### Steps to reproduce

```sh
cd "$(mktemp -d)"

cat > Dockerfile <<'EOF'
FROM debian:stable-slim
COPY . .
RUN cat Dockerfile
EOF

container build
```

The build fails with:

```text
cat: Dockerfile: No such file or directory
```

The same happens with:

```sh
container build .
```

But this succeeds:

```sh
container build "$PWD"
```

In my specific example reproduction, the `mktemp -d` resolved to `/var/folders/qf/ng024_6s5kn4z_sczdcdbbkh0000gn/T/tmp.3tuwoUOXS3`.

### Problem description

When building from a temporary directory created with `mktemp -d`, the build context appears to be empty if the context is specified implicitly or as `.`.

The following commands all produce an empty build context:

`container build`
`container build .`

However, providing the exact same directory as an absolute path works:

`container build "$PWD"`

The behavior does not occur when building from a normal directory such as `~/test`.

### Environment

```markdown
- OS: ProductName:            macOS
      ProductVersion:         26.5.2
      BuildVersion:           25F84
```

### Code of Conduct

lakshitsoni26:
Fixes #1899
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This PR resolves two critical regressions in the build system:
1. **`ADD`/`COPY` Symlink Context Resolution:** Fixes an issue where `tar` buildkit generation was failing to properly include and resolve symlinked files in the build context. In `URL+Extensions.swift`, the path hierarchy functions (`relativeChildPath` and `parentOf`) were overly aggressively resolving symlinks via `.standardizedFileURL`. This is fixed by implementing a safe `stripPrivatePrefix` function. In `BuildFSSync.swift`, `FileInfo.init` was incorrectly attempting to resolve paths instead of reading the raw target from disk, which has now been fixed using `destinationOfSymbolicLink(atPath:)`.
2. **XPC Concurrency Hangs:** Fixes a concurrency deadlock/hang issue within `XPCClient` when iterating over asynchronous stream states under high loads.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

**Local Test Output:**
Ran `make cli && swift test --filter TestCLIBuilderSerial` end-to-end:
`✔ Suite TestCLIBuilderSerial passed after 213.792 seconds.`
`✔ Test run with 44 tests in 1 suite passed after 213.792 seconds.`

Screenshot 2026-07-09 at 12 19 39 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -285,9 +285,9 @@ define RUN_INTEGRATION
 		echo "==> Warmup pass" && \
 		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter "$(WARMUP_FILTER)" && \
 		echo "==> Concurrent pass (width=$(PARALLEL_WIDTH))" && \
-		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --experimental-maximum-parallelization-width $(PARALLEL_WIDTH) --filter "$(CONCURRENT_FILTER)" && \
+		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --parallel --num-workers $(PARALLEL_WIDTH) --filter "$(CONCURRENT_FILTER)" && \
 		echo "==> Global pass (serial)" && \
-		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --experimental-maximum-parallelization-width 1 --filter "$(SERIAL_FILTER)" ; \
+		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --no-parallel --filter "$(SERIAL_FILTER)" ; \
 		exit_code=$$? ; \
 		$(INTEGRATION_POST_TEST) \
 		echo Ensuring apiserver stopped after the CLI integration tests ; \
--- a/Sources/ContainerBuild/URL+Extensions.swift
+++ b/Sources/ContainerBuild/URL+Extensions.swift
@@ -54,9 +54,16 @@ extension URL {
         self.path.fs_cleaned
     }
 
+    private func stripPrivatePrefix(_ path: String) -> String {
+        if path.hasPrefix("/private/") {
+            return String(path.dropFirst(8))
+        }
+        return path
+    }
+
     func parentOf(_ url: URL) -> Bool {
-        let parentPath = self.absoluteURL.cleanPath
-        let childPath = url.absoluteURL.cleanPath
+        let parentPath = stripPrivatePrefix(self.absoluteURL.cleanPath)
+        let childPath = stripPrivatePrefix(url.absoluteURL.cleanPath)
 
         guard parentPath.fs_isAbsolute else {
             return true
@@ -74,15 +81,15 @@ extension URL {
             throw BuildFSSync.Error.pathIsNotChild(cleanPath, context.cleanPath)
         }
 
-        let ctxParts = context.cleanPath.fs_components
-        let selfParts = cleanPath.fs_components
+        let ctxParts = stripPrivatePrefix(context.absoluteURL.cleanPath).fs_components
+        let selfParts = stripPrivatePrefix(self.absoluteURL.cleanPath).fs_components
 
         return selfParts.dropFirst(ctxParts.count).joined(separator: "/")
     }
 
     func relativePathFrom(from base: URL) -> String {
-        let destParts = cleanPath.fs_components
-        let baseParts = base.cleanPath.fs_components
+        let destParts = stripPrivatePrefix(self.absoluteURL.cleanPath).fs_components
+        let baseParts = stripPrivatePrefix(base.absoluteURL.cleanPath).fs_components
 
         let common = zip(destParts, baseParts).prefix { $0 == $1 }.count
         guard common > 0 else { return cleanPath }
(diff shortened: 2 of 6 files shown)


lakshitsoni26:
### 🐛 Root Cause Analysis: `ADD/COPY` Symlink Context Resolution

For context, this PR fixes a deep path-resolution bug that was causing the `buildkit` tarball generation to silently strip symlinks from the build context, resulting in `No such file or directory` errors during `Dockerfile` execution.

**The Bug:**
During build context generation, `BuildFSSync.swift` relies on methods in `URL+Extensions.swift` (specifically `relativeChildPath` and `parentOf`) to map file hierarchy. Previously, these methods unconditionally evaluated paths using `.standardizedFileURL`.

Because `.standardizedFileURL` triggers the system's `realpath` equivalent, it immediately resolves symlinks on the disk. 
1. When a symlink like `Test3Source2/Dest` (pointing to `../Test3Source/Source`) was passed in, `.standardizedFileURL` resolved it instantly.
2. The symlink's actual name (`Dest`) was discarded in memory, and the path was mapped as its destination (`Test3Source/Source`).
3. Consequently, the tarball generated for `buildkit` completely omitted the symlink, replacing it with a duplicate of its target directory.
4. When `RUN cat Test3Source2/Dest/test.txt` executed in the container, it failed because `Dest` was never sent to the daemon.

**The Fix:**
The original usage of `.standardizedFileURL` was a heavy-handed workaround to handle macOS temporary directory mismatches (i.e., `/var/` vs `/private/var/`). 

To safely fix the mismatch without aggressively resolving workspace symlinks:
1. Replaced `.standardizedFileURL` in `URL+Extensions.swift` with a targeted `stripPrivatePrefix` helper that only normalizes the root system prefixes.
2. Updated `FileInfo.init` in `BuildFSSync.swift` to stop attempting manual path resolution for symlinks, and instead directly extract the raw target string from the filesystem using `FileManager.default.destinationOfSymbolicLink(atPath:)`.

This ensures the exact symlink structures are preserved in the `tar` payload sent to `buildkit`.

svelez:
Hey... just wanted to point out https://github.com/apple/container/issues/1899#issuecomment-5039231087

Not sure if you would like to include that in this PR or if I should work on addressing independenly

Sources/ContainerBuild/URL+Extensions.swift
```diff
@@ -54,9 +54,16 @@ extension URL {
         self.path.fs_cleaned
     }
 
+    private func stripPrivatePrefix(_ path: String) -> String {
```
jglogan, line 57: What is the rationale for this? Stripping `/private/` is quite arbitrary and likely has nothing to do with what the perceived problem is.

Makefile
```diff
@@ -285,9 +285,9 @@ define RUN_INTEGRATION
 		echo "==> Warmup pass" && \
 		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --filter "$(WARMUP_FILTER)" && \
 		echo "==> Concurrent pass (width=$(PARALLEL_WIDTH))" && \
-		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --experimental-maximum-parallelization-width $(PARALLEL_WIDTH) --filter "$(CONCURRENT_FILTER)" && \
+		$(SWIFT) test $(INTEGRATION_SWIFT_EXTRA) -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --parallel --num-workers $(PARALLEL_WIDTH) --filter "$(CONCURRENT_FILTER)" && \
```
jglogan, line 288: Don't change this. `--parallel` does not do the same thing as `--experimental-maximum-parallelization-width`.

Sources/ContainerXPC/XPCClient.swift
```diff
@@ -95,45 +95,61 @@ extension XPCClient {
         XPCClientSession(client: self)
     }
 
+    private actor ContinuationResolver {
```
jglogan, line 98: Why have a PR for two (three?) completely unrelated things?

lakshitsoni26:
> Hey... just wanted to point out [#1899 (comment)](https://github.com/apple/container/issues/1899#issuecomment-5039231087)
> 
> Not sure if you would like to include that in this PR or if I should work on addressing independenly

I apologize for bundling these together! I will split this up. I'll revert the Makefile changes immediately, and I will separate the XPC Concurrency fix and the Symlink Context fix into two independent PRs so they are easier to review.

lakshitsoni26:
Hi @svelez and @jglogan, thank you both for the helpful feedback on this!

You are completely right—bundling these three unrelated changes into a single PR made this too difficult to review and caused unnecessary conflicts. To make your lives easier, I have completely split this work up:

1. **Symlink Context Resolution (Fixes #1899):** I see this was recently resolved internally upstream, so I am dropping my commit for it here to avoid conflicts.
2. **XPC Concurrency Deadlock:** I have extracted this into a clean, independent PR here: #1987
3. **Makefile Test Flags:** I have extracted the SwiftPM flag updates into a clean, independent PR here: #1988

I am closing this bundled PR in favor of the clean ones above. Thanks again for the guidance on keeping PRs perfectly scoped!

Makefile
lakshitsoni26: Hi @jglogan, thanks for taking a look at this! 

The rationale for the change was that `--experimental-maximum-parallelization-width` is an older flag that can sometimes trigger "unknown argument" or deprecation warnings on newer Swift toolchains (Swift 5.10 / Swift 6). I had assumed that `--parallel --num-workers` was the modern, direct replacement for it.

However, if `--experimental-...` is triggering a very specific internal behavior for this project's test harness that `--parallel` lacks, I completely defer to your judgment! 

My only goal was to clean up the integration test commands and remove some of the verbose JSON stream outputs. If you prefer we keep the flags exactly as they are to maintain test stability, just let me know and I am more than happy to close this PR.

PR #1926, opened 2026-07-09
Fix exec hang and stop timeout

### LINKED ISSUE #1916: [Bug]: container exec/stop hang indefinitely on SSH reconnect, requiring full macOS restart to recover
### I have done the following

### Steps to reproduce

1. Start a container with a bind mount: container run -d --name test0 -v <host-path>:/code ubuntu:latest
2. Install openssh-server inside it and configure key-based SSH access.
3. Connect via SSH (e.g. VS Code Remote-SSH, or plain `ssh`). This first connection works fine.
4. Disconnect (e.g. restart VS Code, or point Remote-SSH at a different folder within the same container).
5. Attempt to reconnect via SSH, or run `container exec -it test0 bash` directly.
6. The exec/SSH session hangs indefinitely — no prompt, no error, no timeout.
7. `container stop test0` also hangs with no output.
8. `container list --all` continues to report the container as "running" throughout.

In a second occurrence the same session, `container system stop` also hung (log showed `stopTimeoutSeconds=5 ... stopping containers` then nothing further — the stated timeout was not enforced), and `launchctl kickstart -k <container-apiserver-label>` did not recover it either. A full macOS restart was the only thing that restored a working state.

Separately, `ps aux` showed multiple stuck `container exec` client processes accumulated from repeated attempts (none had exited on their own), while `container-runtime-linux` and `container-apiserver` processes appeared healthy — suggesting the hang is in the exec/PTY session layer, not the VM or daemon itself. `ping` to the affected container's IP succeeded instantly while `exec` was hung, ruling out a network-layer cause.

### Problem description

container exec (and subsequently container stop / container system stop) can hang indefinitely, specifically triggered by reconnecting to a container via SSH/exec — the first connection after starting/configuring the container works fine, but a second or later connection attempt reliably triggers the hang. This suggests something isn't being cleaned up when an exec/SSH session ends, and the leftover state blocks the next session.

Separately, `container list --all` reported a container as "running" while `container exec` on that same container immediately failed with `cannot exec: container is not running` — a state-tracking inconsistency between the two commands.

Expected behavior: exec/SSH sessions should be able to reconnect repeatedly without hanging, and container stop should always respect its stated timeout rather than hanging indefinitely and requiring a full OS restart to recover.

### Environment

```markdown
* OS: macOS 26.5.2 (Tahoe)
* Hardware: Apple Silicon (M4)
* Container: Container CLI version 1.1.0 (installed via Homebrew)
* Image: docker.io/library/ubuntu:latest (Ubuntu 26.04 LTS, arm64)
```

### Code of Conduct

lakshitsoni26:
Fixes #1916 

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
This PR resolves a critical bug where `container exec` and `container stop` commands hang indefinitely when an SSH session disconnects abruptly (or the terminal window is closed without cleanly exiting). 

The underlying issues were identified and addressed in two distinct commits:

1. **PTY Resource Leak / Orphaned Process:** When an XPC client connection dropped unexpectedly (e.g., SIGHUP, broken pipe), the corresponding exec process inside the container remained orphaned. Subsequent `exec` attempts would try to acquire the same resources, hit a blocking read/write on a dead file descriptor, and hang indefinitely. This was fixed by registering an `onDisconnect` listener in `ContainersHarness` to forcefully kill the specific orphaned process upon a client disconnect.
2. **Ignored Stop Timeout:** The `container stop` command completely ignored the `stopTimeoutSeconds` flag if an exec session was blocking the `RuntimeService` actor lock. This PR wraps the lock acquisition and the graceful stop sequence inside a `withThrowingTaskGroup`. This ensures that even if an exec session is infinitely holding the lock, the container is forcefully killed if it doesn't shut down within the designated 5-second timeout window.

Fixing these primary issues also inherently prevents the secondary symptom where `container list` incorrectly reports the container as `running` while `container exec` fails.

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

**Local Test Verification:**
```text
# Simulated a SIGHUP abrupt disconnect to an active SSH/bash session, 
# then immediately started a new exec session. It connected instantly without hanging:
./.build/debug/container exec -it test0 bash
root@test0:/# echo SUCCESS_EXEC
SUCCESS_EXEC

# Verified the stop timeout enforcement triggers gracefully:
./.build/debug/container stop test0
test0
./.build/debug/container stop test0  0.02s user 0.01s system 0% cpu 5.875 total

# Verified correct state sync:
./.build/debug/container list --all
ID     IMAGE                            OS     ARCH   STATE    STARTED
test0  docker.io/library/ubuntu:latest  linux  arm64  stopped  2026-07-09T11:22:51Z

### DIFF
--- a/Sources/APIServer/APIServer+Start.swift
+++ b/Sources/APIServer/APIServer+Start.swift
@@ -298,7 +298,7 @@ extension APIServer {
             routes[XPCRoute.containerDial] = XPCServer.route(harness.dial)
             routes[XPCRoute.containerStop] = XPCServer.route(harness.stop)
             routes[XPCRoute.containerStartProcess] = XPCServer.route(harness.startProcess)
-            routes[XPCRoute.containerCreateProcess] = XPCServer.route(harness.createProcess)
+            routes[XPCRoute.containerCreateProcess] = harness.createProcess
             routes[XPCRoute.containerResize] = XPCServer.route(harness.resize)
             routes[XPCRoute.containerWait] = XPCServer.route(harness.wait)
             routes[XPCRoute.containerKill] = XPCServer.route(harness.kill)
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersHarness.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersHarness.swift
@@ -203,7 +203,7 @@ public struct ContainersHarness: Sendable {
     }
 
     @Sendable
-    public func createProcess(_ message: XPCMessage) async throws -> XPCMessage {
+    public func createProcess(_ message: XPCMessage, session: XPCServerSession) async throws -> XPCMessage {
         let id = message.string(key: .id)
         guard let id else {
             throw ContainerizationError(
@@ -227,6 +227,11 @@ public struct ContainersHarness: Sendable {
             config: config,
             stdio: stdio
         )
+
+        await session.onDisconnect { [service] in
+            // Clean up the process if the client connection drops unexpectedly
+            try? await service.kill(id: id, processID: processID, signal: "SIGKILL")
+        }
 
         return message.reply()
     }
--- a/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
+++ b/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
@@ -583,7 +583,7 @@ public actor RuntimeService {
         let id = try message.id()
         let signal = try Signal(message.signal())
 
-        try await self.lock.withLock { [self] _ in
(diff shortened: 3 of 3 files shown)


lakshitsoni26:
Fix clientIsStopped error when sending signal to dead agent

When the guest agent disconnects unexpectedly (e.g. because of the PTY closing on the host), the host tries to forward SIGKILL to the agent but hits a `clientIsStopped` exception from GRPC, leaving waiters hung. This catches that exception and cleanly releases waiters.

Credit: Bug found by @stephenlclarke.

Issue #1936, opened 2026-07-11
[Bug]: debug flag not accepted


larrypl:
### I have done the following

- [x] I have searched the existing issues
- [x] If possible, I've reproduced the issue using the 'main' branch of this project

### Steps to reproduce

```
$ container s status --debug
Error: Unknown option '--debug'. Did you mean '--debug'?
Usage: container system status [--prefix <prefix>] [--format <format>] [--debug]
  See 'container system status --help' for more information.
```

### Problem description

debug flag not accepted

### Environment

```markdown
- OS: MacOS 26
- Xcode: 
- Container:
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1938, opened 2026-07-12
Fix --debug rejected on many subcommands

### LINKED ISSUE #360: [Bug]: "Error: Unknown option '--debug'. Did you mean '--debug'?"

mliem2k:
## Summary

Fixes #360.

`container system status --debug` (and many other subcommands) rejected `--debug` with:

```
Error: Unknown option '--debug'. Did you mean '--debug'?
Usage: container system status [--prefix <prefix>] [--format <format>] [--debug]
```

even though `--debug` is listed in the command's own usage text.

**Root cause:** `Application`'s `CommandConfiguration` set `defaultSubcommand: DefaultCommand.self` (used to dispatch unrecognized top-level input to plugins). `DefaultCommand` declares `@Argument(parsing: .captureForPassthrough) var remaining: [String]`. That specific combination — `defaultSubcommand` pointing at a command with a `.captureForPassthrough`/`.allRemainingInput` argument — breaks ArgumentParser's flag recognition for unrelated, legitimately-matched subcommands throughout the whole tree, not just the default/plugin path.

I confirmed this against a minimal, isolated `swift-argument-parser` reproduction (root command with `defaultSubcommand` + `.captureForPassthrough`, and nested nested subcommands several levels down) — the bug reproduces on ArgumentParser 1.7.0 through the latest 1.8.2, and disappears immediately when `defaultSubcommand` is removed. It is *not* about the leaf command's own option declarations — commands with `logOptions` declared in different orders/positions, with or without extra flags, are affected identically as long as `defaultSubcommand` is present at the root.

In `apple/container` this affected (at least): `system status/version/df`, `builder status/delete`, `image inspect/delete/pull/tag/save`, `network inspect`, `volume inspect` — all rejected `--debug` when it appeared after the subcommand name.

## Fix

- Remove `defaultSubcommand` from `Application`'s `CommandConfiguration`.
- Add `Application.pluginDispatchArguments(_:)`, which inspects the raw args *before* calling `parseAsRoot` and decides whether they should go to `DefaultCommand` (plugin dispatch, or the plugin-aware bare/`--debug`-only help text) or be parsed normally as one of `Application`'s own subcommands. This preserves `DefaultCommand`'s existing plugin-dispatch behavior — including passing a plugin's own `--debug` flag through to it unchanged, rather than swallowing it (which is what I found happens if you instead switch `DefaultCommand` to ArgumentParser's `.allUnrecognized` strategy — a plugin invoking `container my-plugin --debug` would silently lose its own `--debug` to the root's global flag).
- `main()` also replicates the `Application.validate()` side effects (debug log level, Rosetta check) that would otherwise have run on the way to the default subcommand, so plugin dispatch keeps the same behavior for `container --debug some-plugin ...`.

## Before / after

```
$ container s status --debug
Error: Unknown option '--debug'. Did you mean '--debug'?
Usage: container system status [--prefix <prefix>] [--format <format>] [--debug]
  See 'container system status --help' for more information.
```

After the fix (daemon not running in this sandbox, but the flag is now recognized and parsing succeeds):

```
$ container s status --debug
apiserver is not running and not registered with launchd
```

Verified unchanged behavior for: bare `container` (still shows the plugin-aware help with the `PLUGINS:` section), `container --help`/`-h`, `container --version`, `container --debug`, unknown-plugin dispatch (`container some-plugin --flag arg`, including a plugin's own `--debug` passed through untouched), and normal subcommands like `container image list --debug` / `container list --debug`.

## Testing

- `swift build` succeeds.
- Added `Tests/ContainerCommandsTests/PluginDispatchArgumentsTests.swift` (unit tests for the new dispatch-routing logic).
- Added `Tests/IntegrationTests/System/TestCLIDebugFlag.swift` (regression coverage running the actual CLI for the commands reported in #1936).
- Ran `swift test` for the affected suites: `ContainerCommandsTests`, `TestCLIHelp`, `TestCLIStatus`, `TestCLIPluginErrors`, `TestCLIDebugFlag`, `PluginDispatchArgumentsTests` — all 47 tests pass.

## Test plan

- [x] `swift build` succeeds
- [x] `container system status --debug` no longer errors on `--debug`
- [x] `container --help`, `-h`, `--version`, bare `container` unchanged
- [x] Plugin dispatch unchanged, including a plugin's own `--debug` passed through
- [x] New and existing relevant tests pass (`swift test --filter ...`)

### DIFF
--- a/Sources/ContainerCommands/Application.swift
+++ b/Sources/ContainerCommands/Application.swift
@@ -94,9 +94,14 @@ public struct Application: AsyncLoggableCommand {
                 name: "Other",
                 subcommands: Self.otherCommands()
             ),
-        ],
-        // Hidden command to handle plugins on unrecognized input.
-        defaultSubcommand: DefaultCommand.self
+        ]
+        // Note: intentionally no `defaultSubcommand`. Routing unrecognized
+        // top-level input to `DefaultCommand` via `defaultSubcommand` combined
+        // with its `.captureForPassthrough` argument breaks flag parsing for
+        // unrelated, legitimately-matched subcommands throughout the tree
+        // (e.g. `container system status --debug` would fail to recognize
+        // `--debug`). See https://github.com/apple/container/issues/1936.
+        // Plugin dispatch is instead handled manually in `main()` below.
     )
 
     public static func main() async throws {
@@ -116,6 +121,30 @@ public struct Application: AsyncLoggableCommand {
 
         let fullArgs = CommandLine.arguments
         let args = Array(fullArgs.dropFirst())
+
+        // If the input doesn't correspond to one of our own subcommands, hand
+        // it off to `DefaultCommand` for plugin dispatch (or to show the
+        // plugin-aware help text) before ever invoking `Application.parseAsRoot`.
+        // This intentionally avoids ArgumentParser's `defaultSubcommand`
+        // mechanism; see the note on `configuration` above.
+        if let dispatch = Self.pluginDispatchArguments(args) {
+            do {
+                // Mirror the `validate()` side effects (debug log level,
+                // Rosetta check) that `Application.parseAsRoot` would
+                // otherwise have applied on its way to the default subcommand.
+                var app = Application()
+                app.logOptions = dispatch.logOptions
+                try app.validate()
+
+                var defaultCommand = DefaultCommand()
+                defaultCommand.logOptions = dispatch.logOptions
+                defaultCommand.remaining = dispatch.remaining
+                try await defaultCommand.run()
+            } catch {
+                Application.exit(withError: error)
+            }
+            return
+        }
 
         do {
             var command = try Application.parseAsRoot(args)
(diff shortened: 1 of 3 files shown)


jglogan:
@mliem2k We're still working on some issues that are preventing us from spending more time on contributor PRs, but in the meantime I closed your issue as a dupe of #360 and updated your PR description to match.

There is another PR open for this but at first glance, what you've done seems like it could be a better approach.

PR #1956, opened 2026-07-16
Validate container ID from XPC requests

katiewasnothere:
## Type of Change
- [x] Bug fix

## Motivation and Context
Validate XPC message provided container IDs in ContainersHarness to prevent file path traversal when container ID is invalid. 

## Testing
- [x] Tested locally

### DIFF
--- a/Sources/ContainerResource/Container/ManagedContainer.swift
+++ b/Sources/ContainerResource/Container/ManagedContainer.swift
@@ -41,8 +41,7 @@ public struct ManagedContainer: ManagedResource {
     /// not the protocol's 64-hex default.
     public static func generateId() -> String { UUID().uuidString.lowercased() }
 
-    /// Container name rule (mixed case, dots, hyphens, underscores). Duplicated from
-    /// Utility.validEntityName
+    /// Container name rule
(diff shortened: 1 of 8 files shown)


Sources/ContainerResource/Container/ManagedContainer.swift
```diff
@@ -41,9 +41,11 @@ public struct ManagedContainer: ManagedResource {
     /// not the protocol's 64-hex default.
     public static func generateId() -> String { UUID().uuidString.lowercased() }
 
-    /// Container name rule (mixed case, dots, hyphens, underscores). Duplicated from
-    /// Utility.validEntityName
+    /// Container name rule
     public static func nameValid(_ name: String) -> Bool {
+        guard name.count < 63 else {
```
jglogan:
```suggestion
        // Maximum Linux hostname length is 64, but limit to maximum DNS label length
        guard name.count =< 63 else {
```

Do we have a unit test that checks the length constraint?

jglogan (changes requested):

jglogan (approved):

PR #1957, opened 2026-07-16
Do not follow destination symlink when copying user configuration

katiewasnothere:
## Type of Change
- [x] Bug fix

## Motivation and Context
This PR fixes a narrow case where a user may have modified the user configuration file in the application root to use a symlink. In this case, upon restarting the service, we would follow the symlink, potentially replacing files unintentionally. This PR fixes this issue by using the `rename(2)` which will override the `destPath` regardless of its file permissions and status as a symlink or otherwise. 

## Testing
- [x] Tested locally
- [x] Added/updated tests

### DIFF
--- a/Sources/ContainerPersistence/ConfigurationLoader.swift
+++ b/Sources/ContainerPersistence/ConfigurationLoader.swift
@@ -17,6 +17,7 @@
 import Configuration
 import ConfigurationTOML
 import ContainerizationError
+import Darwin
 import Foundation
 import SystemPackage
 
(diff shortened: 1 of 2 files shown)


Sources/ContainerPersistence/ConfigurationLoader.swift
```diff
@@ -198,21 +198,28 @@ public enum ConfigurationLoader {
         let destPath = configurationFile(in: destBase)
 
         let fm = FileManager.default
-
-        if fm.fileExists(atPath: destPath.string) {
```
jglogan, line 202: I'm not sure I understand. Was removeItem following the symlink and removing the file at the target path?

jglogan: Or that the symlink was pointing nowhere and fileExists follows the symlink and then returns false?

jglogan (approved):

PR #1987, opened 2026-07-22
Fix XPCClient ContinuationResolver deadlock

### LINKED ISSUE #2007: [Bug]: XPCClient causes fatal SWIFT TASK CONTINUATION MISUSE crash on delayed replies
### I have done the following

### Steps to reproduce

This race condition can be reliably reproduced by injecting an artificial delay into the C-level libxpc callback while calling XPCClient.send().

1. Check out the main branch.
2. In any XPC test file (e.g., Tests/ContainerXPCTests/XPCClientTests.swift), add a test that configures the mock XPC connection to delay its reply for 3 seconds:

    func testXPCClientTimeoutRaceCondition() async throws {
        // 1. Setup mock connection to sleep for 3.0s before replying
        let client = XPCClient(connection: slowMockConnection) 
        
        do {
            // 2. Call with a shorter timeout (1.0s)
            _ = try await client.send(message, timeout: 1.0)
        } catch {
            print("Task successfully timed out") 
        }
        
        // 3. Keep the process alive to wait for the delayed C callback to finally fire
        try await Task.sleep(nanoseconds: 3_000_000_000)
        
        // -> FATAL CRASH occurs here: SWIFT TASK CONTINUATION MISUSE
    }

3. Run the test: swift test --filter testXPCClientTimeoutRaceCondition
4. Observe that the Swift task correctly times out after 1 second, but 2 seconds later the libxpc callback fires and crashes the entire process.

### Problem description

When XPCClient.send() races a Swift timeout task against a libxpc callback, the Swift timeout can cancel the task group. However, because C callbacks cannot be cancelled natively in Swift Concurrency, the delayed XPC callback will eventually fire after the timeout task has completed. It then attempts to resume a CheckedContinuation that was already torn down, causing a fatal SWIFT TASK CONTINUATION MISUSE crash (resuming a continuation more than once).

The correct behavior should be to wrap the CheckedContinuation in a private thread-safe Actor (like ContinuationResolver) so that the continuation can only be resumed exactly once, and late XPC replies are safely swallowed as no-ops instead of crashing the daemon.

### Environment

```markdown
- OS: macOS 26.3.1 (25D2128)
- Xcode:  26.6 (17F113)
- Container: container CLI version 1.1.0
```

### Code of Conduct

lakshitsoni26:
Fix #2007

### Motivation and Context
This PR addresses a fatal concurrency issue in `XPCClient.swift` where a race condition between an XPC timeout and a delayed XPC reply can cause a Swift task continuation misuse crash or a permanent task group hang.

Previously, the `XPCClient.send()` method utilized a `withThrowingTaskGroup` to race a timeout task against an XPC callback. However, because Swift Task cancellation cannot cancel registered C `libxpc` callbacks, if the timeout elapsed first, the XPC callback could still fire later. This led to a fatal `SWIFT TASK CONTINUATION MISUSE` crash (resuming a continuation more than once). Furthermore, because `waitForAll()` was used, certain XPC hanging behaviors could cause the task group to deadlock indefinitely.

### Description
This PR refactors `XPCClient.send()` to safely bridge the C-level XPC callback into Swift Concurrency:
- Introduces a private `ContinuationResolver` actor to wrap the `CheckedContinuation`.
- The actor guarantees thread-safe, atomic resolution, ensuring the continuation is resumed exactly once. If both the timeout and the XPC reply race to complete, the loser's call safely becomes a no-op instead of causing a fatal crash.
- Replaces `withThrowingTaskGroup` with a single `withCheckedThrowingContinuation` and a detached `timeoutTask`. This removes the `group.waitForAll()` dependency, preventing the deadlock entirely while cleanly isolating the C-callback lifecycle.

### Testing
- [x] Verified `make check` passes with zero formatting or header issues.
- [x] Verified `swift test` passes locally without deadlocks.
- [x] Tested against simulated XPC hang environments to confirm the timeout resolves instantly without crashing when the delayed reply eventually arrives.

### DIFF
--- a/Sources/ContainerXPC/XPCClient.swift
+++ b/Sources/ContainerXPC/XPCClient.swift
@@ -95,45 +95,69 @@ extension XPCClient {
         XPCClientSession(client: self)
     }
 
+    /// An actor that ensures a `CheckedContinuation` is resumed at most once.
+    ///
+    /// When both a timeout and the XPC reply race to resume the same continuation,
+    /// the loser's call becomes a safe no-op instead of a fatal double-resume.
+    private actor ContinuationResolver {
+        var continuation: CheckedContinuation<XPCMessage, Error>?
+
+        init(_ continuation: CheckedContinuation<XPCMessage, Error>) {
+            self.continuation = continuation
+        }
+
+        func resume(returning value: XPCMessage) {
+            continuation?.resume(returning: value)
+            continuation = nil
+        }
+
+        func resume(throwing error: Error) {
+            continuation?.resume(throwing: error)
+            continuation = nil
+        }
+    }
+
     /// Send the provided message to the service.
     @discardableResult
     public func send(_ message: XPCMessage, responseTimeout: Duration? = nil) async throws -> XPCMessage {
-        try await withThrowingTaskGroup(of: XPCMessage.self, returning: XPCMessage.self) { group in
-            if let responseTimeout {
-                group.addTask {
-                    try await Task.sleep(for: responseTimeout)
-                    let route = message.string(key: XPCMessage.routeKey) ?? "nil"
-                    throw ContainerizationError(
-                        .internalError,
-                        message: "XPC timeout for request to \(self.service)/\(route)"
-                    )
-                }
-            }
+        try await withCheckedThrowingContinuation { continuation in
+            let resolver = ContinuationResolver(continuation)
 
-            group.addTask {
-                try await withCheckedThrowingContinuation { cont in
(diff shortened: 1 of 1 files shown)


jglogan:
@lakshitsoni26 Can you create an issue for this fix that provides concrete reproduction steps for the bug?

lakshitsoni26:
> @lakshitsoni26 Can you create an issue for this fix that provides concrete reproduction steps for the bug?

Done! I have created issue #2007 with the concrete reproduction steps and test case for this crash: https://github.com/apple/container/issues/2007

I have also updated the PR description to link directly to #2007.

PR #1988, opened 2026-07-22
Update Makefile to use modern SwiftPM parallel test flags

lakshitsoni26:
### Motivation and Context
The current `Makefile` uses the deprecated `--experimental-maximum-parallelization-width` SwiftPM flag, which can cause warnings or integration issues on newer toolchains. Additionally, there were verbose JSON event stream outputs and hostname logs that cluttered local testing.

As requested in previous PR reviews, these `Makefile` quality-of-life updates have been separated into this independent PR.

### Description
- Replaced `--experimental-maximum-parallelization-width <N>` with the modern, standard `--parallel --num-workers <N>` flags.
- Replaced `--experimental-maximum-parallelization-width 1` with `--no-parallel` for the serial integration pass.
- Removed the `CONCURRENT_EVENT_STREAM_OPTS` and `GLOBAL_EVENT_STREAM_OPTS` JSON output configurations to streamline the integration command.
- Removed `-Xswiftc -enable-testing` from the baseline `SWIFT_CONFIGURATION` as it is redundant for modern standard test targets.
- Removed local hostname logging.

### Testing
- [x] Verified `make check` passes with zero formatting or header issues.
- [x] Verified `swift test` builds and passes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 # Version and build configuration variables
 BUILD_CONFIGURATION ?= debug
 WARNINGS_AS_ERRORS ?= true
-SWIFT_CONFIGURATION := $(if $(filter-out false,$(WARNINGS_AS_ERRORS)),-Xswiftc -warnings-as-errors) -Xswiftc -enable-testing
+SWIFT_CONFIGURATION := $(if $(filter-out false,$(WARNINGS_AS_ERRORS)),-Xswiftc -warnings-as-errors)
 # Code-coverage instrumentation, layered onto the shared build stages. Empty for
 # ordinary builds; the coverage-* targets opt in via a target-specific value so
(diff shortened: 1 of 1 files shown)


jglogan:
@lakshitsoni26 Your analysis is incorrect.

> deprecated `--experimental-maximum-parallelization-width`

This flag was introduced eight months ago. See https://github.com/swiftlang/swift-testing/pull/1390.

swift-testing uses structured concurrency to run tests. The flags you propose manage test *processes*, which work for XCTest, but not swift-testing.

lakshitsoni26:
Ah, that makes perfect sense! Thank you for taking the time to explain how it interacts with the new `swift-testing` concurrency model vs XCTest. I really appreciate the clarification!

PR #1997, opened 2026-07-23
Send signal as a string in ClientProcess.kill (fixes "missing signal in xpc message")

### LINKED ISSUE #1747: [Bug]: Terminal resize SIGWINCH forwarding prints missing signal in xpc message

### LINKED ISSUE #1876: [Bug]: container hangs and cant terminate (^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2])
### I have done the following

### Steps to reproduce

```
 container run -t  nixos/nix
bash-5.3# ls


^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
```

```
> container run --tty ubuntu:resolute
root@c30d6ce8-2bcf-4e81-ae1b-6009ad4ed69f:/# ls
^Cfailed to send signal: ["error": invalidArgument: "missing signal in xpc message", "signal": 2]
^Cfailed to send signal: ["signal": 2, "error": invalidArgument: "missing signal in xpc message"]
^Cfailed to send signal: ["signal": 2, "error": invalidArgument: "missing signal in xpc message"]
^Cfailed to send signal: ["signal": 2, "error": invalidArgument: "missing signal in xpc message"]
^Cfailed to send signal: ["signal": 2, "error": invalidArgument: "missing signal in xpc message"]
```

### Problem description

Run any container in tty mode. Run any command.  It freezes and can not exit anymore.

### Environment

```markdown
- OS: 26.5.1 (25F80)
- Xcode: 
- Container: 1.0.0
```

### Code of Conduct

### LINKED ISSUE #1941: [Bug]: ClientProcess.kill always fails with "missing signal in xpc message" - client sends the signal as an integer, server reads it as a string
### I have done the following

### Steps to reproduce

The failure is in the client library (ContainerAPIClient), so the reproduction is a small Swift program rather than a CLI command.

1. Start the engine and create a throwaway container to exec into:

```sh
container system start
container run --detach --name kill-repro alpine sleep 600
```

2. Build and run this program in a package that depends on `github.com/apple/container` (product `ContainerAPIClient`):

```swift
import ContainerAPIClient
import Foundation

let client = ContainerClient()
let container = try await client.get(id: "kill-repro")

// Start an exec'd process (a long sleep) so there is a live process to signal.
var processConfig = container.configuration.initProcess
processConfig.executable = "/bin/sleep"
processConfig.arguments = ["300"]
processConfig.terminal = false

let process = try await client.createProcess(
    containerId: container.id,
    processId: UUID().uuidString.lowercased(),
    configuration: processConfig,
    stdio: [nil, nil, nil]
)
try await process.start()

do {
    try await process.kill(SIGKILL)
    print("kill succeeded")
} catch {
    print("kill failed: \(error)")
}
```

3. Observe that `kill` throws and the sleep keeps running:

```
kill failed: invalidArgument: "missing signal in xpc message"
```

The same happens with every signal value, not only SIGKILL.

4. Clean up: `container delete --force kill-repro`

### Problem description

`ClientProcess.kill(_: Int32)` never delivers a signal. Every call fails with `invalidArgument: "missing signal in xpc message"`.

The cause is a type mismatch in the XPC message between the client and the API server. The client writes the signal as an integer (`Sources/Services/ContainerAPIService/Client/ClientProcess.swift`):

```swift
request.set(key: .signal, value: Int64(signal))
```

but the server's handler for the `.containerKill` route reads that key as a string, and rejects the request when the string is absent (`Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift`, the `signal()` helper):

```swift
guard let signal = self.string(key: .signal) else {
    throw ContainerizationError(.invalidArgument, message: "missing signal in xpc message")
}
```

The container-level kill shows the intended contract: `ContainerClient.kill(id:signal:)` takes the signal as a `String`, writes it as a string, and works. `ClientProcess.kill` is the odd one out. Since it is also the only API that targets a specific process, there is currently no working way to signal an exec'd process through the client library at all. In practice that means a timeout or cancel feature built on `createProcess` can stop waiting for a command, but cannot terminate it; the command keeps running until the container stops. The failure is also easy to miss, because `kill` is the kind of call that often gets wrapped in `try?`.

What I would expect: the client and server agree on the wire type, and the signal is delivered.

I observed the failure at runtime on container 1.0.0. The client write and the server read are both unchanged at tag 1.1.0 and on current main (checked at 5f277a9), so I expect it reproduces there as well.

Related work I found while searching: #1747 reports this same root cause surfacing through terminal-resize signal forwarding, and #1778 fixes the client-side encoding. The approach in #1778 (send the signal name rather than a number) looks right to me: Darwin and Linux disagree on some signal numbers (USR1 is 30 on Darwin and 10 on Linux), so a numeric passthrough could deliver the wrong signal even if the server accepted numbers.

### Environment

- OS: macOS 26.5.2 (25F84)
- Xcode: 26.6 (17F113)
- Container: container CLI 1.0.0 (build: release, commit: ee848e3)

### Code of Conduct

chengboonrong:
## Summary

`ClientProcess.kill(_:)` never delivered a signal: every call failed with
`invalidArgument: "missing signal in xpc message"`.

Signals cross the client→server boundary as an XPC dictionary value. There are
two client senders and one server reader:

- `ContainerClient.kill` (the `container kill` CLI) sends the signal as a
  **string** name → the server reads a **string** → works.
- `ClientProcess.kill` — used by the public `ContainerAPIClient` *and* by the
  foreground CLI's signal forwarding — sent an **`Int64`**. The server reads
  `.signal` with `xpc_dictionary_get_string`, which returns `nil` for an integer
  value, so the request was always rejected.

Because `ClientProcess.kill` also backs the foreground CLI's `^C` / `SIGWINCH`
forwarding, this broke interactive `-t` containers (`^C` couldn't stop them) and
spammed stderr with the same error on terminal resize.

## Fix

Send the canonical signal **name**, matching the already-working
`ContainerClient.kill` path:

```swift
request.set(key: .signal, value: Signal.platformName(signal) ?? "\(signal)")
```

The server resolves signal strings against the **Linux** signal table, so
sending the name (rather than the raw macOS number) also keeps signals whose
numbers differ between macOS and Linux correct — e.g. host `SIGUSR1`=30 →
`"USR1"` → Linux 10, instead of being misdelivered as Linux `SIGPWR`=30. Falls
back to the numeric string for real-time signals that have no platform name.

## Tests

Adds `SignalWireFormatTests`, which replicates the exact value the client puts
on the wire and the exact resolution the server performs, covering common
signals, the platform-divergent case (`SIGUSR1`), and the non-empty-string
contract.

Verified locally: `swift build`, `swift format lint --strict`, and the new
tests all pass.

Fixes #1941
Fixes #1876
Fixes #1747


cameronehrlich (approved):
I also encountered this issue.  This looks like an appropriately targeted fix.  I don't have the power to actually approve, but approving anyway in hopes that it keeps the ball rolling

gvenzl:
+1, I also just ran into this in version `1.3.1`:

```
% container system version
COMPONENT            VERSION                                                              BUILD    COMMIT
container            1.3.1                                                                release  unspecified
container-apiserver  container-apiserver version 1.3.1 (build: release, commit: unspeci)  release  unspecified
```

Would be great if this could be fixed.

PR #2000, opened 2026-07-23
Fix logs -n N truncating the oldest line at the read-chunk boundary

### LINKED ISSUE #1967: [Bug]: `container logs -n N` truncates the oldest line when it exceeds the 1024-byte read chunk
### Steps to reproduce

1. Start a container that logs a single line longer than 1024 bytes:
   ```
   container run -d --name logbug alpine \
     sh -c 'printf "START"; i=0; while [ $i -lt 3000 ]; do printf "X"; i=$((i+1)); done; printf "END\n"; sleep 120'
   ```
2. `container logs -n 1 logbug | wc -c`

### Current behavior

Returns **1024 bytes**, and the output starts in the middle of the line — the leading `START` is gone. The oldest (here, only) returned line is truncated.

### Expected behavior

Returns the full line (~3009 bytes), matching `container logs logbug` (without `-n`), which returns the complete line.

More generally, `container logs -n N` truncates the oldest of the N returned lines whenever the last N lines span more than one 1024-byte read chunk (e.g. three ~800-byte lines with `-n 2`).

### Root cause

`Application.ContainerLogs.tail(fh:n:follow:)` reads the file backwards in 1024-byte chunks and stops as soon as `lines.count >= n`. When the last N lines span more than one chunk, the first line in the buffer is a fragment truncated at the chunk boundary (the buffer starts mid-line while `offset > 0`), and `Array(lines.suffix(n))` keeps it. It should keep reading until it has *more* than N non-empty lines (so the partial leading line is dropped) or reaches the start of the file.

### Environment

- container 1.1.0
- macOS 26.5 (Tahoe), Apple Silicon (M-series)

chengboonrong:
## Summary

`container logs -n N` truncated the oldest of the N returned lines whenever those
lines spanned more than one 1024-byte read chunk. For example, a single line
longer than 1024 bytes:

```
$ container logs -n 1 logbug | wc -c
1024        # expected ~3009; output also starts mid-line (leading "START" gone)
```

## Root cause

`Application.ContainerLogs.tail(fh:n:follow:)` reads the file backward in
1024-byte chunks and stopped as soon as `lines.count >= n`. While `offset > 0`
the buffer starts in the middle of a line, so the first line in the buffer is a
fragment truncated at the chunk boundary — and `Array(lines.suffix(n))` kept it.

## Fix

Keep reading until there are **more** than `n` non-empty lines (so the first of
the `n` kept lines is guaranteed complete) or the start of the file is reached.
The backward read is extracted into a testable `lastLines(fh:n:)`; the `follow`
and full-file paths are unchanged.

## Tests

Adds `ContainerLogsTests` covering a single line longer than one chunk, N lines
spanning multiple chunks (the general case), the ordinary short-line case, and
`N` larger than the file. Verified locally: `swift build`,
`swift format lint --strict`, and the new tests all pass.

Fixes #1967

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerLogs.swift
+++ b/Sources/ContainerCommands/Container/ContainerLogs.swift
@@ -63,27 +63,7 @@ extension Application {
             follow: Bool
         ) async throws {
             if let n {
-                var buffer = Data()
-                let size = try fh.seekToEnd()
-                var offset = size
-                var lines: [String] = []
-
-                while offset > 0, lines.count < n {
-                    let readSize = min(1024, offset)
-                    offset -= readSize
-                    try fh.seek(toOffset: offset)
-
-                    let data = fh.readData(ofLength: Int(readSize))
-                    buffer.insert(contentsOf: data, at: 0)
-
-                    if let chunk = String(data: buffer, encoding: .utf8) {
-                        lines = chunk.components(separatedBy: .newlines)
-                        lines = lines.filter { !$0.isEmpty }
-                    }
-                }
-
-                lines = Array(lines.suffix(n))
-                for line in lines {
+                for line in try Self.lastLines(fh: fh, n: n) {
                     print(line)
                 }
             } else {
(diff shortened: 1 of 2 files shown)


chengboonrong:
Thanks @OrtegaMatias — you're right, #1968 came first and takes the same approach (extracting `lastLines(fh:n:)` and reading one line past N so the truncated leading fragment is dropped). Closing this in favor of #1968 to save reviewer time. Happy to review yours if that's helpful.

PR #2006, opened 2026-07-23
fix: Increase XPC timeout for Machine API operations

dev-kvt:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
In constrained CI environments, heavy operations such as creating/unpacking images (`createMachine`) and booting virtual machines (`bootMachine`) can take longer than the default 10-second XPC timeout. This causes intermittent test failures (e.g. `XPC timeout for request to com.apple.container.core.machine-apiserver/createMachine`).

This PR increases the XPC request timeouts in `MachineClient.swift` for `create` and `boot` operations to 120 seconds, and `stop`/`delete` operations to 60 seconds. It also cleans up `TestCLIMachineCommand.swift` by removing the `withKnownIssue` workaround wrapper on `testCreateNameLongestValid`, as the increased timeout resolves the underlying issue.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/MachineAPIService/Client/MachineClient.swift
+++ b/Sources/Services/MachineAPIService/Client/MachineClient.swift
@@ -106,7 +106,7 @@ public struct MachineClient: Sendable {
 
             let response = try await xpcSend(
                 message: request,
-                timeout: .seconds(10)
+                timeout: nil
             )
             let data = response.dataNoCopy(key: MachineKeys.machines.rawValue)
             guard let data else {
@@ -142,7 +142,7 @@ public struct MachineClient: Sendable {
             let bootData = try JSONEncoder().encode(bootConfig)
             request.set(key: MachineKeys.bootConfig.rawValue, value: bootData)
 
-            let _ = try await xpcSend(message: request)
+            let _ = try await xpcSend(message: request, timeout: nil)
         } catch {
             throw ContainerizationError(
                 .internalError,
@@ -158,7 +158,7 @@ public struct MachineClient: Sendable {
             let request = XPCMessage(route: MachineRoutes.deleteMachine.rawValue)
             request.set(key: MachineKeys.id.rawValue, value: id)
 
-            let _ = try await xpcSend(message: request, timeout: .seconds(15))
+            let _ = try await xpcSend(message: request, timeout: .seconds(60))
         } catch {
             throw ContainerizationError(
                 .internalError,
@@ -216,7 +216,7 @@ public struct MachineClient: Sendable {
             let dynamicEnvData = try JSONEncoder().encode(dynamicEnv)
             request.set(key: MachineKeys.dynamicEnv.rawValue, value: dynamicEnvData)
 
-            let response = try await xpcSend(message: request)
+            let response = try await xpcSend(message: request, timeout: nil)
             guard let data = response.dataNoCopy(key: MachineKeys.snapshot.rawValue) else {
                 throw ContainerizationError(
                     .internalError,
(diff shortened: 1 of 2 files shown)


Sources/Services/MachineAPIService/Client/MachineClient.swift
```diff
@@ -142,7 +142,7 @@ public struct MachineClient: Sendable {
             let bootData = try JSONEncoder().encode(bootConfig)
             request.set(key: MachineKeys.bootConfig.rawValue, value: bootData)
 
-            let _ = try await xpcSend(message: request)
+            let _ = try await xpcSend(message: request, timeout: .seconds(120))
```
jglogan: @dev-kvt I did a little more looking into this. The thing I was puzzled by is that we never see this sort of thing for `container run` even if image pulls and unpacks take minutes.

For example, for a pull, no timeout parameter: https://github.com/jglogan/container/blob/main/Sources/Services/ContainerAPIService/Client/ClientImage.swift#L279

You're on the right track, but the answer isn't to increase the timeout, but to get rid of it entirely. For each of the `xpcSend` calls in this file, have a look at what's happening in corresponding code in the machine-apiserver:

- If the machine-apiserver code makes any `xpcSend` call down to the container-apiserver, remove the timeout from the MachineClient call, regardless of whether the container-apiserver call has a timeout or not:
  - If it doesn't have a timeout, then the container-apiserver call is expected to take an arbitrarily long amount of time, meaning the MachineClient call will also.
  - If it does have a timeout, then MachineClient call doesn't need a timeout because the container-apiserver call timeout should cascade back to MachineClient.
- If the machine-apiserver code makes no calls to container-apiserver (I haven't looked to see which do this, but perhaps `ls` is one), then choose a timeout based on what the call is doing and/or what similar calls in the container-apiserver use.

```suggestion
            let _ = try await xpcSend(message: request)
```

jglogan (changes requested):

jglogan:
@dev-kvt Your commits are signed but unverified. When you update your PR, make sure you've got your signing key set up so GH can verify the commits.

Sources/Services/MachineAPIService/Client/MachineClient.swift
dev-kvt: notes 

> @dev-kvt I did a little more looking into this. The thing I was puzzled by is that we never see this sort of thing for `container run` even if image pulls and unpacks take minutes.
> 
> For example, for a pull, no timeout parameter: https://github.com/jglogan/container/blob/main/Sources/Services/ContainerAPIService/Client/ClientImage.swift#L279
> 
> You're on the right track, but the answer isn't to increase the timeout, but to get rid of it entirely. For each of the `xpcSend` calls in this file, have a look at what's happening in corresponding code in the machine-apiserver:
> 
> * If the machine-apiserver code makes any `xpcSend` call down to the container-apiserver, remove the timeout from the MachineClient call, regardless of whether the container-apiserver call has a timeout or not:
>   
>   * If it doesn't have a timeout, then the container-apiserver call is expected to take an arbitrarily long amount of time, meaning the MachineClient call will also.
>   * If it does have a timeout, then MachineClient call doesn't need a timeout because the container-apiserver call timeout should cascade back to MachineClient.
> * If the machine-apiserver code makes no calls to container-apiserver (I haven't looked to see which do this, but perhaps `ls` is one), then choose a timeout based on what the call is doing and/or what similar calls in the container-apiserver use.

noted !

dev-kvt:
@jglogan I have updated MachineClient.swift to remove the timeouts (timeout: nil) for list, create, boot, stop, and inspect, since their underlying handlers delegate to the container-apiserver (allowing those long-running tasks like pulling and unpacking to finish without hitting XPC timeouts).

For the local-only operations that do not call the container-apiserver, I kept sensible timeouts (the .seconds(60) timeout for deleting the local bundle directory in delete, and the default 10s timeout for simple get/set operations).

Also, thank you for pointing out the unverified commits issue! I realized I had not uploaded my local SSH key specifically as a "Signing Key" on GitHub. I have corrected that in my GitHub settings and amended the commit, so the changes in this PR should now be fully verified.

Please let me know if you would like me to adjust anything else!
#

jglogan (approved):
@dev-kvt Changes look good. Thanks, this will get rid of some silly CI build fails!

PR #2019, opened 2026-07-25
Close a UDP proxy backend that becomes active after eviction

### LINKED ISSUE #2015: [Bug]: UDP proxy socket leaks when a backend is evicted before its channel becomes active
### I have done the following

### Steps to reproduce

Publish a UDP port and send datagrams from more than 256 distinct source ports in quick succession, faster than the backend channels finish binding:

```
container run --rm --detach --publish 5000:5000/udp <image running a UDP echo service>

# from the host, spray from many ephemeral source ports
for i in $(seq 1 600); do
  echo probe | nc -u -w0 -p $((20000 + i)) 127.0.0.1 5000
done
```

Sockets held by the forwarder process grow and do not return to the earlier level once the traffic stops. `lsof -p <apiserver pid> | grep UDP | wc -l` stays elevated.

### Problem description

`UDPProxyFrontend` keeps at most 256 backends in an LRU cache and closes whichever one is evicted:

https://github.com/apple/container/blob/main/Sources/SocketForwarder/UDPForwarder.swift#L151

```swift
if let (_, evictedContext) = proxies.put(key: key, value: context) {
    self.log?.trace("frontend - closing evicted backend")
    evictedContext.proxy.close()
}
```

`UDPProxyBackend.close()` returns early when the channel is not set yet:

https://github.com/apple/container/blob/main/Sources/SocketForwarder/UDPForwarder.swift#L82

```swift
func close() {
    guard let channel = state.channel else {
        self.log?.warning("backend - close on inactive channel")
        return
    }
    _ = channel.close()
}
```

`state.channel` is nil from construction until `channelActive` runs, which happens after `bind` completes. A backend evicted inside that window is never closed. `channelActive` then assigns `state.channel` and the socket becomes active, but the `ProxyContext` has already been dropped from the cache, so no reference remains and nothing closes it. The socket stays open for the life of the process.

The existing log line records the condition as a warning, so the case was anticipated, but the early return leaves the socket open rather than deferring the close.

A fix would record the close request in the backend state and have `channelActive` honour it, closing the channel immediately instead of adopting it.

### Environment

- OS: macOS 26.5.2 (25F84)
- Xcode: 26.6 (17F113)
- Container: main at 07ff3c0 (also present in 1.1.0)

### Code of Conduct

devops-thiago:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Fixes #2015.

`UDPProxyFrontend` keeps at most 256 backends in an LRU cache and closes whichever one is evicted. `UDPProxyBackend.close()` returned early when the backend channel had not finished binding:

```swift
func close() {
    guard let channel = state.channel else {
        self.log?.warning("backend - close on inactive channel")
        return
    }
    _ = channel.close()
}
```

`state.channel` is nil from construction until `channelActive` runs, which happens after `bind` completes. A backend evicted inside that window is never closed. `channelActive` then adopts the channel, and because the `ProxyContext` has already been dropped from the cache, no reference remains that could close it. The socket stays open for the lifetime of the process.

The existing warning shows the case was anticipated, but the early return leaves the socket open rather than deferring the close.

This change records the close request in the backend state and has `channelActive` honour it, closing the channel instead of adopting it and discarding any queued payloads. Both paths run on the event loop the backend is bootstrapped on, since `DatagramBootstrap(group: context.eventLoop)` and `NIOLoopBound(proxy, eventLoop: context.eventLoop)` bind them to that loop, so the flag needs no further synchronisation.

`UDPProxyBackend` changes from file-private to internal so the test can drive the ordering directly. Nothing outside the module references it.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

`testBackendClosedBeforeChannelActiveClosesTheChannel` calls `close()` before the channel becomes active, then activates it through an `EmbeddedChannel` and asserts the channel is no longer active. The test fails against the current code and passes with this change. `testUDPForwarder` and `testLRUCache` continue to pass.

### DIFF
--- a/Sources/SocketForwarder/UDPForwarder.swift
+++ b/Sources/SocketForwarder/UDPForwarder.swift
@@ -22,7 +22,7 @@ import NIOFoundationCompat
 import Synchronization
 
 // Proxy backend for a single client address (clientIP, clientPort).
-private final class UDPProxyBackend: ChannelInboundHandler {
+final class UDPProxyBackend: ChannelInboundHandler {
     typealias InboundIn = AddressedEnvelope<ByteBuffer>
     typealias OutboundOut = AddressedEnvelope<ByteBuffer>
 
@@ -35,6 +35,7 @@ private final class UDPProxyBackend: ChannelInboundHandler {
     private struct State {
         var queuedPayloads: Deque<ByteBuffer>
         var channel: (any Channel)?
+        var closed: Bool
     }
 
     private let clientAddress: SocketAddress
@@ -48,7 +49,7 @@ private final class UDPProxyBackend: ChannelInboundHandler {
         self.serverAddress = serverAddress
         self.frontendChannel = frontendChannel
         self.log = log
-        let initialState = State(queuedPayloads: Deque(), channel: nil)
+        let initialState = State(queuedPayloads: Deque(), channel: nil, closed: false)
         self.state = initialState
     }
 
@@ -61,6 +62,16 @@ private final class UDPProxyBackend: ChannelInboundHandler {
     }
 
     func channelActive(context: ChannelHandlerContext) {
+        guard !state.closed else {
+            // close() ran while this channel was still binding, so there was no channel to
+            // close at the time. Close it now rather than adopting it: the backend has
+            // already been evicted from the proxy cache, so no reference remains that could
+            // close it later.
+            self.log?.trace("backend - closing channel that became active after close")
+            state.queuedPayloads.removeAll()
+            context.channel.close(promise: nil)
+            return
+        }
         if !state.queuedPayloads.isEmpty {
             self.log?.trace("backend - writing \(state.queuedPayloads.count) queued datagrams to server")
             while let queuedData = state.queuedPayloads.popFirst() {
@@ -88,8 +99,9 @@ private final class UDPProxyBackend: ChannelInboundHandler {
     }
 
     func close() {
+        state.closed = true
         guard let channel = state.channel else {
-            self.log?.warning("backend - close on inactive channel")
+            self.log?.trace("backend - close requested before the channel became active")
             return
         }
         _ = channel.close()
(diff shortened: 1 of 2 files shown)


devops-thiago:
@katiewasnothere I noticed you labeled #2015 `stability` — thanks for looking. This PR fixes it: a backend evicted from the LRU while its channel is still binding never gets closed, so the socket leaks once `channelActive` fires. The fix records the close request and honours it in `channelActive`; both run on the same event loop, so no extra synchronisation is needed. Regression test included.

I have three other small parser fixes open with the same setup — #2016, #2017, #2018 — each one line plus a test, each with a linked issue. All four are rebased onto current main, signed, and pass the unit suite. Since I don't have a merged PR yet, the workflows are waiting on approval to run. If you're able to approve them, that would unblock CI on the set. Happy to adjust anything.

devops-thiago:
@egernst Pinging you since you fixed a closely related bug in #2260 (`ConnectHandler.swift` closing the wrong channel reference). This PR fixes the same class of issue in the UDP path: a backend evicted from the LRU while its channel is still binding never gets closed, so the socket leaks once `channelActive` fires later. The fix records the close request and honours it in `channelActive`, on the same event loop so no extra synchronisation is needed. Regression test included.

Three other small parser fixes are open with the same setup — #2016, #2017, #2018 — each one line plus a test, each with a linked issue. All four (plus four unrelated fixes: #2090, #2091, #2092, #2093) are rebased onto current main, signed, and passing the unit suite. I don't have a merged PR yet, so the workflows are stuck waiting on approval to run. An approve on any of these would unblock CI for the rest.

devops-thiago:
@jglogan I pinged @katiewasnothere on Sep 2 and @egernst on Sep 14 about this same batch with no reply, so per the contributing guide I'm trying someone else.

Eight small PRs are open, all rebased onto current main, signed, and passing the unit suite: #2016, #2017, #2018, #2019 (parser/proxy fixes, each one line plus a test) and #2090, #2091, #2092, #2093 (exit code persistence, XPC peer validation, image list filtering, build context error reporting). None have a review yet, so CI is stuck waiting on approval to run. An approve on any one would unblock the workflows for the rest.

Happy to split these up, add more tests, or adjust anything that doesn't fit.

PR #2020, opened 2026-07-26
feat: add build-context support

### LINKED ISSUE #1930: [Request]: Support Docker/BuildKit named build contexts in container build

`container build` should support Docker/BuildKit named build contexts, equivalent to Docker/Buildx and Podman `--build-context name=value`.

This is a Dockerfile/BuildKit feature used by higher-level build tooling to pass additional local directories, Git contexts, URLs, OCI layouts, or image references into a build without copying them under the primary build context. Typical Dockerfile usage is:

```Dockerfile
FROM alpine
COPY --from=deps /some/file /some/file
```

with a CLI invocation like:

```sh
container build \
  --build-context deps=/path/to/deps \
  -t example:latest \
  .
```

Today `container build --help` does not expose `--build-context`, and the public build RPC/config path appears to carry only one primary context directory. That means tools that rely on named contexts cannot transparently use Apple `container` as a Docker-compatible builder even though the underlying builder is BuildKit-based.

### Why this matters

Named contexts are not only a convenience feature. They are used by libraries and build systems that need to:

- avoid copying large generated/source trees into the main context;
- keep Dockerfile context boundaries explicit;
- consume additional local source roots in monorepos;
- pin or lock base images via `docker-image://...` contexts;
- match Docker/Buildx and Podman behavior for existing consumers.

For example, AWS CDK-style container asset builders can receive a user-supplied map of build contexts. Without this flag, such libraries must either reject Apple `container` for those builds or switch to a separate BuildKit path and only load the final OCI image into `container` afterward.

### Existing ecosystem support

Docker/Buildx supports named build contexts:

```sh
docker buildx build --build-context deps=/path/to/deps .
docker buildx build --build-context base=docker-image://alpine:latest .
```

Podman supports the same concept:

```sh
podman build --build-context deps=/path/to/deps .
podman build --build-context base=docker-image://alpine:latest .
```

Direct BuildKit supports the underlying mechanism through Dockerfile frontend options. For local named contexts this is roughly:

```sh
buildctl build \
  --frontend=dockerfile.v0 \
  --local context=. \
  --local dockerfile=. \
  --local deps=/path/to/deps \
  --opt context:deps=local:deps
```

For image/OCI contexts BuildKit frontend attrs use forms such as:

```sh
--opt context:base=docker-image://alpine:latest
--opt context:base=oci-layout:/path/to/oci-layout@sha256:...
```

So the underlying BuildKit machinery already supports this. The missing piece seems to be Apple `container` CLI/API/shim plumbing for multiple named contexts and the corresponding context transfer/serving logic.

### Suggested behavior

Add support for one or more `--build-context name=value` flags to `container build`, matching Docker/Podman syntax where practical.

At minimum, support these common values:

- `name=/absolute/or/relative/local/path`
- `name=docker-image://registry/name:tag`
- `name=docker-image://registry/name@sha256:...`
- `name=oci-layout:/path/to/layout@sha256:...`
- URL/Git contexts if they are already supported by the selected Dockerfile frontend

The local path case likely needs the build request/shim protocol to carry more than the current primary context and expose each named local context to BuildKit as a separate local source.

### Current workaround

The practical workaround is to bypass `container build`, run BuildKit directly with named context frontend options, export an OCI archive, and then load the result into Apple `container`:

```sh
buildctl build ... --output type=oci,dest=image.tar
container image load -i image.tar
container image tag <loaded-ref> example:latest
```

That works, but it loses the simplicity of `container build` and requires users/tools to manage a separate BuildKit daemon, auth, cache, and networking behavior.

### Code of Conduct

0Itsuki0:
# Add support for named build contexts, equivalent to Docker/Buildx and Podman --build-context name=value.

## Type of Change
- New feature  

## Motivation and Context
Supporting named build contexts : resolve #1930 

### Supported formats:

- name=/absolute/or/relative/local/path
- name=docker-image://xxx
- name=oci-layout:/path/to/layout (with or without digest)
- URL/Git

## Testing
- Tested locally

1. Check out the container fork for enabling build-context option
2. build the container by running `make all && make install`
3. Check out [builder shim fork](https://github.com/apple/container-builder-shim/pull/88)
4. build the shim image by running `container build -t builder .`
5. update build image in container config at ~/.config/container/config.toml 

```
[build]
image = "builder:latest"
```

6. Use the [BuildContextTest.zip](https://github.com/user-attachments/files/30383639/BuildContextTest.zip) to test out all 4 scenarios above + when a regular build where no named context is specified.

### DIFF
--- a/Sources/ContainerBuild/BuildFSSync.swift
+++ b/Sources/ContainerBuild/BuildFSSync.swift
@@ -133,30 +133,34 @@ actor BuildFSSync: BuildPipelineHandler {
         _ buildID: String
     ) async throws {
         let wantsTar = packet.mode() == "tar"
+        // context directory -> coming in as resolved (absolute) path
+        // build-context -> relative path to the current working directory
+        let root = URL(filePath: packet.source)
 
         var entries: [String: Set<DirEntry>] = [:]
         let followPaths: [String] = packet.followPaths() ?? []
 
-        let followPathsWalked = try walk(root: self.contextDir, includePatterns: followPaths)
+        let followPathsWalked = try walk(root: root, includePatterns: followPaths)
+
         for url in followPathsWalked {
-            guard self.contextDir.absoluteURL.cleanPath != url.absoluteURL.cleanPath else {
+            guard root.absoluteURL.cleanPath != url.absoluteURL.cleanPath else {
                 continue
             }
-            guard self.contextDir.parentOf(url) else {
+            guard root.parentOf(url) else {
                 continue
             }
 
-            let relPath = try url.relativeChildPath(to: contextDir)
-            let parentPath = try url.deletingLastPathComponent().relativeChildPath(to: contextDir)
+            let relPath = try url.relativeChildPath(to: root)
+            let parentPath = try url.deletingLastPathComponent().relativeChildPath(to: root)
             let entry = DirEntry(url: url, isDirectory: url.hasDirectoryPath, relativePath: relPath)
             entries[parentPath, default: []].insert(entry)
 
             if url.isSymlink {
                 let target: URL = url.resolvingSymlinksInPath()
-                if self.contextDir.parentOf(target) {
-                    let relPath = try target.relativeChildPath(to: self.contextDir)
+                if root.parentOf(target) {
+                    let relPath = try target.relativeChildPath(to: root)
                     let entry = DirEntry(url: target, isDirectory: target.hasDirectoryPath, relativePath: relPath)
-                    let parentPath: String = try target.deletingLastPathComponent().relativeChildPath(to: self.contextDir)
+                    let parentPath: String = try target.deletingLastPathComponent().relativeChildPath(to: root)
                     entries[parentPath, default: []].insert(entry)
                 }
             }
@@ -167,7 +171,7 @@ actor BuildFSSync: BuildPipelineHandler {
 
         if !wantsTar {
             let fileInfos = try fileOrder.map { rel -> FileInfo in
-                try FileInfo(path: contextDir.appendingPathComponent(rel), contextDir: contextDir)
+                try FileInfo(path: root.appendingPathComponent(rel), contextDir: root)
             }
 
             let data = try JSONEncoder().encode(fileInfos)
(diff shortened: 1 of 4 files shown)


MayCXC:
I have been building on this branch for a few weeks and would like to offer what came out of it. The long version, including the one design point, is on the shim side at apple/container-builder-shim#88; the short version of that point is that BuildKit's own filesync has the requester send a NAME and the server hold the PATHS (`FSSync()` sets `dir-name` on every request, `fsSyncProvider.handle()` resolves it with `LookupDir` and refuses an unregistered name), whereas here the shim sends a path and the host serves what arrives. That matters on this side specifically, because `BuildPipelineHandler` documents a containment guarantee (`openat(O_NOFOLLOW)` descent, canonical-path containment inside the context root) and a path arriving from the builder VM moves that boundary into the VM. With one root that is invisible; with named contexts there are several roots and the guarantee has to hold for each.

What I have on top of your commit, useful only once it lands:

- The host owning the name-to-path mapping: a validated `[String: URL]` built from the command line, a lookup, and a refusal for anything else, with the containment checks applied per root and tests pinning both.
- A missing `oci-layout` directory refused CLI-side with its real cause rather than a downstream resolution error.
- The six sandbox keys the frontend applies to every `RUN`: `--add-host`, `--hostname`, `--shm-size`, `--ulimit`, `--cgroup-parent`, `--network`. `--shm-size` takes the suffixed sizes people expect from docker and resolves them to the plain byte count the frontend wants, using the same `Parser.memoryStringAsBytes` that `--memory` uses.

I went through `frontend/dockerui`'s `config.go`, `context.go` and `namedcontext.go` at v0.29.0 key by key and verified each value form end to end, with `input:<stage>` the one row that is not applicable here, since only bake has a producer to bind it to. Related: #1930 asks for this feature and #2037 is the empty-context report next to it.

Happy to open these against your branch, hold them until this merges, or fold the mapping change into a review comment if you would rather own it. This is your work and the extension only makes sense on top of it; I would rather not land a second rendition of the same feature.

0Itsuki0:
@MayCXC 

Hey,

Thank you very much for building on this branch and making improvements to it (also the builder-shim side! Thank you.)
I have seen that you have opened a couple of new PRs that sit on top of this. 
I am totally cool with that so unless Apple Container's maintainer would like to make it into one PR or something, I think we can just keep those as they are for now.

PR #1662, opened 2026-06-08
Add `container machine` for managing persistent Linux environments

realrajaryan:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
`container` runs each workload in an ephemeral VM, so there's no built-in way to keep a persistent Linux environment you can log into and work in. `container machine` adds one.

A container machine is a lightweight, persistent, and integrated Linux environments that feel like an extension of your Mac, created from standard OCI images with a familiar UX. The login user matches your host account with passwordless `sudo`, your home directory is mounted inside the VM, and each machine keeps its filesystem and runs the image's own init system (such as`systemd` or `openrc`).

```bash
container machine create alpine:3.22 --name my-machine
container machine run -n my-machine # interactive shell
container machine set -n my-machine cpus=4 memory=8G
```

Subcommands: `create`, `run`, `list` (`ls`), `inspect`, `set`, `set-default`, `logs`, `stop`, `delete` (`rm`); `m` aliases `machine`. Docs added to `docs/command-reference.md` (Machine Management) and `docs/how-to.md` ("Use container machines").

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

PR #2042, opened 2026-07-30
Start default machine shell as a login shell

### LINKED ISSUE #1991: [Request]: container machine run should spawn a login shell

Default behaviour of `container machine run` is to spawn an interactive shell inside container machine. Help info on `container machine run --help` even states following:

```
ARGUMENTS:
  <executable>            Command to run (default: login shell)
```

My observed behaviour is that interactive shell spawned is not a login shell. This has a number of drawbacks how interactive shell is initialised (both bash and zsh will not source "profile" rc files). This is in contrast to say connection via SSH to a running machine instances.

I would argue that shell should be login shells, or at least there should be a way to configure this per container machine.

```
container CLI version 1.0.0 (build: release, commit: ee848e3)
```

### Code of Conduct

Dave861:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

`container machine run` documents its default behavior as starting a login shell, but the machine init helper invoked the resolved shell without its login option. As a result, startup files such as `~/.profile` were not sourced.

Invoke the default shell with `-l` while leaving the explicit command path unchanged.

Fixes #1991.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Added a focused integration regression that verifies the default shell sources `~/.profile`. Confirmed it failed before the change and passed afterward.

`make check` passes.

### DIFF
--- a/Sources/Plugins/MachineAPIServer/Resources/init
+++ b/Sources/Plugins/MachineAPIServer/Resources/init
@@ -51,7 +51,7 @@ if [ "$1" = "-s" ]; then
     if [ $# -gt 0 ]; then
         exec "${USER_SHELL:-${SHELL}}" -c "$*"
     else
-        exec "${USER_SHELL:-${SHELL}}"
+        exec "${USER_SHELL:-${SHELL}}" -l
     fi
 elif [ "$1" = "-u" ]; then
     # DEPRECATED 0.11.0.0 - use `id` instead of checking `${MACHINE_INITIALIZED}` for backward compatibility, remove in 0.13.0.0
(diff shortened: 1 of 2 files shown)


Dave861:
@realrajaryan, could you take a look at this, or point me to the right reviewer? It touches the default shell path from your persistent-machine work in #1662. The fix adds `-l` only when no command is supplied, with integration coverage, and `make check` passes locally. The PR workflows are also awaiting maintainer approval.

PR #2044, opened 2026-07-30
K8s plugin

### LINKED ISSUE #2043: [Request]:  k8s plugin

# Feature Request: k8s plugin — local Kubernetes cluster support

> [!IMPORTANT]
>
> This is an experimental feature. CLI options and configuration API will be unstable as the plugin evolves.

## Summary

Add a `k8s` plugin to the `container` CLI that lets users spin up and manage local Kubernetes development clusters. Cluster nodes run upstream Kubernetes bootstrapped via `kubeadm`, using the [kindest/node](https://hub.docker.com/r/kindest/node) image from the [KIND](https://kind.sigs.k8s.io) project as the base, running directly on the Apple container runtime.

## Motivation

Local Kubernetes development on Apple silicon today requires third-party tools that either ship their own VMs or rely on Docker. Since `container` already manages lightweight Linux containers natively on macOS, it is a natural fit for running a local k8s cluster with minimal overhead and no extra dependencies.

### Initial feature set — `create`, `start`, `delete`, `load-image`, `write-config`, `list`

The initial feature set covers the core day-to-day workflow: boot a single-node cluster, load images into it, point `kubectl` at it, and tear it down when finished.

**Node image** — The plugin uses `kindest/node` as the cluster node image. This image ships with kubeadm, kubelet, kubectl, and containerd pre-installed. On first use the plugin pulls the image automatically if it is not already present.

**create** — Creates and starts a single control-plane cluster. The plugin pulls the node image if needed, boots the container, runs a prep script to configure the native containerd snapshotter and required sysctl values, runs `kubeadm init`, applies the kindnet CNI, removes the control-plane taint so pods can schedule on the single node, and merges a kubeconfig entry into `~/.kube/config` automatically.

```bash
container k8s create --name my-cluster
```

**start** — Starts a previously stopped cluster and re-merges the kubeconfig into `~/.kube/config`, updating the API server address in case the container IP changed on restart.

```bash
container k8s start --name my-cluster
```

**delete** — Stops and removes the cluster and cleans up its entries from `~/.kube/config`.

```bash
container k8s delete --name my-cluster
```

**load-image** — Loads a locally-stored image into the cluster's containerd so kubelet can schedule pods using it without an external registry. The image is saved from the `container` image store to a temporary OCI tar and imported into the node via `ctr images import`. Short references (`alpine`) are automatically qualified to `docker.io/library/alpine:latest` and also tagged back under the short name so kubelet can find them. An optional `--platform` flag controls which platform variant is selected from the local image store when the image has multiple architectures (default: host architecture).

```bash
container k8s load-image --name my-cluster my-app:latest
container k8s load-image --name my-cluster --platform linux/amd64 my-app:latest
```

**write-config** — Merges the cluster's kubeconfig into `~/.kube/config` (or a custom path via `--kubeconfig`). Rewrites the API server address to `127.0.0.1:<host-port>` or to the machine's FQDN when one is detected. Does not change `current-context` — use `kubectl config use-context` to switch contexts.

```bash
container k8s write-config --name my-cluster
container k8s write-config --name my-cluster --kubeconfig ~/.kube/my-cluster.kubeconfig
```

**list** — Lists all clusters and their nodes in a table, discovered via the `plugin=k8s` container label.

```bash
container k8s list
# CLUSTER     NODE        ROLE           STATE    CPUS  MEMORY   ADDR          PORTS
# my-cluster  my-cluster  control-plane  running  4     4096 MB  192.168.64.5  6445->6443
```

## Architecture

Image

All components run on macOS. The `container k8s` CLI drives the `container` service to create, start, and exec into the cluster node container. Images are transferred out-of-band: the CLI reads from the `container` image store and streams an OCI tar into the node's containerd (k8s.io namespace). `kubectl` reaches the API server through a published host port (`6445 → 6443`). `write-config` merges credentials directly into `~/.kube/config`.

## Future Work

**API support** — Cluster management API calls use a well defined `K8sClusterConfiguration` object for creation, and a `K8sClusterResource` conforming to `ManagedResource` for existing clusters. `k8s create` accepts a `--config config-path` options as well as existing cluster configuration options.

**Multi-worker clusters** — `create` will accept a `--workers N` flag to start N worker nodes alongside the control plane. Each worker joins using a token fetched from the control plane. `load-image` will target all nodes by default so every kubelet can schedule pods regardless of placement.

**HA control plane** — `create` accepts a `--control-planes N` flag to start N control plane nodes, with a load balancing implementation that provides a single HA cluster API endpoint. 

**Service load balancing** — Allow clusters to expose services through a host-accessible endpoint without manual port-forwarding, designed to map cleanly to the `container` networking model.

**Registries** — Let users point a cluster at a local registry (e.g. one started with `container run`) so that images pushed there are immediately pullable inside the cluster without running `load-image` manually.

**Alternate cluster providers** — Refine cluster and node provisioning into solid abstractions, so that nodes can be created and joined using distributions other than kind.

## Questions for the Community

1. Is this something you'd use? Would this be valuable to your workflow?
2. What features are must-haves?
3. What use cases would this unlock for you?

---
Please share your thoughts and feedback. We want to hear from you!

### Code of Conduct

jshi991:
- closes #2043 
- introduces the k8s plugin, allowing users to make single node clusters with the kind base image 
- other functionality is included as well such as creation, deletion, and loading custom images

### DIFF
--- a/Sources/Plugins/K8s/K8sCommand.swift
+++ b/Sources/Plugins/K8s/K8sCommand.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerVersion
+
+@main
+struct K8sCommand: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "k8s",
+        abstract: "Manage local Kubernetes development clusters (EXPERIMENTAL)",
+        discussion: """
+            EXAMPLES:
+              Create a cluster by name and list clusters:
+                $ container k8s create --name my-cluster
+                $ container k8s list
+
+              Switch between clusters:
+                $ container k8s create --name second-cluster
+                $ kubectl config use-context second-cluster
+                $ kubectl config use-context my-cluster
+
+              Write the cluster context to an alternate configuration file:
+                $ container k8s write-config --name my-cluster --kubeconfig ~/.kube/my-cluster.kubeconfig
+                $ KUBECONFIG=~/.kube/my-cluster.kubeconfig kubectl cluster-info
+
+              Load a local image into the cluster and run it:
+                $ container image pull docker.io/library/hello-world:latest
+                $ container image tag docker.io/library/hello-world:latest my-hello-world:latest
+                $ container k8s load-image --name my-cluster my-hello-world:latest
+                $ kubectl run hello-job --image=my-hello-world:latest --restart=Never --attach --rm -i
+
+              Stop and delete the cluster:
+                $ container k8s delete --name my-cluster
+            """,
+        version: ReleaseVersion.singleLine(appName: "k8s"),
+        subcommands: [
+            K8sCreate.self,
+            K8sDelete.self,
+            K8sList.self,
+            K8sLoadImage.self,
+            K8sStart.self,
+            K8sWriteConfig.self,
+        ]
+    )
+}
--- a/Sources/Plugins/K8s/K8sCreate.swift
+++ b/Sources/Plugins/K8s/K8sCreate.swift
@@ -1,0 +1,187 @@
+//===----------------------------------------------------------------------===//
+// Copyright © 2026 Apple Inc. and the container project authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerAPIClient
+import ContainerLog
+import ContainerPersistence
+import ContainerResource
+import ContainerizationError
+import Darwin
+import Foundation
+import Logging
+import TerminalProgress
+
+struct K8sCreate: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "create",
+        abstract: "Create and start a local Kubernetes cluster"
+    )
+
+    @Option(name: .long, help: "Cluster name (default: \(K8sHelper.defaultName))")
+    var name: String = K8sHelper.defaultName
+
+    @Flag(name: [.customLong("rm"), .long], help: "Remove the cluster container after it stops")
+    var remove: Bool = false
+
+    @OptionGroup(title: "Resource options")
+    var resourceFlags: Flags.Resource
+
+    @OptionGroup(title: "Registry options")
+    var registryFlags: Flags.Registry
+
+    @OptionGroup(title: "Image fetch options")
+    var imageFetchFlags: Flags.ImageFetch
+
+    @Option(help: "Node image reference (default: \(K8sHelper.nodeImage))")
+    var nodeImage: String = K8sHelper.nodeImage
+
+    func run() async throws {
+        LoggingSystem.bootstrap { _ in StderrLogHandler() }
+        let log = Logger(label: K8sHelper.pluginName)
+
+        guard ManagedContainer.nameValid(name) else {
+            throw ContainerizationError(.invalidArgument, message: "cluster name \(name) is not a valid container ID")
+        }
+
+        let isTTY = isatty(FileHandle.standardError.fileDescriptor) == 1
+        let progressConfig = try ProgressConfig(
+            showSpinner: isTTY,
+            showTasks: true,
+            showItems: true,
+            ignoreSmallSize: true,
+            totalTasks: 2,  // fetch image, unpack image
+            clearOnFinish: isTTY,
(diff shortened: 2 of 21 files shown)


jglogan:
@jshi991 Please resubmit with verified commit signatures. Thanks!

Sources/Plugins/K8s/K8sHelper.swift
```diff
@@ -0,0 +1,658 @@
...
+        }
+
+        existing.clusters.removeAll { $0.name == containerId }
+        existing.contexts.removeAll { $0.name == containerId }
+        existing.users.removeAll { $0.name == containerId }
+
+        existing.clusters.append(contentsOf: config.clusters)
+        existing.contexts.append(contentsOf: config.contexts)
+        existing.users.append(contentsOf: config.users)
+        if setCurrentContext && existing.currentContext.isEmpty {
```
jglogan:
```suggestion
        if setCurrentContext && existing.currentContext == nil {
```

It seems like the logic for this is "set the current context only if it isn't already set in the current config?"

Should we always update `current-config` whether it's defined or not?

Sources/Plugins/K8s/K8sHelper.swift
```diff
@@ -0,0 +1,658 @@
...
+        }
+
+        var updated = existing
+
+        updated.clusters.removeAll { $0.name == containerId }
+        updated.contexts.removeAll { $0.name == containerId }
+        updated.users.removeAll { $0.name == containerId }
+
+        if updated.currentContext == containerId {
+            updated.currentContext = ""
```
jglogan:
```suggestion
            updated.currentContext = nil
```

Sources/Plugins/K8s/KubeConfig.swift
```diff
@@ -0,0 +1,156 @@
+//===----------------------------------------------------------------------===//
```
jglogan, line 1: Swift code is not a resource, this file should be directly under K8s I think.

Sources/Plugins/K8s/Resources/KubeConfig.swift
```diff
@@ -0,0 +1,156 @@
...
+        case apiVersion, kind, clusters, contexts, users
+        case currentContext = "current-context"
+    }
+
+    var apiVersion: String = "v1"
+    var kind: String = "Config"
+    var clusters: [NamedCluster] = []
+    var contexts: [NamedContext] = []
+    var users: [NamedAuthInfo] = []
+    var currentContext: String = ""
```
jglogan:
```suggestion
    var currentContext: String?
```

We need this to be an optional, to support `current-context` not present in the config.

Sources/Plugins/K8s/Resources/KubeConfig.swift
```diff
@@ -0,0 +1,156 @@
...
+        case server
+        case tlsServerName = "tls-server-name"
+        case insecureSkipTLSVerify = "insecure-skip-tls-verify"
+        case certificateAuthority = "certificate-authority"
+        case certificateAuthorityData = "certificate-authority-data"
+        case proxyURL = "proxy-url"
+    }
+
+    var server: String
+    var tlsServerName: String? = nil
```
jglogan:
```suggestion
    var tlsServerName: String?
```

`= nil` for all of these stored properties on optionals are unnecessary boilerplate, remove.

jglogan:
@jshi991 In the PR description, please remove the alert regarding signed commits, and replace it with a short bullet list:

```
- Closes #2043.
- Any brief description bullet items you want to make about the change.
...
```

perplexiphone-cmd (approved):

docs/command-reference.md
```diff
@@ -1578,3 +1578,172 @@ container system property list
 # output as JSON for scripting
 container system property list --format json
 ```
+
+## Kubernetes Cluster Management
+
+`container k8s` manages local single-node Kubernetes clusters backed by container VMs. Each cluster runs a Kubernetes control-plane node inside a container using `kindest/node` and `kubeadm`.
+
```
jglogan, line 1590:
```suggestion

> [!IMPORTANT]
> The `k8s` command is an experimental feature and its subcommands and options are subject to change.
```

Sources/Plugins/K8s/K8sCommand.swift
```diff
@@ -0,0 +1,59 @@
...
+//===----------------------------------------------------------------------===//
+
+import ArgumentParser
+import ContainerVersion
+
+@main
+struct K8sCommand: AsyncParsableCommand {
+    static let configuration = CommandConfiguration(
+        commandName: "k8s",
+        abstract: "Manage local Kubernetes development clusters",
```
jglogan:
```suggestion
        abstract: "Manage local Kubernetes development clusters (EXPERIMENTAL)",
```

jglogan (changes requested):

Sources/Plugins/K8s/K8sHelper.swift
jshi991: Yup config should be setup like this

jglogan (changes requested):
Two more change requests:

- Rebase your PR onto latest main to pick up the fix from @JaewonHur to move the ssh build test to the serial tests.
- Increase the timeout in the `.github/workflows/common.yaml` file from 70 to 90 minutes, as we don't know how much additional time your tests will need.

jglogan (changes requested):
@jshi991 You may be missing a feature needed for running up on GH. Anything that goes to the network needs to have proxy variables set.

- https://kind.sigs.k8s.io/docs/user/quick-start/#configure-kind-to-use-a-proxy
- https://github.com/kubernetes-sigs/kind/blob/main/pkg/cluster/internal/providers/docker/provision.go#L184

Sources/Plugins/K8s/K8sCreate.swift
```diff
@@ -0,0 +1,186 @@
...
+            ssh: false,
+            shmSize: nil,
+            tmpFs: [],
+            useInit: false,
+            virtualization: false,
+            volumes: []
+        )
+
+        let updatedResource = K8sHelper.defaultedResourceFlags(resourceFlags)
+        let proxyEnvVars = ["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy"]
```
jglogan: nit: define this once as a package-scoped static in K8sHelper, don't repeat in both create and run

jglogan (changes requested):

Sources/Plugins/K8s/K8sCreate.swift
jshi991: resolved

jglogan (approved):

Issue #1514, opened 2026-05-06
[Request]: Be able to use containers from LaunchDaemon-based CI runners


thomasjm:
### Feature or enhancement request details

When you run `container system start`, it installs `container-apiserver` as a per-user LaunchAgent in `gui/<uid>`, with state under `~/Library/Application Support/com.apple.container/`. This is awkward for CI: GitHub's `actions/runner` and most macOS CI runners are designed to install as LaunchDaemons (no GUI session needed). A daemon-mode runner can't reach the apiserver's mach service in `gui/<uid>`, so steps that shell out to `container ...` fail.

If you install the CI runner itself as a LaunchAgent, then the processes it spawns end up in the user GUI session, where they become subject to macOS's Local Network Privacy checks. So if you start a container and it ends up with IP address 192.168.64.1, and your test code tries to access it, you can run into LNP issues and get "No route to host" errors.

Apple's support itself actually recommends [using a LaunchDaemon](https://developer.apple.com/forums/thread/770473) as a workaround for LNP issues.

So there are seemingly blockers for using containers in CI via either LaunchAgent or LaunchDaemon runners. In general I'd prefer to use LaunchDaemons since LaunchAgents are more cumbersome. Is there some way to work with containers from within a LaunchDaemon? If not, I'd like to request this functionality. Thanks!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2045, opened 2026-07-30
Fix launchctl bootstrap failures being discarded as XPC errors

### LINKED ISSUE #2008: [Bug]: `container system start` discards launchctl bootstrap failures, yielding a misleading XPC error in non-login (CI) sessions
### I have done the following

### Steps to reproduce

On a session whose launchd context is not a logged-in user session — a GitHub Actions `macos-26` runner reproduces this deterministically:

```sh
sudo installer -pkg container-installer-unsigned.pkg -target /
sudo container system start --enable-kernel-install
```

Output (instant, no retry helps — 15 consecutive attempts fail identically):

```
Launching container-apiserver...
Testing access to container-apiserver...
Error: internalError: "failed to get a response from apiserver: interrupted: "XPC connection error: Connection invalid""
Ensure container system service has been started with `container system start`.
```

`log show --predicate 'subsystem == "com.apple.container"'` over the failure window is **completely empty** — `container-apiserver` never launched at all.

### Problem description

Two layered problems:

**1. The launchctl bootstrap failure is silently discarded.** `ServiceManager.register` (Sources/ContainerPlugin/ServiceManager.swift) pipes launchctl's stdout/stderr to `/dev/null` *and* ignores its exit status:

```swift
public static func register(plistPath: String) throws {
    let domain = try Self.getDomainString()
    _ = try runLaunchctlCommand(args: ["bootstrap", domain, plistPath])
}
```

When the bootstrap fails, `SystemStart` proceeds to "Testing access to container-apiserver..." and dies with a generic XPC error that points the user back at the very command that just failed ("Ensure container system service has been started with `container system start`"). The actual launchctl error — which names the domain and the real reason — is never shown. This cost us several debugging round-trips chasing the XPC layer before reading the source revealed the ping was doomed before it ran.

**2. The computed domain is wrong for sudo-in-a-background-session.** `getDomainString()` maps `launchctl managername` + `getuid()`: Aqua → `gui/<uid>`, Background → `user/<uid>`. Under `sudo` in a CI runner's session, that computes `user/0`/`gui/0`, which can't be bootstrapped into — so registration always fails there. This is the same class of problem as #1514, but note the service plist already includes `System` in `LimitLoadToSessionType`, so the system domain looks intended to be supported.

**Suggested behavior:**
- At minimum: `register()` should check launchctl's exit status and surface its stderr in the thrown error. That one change would have made the root cause obvious from the first failure.
- Ideally: when running as root (euid 0) outside an Aqua session, bootstrap into the `system` domain. We verified this works end to end as a workaround — after `container system start` writes the plist, running `sudo launchctl bootstrap system <plist>` and re-running `container system start` brings everything up (apiserver, machine-apiserver, plugins, kernel install), and Mach lookups chain gui → user → system so the service is reachable from any session. Working example: https://github.com/cap10morgan/whalebridge/blob/7de7456/.github/workflows/ci.yml (the "Start apple/container services" step; also being adopted by socktainer/socktainer#325).

### Environment

- OS: macOS 26 (GitHub Actions `macos-26` runner image, Apple silicon)
- Xcode: Xcode 26 (runner default)
- Container: container CLI version 1.1.0 (release pkg)

### Code of Conduct

rogerneal:
## Summary
- Surface `launchctl bootstrap` exit status and stderr from `ServiceManager.register` instead of continuing into a misleading apiserver XPC error.
- Keep `container system start` idempotent when the service is already registered.
- Add unit tests for domain selection helpers and a real `launchctl` failure path for error surfacing.

The system-domain bootstrap change for root outside Aqua is intentionally left to a follow-up PR so this half can merge on its own evidence.

Fixes #2008

### Tested
- [x] `swift test --filter ServiceManagerTests` (5 tests, all passing after rebase)
- [x] Real `launchctl` failure via `ServiceManager.register` with a missing plist; error includes `launchctl bootstrap` and status
- [x] Local Aqua `system start`: `register()` loaded `com.apple.container.apiserver` and the process started

### Not in this PR
- [ ] Root outside Aqua bootstrapping into the `system` domain (follow-up)
- [ ] Exact GitHub Actions `macos-26` Background / non-login `sudo container system start` reproduction

### DIFF
--- a/Sources/ContainerPlugin/ServiceManager.swift
+++ b/Sources/ContainerPlugin/ServiceManager.swift
@@ -18,25 +18,45 @@ import ContainerizationError
 import Foundation
 
 public struct ServiceManager {
-    private static func runLaunchctlCommand(args: [String]) throws -> Int32 {
+    private static func runLaunchctlCommand(args: [String]) throws -> (status: Int32, stderr: String) {
         let launchctl = Foundation.Process()
         launchctl.executableURL = URL(fileURLWithPath: "/bin/launchctl")
         launchctl.arguments = args
 
         let null = FileHandle.nullDevice
+        let stderrPipe = Pipe()
         launchctl.standardOutput = null
-        launchctl.standardError = null
+        launchctl.standardError = stderrPipe
 
         try launchctl.run()
+        let stderrData = stderrPipe.fileHandleForReading.readDataToEndOfFile()
         launchctl.waitUntilExit()
 
-        return launchctl.terminationStatus
+        let stderr = String(data: stderrData, encoding: .utf8) ?? ""
+        return (launchctl.terminationStatus, stderr)
     }
 
     /// Register a service by providing the path to a plist.
     public static func register(plistPath: String) throws {
         let domain = try Self.getDomainString()
-        _ = try runLaunchctlCommand(args: ["bootstrap", domain, plistPath])
+        let command = "launchctl bootstrap \(domain) \(plistPath)"
+        let (status, stderr) = try runLaunchctlCommand(args: ["bootstrap", domain, plistPath])
+        guard status == 0 else {
+            // `container system start` is idempotent: if the service is already
+            // bootstrapped, launchctl returns non-zero. Treat that as success.
+            // Use try? so a launchctl spawn failure does not replace the bootstrap error.
+            if let label = try? launchdLabel(fromPlistAt: plistPath),
+                (try? isRegistered(fullServiceLabel: label)) == true
+            {
+                return
+            }
+            let detail = stderr.trimmingCharacters(in: .whitespacesAndNewlines)
+            let message =
+                detail.isEmpty
+                ? "command `\(command)` failed with status \(status)"
+                : "command `\(command)` failed with status \(status), message: \(detail)"
+            throw ContainerizationError(.internalError, message: message)
+        }
     }
 
     /// Deregister a service by a launchd label.
@@ -46,7 +66,7 @@ public struct ServiceManager {
 
     /// Deregister a service and pass return status
     public static func deregister(fullServiceLabel label: String, status: inout Int32) throws {
-        status = try runLaunchctlCommand(args: ["bootout", label])
+        status = try runLaunchctlCommand(args: ["bootout", label]).status
     }
 
     /// Restart a service by a launchd label.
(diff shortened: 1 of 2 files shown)


singhible:
Thanks for picking this up @rogerneal, and for being clear about what you did and
did not test. That made this easy to review.

I checked out `f8d17d2` on macOS 26.5.2, Aqua session, non root. Your
`ServiceManagerTests` pass 7/7, `PluginLoaderTest` passes 12/12 twice in a row,
and `swift format lint --strict` is clean on both files. I could not reproduce the
non login root case either.

Three things I would raise:

**The commit is unsigned.** `pr-build.yml` runs a `verify-signatures` job that
fails on any unverified commit, so nothing can go green until that is fixed.

**I would split the domain change out.** The issue proposes two fixes, and this PR
does both. Surfacing the bootstrap failure is small and provable in any session
type. Bootstrapping into `system` moves where the services live, neither of us can
verify it, and it overlaps
[#1514](https://github.com/apple/container/issues/1514). Splitting lets the first
half merge on its own evidence instead of waiting for a `macos-26` runner.

**If the domain change stays, the rest of the code may not agree with it.**
`SystemStatus.swift:74` and `SystemStop.swift:99` both call `launchctl list`
without a domain target, so they resolve against the calling process's launchd
context rather than the `system` domain you bootstrapped into. Your own idempotency
fallback goes through `isRegistered`, so it has the same question. If they do not
line up, a second `system start` throws instead of being a no op and
`system status` reports `unregistered` for a service that is running. Worth
confirming on the runner.

Two small ones while I am here: `try isRegistered(...)` inside the guard will
replace your good error message with a spawn error if launchctl cannot start, so
`try?` is safer, and `LaunchPlist.CodingKeys.label.rawValue` already exists
instead of the hardcoded `"Label"`.

I think the core idea is okay, and it matches what `enumerate()` in the same type
already does.

rogerneal:
@singhible Thanks for the careful review. I took your suggestions:

1. Commit is now signed and verified on GitHub.
2. Split out the system-domain change. This PR only surfaces bootstrap failures and keeps start idempotent.
3. Applied `try?` around `isRegistered` so a spawn failure cannot replace the bootstrap error, and switched the plist key lookup to `LaunchPlist.CodingKeys.label.rawValue`.

Follow-up for the domain half is coming next, including the `SystemStatus` / `SystemStop` / idempotency alignment you called out.

PR #2050, opened 2026-07-31
Fix launchd domain mismatch between `container system start`/`stop`

JaewonHur:
## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
getDomainString() picked gui/<uid> or user/<uid> based on the calling shell's launchd session type (Aqua vs Background), so registering the apiserver from a GUI Terminal and stopping it from tmux/ssh computed different domains, causing bootout to silently target a domain the service was never registered in.

Always resolve to user/<uid> instead: it's the parent of the Aqua session domain for the same UID, so lookups from GUI shells still find it, and start/stop now agree regardless of which shell invokes them.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/ContainerPlugin/ServiceManager.swift
+++ b/Sources/ContainerPlugin/ServiceManager.swift
@@ -126,10 +126,14 @@ public struct ServiceManager {
         switch currentSessionType {
         case LaunchPlist.Domain.System.rawValue:
             return LaunchPlist.Domain.System.rawValue.lowercased()
-        case LaunchPlist.Domain.Background.rawValue:
+        case LaunchPlist.Domain.Background.rawValue, LaunchPlist.Domain.Aqua.rawValue:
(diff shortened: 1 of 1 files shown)


JaewonHur:
Has some issue when `container system` needs a GUI alert for user, like asking keychain password.

jglogan:
@JaewonHur what issue does this fix?

JaewonHur:
I often do `container system start` on normal terminal, or tmux terminal. Then, every time, the container launchd services are started under different domain (i.e., `gui/501` when starting from normal terminal, `user/501` when from tmux). It gets confusing when stopping container system because stopping it from different domain tricks me as if the system is stopped correctly, but not. I often wasted my time debugging leftover launchd services.

PR #2055, opened 2026-08-02
Adjust overcommit and max_map_count vm defaults in guest VMs

### LINKED ISSUE #2053: [Request]: Set container-friendly defaults for `vm.overcommit_memory` and `vm.max_map_count` in the guest VM

# 

### Summary

Container guest VMs currently boot with Linux's stock kernel defaults for two memory-related sysctls that container tooling commonly relies on being tuned: `vm.overcommit_memory` (default `0`, heuristic overcommit) and `vm.max_map_count` (default `65530`, the per-process memory-mapping ceiling). Because each container's guest VM is sized to a small, fixed amount of RAM with no swap configured, both of these stock limits are hit far more easily than on a typical bare-metal host or general-purpose VM — causing common oversubscribed-mmap and many-mapping workloads to fail inside a container even though they work fine elsewhere.

---

### Issue 1: `vm.overcommit_memory`

**Steps to Reproduce**

Run a container with a modest memory limit, then call `mmap()` for a size larger than that limit, writing to only a small portion of it:

```
container run --rm --memory=4G python:3-alpine python3 -c "
import mmap
size = 16*1024*1024*1024
m = mmap.mmap(-1, size)
print('mmap OK for', size, 'bytes')
m.close()
"
```

**Expected Result**

The `mmap` call succeeds. The kernel should only fail an allocation when the process actually tries to use more memory than is available (at page-fault time), not merely because the initial address-space reservation exceeds the memory limit.

**Actual Result**

```
mmap FAILED: [Errno 12] Out of memory
```

**Root Cause**

The guest kernel boots with the Linux default `vm.overcommit_memory=0` ("heuristic overcommit"). In this mode, the kernel rejects a new anonymous mapping outright whenever the requested size exceeds the guest's currently free RAM plus swap, regardless of whether the process will ever touch that memory. Since the guest's total RAM is sized to roughly match `--memory` (with no swap device configured), that free-memory-plus-swap figure is small relative to the limit, so any oversubscribed mmap request fails immediately.

---

### Issue 2: `vm.max_map_count`

**Steps to Reproduce**

Run a workload that creates a large number of distinct memory mappings — for example, Elasticsearch, many JVM-based applications, or a test program that repeatedly `mmap`s small anonymous regions without a corresponding `munmap`, until the process's total mapping count exceeds 65530.

**Expected Result**

Mapping creation continues to succeed well past the stock Linux default, matching the higher ceiling that most container hosts and general-purpose Linux distributions configure out of the box (commonly `262144`).

**Actual Result**

Once the process's map count reaches the default limit of `65530`, further `mmap` calls fail with `ENOMEM`, even though the workload's actual memory usage may be nowhere near the container's memory limit.

**Root Cause**

The guest kernel boots with the Linux kernel's stock default `vm.max_map_count=65530`. Nothing in the current guest boot path raises this value, so any workload that legitimately needs a higher mapping count — a well-documented requirement for several common memory-mapping-heavy applications — fails inside a container.

---

### Proposed Fix

Set container-friendly defaults for both sysctls at guest VM boot:

- `vm.overcommit_memory=1` ("always overcommit"), so the kernel skips its up-front admission check and instead relies on the existing cgroup memory limit and OOM killer to enforce memory usage at page-fault time.
- `vm.max_map_count=262144`, matching the higher ceiling many other container hosts and Linux distributions already ship with by default.

### Workaround

Until this is addressed, users can set both sysctls themselves at container startup. Writing to `/proc/sys/vm/*` only requires running as root (the default for most images) and does not require `--privileged` or extra capabilities:

```
container run --rm --memory=4G <image> sh -c "echo 1 > /proc/sys/vm/overcommit_memory && echo 262144 > /proc/sys/vm/max_map_count && <your command>"
```

### Code of Conduct

adityabagchi24:
## Type of Change

- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Description

Each container runs in its own guest VM sized to `--memory` with no swap, so the guest kernel's stock `vm` sysctl defaults are hit far too easily:

- `vm.overcommit_memory=0` (heuristic overcommit) rejects an oversized `mmap()` upfront whenever the reservation exceeds the small, swap-less VM's free RAM — even if the memory is never touched — returning `ENOMEM`.
- `vm.max_map_count=65530` caps per-process mapping count, which mapping-heavy applications (e.g. Elasticsearch, many JVMs) legitimately exceed.

Fixes #2053

## Motivation and Context

These stock defaults cause common oversubscribed-`mmap` and many-mapping workloads to fail inside a container even though they work on a typical host. This PR applies container-friendly defaults at container start so these workloads succeed, while actual memory usage remains enforced by the container's existing `--memory` limit (via cgroup memory limit and the OOM killer).

## Implementation

- `Sources/Services/RuntimeLinux/Server/RuntimeService.swift`: set `vm.overcommit_memory=1` and `vm.max_map_count=262144` as defaults when configuring a container's guest sysctls.

### DIFF
--- a/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
+++ b/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
@@ -997,9 +997,12 @@ public actor RuntimeService {
         czConfig.cpus = config.resources.cpus
         czConfig.cpuOverhead = config.resources.cpuOverhead
         czConfig.memoryInBytes = config.resources.memoryInBytes
-        czConfig.sysctl = config.sysctls.reduce(into: [String: String]()) {
-            $0[$1.key] = $1.value
-        }
+        // Overcommit memory and allow more memory mappings than the kernel default
+        // so workloads inside swap-less guest VMs hit limits less easily.
+        var sysctls = config.sysctls
+        sysctls["vm.overcommit_memory"] = "1"
+        sysctls["vm.max_map_count"] = "262144"
+        czConfig.sysctl = sysctls
         // If the host doesn't support this, we'll throw on container creation.
         czConfig.virtualization = config.virtualization
         czConfig.useInit = config.useInit


jglogan (approved):
@adityabagchi24 Thanks for taking care of this!

adityabagchi24:
@jglogan Thank you so much for merging my PR.

PR #2058, opened 2026-08-02
Bootstrap into system domain as root outside Aqua

rogerneal:
## Summary
- When running as root (`euid == 0`) outside an Aqua session, bootstrap into the `system` domain instead of `user/0` or `gui/0`.
- Align launchd helpers with that domain:
  - `isRegistered` uses `launchctl print` for domain-qualified targets
  - `SystemStatus` checks `domain/label`
  - `register` idempotency queries the same domain it bootstrapped into
  - `SystemStop` compares bare labels from `enumerate` correctly before re-qualifying for bootout

Stacked on #2045. Related to #2008 and #1514.

### Tested
- [x] Unit tests for root-outside-Aqua → `system` domain selection

### Not tested
- [ ] Exact GitHub Actions `macos-26` reproduction: `sudo container system start --enable-kernel-install`
- [ ] Confirm start succeeds without `XPC connection error: Connection invalid`
- [ ] `container system status` reports registered/running in the same sudo/root context
- [ ] Second `container system start` remains idempotent

Please merge #2045 first; this branch is based on it.

### DIFF
--- a/Sources/ContainerCommands/System/SystemStatus.swift
+++ b/Sources/ContainerCommands/System/SystemStatus.swift
@@ -71,7 +71,10 @@ extension Application {
         }
 
         public func run() async throws {
-            let isRegistered = try ServiceManager.isRegistered(fullServiceLabel: "\(prefix)apiserver")
+            // Use the same launchd domain target as register/stop so a system-domain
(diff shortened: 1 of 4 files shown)


jglogan:
@rogerneal What is your use case for running as root?

PR #2060, opened 2026-08-03
Reap completed exec host state in the Linux runtime helper

### LINKED ISSUE #2057: [Bug]: completed exec state is retained in container-runtime-linux
### Description

In Apple Container 1.2.0 (`6e65319`), a long-lived
`container-runtime-linux` process retains host-side state for every completed
exec. Repeated sequential `container machine run ... /usr/bin/true` calls make
the runtime's live allocations and physical footprint grow approximately in
proportion to the completed command count. Stopping the machine reclaims the
state because it terminates the runtime helper, but a long-lived helper does
not plateau.

### Sanitized reproduction

1. Start one machine and keep the same `container-runtime-linux` PID and guest
   boot for the entire test.
2. Record the helper's live malloc allocation count and physical footprint.
3. Run thousands of sequential, concurrency-one `/usr/bin/true` commands with
   `container machine run`.
4. Allow an idle settle period and measure the same helper again.
5. Repeat another equal command batch.

The completed-command count predicts the retained allocation growth, and
pausing the command source stops that growth. A 100-command probe left 13,589
additional live allocations in the helper in one measured run.

### Source trace

The current exec path appears to leave four per-exec objects reachable:

- `RuntimeService.processes[id]`
- `RuntimeService.waiters[id]`
- `ExitMonitor.exitCallbacks[id]`
- `ExitMonitor.runningTasks[id]`

`RuntimeService.createProcess` inserts the process and waiter and registers an
exit callback. The exit callback releases the waiter, deletes the underlying
process, and marks the process stopped, but it does not remove the process or
waiter entries or call `ExitMonitor.stopTracking`. `ExitMonitor.track` also
does not remove its completed callback/task entries.

Simply deleting the entries in the callback is racy because the current
protocol does not declare whether an exec will later be waited or is detached.
An exec can exit before a normal caller sends `wait`, while a detached exec
will never send it.

### Expected behavior

Completed normal, detached, failed-start, canceled-client, signal, and shutdown
paths should release all per-exec host state and return to a stable allocation
baseline without requiring a machine restart. Completion ownership should be
declared before start (for example, atomic wait-and-reap versus reap-on-exit),
so cleanup does not depend on whether a late wait happens to arrive.

I am preparing a focused patch and deterministic normal/detached/failed-start
stress coverage against this upstream baseline.

rogerneal:
## Summary
- `ExitMonitor.track` now drops callback/task entries after the exit path finishes, so completed work does not retain host state forever.
- `RuntimeService` reaps per-exec `processes` / `waiters` on exit, failed start, and container cleanup. Late `wait` still works via a compact `completedExits` cache when no waiter was present yet.
- Failed `createProcess` registration and failed `startExecProcess` also reap instead of leaving orphaned entries.

Fixes #2057

### Tested
- [x] `swift test --filter ExitMonitorTests` (3/3 passed)
  - releases entries after successful exit
  - releases entries after WaitHandler failure
  - 200 sequential track/exit cycles still allow re-registration
- [x] Existing `MultiWriterTests` still compile against `RuntimeService` changes

### Not tested
- [ ] Full long-lived `container-runtime-linux` malloc growth repro with thousands of `container machine run ... /usr/bin/true` (needs installing this build into a running daemon session and measuring helper allocations)
- [ ] Detached-exec path under allocation profiling

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -412,6 +412,12 @@ let package = Package(
             ],
             path: "Sources/Services/Runtime/RuntimeClient"
         ),
+        .testTarget(
+            name: "ContainerRuntimeClientTests",
+            dependencies: [
+                "ContainerRuntimeClient",
+            ]
+        ),
         .target(
             name: "ContainerResource",
             dependencies: [
--- a/Sources/Services/Runtime/RuntimeClient/ExitMonitor.swift
+++ b/Sources/Services/Runtime/RuntimeClient/ExitMonitor.swift
@@ -48,8 +48,7 @@ public actor ExitMonitor {
         if let task = self.runningTasks[id] {
             task.cancel()
         }
-        exitCallbacks.removeValue(forKey: id)
-        runningTasks.removeValue(forKey: id)
+        finishTracking(id: id)
     }
 
     /// Register long running work so that the monitor invokes
@@ -86,6 +85,14 @@ public actor ExitMonitor {
                 self.log?.error("WaitHandler for \(id) threw error \(String(describing: error))")
                 try? await onExit(id, ExitStatus(exitCode: -1))
             }
+            // Drop callback/task entries once the exit path finishes so completed
+            // work does not retain host state indefinitely (apple/container#2057).
+            await self.finishTracking(id: id)
         }
+    }
+
+    private func finishTracking(id: String) {
+        exitCallbacks.removeValue(forKey: id)
+        runningTasks.removeValue(forKey: id)
     }
 }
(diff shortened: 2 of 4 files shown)


stephenlclarke:
I found two related exit-delivery cases while reviewing this.

A repeated exit callback can attempt to resume the same checked continuation twice. Also, the internal wait used by SIGKILL can consume the cached status before the client calls `wait`.

I’ve put the fixes and three focused regressions on `upstream/exit-waiter-status`, based directly on this PR’s current head:
https://github.com/stephenlclarke/container/commit/cc1041b4c21601ac21a30eafca21aed04b2d7b8d

The idempotent-exit part came from Bhavesh Varma’s earlier #1861, so the commit credits @radheradhe01 as co-author. All three focused tests pass.

Please feel free to cherry-pick the commit or take the branch. I haven’t opened a competing PR.

PR #2064, opened 2026-08-03
Add --sysctl and --privileged flags to run/create

### LINKED ISSUE #2041: [Bug]: /proc/sys/net/ipv4/ip_forward is no longer writable (regression)
### I have done the following

### Steps to reproduce

```bash
container run --rm -it --cap-add ALL busybox sh -c "echo 1 > /proc/sys/net/ipv4/ip_forward"
```

Outputs
```txt
sh: line 0: can't create /proc/sys/net/ipv4/ip_forward: Read-only file system
```
The file is actually writable by root.  It just mounted as a read-only mount now (it seems).

I found this because I typically run dagger-engine within an apple container.  That worked with 1.1.0 but no longer works with 1.2.0.

My actual use case is more closely

```bash
container run --rm -it --cap-add ALL registry.dagger.io/engine:v0.21.7
```

```txt
#!/bin/sh
set -e

cat $0

# cgroup v2: enable nesting
# see https://github.com/moby/moby/blob/38805f20f9bcc5e87869d6c79d432b166e1c88b4/hack/dind#L28
if [ -f /sys/fs/cgroup/cgroup.controllers ]; then
        # move the processes from the root group to the /init group,
        # otherwise writing subtree_control fails with EBUSY.
        # An error during moving non-existent process (i.e., "cat") is ignored.
        mkdir -p /sys/fs/cgroup/init
        xargs -rn1 < /sys/fs/cgroup/cgroup.procs > /sys/fs/cgroup/init/cgroup.procs || :
        # enable controllers
        sed -e 's/ / +/g' -e 's/^/+/' < /sys/fs/cgroup/cgroup.controllers \
                > /sys/fs/cgroup/cgroup.subtree_control
fi

# expect more open files due to per-client SQLite databases
# many systems default to 1024 which is far too low
ulimit -n 1048576 || echo "cannot increase open FDs with ulimit, ignoring"

exec tini -- /usr/local/bin/dagger-engine --config /etc/dagger/engine.toml "$@"
INFO[0000] starting dagger engine version:v0.21.7 v0.21.7 linux/arm64/v8 
INFO[0000] configuring iptables backend: legacy         
INFO[0000] set /usr/bin/iptables -> /usr/bin/xtables-legacy-multi 
INFO[0000] set /usr/bin/iptables-save -> /usr/bin/xtables-legacy-multi 
INFO[0000] set /usr/bin/iptables-restore -> /usr/bin/xtables-legacy-multi 
INFO[0000] set /usr/bin/ip6tables -> /usr/bin/xtables-legacy-multi 
INFO[0000] set /usr/bin/ip6tables-save -> /usr/bin/xtables-legacy-multi 
INFO[0000] set /usr/bin/ip6tables-restore -> /usr/bin/xtables-legacy-multi 
INFO[0000] using ipMasqBackend: iptables                
INFO[0000] detected mtu 1280 via interface eth0         
INFO[2026-07-29T11:10:44Z] engine name: 84c39228-0023-4137-94d2-2d9b84e872d4 
DEBU[2026-07-29T11:10:44Z] creating engine GRPC server                  
DEBU[2026-07-29T11:10:44Z] creating engine lockfile                     
DEBU[2026-07-29T11:10:44Z] creating engine server                       
INFO[2026-07-29T11:10:44Z] auto snapshotter: using overlayfs            
DEBU[2026-07-29T11:10:44Z] could not read "/var/lib/dagger/net/cni" for cleanup: open /var/lib/dagger/net/cni: no such file or directory 
shutting down telemetry...
telemetry shut down complete
dagger-engine: failed to create engine: failed to create network providers: CNI setup error: plugin type="bridge" failed (add): failed to enable forwarding: open /proc/sys/net/ipv4/ip_forward: read-only file system
```

I like running dagger engine in an apple container because I get more isolation than just running it in podman as a privileged container.  `--cap ALL` has been enough in the past to make this work fine.

### Problem description

dagger-engine no longer works in apple containers.  Worked in 1.1.0 but not in 1.2.0.

### Environment

```markdown
- OS: 26.5.2
- Xcode: 26.6
- Container: 1.2.0
```

### Code of Conduct

ronandi:
Fixes #2041.

## Problem

container 1.2.0 picked up containerization's secure-by-default changes (default `readonlyPaths` includes `/proc/sys`, restricted capability set) with no CLI opt-out. Guests can no longer write sysctls or remount mounts, so Kubernetes node images (`kindest/node`, `rancher/k3s`) and docker-in-docker engines like dagger fail to boot.

## Change

Two Docker-compatible flags on `run`/`create`:

- `--sysctl key=value` (repeatable): sets guest sysctls at creation. vminitd applies them before the read-only remount, so they compose with the default readonly paths.
- `--privileged`: grants the full capability set and clears the default masked/readonly paths. Every container is its own VM with its own kernel, so an explicit opt-in escape hatch is proportionate and matches Docker's UX.

The API plumbing (`sysctls`, `readonlyPaths`, `maskedPaths` on `ContainerConfiguration`) already existed — this exposes it.

## Verification (against 1.2.0 services)

- `--sysctl net.ipv4.ip_forward=1`: value lands in the guest (`cat` = 1)
- `--privileged` ubuntu: `sysctl -w` succeeds, `mount -o remount,ro /sys` succeeds
- `--privileged` kindest/node:v1.36.1: boots cleanly (containerd active, kubelet starts) — fails hard without the flag
- Added `Parser.sysctls` unit tests (note: `swift test` needs full Xcode; I only had CLT locally, so tests are unrun locally — CI should cover)

Defaults are unchanged: without the flags, behavior is exactly 1.2.0.

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -191,6 +191,8 @@ public struct Flags {
             runtime: String?,
             ssh: Bool,
             shmSize: String?,
+            sysctl: [String] = [],
+            privileged: Bool = false,
             tmpFs: [String],
             useInit: Bool,
             virtualization: Bool,
@@ -221,6 +223,8 @@ public struct Flags {
             self.runtime = runtime
             self.ssh = ssh
             self.shmSize = shmSize
+            self.sysctl = sysctl
+            self.privileged = privileged
             self.tmpFs = tmpFs
             self.useInit = useInit
             self.virtualization = virtualization
@@ -344,6 +348,15 @@ public struct Flags {
 
         @Option(name: .customLong("shm-size"), help: "Size of /dev/shm (e.g. 64M, 1G)")
         public var shmSize: String?
+
+        @Option(
+            name: .long,
+            help: .init("Set a sysctl in the container (format: key=value, repeatable)", valueName: "key=value")
+        )
+        public var sysctl: [String] = []
+
+        @Flag(name: .long, help: "Give the container the full capability set and disable the default masked and read-only paths")
+        public var privileged = false
 
         @Option(name: .customLong("tmpfs"), help: "Add a tmpfs mount to the container at the given path")
         public var tmpFs: [String] = []
--- a/Sources/Services/ContainerAPIService/Client/Parser.swift
+++ b/Sources/Services/ContainerAPIService/Client/Parser.swift
@@ -1024,6 +1024,18 @@ public struct Parser {
 
     /// Parse and validate --cap-add / --cap-drop arguments.
     /// Returns normalized uppercase CAP_* strings.
+    public static func sysctls(_ rawSysctls: [String]) throws -> [String: String] {
+        var result: [String: String] = Dictionary(minimumCapacity: rawSysctls.count)
+        for item in rawSysctls {
+            let parts = item.split(separator: "=", maxSplits: 2)
+            guard parts.count == 2, !parts[0].isEmpty, !parts[1].isEmpty else {
+                throw ContainerizationError(.invalidArgument, message: "invalid sysctl format \(item) — expected key=value")
+            }
+            result[String(parts[0])] = String(parts[1])
+        }
+        return result
+    }
+
     public static func capabilities(capAdd: [String], capDrop: [String]) throws -> (capAdd: [String], capDrop: [String]) {
         var normalizedAdd: [String] = []
         normalizedAdd.reserveCapacity(capAdd.count)
--- a/Sources/Services/ContainerAPIService/Client/Utility.swift
+++ b/Sources/Services/ContainerAPIService/Client/Utility.swift
@@ -257,6 +257,18 @@ public struct Utility {
         config.capDrop = caps.capDrop
         config.stopSignal = imageConfig?.stopSignal
 
+        config.sysctls = try Parser.sysctls(management.sysctl)
+
+        if management.privileged {
+            // Docker-compatible privileged mode: the full capability set and
+            // none of the default masked/read-only paths. Proportionate here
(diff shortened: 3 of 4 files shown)


jglogan:
@ronandi Have you tried this sort of thing for sysctl:

```
--kernel-arg sysctl.net.ipv4.ip_forward=1
```

ronandi:
Thank you for that @jglogan!  `--kernel-arg sysctl.net.ipv4.ip_forward=1` does work on 1.2.0 (verified: guest reads 1). That covers the sysctl half

I'm trying to run a kube cluster and `kindest/node`'s entrypoint dies at `mount -o remount,ro /sys` with EPERM, and kernel args can't help there. With `--privileged` it boots cleanly (containerd + kubelet come up).

Use case, for context: running a long-lived local Kubernetes cluster on apple/container (kubeadm/k3s node images). Kubernetes nodes fundamentally need to write sysctls and remount /sys at boot, so 1.2.0's defaults make node images unbootable

prskr:
I don't mean to hijack this discussion but how would that apply to container machines? From what I can tell the `--kernel-arg` flag is not exposed for any `container machine` sub-command but they're facing the same issue that `/proc/sys` is not writable and therefore - if you'd use a container machine for instance as a secure dev machine - parameters like `fs.inotify.max_user_instances` are quickly becoming a limitation for editors connecting remotely, let alone a Docker / Podman setup 🙈

jglogan:
@ronandi Thanks for the contribution, but we're going to close it as we're working on getting a different fix out quickly for this one. We're not ready to a `--privileged` flag for container at this time; rather, we'll expose the fine grained controls for power users.

As a quick fix we're going to make the OCI path controls manageable as experimental CLI options that we'll deprecate once we can get a more general solution in place. We'll do a patch release at the end of the week that includes the fix.

PR #2069, opened 2026-08-04
Add `--read-only-path` and `--masked-path` option to container run / create

### LINKED ISSUE #2041: [Bug]: /proc/sys/net/ipv4/ip_forward is no longer writable (regression)

adityaramani:
Closes https://github.com/apple/container/issues/2041
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
The 1.2.0 release of container updated the underlying containerization dependency. This caused the default runtime service to mark a set of default paths as masked / readonly inside of the container. This change adds the option to override these paths if needed while creating a new container

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Flags.swift
+++ b/Sources/Services/ContainerAPIService/Client/Flags.swift
@@ -178,6 +178,7 @@ public struct Flags {
             kernel: String?,
             kernelArgs: [String],
             labels: [String],
+            maskedPaths: [String],
             mounts: [String],
             name: String?,
             networks: [String],
@@ -186,6 +187,7 @@ public struct Flags {
             publishPorts: [String],
             publishSockets: [String],
             readOnly: Bool,
+            readonlyPaths: [String],
             remove: Bool,
             rosetta: Bool,
             runtime: String?,
@@ -208,6 +210,7 @@ public struct Flags {
             self.kernel = kernel
             self.kernelArgs = kernelArgs
             self.labels = labels
+            self.maskedPaths = maskedPaths
             self.mounts = mounts
             self.name = name
             self.networks = networks
@@ -216,6 +219,7 @@ public struct Flags {
             self.publishPorts = publishPorts
             self.publishSockets = publishSockets
             self.readOnly = readOnly
+            self.readonlyPaths = readonlyPaths
             self.remove = remove
             self.rosetta = rosetta
             self.runtime = runtime
@@ -291,6 +295,16 @@ public struct Flags {
         @Option(name: [.short, .customLong("label")], help: "Add a key=value label to the container")
         public var labels: [String] = []
 
+        /// EXPERIMENTAL: The flag is subject to change.
+        @Option(
+            name: .customLong("masked-path"),
+            help: .init(
+                "[EXPERIMENTAL] Hide a path inside the container, in addition to the runtime defaults (or NONE to clear prior values and the defaults)",
+                valueName: "path"
+            )
+        )
+        public var maskedPaths: [String] = []
+
         @Option(name: .customLong("mount"), help: "Add a mount to the container (format: type=<>,source=<>,target=<>,readonly)")
         public var mounts: [String] = []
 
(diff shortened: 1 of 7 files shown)


jglogan (approved):

jglogan:
Closes #2077.

PR #2071, opened 2026-08-05
Simplify correlation of test prefix to test ID.

jglogan:
- At present in the integration tests, it's difficult for humans to look at logs and correlate the test prefix value we use to keep resource names short to the test suite and method.
- Adds the test name to the log metadata so both are readily at hand.
- Add test lifecycle messages to ContainerFixture for more accurate timing measurement.

### DIFF
--- a/Sources/ContainerTestSupport/ContainerFixture.swift
+++ b/Sources/ContainerTestSupport/ContainerFixture.swift
@@ -141,10 +141,12 @@ public final class ContainerFixture: Sendable {
             let result = try await body(fixture)
             logger.info("test end", metadata: ["result": "pass"])
             await fixture.runCleanup()
+            logger.info("test cleaned up")
             return result
         } catch {
             logger.info("test end", metadata: ["result": "fail", "error": "\(error)"])
             await fixture.runCleanup()
+            logger.info("test cleaned up")
             throw error
         }
     }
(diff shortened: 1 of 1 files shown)


Sources/ContainerTestSupport/ContainerFixture.swift
```diff
@@ -80,38 +80,54 @@ public final class ContainerFixture: Sendable {
...
+        // — the same identifier format used in the swift-testing event-stream JSON.
+        let testIdentifier = Test.current.map { "\($0.id)" }
+        let suiteName = testIdentifier?.split(separator: "/", maxSplits: 1).first.map(String.init) ?? "unknown"
+
+        // Swift Testing doesn't expose a stable per-case identifier or the case's arguments
+        // publicly, only `isParameterized`. Parameterized tests share one `testName` across all
+        // their concurrently-running cases, so fall back to the per-invocation `testID` to keep
+        // each case's log file distinct.
+        let isParameterized = Test.Case.current?.isParameterized ?? false
+        let logFileName = isParameterized ? "\(testName)-\(testID).log" : "\(testName).log"
```
JaewonHur, line 98: We can use `Test.Case.current?.arguments` to append passed argument in log file name.

JaewonHur (approved):

PR #2079, opened 2026-08-06
refactor: extract K8s logic into ContainerK8s library target

### LINKED ISSUE #2078: [k8s] Extract K8s commands into a ContainerK8s library target
### Summary

  Extract all K8s logic out of the `container-k8s` plugin executable and into a new `ContainerK8s` library target. The plugin binary in `Sources/Plugins/K8s/` becomes a thin entry point that calls `K8sCommand.main()`. No behavior changes are expected

  ---

  ### Background

Currently all K8s logic — commands, kubeadm provisioning, kubeconfig management, cluster helpers — lives directly inside the `container-k8s` plugin executable target. This makes it impossible for other targets to import and extend K8s commands without forking the binary. A library extraction gives K8s a clean, importable surface while keeping the plugin pattern intact.

  ---

  ### Changes

  #### 1. New `Sources/ContainerK8s/` library

  Move all Swift source files from `Sources/Plugins/K8s/` (except the entry point) into `Sources/ContainerK8s/`. File names and contents are unchanged.

Sources/ContainerK8s/
 ├── Commands/
            ├── K8sCommand.swift
            ├── K8sCreate.swift
            ├── K8sDelete.swift
            ├── K8sList.swift
            ├── K8sLoadImage.swift
            ├── K8sStart.swift
      └── K8sWriteConfig.swift
  ├── K8sHelper.swift
  ├── KubeConfig.swift
  └── Resources/
      └── kindnet.yaml

  All types used by the entry point must be `public`.

  #### 2. `Sources/Plugins/K8s/` becomes a thin entry point

  Only the entry point Swift file and `config.toml` remain:

  Sources/Plugins/K8s/
  ├── config.toml          (unchanged)
  └── K8sMain.swift

  3. Package.swift

  Add the new library target and update the plugin executable to depend on it:

  ---
  Acceptance criteria
  
  - swift build succeeds
  - container k8s create, start, delete, list, load-image, write-config behave identically
  - Sources/Plugins/K8s/ contains only config.toml and the entry point file
  - ContainerK8s compiles independently of the plugin target
  - ContainerK8s has no dependency on targets unrelated to K8s (e.g. ContainerBuild, ContainerXPC)

### Code of Conduct

jshi991:
Closes #2078
Move all K8s sources from the container-k8s executable into a new ContainerK8s library target. Sources/Plugins/K8s/ becomes a thin entry point (K8sMain.swift) that calls K8sCommand.main().


Sources/ContainerTestSupport/WarmupImage.swift
```diff
@@ -23,6 +23,7 @@ public enum WarmupImage: String, CaseIterable, Sendable {
     case alpine320 = "ghcr.io/linuxcontainers/alpine:3.20"
     case alpine318 = "ghcr.io/linuxcontainers/alpine:3.18"
     case busybox136 = "ghcr.io/containerd/busybox:1.36"
+    case kindestNodeV1_35_5 = "docker.io/kindest/node:v1.35.5@sha256:ce977ae6d65918d0b58a5f8b5e940429c2ce42fa3a5619ec2bbc60b949c0ac95"
```
jglogan, line 26: Were you going to adapt your serial tests to restore the warmup image for each one that runs a k8s cluster?

Sources/Plugins/K8s/K8sCommand.swift
```diff
@@ -17,9 +17,8 @@
 import ArgumentParser
 import ContainerVersion
 
-@main
-struct K8sCommand: AsyncParsableCommand {
-    static let configuration = CommandConfiguration(
+public struct K8sCommand: AsyncParsableCommand {
```
jglogan, line 21: This one can go back in Plugins/K8s.

All of the leaf commands in this library need to be made public.

Sources/ContainerK8s/K8sHelper.swift
```diff
@@ -291,8 +291,8 @@ struct K8sHelper {
         sysctl -w net.bridge.bridge-nf-call-ip6tables=1 2>/dev/null || true
         systemctl restart containerd
         ctr -n k8s.io images tag registry.k8s.io/pause:3.10 registry.k8s.io/pause:3.10.1 2>/dev/null || true
-        iptables -t mangle -A OUTPUT  -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss 1220
-        iptables -t mangle -A FORWARD -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss 1220
+        /usr/sbin/iptables-nft -t mangle -A OUTPUT  -p tcp --tcp-flags SYN,RST SYN -j TCPMSS --set-mss 1220 2>/dev/null || true
```
jglogan: Why dump the diagnostic output to /dev/null? I think we'd want to see that.

Why don't we let the command just fail if it doesn't work, instead of ignoring failure?

jglogan (changes requested):
Just a few changes, see comments. Thanks @jshi991!

Sources/ContainerK8s/K8sHelper.swift
egernst: since this is different than the rest of the commit, you mind just adding this one as its own commit and describe the rationale on it?

jshi991: yup I can do that do you want this to be left with just the two commits

Sources/Plugins/K8s/K8sCommand.swift
jshi991: got it

Sources/ContainerTestSupport/WarmupImage.swift
jshi991: yup

Sources/ContainerK8s/K8sHelper.swift
jshi991: yup ill fail and log

jglogan: @jshi991 Yes. Ideally three commits:

- Library extraction
- nftables
- test optimization (see https://github.com/apple/container/blob/main/Tests/IntegrationTests/Images/TestCLIImagePruneSerial.swift#L62 for restoreWarmupImage example)

jglogan (approved):

PR #2080, opened 2026-08-06
Update sudoers file to not include dots

tumd:
Files under /etc/sudoers.d/ must not contain dots as these will be ignored as "backup-files".

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Many have usernames formated as "first.lastname". These users will currently be ignored by sudoers.

See https://superuser.com/questions/1820104/why-does-sudoers-d-ignore-filenames-with-dots-in-them

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs


jglogan (approved):

jglogan:
@tumd Looks good! Thank you for the fix!

PR #2095, opened 2026-08-07
Return nil if no local image of matching label found

### LINKED ISSUE #2098: [Bug]: `image push` can push a different image tag than `latest`.
### I have done the following

### Steps to reproduce

```console
container build -t <imageURL>:latest .
container tag <imageURL>:latest <imageURL>:123456
container image rm <imageURL>:latest

container image push <imageURL>:latest
```

### Problem description

Last image push command should throw not found error, but it pushes <imageURL>:123456 instead.
The built image has an annotation of <imageURL>:latest, so it's filtered in localImageMatches there, and the guard branch returns localImageMatches.first.

However, <imageURL>:latest tag is already deleted, and it should actually return nil.

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

JaewonHur:
- Closes #2098.

Locally built images have annotation field `com.apple.containerization.image.name` whose value is the tag given when building the image. Image management checks that annotation field to prioritize locally built images over the fetched ones. However, current implementation was wrongly returning the locally built image even when that image is retagged to another reference. This PR fixes it by removing the code block that blindly returns the image, and forces it to always check the current tag.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Testing
- [X] Tested locally
- [ ] Added/updated tests
- [ ] Added/updated docs

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/ClientImage.swift
+++ b/Sources/Services/ContainerAPIService/Client/ClientImage.swift
@@ -185,7 +185,7 @@ extension ClientImage {
     public static func get(reference: String, containerSystemConfig: ContainerSystemConfig) async throws -> ClientImage {
         let all = try await self.list()
         guard let found = try self._search(reference: reference, in: all, containerSystemConfig: containerSystemConfig) else {
-            throw ContainerizationError(.notFound, message: "image with reference \(reference)")
+            throw ContainerizationError(.notFound, message: "image with reference \(reference) not found")
         }
         return found
     }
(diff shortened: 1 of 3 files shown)


adityaramani (commented):
Want to understand a bit more about this change.. 

The `locallyBuiltImage` block exists to handle the case where the user builds as image without the registry host in the reference, since buildkit auto adds the hostname `docker.io`

But the `container-builder-shim` adds an annotation to the image which contains the true value of the reference that the user asked us to build.

Just want to see if nothing in that flow breaks cause of this

JaewonHur:
@adityaramani Could you give example scenario for this?

adityaramani (approved):

PR #2097, opened 2026-08-07
Refactor `container k8s` to use plugin resource management.

### LINKED ISSUE #2096: [Bug]: `container k8s` resource load doesn't work from installation package.
### I have done the following

### Steps to reproduce

```console
% container k8s create
Waiting for node to boot: ["node": k8s-dev]
Preparing node: ["id": k8s-dev]t [4s]
Running kubeadm init: ["node": k8s-dev]
Removing control-plane taint for single-node scheduling: ["node": k8s-dev]
Applying kindnet CNI: ["node": k8s-dev]
k8s/resource_bundle_accessor.swift:12: Fatal error: could not load resource bundle: from /usr/local/libexec/container/plugins/k8s/bin/container_k8s.bundle or /Users/runner/actions-runner/_work/container/container/.build/arm64-apple-macosx/release/container_k8s.bundle
zsh: trace trap  container k8s create
```

### Problem description

Regrettably this bug landed in the 1.2.1 release as integration tests didn't catch it.

The K8s plugin used a macOS resource bundle for the kindnet.yaml resource, but that doesn't work in the unix-layout pkg file produced for the release.

We'll fix this and put out a patch release ASAP!

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

jglogan:
- Closes #2096.
- Storing resources in a macOS resource bundle won't work for unix-layout plugins.
- Rework the kindnet resource fetch to create a plugin loader, find the plugin based on the plugin executable path, and use the plugin's resourceURL to locate the resource.
- Add a findPlugin function to locate the plugin whose executable path matches the supplied path.
- Fix help text to insure example runs regardless of the default registry configuration.

## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Needed to fix released version of k8s plugin.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs


Sources/ContainerPlugin/PluginLoader.swift
```diff
@@ -198,6 +199,32 @@ extension PluginLoader {
...
+                continue
+            }
+            if binaryPath == resolvedPath {
+                return plugin
+            }
+        }
+        return nil
+    }
+
+    private static func resolveSymlinks(_ path: String) -> String? {
```
JaewonHur, line 221: Can't we use `resolvingSymlinks` defined in `FilePath+Symlink.swift`?
Though we need to add dependency on `ContainerPersistence` for that.

JaewonHur (approved):

JaewonHur (approved):

Sources/ContainerPlugin/PluginLoader.swift
jglogan: 100% the right idea.

We can live with this bit of non-DRY code for now.

How we're going to tackle this tech debt is to coalesce the different FilePath helpers we have to https://github.com/apple/containerization/blob/main/Sources/ContainerizationOS/FilePathOps.swift.

Then we address stuff like this as we migrate the parts of container that haven't yet been migrated from String/URL paths to FilePath.

PR #398, opened 2025-07-30
Apple Container's implementation of compose functionality using familiar docker-compose YAML syntax.

mazdak:
Container Compose brings the convenience of docker-compose to Apple Container, enabling the user to:
- Define multi-container applications in a single YAML file
- Manage service dependencies automatically
- Start, stop, and manage entire application stacks with simple commands
- Use environment variable interpolation for flexible configurations

Not all features of the docker compose tool are supported due to limitations of the Apple Container system.

Issue #235, opened 2025-06-18
[Request]: Support Docker Compose Files


stevenanthonyrevo:
### Feature or enhancement request details

I'd appreciate the ability to run compose files directly in some capacity and please correct me if compose is already supported.

Supporting the command for `docker-compose.yml` files.

I'm requesting/asking for support for [Compose](https://docs.docker.com/compose/intro/compose-application-model/)

```
container compose up
```

Excellent work, Thanks! 

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

andreybleme:
This is what you're looking for https://github.com/noghartt/container-compose

thinkverse:
This request can be closed, this has been brought up before. Refer to the discussions in https://github.com/apple/container/discussions/194.

PR #2123, opened 2026-08-15
feat(compose): add persistent Docker Compose support

RobertDeRose:
## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

Inspired by the **k8s' plugin** implementation and `container machines`, this feels like the correct way to finally support docker-compose.yml workflows in an efficient way that doesn't require a full compose-to-container translation layer.

This adds `container compose` support backed by a persistent Linux container machine
running Docker Engine and the real Docker Compose CLI.

This provides a Docker-compatible Compose workflow without Docker Desktop while
keeping Compose state isolated inside a managed `compose` machine.

## What Changed

- Added the `container compose` plugin.
- Automatically builds and provisions the bundled Compose machine image.
- Persists the `compose` machine and its Docker data across invocations.
- Forwards Compose arguments unchanged to `/usr/bin/docker compose`.
- Added `container machine start` for generic stopped-machine recovery.
- Added optional Docker socket access through `--socket-path`.
- Added machine DNS aliases and published-port diagnostics.
- Added configurable CPU, memory, and idle-shutdown settings.
- Added ownership tokens and safe orphan/cancellation handling.
- Added daemon-root-aware configuration loading.
- Documented lifecycle, filesystem, networking, socket, and configuration behavior.

## Important Behavior

- `container compose down` removes Compose resources but does not delete or stop the machine.
- All Compose projects share the persistent `compose` machine.
- Bind mounts must be below the mounted macOS home directory.
- Published ports are reachable through the Compose machine address, not automatically through macOS loopback.
- `--socket-path` reports the endpoint only; it never starts the machine.
- Configuration changes require restarting the container service.

## Testing

- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

Validation completed:

- 814 unit tests across 91 suites passed.
- Machine lifecycle integration tests passed, including first boot and starting a stopped machine.
- Warnings-as-errors build passed.
- `git diff --check` passed.

## Related Issues and Discussions

 - #55 — “compose support?” — closed request for Docker Compose file support.
 - #235 — “Support Docker Compose Files” — closed duplicate/related request.
 - Discussion #320 — discusses implementing Compose as a plugin.
 - Discussion #1410 — ongoing plugin/API compatibility discussion.
 - PR #239 and PR #398 — prior Compose-related implementations.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,6 +135,8 @@ $(STAGING_DIR):
 	@mkdir -p "$(join $(STAGING_DIR), libexec/container/plugins/machine-apiserver/resources)"
 	@mkdir -p "$(join $(STAGING_DIR), libexec/container/plugins/k8s/bin)"
 	@mkdir -p "$(join $(STAGING_DIR), libexec/container/plugins/k8s/resources)"
+	@mkdir -p "$(join $(STAGING_DIR), libexec/container/plugins/compose/bin)"
+	@mkdir -p "$(join $(STAGING_DIR), libexec/container/plugins/compose/resources)"
 
 	@install "$(BUILD_BIN_DIR)/container" "$(join $(STAGING_DIR), bin/container)"
 	@install "$(BUILD_BIN_DIR)/container-apiserver" "$(join $(STAGING_DIR), bin/container-apiserver)"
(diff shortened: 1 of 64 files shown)


stephenlclarke:
Nice approach. Could the top-level docs make the dependency boundary explicit: Docker Desktop and a host Docker install aren’t required, but the plugin provisions Docker Engine/Compose inside the managed machine (and builds that image on first use)? That distinction will help set expectations.

RobertDeRose:
> Nice approach. Could the top-level docs make the dependency boundary explicit: Docker Desktop and a host Docker install aren’t required, but the plugin provisions Docker Engine/Compose inside the managed machine (and builds that image on first use)? That distinction will help set expectations.

Updated based on your comment.

jglogan:
Hi Robert, thank you for your contribution, but embedding docker/compose is not within the scope of this project. It requires far more effort to maintain than the `k8s` capability.

We've answered it in many prior issues, PRs, and discussion topics, but the TL;DR is that we'd prefer Compose to be an external plugin (see the work of @stephenlclarke, @Mcrich23, and others).

RobertDeRose:
> Hi Robert, thank you for your contribution, but embedding docker/compose is not within the scope of this project. It requires far more effort to maintain than the `k8s` capability.
> 
> We've answered it in many prior issues, PRs, and discussion topics, but the TL;DR is that we'd prefer Compose to be an external plugin (see the work of @stephenlclarke, @Mcrich23, and others).

@jglogan All the external solutions fail to actually work while this design works perfectly. I am not sure how it's more than k8s and it's essentially the same thing just that I am using the container machine feature which honestly the k8s feature could benefit from as well.

But that's fine, but there are still things about this PR worth considering that I would like to pull out and turn into separate PRs, but only if you are open to them.

1. Add the `stop` command to the `container machine` list of commands.
2. The automatic shutdown feature. This is truly useful, especially for buildkit.
  - add a generic way to monitor a service for liveliness and if it is not in use for a configurable amount of time, shut it down.
  - this allows you to over provision with a little more latitude because it allows macOS to actually reclaim the RAM once shutdown. So, if you configure buildkit to use 8GB for a large set of builds, instead of that RAM being held by the microvm forever, the auto-shutdown reclaims it.
  - even if this was not a generic feature an just how buildkit worked by default would be better because the startup is milliseconds, but the RAM usage is forever, unless you manually stop it.

While I can live without 2 even though I think it a real missing feature, 1 would be genuinely helpful I think to all, though I guess it is just a `container machine run poweroff` convenience, having it request a gracefully shutdown generically seems more natural.

thanks for the consideration.

PR #2125, opened 2026-08-15
Fix supplementary groups for machine users

### LINKED ISSUE #2108: [Bug]: `usermod -aG docker $USER` has no effect in `container machine`
### I have done the following

### Steps to reproduce

Create a container machine.
Add user to a group using `usermod -aG <group> <user>`.

Reopen shell using `container machine run`.
Groups of the user is not updated.

### Problem description

This is because `sbin.machine/init -s` (which is invoked on `machine run`) just spawns the shell, not running user login procedures.

Appending `-l` to the shell command doesn't solve this issue.

To fix this, we need to 
1. Run `exec su -l "$USER"` in `sbin.machine/init` instead of just running `exec "${USER_SHELL:-${SHELL}}"`
2. Update `MachineRun.swift` to run `sbin.machine/init` process as root to avoid prompting password for `su -l`
3. Pass username as $USER env variable

Related to #1991 

### Environment

```markdown
- OS: 
- Xcode: 
- Container:
```

### Code of Conduct

rupinajay:
### Summary

Fixes #2108.

`container machine run` was constructing the default machine user as a numeric `.id(uid:gid:)`, which prevented the guest containerization runtime from resolving supplementary groups from `/etc/group`.

Updating `UserSetup.user` to return `.raw(userString: username)` when a username is available allows the existing guest credential-resolution mechanism in the containerization runtime to look up `/etc/passwd` and `/etc/group` inside the guest VM. Consequently, supplementary groups added to the machine user (e.g. via `usermod -aG <group> <user>` or `addgroup`) are dynamically inherited on subsequent `container machine run` invocations.

This fix uses the existing `ProcessConfiguration.User.raw` abstraction designed for username lookup and avoids running `/sbin.machine/init` as root or introducing `su -l`.

### Testing

- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Production build:
`swift build --target ContainerBuild` completed successfully.

Formatting & linting:
`make fmt` and `git diff --check` completed cleanly with no warnings or errors. Pre-commit hooks installed.

Unit test suite:
`make test` executed locally and passed all 757 unit tests (0 failures).

Machine integration test:
Added `testMachineUsermodSupplementaryGroups` in `Tests/IntegrationTests/Machine/TestCLIMachineRuntimeSerial.swift`. Verified end-to-end inside an actual VM that supplementary groups created with `addgroup` and assigned to the user are dynamically reflected in subsequent `container machine run` commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,8 +186,15 @@ dsym:
 	@echo Packaging the debug symbols...
 	@(cd "$(dir $(DSYM_DIR))" ; zip -r $(notdir $(DSYM_PATH)) $(notdir $(DSYM_DIR)))
 
+.PHONY: test-scripts
+# Shell-level tests for the guest provisioning scripts, which are shipped as
+# resources rather than compiled and so are not covered by `swift test`.
+test-scripts:
+	@echo Running script tests...
+	@Tests/ScriptTests/TestCreateMachineUser.sh
+
 .PHONY: test
-test: build-tests
+test: build-tests test-scripts
 	@$(SWIFT) test --skip-build -c $(BUILD_CONFIGURATION) $(SWIFT_CONFIGURATION) --skip TestCLI --skip IntegrationTests
 
 .PHONY: install-kernel
--- a/Package.swift
+++ b/Package.swift
@@ -161,8 +161,10 @@ let package = Package(
         .testTarget(
             name: "ContainerCommandsTests",
             dependencies: [
+                .product(name: "ContainerizationOCI", package: "containerization"),
                 "ContainerCommands",
                 "ContainerResource",
+                "MachineAPIClient",
             ]
         ),
         .testTarget(
--- a/Sources/ContainerCommands/Machine/MachineRun.swift
+++ b/Sources/ContainerCommands/Machine/MachineRun.swift
@@ -99,7 +99,7 @@ extension Application {
 
             // Build environment with HOME set correctly
             let envVars = try Parser.allEnv(
-                imageEnvs: ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],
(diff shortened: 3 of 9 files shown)


stephenlclarke:
There is one duplicate-UID case behind this change. An image may already use the requested numeric UID under another account name. The current setup script then skips user creation, leaving the new `.raw` username unresolved.

I’ve put a follow-up on `upstream/named-uid-aliases`, based directly on this PR’s current head:
https://github.com/stephenlclarke/container/commit/37656b9714a73324dd5616de07675c2689f52177

It creates the requested username as an alias for the existing UID, preserves that name when running the machine, and selects the alias’s shell rather than whichever username happens to appear first for the UID.

The focused Swift test and isolated user-creation script test both pass.

Please feel free to cherry-pick the commit or take the branch.

rupinajay:
Yup confirmed. Keying the skip on `CONTAINER_UID` left the `.raw` username unresolvable whenever an image already held that UID under another name, and the `init -u` gate keyed on the name alone, so setup re-ran and re-skipped on every boot instead of recovering.

Fixed in 80355bc along the lines you suggested: `create-user.sh` now keys on `CONTAINER_USER` and adds the entry as a second name for the UID when it's already taken, and `init` resolves the shell by `CONTAINER_USER` rather than `id -un` and re-runs setup unless the name resolves to the expected UID. Existing machines recover on their next boot.

One deliberate difference: `machine run` passes only `CONTAINER_USER` through, rather than the whole `processEnvironment`, to keep `CONTAINER_MACHINE_ID`/`UID`/`GID`/`HOME` out of every command's environment.

Tests landed at the same path as yours, covering the alias case, idempotency across boots, the conflicting-name failure, and group reuse, plus unit tests for the environment selection. Your script test passes unmodified against it.

Thanks for the branch, made this easy to verify. @stephenlclarke

PR #2126, opened 2026-08-15
fix(os): close directory watcher descriptors after failed startup

### LINKED ISSUE #2034: [Bug]: Directory watcher leaks descriptors when initial handlers fail
### I have done the following

- [x] I have reproduced the issue using the `main` branch of this project

### Steps to reproduce

During testing of my Container Compose plugin, I found a deterministic descriptor leak on stock `apple/container` `main` at `48145ac7fb177d9fb14e015a0bdea4c642b36729`.

1. Create a temporary directory and a `DirectoryWatcher` for that path.
2. Invoke the watcher's initial-start path with a handler that throws.
3. Repeat the failing start 32 times, matching the public watcher's retry behaviour without waiting 32 seconds.
4. Enumerate `/dev/fd`, resolve each descriptor with `fcntl(F_GETPATH)`, and count only descriptors whose path is the watched directory.
5. Observe exactly 32 leaked descriptors on stock `main`.

The public `startWatching` loop retries once per second while no dispatch source exists, so a persistently failing initial handler leaks one descriptor per retry.

The same audit found that `_startWatching` treats descriptor `0` as an open failure even though `open` succeeds for every non-negative descriptor. If standard input is closed and `open` returns `0`, the watcher rejects and leaks that valid descriptor.

### Problem description

`DirectoryWatcher._startWatching` opens the watched directory before its initial enumeration and handler call. The dispatch source that normally owns and closes the descriptor is created only after those operations succeed. If either operation throws, the function returns without transferring ownership or closing the descriptor.

The failure path should close the descriptor before throwing. Descriptor `0` should be accepted as a successful `open` result. A regression test should count descriptors for the exact watched path rather than relying on a process-wide threshold, which can pass despite a leak or fail because unrelated tests opened files.

This is the focused `DirectoryWatcher` leak described by #1773. It is independently reproducible, but it does not by itself establish the VM virtiofs behaviour reported in #1097, so the proposed fix should not close #1097.

### Environment

- OS: macOS 26.5.1 (25F80)
- Xcode: 26.6 (17F113)
- Swift: 6.3.3
- Container: source `main` at `48145ac7fb177d9fb14e015a0bdea4c642b36729`

### Code of Conduct

stephenlclarke:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #2034.

The initial watch path opens the directory before enumerating it and invoking the handler, but only transfers descriptor ownership to a dispatch source afterward. Close the descriptor when either initial operation fails, and accept descriptor 0 as a successful open result.

This supersedes the earlier closed #2036 with the same focused fix rebased onto current `main`.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Negative control on current `main`: 32 failing starts leaked 32 descriptors for the exact watched path.

- DirectoryWatcher focused suite: 5/5 passed
- Leak regression repeated 10 times: passed
- Full non-integration suite: 757 tests passed
- Commit signature: verified

### DIFF
--- a/Sources/ContainerOS/DirectoryWatcher.swift
+++ b/Sources/ContainerOS/DirectoryWatcher.swift
@@ -105,11 +105,11 @@ public actor DirectoryWatcher {
         }
     }
 
-    private func _startWatching(
-        handler: @escaping ([FilePath]) throws -> Void
+    func _startWatching(
+        handler: @Sendable @escaping ([FilePath]) throws -> Void
     ) throws {
         let descriptor = open(directoryPath.string, O_EVTONLY)
-        guard descriptor > 0 else {
+        guard descriptor >= 0 else {
             throw ContainerizationError(.internalError, message: "cannot open \(directoryPath.string), descriptor=\(descriptor)")
         }
 
@@ -117,6 +117,7 @@ public actor DirectoryWatcher {
             let files = try FileManager.default.contentsOfDirectory(atPath: directoryPath.string)
             try handler(files.map { directoryPath.appending($0) })
         } catch {
+            close(descriptor)
             throw ContainerizationError(.internalError, message: "failed to run handler for \(directoryPath.string)")
         }
 
--- a/Tests/ContainerOSTests/DirectoryWatcherTest.swift
+++ b/Tests/ContainerOSTests/DirectoryWatcherTest.swift
@@ -14,12 +14,15 @@
 // limitations under the License.
 //===----------------------------------------------------------------------===//
 
-import ContainerOS
 import ContainerizationError
 import DNSServer
+import Darwin
 import Foundation
+import Synchronization
 import SystemPackage
 import Testing
+
+@testable import ContainerOS
 
 struct DirectoryWatcherTest {
     let testUUID = UUID().uuidString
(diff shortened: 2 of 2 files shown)


stephenlclarke:
Refresh 2026-08-26: rebased the single signed fix onto current main (d96dfa31). New head: 2ce11c93. Focused DirectoryWatcherTest passed 5/5, and git diff --check passed.

stephenlclarke:
Rebased cleanly onto current Apple `main` (`d925dab865cf`). The updated head is signed and remains one commit over the base: `c8382b46a3d18731e578bcc700e5264737e25dad`.

Exact-head focused verification: `swift test --jobs 4 --filter DirectoryWatcherTest` — 5 focused tests passed; `git diff --check` passed and the worktree is clean. The broader test evidence already recorded in this PR remains unchanged; this update only incorporates the newer Apple base.

@codex review

stephenlclarke:
Refresh 2026-09-05: rebased the single signed fix onto current Apple `main` (`eee7ad097079`). The new head is `6540619ba679e14fbf00a5d99638bb363f26c72f`; `swift test --jobs 4 --filter DirectoryWatcherTest` passed 5/5 and `git diff --check` passed. The PR is mergeable; fork-originated CI still requires Apple approval.

PR #2127, opened 2026-08-15
fix(persistence): preserve unloadable container bundles

### LINKED ISSUE #2033: [Bug]: Failed startup loads can delete bundles and publish unusable state
### I have done the following

- [x] I have reproduced the issue using the `main` branch of this project

### Steps to reproduce

During testing of my Container Compose plugin, I found two related startup persistence failures on stock `apple/container` `main` at `48145ac7fb177d9fb14e015a0bdea4c642b36729`.

#### Unloadable metadata

1. Create a directory under the container bundle root.
2. Put malformed JSON in both `config.json` and the legacy fallback `runtime-configuration.json`.
3. Call `ContainersService.loadAtBoot`.
4. Observe that the entire bundle directory is deleted by the catch block.

#### Missing runtime plugin

1. Create a valid container bundle whose configuration names a runtime plugin that is not currently registered.
2. Call `ContainersService.loadAtBoot`.
3. Observe that a stopped `ContainerState` is inserted before runtime availability is checked.
4. The subsequent error is caught and the bundle is deleted, but the unusable in-memory state remains in the returned dictionary.

The same bundle writer persists JSON directly to `config.json`, `options.json`, `kernel.json`, and `rootfs.json`. A process interruption during one of these writes can therefore leave truncated metadata that triggers the destructive startup path.

### Problem description

`ContainersService.loadAtBoot` deletes a user's complete on-disk container bundle for every load error, including transient runtime-plugin availability and recoverable metadata errors. In the missing-runtime case, it also returns an in-memory state for the bundle it just deleted because state insertion happens before validation.

Startup should preserve unloadable bundles for diagnosis or a later successful load, reject them from the returned in-memory state, and log the failure. Bundle metadata replacements should be atomic so an interrupted write cannot expose partially written JSON.

### Environment

- OS: macOS 26.5.1 (25F80)
- Xcode: 26.6 (17F113)
- Swift: 6.3.3
- Container: source `main` at `48145ac7fb177d9fb14e015a0bdea4c642b36729`

### Code of Conduct

stephenlclarke:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #2033.

Preserve malformed or temporarily unloadable container bundles instead of deleting user state during startup. Validate runtime availability before publishing in-memory state, and atomically replace bundle metadata to avoid exposing interrupted JSON writes.

This supersedes the earlier closed #2035 with strengthened malformed-current-and-legacy configuration coverage, rebased onto current `main`.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Negative control on current `main`: malformed bundles were deleted and missing-runtime state was returned before its bundle was deleted.

- Focused startup persistence regressions: 2/2 passed
- Full non-integration suite: 758 tests passed
- Commit signature: verified

### DIFF
--- a/Sources/ContainerResource/Container/Bundle.swift
+++ b/Sources/ContainerResource/Container/Bundle.swift
@@ -137,7 +137,7 @@ extension Bundle {
 
     public func setContainerRootFs(fs: Filesystem) throws {
         let fsData = try JSONEncoder().encode(fs)
-        try fsData.write(to: self.containerRootfsConfig)
+        try fsData.write(to: self.containerRootfsConfig, options: .atomic)
     }
 
     public func cloneContainerRootFs(cloning fs: Filesystem, readonly: Bool = false) throws {
@@ -160,7 +160,7 @@ extension Bundle {
 
     private static func write(_ path: URL, value: Encodable) throws {
         let data = try JSONEncoder().encode(value)
-        try data.write(to: path)
+        try data.write(to: path, options: .atomic)
     }
 
     public func load<T>(filename: String) throws -> T where T: Decodable {
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -132,6 +132,12 @@ public actor ContainersService {
                     continue
                 }
 
+                guard runtimePlugins.first(where: { $0.name == config.runtimeHandler }) != nil else {
+                    throw ContainerizationError(
+                        .internalError,
+                        message: "failed to find runtime plugin \(config.runtimeHandler)"
+                    )
+                }
                 let state = ContainerState(
                     snapshot: .init(
                         configuration: config,
@@ -141,16 +147,9 @@ public actor ContainersService {
                     ),
                 )
                 results[config.id] = state
-                guard runtimePlugins.first(where: { $0.name == config.runtimeHandler }) != nil else {
-                    throw ContainerizationError(
-                        .internalError,
-                        message: "failed to find runtime plugin \(config.runtimeHandler)"
-                    )
-                }
             } catch {
-                try? FileManager.default.removeItem(at: dir)
                 log.warning(
-                    "failed to load container",
+                    "failed to load container; leaving bundle on disk",
                     metadata: [
                         "path": "\(dir.path)",
                         "error": "\(error)",
--- a/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
+++ b/Sources/Services/ContainerAPIService/Server/Containers/ContainersService.swift
@@ -148,9 +148,8 @@ public actor ContainersService {
                     )
                 }
             } catch {
-                try? FileManager.default.removeItem(at: dir)
                 log.warning(
-                    "failed to load container",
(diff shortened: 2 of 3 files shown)


stephenlclarke:
Refresh 2026-08-26: rebased the single signed fix onto current main (d96dfa31). New head: c131869b. Focused ContainerLoadAtBootTests passed 2/2, and git diff --check passed.

MayCXC:
I reached the same defect from the other direction (a plane spawned without an install root resolves no runtime plugins at all, so every container on disk was removed at boot) and wrote a narrower fix before finding this one. Yours covers strictly more: the malformed-metadata half, the atomic bundle replacement, and the tests. I am not filing a competing PR.

One behavioural point worth deciding deliberately rather than by accident, since moving the guard above the `results[config.id] = state` insertion settles it:

A container whose runtime plugin is not registered in this particular spawn is now preserved on disk and absent from the returned state, so it does not appear in `container list`. The bundle is intact and the container becomes visible again the moment a plane that can resolve that plugin starts, but in the meantime `container list` reports a name that `container create` will then refuse as taken, and there is nothing in the listing to explain why.

Contrast the malformed-metadata case, where absence is the only honest answer: nothing can be said about a bundle whose configuration will not parse. Here the configuration parsed, so the id, image, and runtime handler are all known; the only thing missing is a plugin this process cannot see.

Keeping such a container in the returned state as stopped, with the warning you already log, would let the listing account for the name and would not need the caller to guess. Either answer is defensible and this is the maintainers' call, but the two cases differ in what is actually knowable, so it may be worth not treating them identically.

stephenlclarke:
Rebased cleanly onto current Apple `main` (`d925dab865cf`). The updated head is signed and remains one commit over the base: `4fdd7e56f866a9c54c1b641882f056d5f84cf083`.

Exact-head focused verification: `swift test --jobs 4 --filter ContainerLoadAtBootTests` — 2 focused tests passed; `git diff --check` passed and the worktree is clean. The broader test evidence already recorded in this PR remains unchanged; this update only incorporates the newer Apple base.

@codex review

stephenlclarke:
Thanks @MayCXC — I agree that an unavailable runtime differs from malformed metadata because the configuration is still knowable. The updated branch now preserves the bundle **and** retains the container in the returned state as stopped, so it remains visible and its name is accounted for.

That exposed a related safety issue in `bootstrap`: the runtime plugin lookup was force-unwrapped. The same commit now returns a controlled `ContainerizationError` if the plugin is still unavailable. `ContainerLoadAtBootTests` covers persistence, stopped-state visibility, and the safe bootstrap failure; the exact focused suite passes 2/2 on signed head `7b379fa6ff333ccee274ddf9d7fd4050c6148c05`, with `git diff --check` clean.

@codex review

stephenlclarke:
Refresh 2026-09-05: rebased the single signed fix onto current Apple `main` (`eee7ad097079`), including the new clean-command and ContainerService changes. The new head is `c8b4d767a77fd70e01b11129199222661a0f92c0`; `swift test --jobs 4 --filter ContainerLoadAtBootTests` passed 2/2 and `git diff --check` passed. The PR is mergeable; fork-originated CI still requires Apple approval.

PR #2129, opened 2026-08-15
fix(dns): avoid global resolver search domains

### LINKED ISSUE #1917: [Bug]: leftover /etc/resolver/containerization.* 'search' directive hangs ALL bare-hostname DNS lookups system-wide once apiserver stops
### I have done the following

### Steps to reproduce

1. Follow the container DNS tutorial, e.g. `sudo container system dns create test`, then run a container and resolve it via `<name>.test` (see `/etc/resolver/containerization.test` created as described in #1407).
2. Stop the container system entirely: `container system stop` (or just let the apiserver not be running, e.g. after a reboot before starting it again).
3. Do **not** remove `/etc/resolver/containerization.test` — it is left in place.
4. From the host, resolve *any* bare, unqualified hostname unrelated to containers at all, e.g. one handled by a VPN's split-DNS (in my case a Tailscale MagicDNS short name), or any LAN hostname:
   ```
   time dscacheutil -q host -a name some-other-host
   ```

### Problem description

`/etc/resolver/containerization.test` (created by the DNS tutorial / `container system dns create`) contains:

```
domain test
search test
nameserver 127.0.0.1
port 2053
```

The `search test` directive adds `test` to **macOS's system-wide search-domain list** — the list used to expand *any* unqualified single-label hostname, not just ones ending in `.test`. This is independent of the `containerization.test` filename-vs-`domain` mismatch already reported in #1407.

The practical effect: once the container apiserver is not running (stopped, or simply not yet started after boot), `127.0.0.1:2053` has nothing listening. Every bare hostname lookup on the Mac — SSH to a Tailscale MagicDNS short name, a LAN device, anything with no dot in it — now also gets tried as `<name>.test` against that dead nameserver as part of normal search-domain expansion, and that query hangs (no ICMP unreachable, no fast refusal) rather than failing quickly. Observed delay: consistently ~10 seconds per bare-hostname lookup, which is enough to make unrelated tools (ssh, ping, etc.) appear to hang system-wide.

This makes the resolver file dangerous to leave behind: the "harmless" tutorial artifact silently degrades unrelated DNS resolution for the entire machine as soon as the container apiserver stops, until the file is manually deleted (`sudo rm /etc/resolver/containerization.test`) and there's no message/warning to that effect.

### Expected behavior

Either:
- `container system dns create <domain>` should not use a global `search` directive at all — `/etc/resolver` files should only need `domain`/`nameserver`/`port` to scope matching by filename per `resolver(5)`, without contributing to the system-wide search list, or
- The resolver file (and its `search` contamination) should be automatically removed when `container system stop` runs / the apiserver is not registered with launchd, or
- At minimum, `container system status` / `container system stop` should warn if a stale `/etc/resolver/*` file exists pointing at a now-unreachable local port, since its side effects extend well beyond container networking.

### Environment

```markdown
- OS: macOS 26.5.2
- Container: 1.1.0
```

### Code of Conduct

stephenlclarke:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #1917.

Scoped files under `/etc/resolver` only need the domain, nameserver, and port. Removing the `search` directive prevents unrelated single-label hostnames from being expanded through a stopped container DNS server.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Negative control on current `main`: the resolver fixture contained the global search directive.

- HostDNSResolver focused suite: 5/5 passed
- Full non-integration suite: 756 tests passed
- Commit signature: verified

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/HostDNSResolver.swift
+++ b/Sources/Services/ContainerAPIService/Client/HostDNSResolver.swift
@@ -66,7 +66,6 @@ public struct HostDNSResolver {
             } ?? ""
         let resolverText = """
             domain \(name)
-            search \(name)
             nameserver 127.0.0.1
             port \(dnsPort)
             \(options)
--- a/Tests/ContainerAPIClientTests/HostDNSResolverTest.swift
+++ b/Tests/ContainerAPIClientTests/HostDNSResolverTest.swift
@@ -42,13 +42,13 @@ struct HostDNSResolverTest {
         let actualText = try String(contentsOfFile: resolverConfigPath.string, encoding: .utf8)
         let expectedText = """
             domain foo.bar
-            search foo.bar
             nameserver 127.0.0.1
             port 2053
 
             """
 
         #expect(actualText == expectedText)
+        #expect(!actualText.split(separator: "\n").contains { $0.hasPrefix("search ") })
 
         try resolver.createDomain(name: try! DNSName("bar.foo"))
         let domains = resolver.listDomains()


stephenlclarke:
Refresh 2026-08-26: rebased the single signed fix onto current main (d96dfa31). New head: 6404f2a1. Focused HostDNSResolverTest passed 5/5, and git diff --check passed.

PR #2138, opened 2026-08-18
Fix tmpfs mount source field left empty for --mount type=tmpfs

### LINKED ISSUE #2109: [Bug]: --mount type=tmpfs creates a mount with an empty source field, breaking df and mount
### I have done the following

### Steps to reproduce

Environment: container CLI 1.2.2.0 (commit 0190097), macOS 26. 

## Reproduction

Mount a tmpfs with --mount and ask df about it:
```
% container run --rm --mount type=tmpfs,target=/tmpfsmount1,size=512M alpine df -h /tmpfsmount1
Filesystem                Size      Used Available Use% Mounted on
df: tmpfs: No such file or directory
```

### Problem description

 The mount is also absent from a plain df -h with no path argument. Doing the same thing with the --tmpfs flag works as expected:
```
% container run --rm --tmpfs /tmpfsmount1 alpine df -h
...
tmpfs                   556.1M         0    556.1M   0% /tmpfsmount1
```

The inconsistency is visible directly in /proc/mounts. Note the leading space on the first line — field 1, the device/source, is empty:

```
% container run --rm --mount type=tmpfs,target=/tmpfsmount1,size=512M alpine grep tmpfsmount1 /proc/mounts
 /tmpfsmount1 tmpfs rw,relatime,size=524288k 0 0

% container run --rm --tmpfs /tmpfsmount1 alpine grep tmpfsmount1 /proc/mounts
tmpfs /tmpfsmount1 tmpfs rw,relatime 0 0
```

Any tool that parses /proc/mounts positionally shifts every field left by one, so mount also misreports the entry, printing the target as the source and the options as the filesystem type:

```
% container run --rm --mount type=tmpfs,target=/tmpfsmount1,size=512M alpine mount | grep tmpfsmount1
/tmpfsmount1 on tmpfs type rw,relatime,size=524288k (0)
```

The mount itself is correct — size=524288k is the requested 512 MiB, and it's writable. Only its representation in the mount table is wrong. stat -f /tmpfsmount1 reports the right values, since it doesn't go through /proc/mounts.

## Root cause

Two code paths construct tmpfs mounts, and only one of them sets a source. 

Filesystem.init() defaults source to the empty string (Sources/ContainerResource/Container/Filesystem.swift:74), while the Filesystem.tmpfs(destination:options:) factory sets source: "tmpfs" (Filesystem.swift:128). The --tmpfs flag goes through the factory, which is why it's correct. 

Parser.mount instead starts from var fs = Filesystem() and, in its type=tmpfs branch, assigns type, destination, and options but never source (Sources/Services/ContainerAPIService/Client/Parser.swift:419), so the empty default survives into the guest.

Tests/ContainerAPIClientTests/ParserTest.swift already has tmpfs parsing cases from #2103 that a regression test asserting the source field could hang off.

### Environment

```markdown
- OS: 26
- Container: 1.2.2
```

### Code of Conduct

sivasath16:
> [!IMPORTANT]
> All commits must be signed and verified. Pull requests containing unsigned or unverified commits cannot be built or merged. See the [GitHub documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#about-commit-signature-verification) for instructions.
>
> For all but trivial fixes, make sure to first create a GitHub issue that concisely describes the bug or desired enhancement as justification for the change. Large PRs with no justifying issue will be closed.

## Type of Change
- [X] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #2109.

`Filesystem.source` is left empty when a tmpfs mount is created via `--mount type=tmpfs,...`, which corrupts its entry in
`/proc/mounts` and breaks `df` (`df: tmpfs: No such file or directory`) and `mount`, both of which parse that line positionally.

`Parser.mount()` sets `Filesystem.type` for `type=tmpfs` but never sets `Filesystem.source`. The `--tmpfs` flag doesn't have this problem because it builds its `Filesystem` via `Filesystem.tmpfs()`, which already hardcodes `source: "tmpfs"`.

This sets `source` the same way in `Parser.mount()`, so both paths agree.

## Testing
- [X] Tested locally
- [X] Added/updated tests
- [ ] Added/updated docs

### Before / after
Added `testRunCommandMountTmpfs`, which runs a real container with `--mount type=tmpfs,...` and checks `df` and `/proc/mounts`.

**Before the fix:**
image
✘ Test testRunCommandMountTmpfs() failed after 2.168 seconds with 1 issue.

**After the fix:**
image
✔ Test testRunCommandMountTmpfs() passed after 2.124 seconds.

Also ran the full targeted suites on the final diff:
- `swift test --filter ParserTest` — 138/138 passed
- `CONTAINER_CLI_PATH=$(pwd)/bin/container swift test --filter TestCLIRunCommand` — 33/33 passed

### DIFF
--- a/Sources/Services/ContainerAPIService/Client/Parser.swift
+++ b/Sources/Services/ContainerAPIService/Client/Parser.swift
@@ -433,6 +433,7 @@ public struct Parser {
                     fs.type = Filesystem.FSType.virtiofs
                 case "tmpfs":
                     fs.type = Filesystem.FSType.tmpfs
+                    fs.source = "tmpfs"
                 case "volume":
                     isVolume = true
                 default:
--- a/Tests/ContainerAPIClientTests/ParserTest.swift
+++ b/Tests/ContainerAPIClientTests/ParserTest.swift
@@ -459,6 +459,40 @@ struct ParserTest {
     }
 
     @Test
+    func testMountTmpfsWithMode() throws {
+        let result = try Parser.mount("type=tmpfs,dst=/foo,size=64m,mode=1777")
+
+        switch result {
+        case .filesystem(let fs):
+            #expect(fs.destination == "/foo")
+            #expect(fs.options.contains("mode=1777"))
+        case .volume:
+            #expect(Bool(false), "Expected filesystem mount, got volume")
+        }
+    }
+
+    @Test
+    func testMountTmpfsSource() throws {
+        let result = try Parser.mount("type=tmpfs,target=/tmpfsmount1,size=512M")
+        switch result {
+        case .filesystem(let fs):
+            #expect(fs.type == .tmpfs)
+            #expect(fs.source == "tmpfs")
+            #expect(fs.destination == "/tmpfsmount1")
+            #expect(fs.options.contains("size=536870912"))
+        case .volume:
+            #expect(Bool(false), "Expected filesystem mount, got volume")
+        }
+    }
+
+    @Test
+    func testMountTmpfsSourceRejection() throws {
+        #expect(throws: ContainerizationError.self) {
+            _ = try Parser.mount("type=tmpfs,source=tmpfs,target=/tmpfsmount1")
+        }
+    }
+
+    @Test
     func testMountVolumeValidName() throws {
         let result = try Parser.mount("type=volume,src=myvolume,dst=/data")
 
--- a/Tests/IntegrationTests/Run/TestCLIRunCommand.swift
+++ b/Tests/IntegrationTests/Run/TestCLIRunCommand.swift
@@ -295,6 +295,32 @@ struct TestCLIRunCommand {
         }
     }
 
+    @Test func testRunCommandMountTmpfs() async throws {
+        try await ContainerFixture.with { f in
+            let image = alpine.rawValue
+            let c = "\(f.testID)-c"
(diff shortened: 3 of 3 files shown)


sivasath16:
Hi @egernst, this is ready for review. I added the fix for #2109 along with parser regression coverage and an integration test that reproduces the `df` failure before the fix and passes after it.

I also included the before/after results in the PR description as requested.

Thanks!

rupinajay:
LGTM

sivasath16:
Hi @egernst ! When you get a chance, could you please review this? Thank you!

egernst:
@sivasath16, please sign your commits. See:
  - https://github.com/apple/containerization/blob/main/CONTRIBUTING.md#pull-requests
  - https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

sivasath16:
Hi @egernst, Signed the commit. Thanks!

sivasath16:
Hello @egernst, I’ve updated the commit with the required signature. Whenever you have a moment, I’d appreciate it if you could review the changes again. Thanks for your time!

egernst (approved):

PR #2146, opened 2026-08-21
feat(stats): report cpuPercent in the machine-readable formats

### LINKED ISSUE #2144: [Request]: container stats --format json omits the CPU percentage it already computes
`container stats --format json` waits two seconds to collect a CPU delta, then discards it. The table output computes and prints a percentage from the same data; the JSON consumer cannot.

## What happens today

```
$ container stats --no-stream
Container ID  Cpu %    Memory Usage            Net Rx/Tx             Block I/O            Pids
cs-spin       103.23%  2.69 MiB / 256.00 MiB   27.84 KiB / 0.59 KiB  2.51 MiB / 0.00 KiB  1
arango        0.63%    5.37 GiB / 6.00 GiB     74.48 MiB / 9.88 GiB  96.52 MiB / 4.00 KiB 61
```

```
$ container stats --no-stream --format json
{
  "id": "cs-spin",
  "cpuUsageUsec": 5734022776,
  "memoryUsageBytes": 2822144,
  "memoryLimitBytes": 268435456,
  "networkRxBytes": 28508,
  "networkTxBytes": 602,
  "blockReadBytes": 2633728,
  "blockWriteBytes": 0,
  "numProcesses": 1
}
```

`cpuUsageUsec` is a cumulative counter. A single sample of it cannot yield a percentage, so the most useful column of the table is absent from the machine-readable format.

## The cost is already paid

`collectStats` takes two samples unconditionally, sleeping two seconds between them, and the static path is what `--format json` uses:

- `Sources/ContainerCommands/Container/ContainerStats.swift:46` — `if format != .table || noStream { try await runStatic() }`
- `:162-196` — `collectStats` gathers `stats1`, `try await Task.sleep(for: .seconds(2))`, then `stats2`
- `:206-217` — `calculateCPUPercent(cpuUsage1:cpuUsage2:timeInterval:)` turns the pair into the percentage the table shows
- `:105` — `Output.render(payload: statsData.map { $0.stats2 }, format: format)` emits **only the second sample**

So a JSON consumer waits the same two seconds, and then receives neither the computed percentage nor the first sample it would need to compute one itself. To get a CPU percentage from the CLI today you must invoke `container stats` twice and diff `cpuUsageUsec` yourself — four seconds of sleeping for a number the tool already had after two.

## Suggestion

Emit the percentage in the non-table formats. The value already exists at the render site, so the smallest version is a field on the rendered payload:

```
"cpuPercent": 103.23
```

Two details worth your opinion:

1. **Naming and units.** `calculateCPUPercent` documents "100% = one fully utilized core", matching `top` and `docker stats`. Worth stating that in the field's doc comment so consumers do not divide by the container's CPU allocation — I made exactly that mistake against this API before checking `container stats` as ground truth.
2. **Where it belongs.** `ContainerResource.ContainerStats` is documented as "Statistics for a container suitable for CLI display" and models a single sample, so a derived rate may not belong on it. A small wrapper for the rendered payload, or emitting both samples, may be cleaner than widening the model. Happy to follow whichever you prefer.

Related but separate: `container list` reports the CPU allocation (`CPUS`) while `stats` reports `memoryLimitBytes` but no CPU equivalent. Not needed for the percentage above, but it means a consumer wanting "percent of what this container was granted" still needs a second command.

I am happy to send a PR. Environment: macOS 27.0 (Tahoe), Apple silicon, `container` 1.2.2 from the Homebrew formula.

guanchzhou:
Fixes #2144

`container stats` already samples twice, sleeping two seconds between samples, and computes a CPU percentage from the pair for its table. `Output.render` received only `stats2`, so `json`, `yaml` and `toml` consumers paid that latency and got back neither the percentage nor the earlier sample needed to derive it. Obtaining a CPU percentage from the CLI meant invoking the command twice and diffing `cpuUsageUsec` by hand — four seconds of sleeping for a number the tool already had after two.

## Change

A `StatsReport` payload for the non-table formats: the latest sample plus the derived percentage.

`ContainerResource.ContainerStats` is documented as modelling a single sample, and a rate is not a property of one sample, so the derived value lives in the command rather than widening that type. That also keeps the wire shape of the existing fields unchanged — `cpuPercent` is additive.

The two-second interval becomes a single `sampleInterval` constant. `collectStats` sleeps for it and `calculateCPUPercent` divides by it, so the two have to agree; they were separate `.seconds(2)` literals before, one in `collectStats` and one in `statsTable`.

## Verification, end to end with this build

Table (unchanged) and json side by side, same runtime, four containers — one of them a `while :; do :; done` loop allocated a single CPU:

```
$ .build/debug/container stats --no-stream
Container ID  Cpu %    Memory Usage            Pids
cs-spin       101.49%  2.69 MiB / 256.00 MiB   1
cs-mem        0.00%    66.96 MiB / 512.00 MiB  1
arango        0.65%    5.64 GiB / 6.00 GiB     61
cs-idle       0.00%    2.60 MiB / 2.00 GiB     1

$ .build/debug/container stats --no-stream --format json
fields: [blockReadBytes, blockWriteBytes, cpuPercent, cpuUsageUsec, id,
         memoryLimitBytes, memoryUsageBytes, networkRxBytes, networkTxBytes, numProcesses]

cs-spin    cpuPercent = 107.43
cs-mem     cpuPercent = 0
arango     cpuPercent = 0.7456
cs-idle    cpuPercent = 0
```

The percentages agree with the table across independent samplings, and the single-CPU spin loop reads ~100% as expected.

`yaml` and `toml` carry the field too, including through TOML's `[[items]]` wrapper for top-level arrays:

```
- id: cs-spin
  cpuPercent: 1.045919e+2

[[items]]
cpuPercent = 102.27360000000002
```

`swift build --product container` succeeds; `swift test --filter ContainerCommandsTests` → 29 tests in 10 suites passed.

## Two things I left for your call

1. **Units in the name.** `calculateCPUPercent` documents "100% = one fully utilized core", matching `top` and `docker stats`. I noted that on the field, because I made the opposite mistake against this API before checking `container stats` as ground truth — dividing by the container's CPU allocation, which understates a 4-CPU container fourfold. Happy to rename to something like `cpuPercentOfOneCore` if you would prefer the convention in the name.
2. **`stats1` is still dropped.** Emitting both samples would let a consumer compute other rates (network and block I/O per second) the same way. I kept this PR to the one number the table already shows; say the word if the fuller shape is preferable.

Environment: macOS 27.0 (Tahoe), Apple silicon, Swift 6.4.

Per CONTRIBUTING: I used AI assistance while investigating and drafting this, and I can explain and justify every line — the change moves an already-computed value into the rendered payload and de-duplicates one interval literal.

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerStats.swift
+++ b/Sources/ContainerCommands/Container/ContainerStats.swift
@@ -153,58 +153,78 @@ extension Application {
             }
         }
 
-        private struct StatsSnapshot {
+        fileprivate struct StatsSnapshot {
             let container: ContainerSnapshot
             let stats1: ContainerResource.ContainerStats
             let stats2: ContainerResource.ContainerStats
+
+            /// Percent of one core, matching `top` and `docker stats`: 400% is four cores
+            /// saturated. Deliberately not divided by the container's CPU allocation.
+            var cpuPercent: Double? {
+                guard let first = stats1.cpuUsageUsec, let second = stats2.cpuUsageUsec else {
+                    return nil
+                }
+                return ContainerStats.calculateCPUPercent(
+                    cpuUsage1: .microseconds(first),
+                    cpuUsage2: .microseconds(second),
+                    timeInterval: ContainerStats.sampleInterval
+                )
+            }
         }
 
         /// `collectStats` sleeps for this between samples and the percentage divides by it,
         /// so the two must not drift apart.
-        private static let sampleInterval: Duration = .seconds(2)
+        static let sampleInterval: Duration = .seconds(2)
 
         /// The payload for the machine-readable formats: the second sample plus the CPU
         /// percentage derived from both.
         ///
         /// `ContainerResource.ContainerStats` models a single sample, and a rate is not a
         /// property of one sample, so the derived value lives here rather than widening that
-        /// type. Until now only `stats2` was rendered, which discarded both the percentage
-        /// the table computes and the earlier sample a consumer would need to compute it.
-        private struct StatsReport: Encodable {
-            let id: String
-            /// Percent of one core, matching `top` and `docker stats`: 400% is four cores
-            /// saturated. Deliberately not divided by the container's CPU allocation.
+        /// type. The sample is wrapped and flattened on encode so the stored shape stays
+        /// additive (`cpuPercent` only) without copying every sample field onto this type.
+        ///
+        /// Encoding goes through one keyed container. TOMLEncoder replaces `encoder.value`
+        /// on each `container(keyedBy:)` call, so `stats.encode(to:)` followed by a second
+        /// container would drop the sample fields.
+        struct StatsReport: Encodable {
+            let stats: ContainerResource.ContainerStats
             let cpuPercent: Double?
-            let cpuUsageUsec: UInt64?
-            let memoryUsageBytes: UInt64?
-            let memoryLimitBytes: UInt64?
-            let networkRxBytes: UInt64?
-            let networkTxBytes: UInt64?
-            let blockReadBytes: UInt64?
-            let blockWriteBytes: UInt64?
-            let numProcesses: UInt64?
 
-            init(snapshot: StatsSnapshot) {
-                let latest = snapshot.stats2
-                self.id = latest.id
-                self.cpuUsageUsec = latest.cpuUsageUsec
-                self.memoryUsageBytes = latest.memoryUsageBytes
-                self.memoryLimitBytes = latest.memoryLimitBytes
-                self.networkRxBytes = latest.networkRxBytes
-                self.networkTxBytes = latest.networkTxBytes
-                self.blockReadBytes = latest.blockReadBytes
-                self.blockWriteBytes = latest.blockWriteBytes
-                self.numProcesses = latest.numProcesses
+            fileprivate init(snapshot: StatsSnapshot) {
+                self.init(stats: snapshot.stats2, cpuPercent: snapshot.cpuPercent)
+            }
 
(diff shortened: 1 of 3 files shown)


kavix (changes requested):
Thanks for working on this, @guanchzhou! Adding cpuPercent to the machine-readable output is a useful improvement, and consolidating the hardcoded 2-second sleep into sampleInterval also helps prevent sampling drift.

A few small suggestions that could improve maintainability

1. Deduplicate CPU calculation
    Consider moving cpuPercent into a computed property on StatsSnapshot. This would allow both StatsReport and statsTable to share the same calculation logic and avoid duplication.
2. Avoid field duplication in StatsReport
    Rather than manually mirroring all properties from ContainerResource.ContainerStats (and potentially missing fields added in the future), consider wrapping stats2 directly and implementing a custom encode(to:). This would make the implementation more resilient to future changes.
3. Add test coverage
    Please consider adding or updating unit/snapshot tests to verify that --format json includes the expected cpuPercent field.

guanchzhou:
Thanks @kavix — took all three.

1. `cpuPercent` is now a computed property on `StatsSnapshot`. The table and `StatsReport` both read it.

2. `StatsReport` wraps the latest sample instead of copying its fields, and `encode(to:)` flattens through one keyed container. I didn't call `stats.encode(to:)` and then add `cpuPercent` on a second container: that works for JSON and YAML, but TOMLEncoder resets `encoder.value` on each `container(keyedBy:)`, so the sample fields would disappear from toml output.

3. Unit tests cover the formula and the flattened json/yaml/toml shape. The existing `--format json` CLI tests now require `cpuPercent`.

kavix (approved):
LGTM!

PR #2158, opened 2026-08-26
Fix Kubernetes control-plane restart after IP rotation

### LINKED ISSUE #2156: Kubernetes control-plane restart fails after node IP rotation
## Summary

A single-node Kubernetes cluster created by the Apple container K8s plugin
does not restart after its VM receives a different IPv4 address. Address
rotation is expected; the Kubernetes provider does not preserve the kind
bootstrap contract or configure an address-independent in-VM API endpoint.

## Environment

- macOS on Apple Silicon
- apple/container 1.3.0 (`d6de569`)
- `kindest/node:v1.35.5`
- One VM with both control-plane and worker roles
- No workload or custom DNS configuration

## Minimal reproduction

```sh
container k8s create --name k8s-repro
container exec k8s-repro cat /kind/old-ipv4
container stop k8s-repro
container k8s start --name k8s-repro
```

On a clean runtime, the first ordinary restart changed
`192.168.64.2 -> 192.168.64.3` and failed before the K8s plugin could perform
its readiness checks:

```text
INFO: detected IPv4 address: 192.168.64.3
INFO: detected old IPv4 address: 192.168.64.2
error: unable to read config from "/kind/kubeadm.conf":
open /kind/kubeadm.conf: no such file or directory
```

## Root cause

There are two independent restart requirements:

1. The plugin runs kubeadm with `/etc/kubernetes/kubeadm-config.yaml`, while a
   kind node expects its provider to create `/kind/kubeadm.conf` for restart
   reconciliation. The kind-owned file was never created; kubeadm did not
   consume or delete it.
2. Without `controlPlaneEndpoint`, kubeadm persists the VM address in clients
   such as admin.conf and kube-proxy. In an ablation test, adding only the kind
   file let kind repair API-server state, but kube-proxy continued dialing the
   retired address and CoreDNS remained unready.

## Expected behavior

The node should recover after intentional address rotation without requiring a
sticky IP. kind should reconcile address-dependent advertise/certificate
state, while persistent clients use an address-independent endpoint.

## Validated fix

PR #2158 writes and uses `/kind/kubeadm.conf` directly and configures
`controlPlaneEndpoint: 127.0.0.1:6443` for clients inside the current one-VM
cluster. After rotation, kind rewrites the node advertise address to the new
VM address while in-VM clients retain loopback. Host kubeconfig export remains
unchanged.

The fix is intentionally limited to newly created single-node clusters. It
does not add sticky addressing, DNS, a proxy, broad start-time rewriting, or
migration for clusters created by older releases. A future multi-node topology
will need a stable shared DNS, VIP, or load-balancer endpoint.

mxschmitt:
## Summary

Fix restart of newly created Kubernetes nodes after their container VM gets a
new IPv4 address.

Closes #2156.

## Bug before

On a clean 1.3.0 installation, with no workload or custom DNS:

```sh
container k8s create --name k8s-repro
container stop k8s-repro
container k8s start --name k8s-repro
```

The first restart changed `192.168.64.2 -> 192.168.64.3` and failed in the
kind entrypoint:

```text
INFO: detected IPv4 address: 192.168.64.3
INFO: detected old IPv4 address: 192.168.64.2
error: unable to read config from "/kind/kubeadm.conf":
open /kind/kubeadm.conf: no such file or directory
```

There are two independent causes:

1. The plugin ran kubeadm with `/etc/kubernetes/kubeadm-config.yaml`, but a
   kind node expects its provider to create `/kind/kubeadm.conf` for restart
   reconciliation. The file was never created; kubeadm did not consume it.
2. Without `controlPlaneEndpoint`, kubeadm persisted the current vmnet address
   in clients such as admin.conf and kube-proxy. Preserving only the kind file
   let kind repair the API server, but kube-proxy still dialed the retired
   address and CoreDNS stayed unready.

## Fix and behavior afterwards

- Write the generated configuration directly to `/kind/kubeadm.conf` and run
  kubeadm with that same file, matching kind's provider contract.
- Set `controlPlaneEndpoint` to `127.0.0.1:6443` for clients inside the current
  single-node control-plane/worker VM.

On restart, kind rewrites address-dependent advertise and certificate state to
the new VM address. Persistent in-VM clients keep using loopback, so they do not
retain a rotating vmnet address. The node still advertises its real current
address, and host kubeconfig generation remains unchanged.

This intentionally does not add sticky IP allocation, DNS setup, a proxy, or a
start-time reconciliation hook. Existing clusters created without
`/kind/kubeadm.conf` are not migrated by this change.

Loopback is appropriate for the plugin's current one-VM topology. Separate
workers or multiple control planes would require a stable shared DNS, VIP, or
load-balancer endpoint instead.

## Testing

- Built the rebased patch on current `main` with `make container`.
- `swift test --filter K8sBootstrapTests`: 2/2 passed.
- `TestK8sRunSerial.testRestartAfterAddressRotation` passed twice against the
  rebuilt daemon, including `192.168.64.5 -> 192.168.64.7` after the test used
  an intervening container to advance the rotating allocator.
- The regression verifies `/kind/kubeadm.conf`, the loopback endpoint, actual
  address rotation, kind's rewritten `advertiseAddress`, successful
  `k8s start`, and a Ready node. Cleanup removes both test containers.
- A separate restarted-cluster smoke test loaded local Caddy and Alpine images,
  resolved the Service through CoreDNS, and fetched it successfully.
- Strict Swift format lint and `git diff --check` passed.

### DIFF
--- a/Sources/ContainerK8s/Commands/K8sCreate.swift
+++ b/Sources/ContainerK8s/Commands/K8sCreate.swift
@@ -102,6 +102,10 @@ public struct K8sCreate: AsyncParsableCommand {
             try await K8sHelper.prepareNode(nodeID: name, client: client, log: log)
             try await K8sHelper.bootstrapControlPlane(
                 nodeID: name, apiServerSANs: sans, advertiseAddress: vmIP,
+                // All in-VM Kubernetes clients (including host-networked kube-proxy)
+                // can reach the single-node API through loopback. This endpoint is
+                // independent of the container's rotating vmnet address.
+                controlPlaneEndpoint: K8sHelper.nodeLocalControlPlaneEndpoint,
                 schedulable: provisioner.roles.contains(StandardRoles.worker),
                 client: client, log: log)
 
--- a/Sources/ContainerK8s/K8sHelper.swift
+++ b/Sources/ContainerK8s/K8sHelper.swift
@@ -51,6 +51,8 @@ public struct K8sHelper {
     static let proxyEnvVars = ["HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "http_proxy", "https_proxy", "no_proxy"]
 
     public static let clusterContainerPort: UInt16 = 6443
+    /// Endpoint used by clients inside the single-node cluster container.
+    static let nodeLocalControlPlaneEndpoint = "127.0.0.1:\(clusterContainerPort)"
 
     // MARK: - Resource defaults
 
--- a/Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
+++ b/Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
@@ -35,12 +35,15 @@ extension K8sHelper {
 
     static func bootstrapControlPlane(
         nodeID: String, apiServerSANs: [String], advertiseAddress: String,
+        controlPlaneEndpoint: String,
         schedulable: Bool, client: ContainerClient, log: Logger
     ) async throws {
-        let configYAML = initConfigYAML(advertiseAddress: advertiseAddress, certSANs: apiServerSANs)
+        let configYAML = initConfigYAML(
+            advertiseAddress: advertiseAddress, certSANs: apiServerSANs,
+            controlPlaneEndpoint: controlPlaneEndpoint)
         var r = try await execCapture(
             containerId: nodeID, executable: "/bin/sh",
-            arguments: ["-c", "cat > /etc/kubernetes/kubeadm-config.yaml <<'EOF'\n\(configYAML)\nEOF"],
+            arguments: ["-c", "mkdir -p /kind && cat > /kind/kubeadm.conf <<'EOF'\n\(configYAML)\nEOF"],
             client: client)
         guard r.code == 0 else {
             throw ContainerizationError(.internalError, message: "write kubeadm config failed on \(nodeID): \(r.output)")
@@ -50,7 +53,7 @@ extension K8sHelper {
         r = try await execCapture(
             containerId: nodeID, executable: kubeadmPath,
             arguments: [
-                "init", "--config", "/etc/kubernetes/kubeadm-config.yaml",
+                "init", "--config", "/kind/kubeadm.conf",
                 "--ignore-preflight-errors", ignorePreflightErrors,
             ],
             client: client)
@@ -135,7 +138,9 @@ extension K8sHelper {
         """
     }
 
-    private static func initConfigYAML(advertiseAddress: String, certSANs: [String]) -> String {
+    static func initConfigYAML(
+        advertiseAddress: String, certSANs: [String], controlPlaneEndpoint: String
+    ) -> String {
         let sans = certSANs.map { "  - \($0)" }.joined(separator: "\n")
         return """
             apiVersion: kubeadm.k8s.io/v1beta4
@@ -148,6 +153,7 @@ extension K8sHelper {
             ---
             apiVersion: kubeadm.k8s.io/v1beta4
             kind: ClusterConfiguration
+            controlPlaneEndpoint: \(controlPlaneEndpoint)
             kubernetesVersion: \(kubernetesVersion())
             networking:
               podSubnet: \(podSubnet)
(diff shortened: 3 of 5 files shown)


mxschmitt:
## Long-term design question

This PR intentionally fixes the current single-node plugin with the smallest
address-independent bootstrap change: new in-VM Kubernetes clients use
`127.0.0.1:6443`, while kind continues to reconcile the node's changing
advertise address.

I would welcome maintainer guidance on the longer-term abstraction before
multi-control-plane support is added. At that point, loopback would no longer
be sufficient and the plugin would need a stable shared endpoint, for example:

- a container/runtime-provided DNS identity,
- a local API load balancer with a dynamic backend set, or
- a narrowly defined endpoint-reconciliation API.

After a clean-room and ablation retest, I removed the earlier start-time
ConfigMap migration. It did not provide a valid upgrade path for stock legacy
clusters and was not needed by newly created clusters, so keeping it would
have made this fix broader than the demonstrated defect.

mxschmitt:
@jglogan, could you take a look when you have a chance? This fixes a clean
single-node `container k8s` stop/start failure after expected VM IP rotation.
The patch aligns bootstrap with kind's `/kind/kubeadm.conf` contract, uses a
stable in-VM endpoint for the current one-VM topology, and includes the exact
restart regression. Since you reviewed much of the original K8s plugin, your
input would be especially helpful.

guanchzhou:
Independent check of this PR against the kube-proxy half of https://github.com/apple/container/issues/2156, on Apple Silicon, container 1.3.1 daemon, plugin built from this branch (`86735d0`).

Threw away cluster `k8s-2158` (2 CPU / 4 GiB). Did not migrate a pre-existing cluster; that is out of scope here as documented.

Create wrote `/kind/kubeadm.conf` with `controlPlaneEndpoint: 127.0.0.1:6443`. In-cluster kube-proxy ConfigMap was already:

```
server: https://127.0.0.1:6443
```

Stop, run an intervening Alpine container so vmnet could not reuse the address, then `k8s start`:

- node IP `192.168.64.8` → `192.168.64.10`
- kind rewrote `advertiseAddress` to `192.168.64.10`
- `controlPlaneEndpoint` stayed `127.0.0.1:6443`
- kube-proxy ConfigMap still `https://127.0.0.1:6443`
- kube-proxy logs: zero `no route to host`
- `apiserver.crt` still present (this is the crash-loop we hit on an unpatched 1.3.1 restart)
- node Ready; a busybox pod fetched `gitVersion: v1.35.5` from `https://10.96.0.1:443/version`

So for **new** clusters this PR does resolve the silent Service-routing break. It does not, and does not claim to, fix clusters created before `/kind/kubeadm.conf` existed.

Note for anyone reproducing on 1.3.1: this branch still embeds `iptables-nft` in the node prep script (https://github.com/apple/container/issues/2120). The guest kernel here has no nftables, so create needed the same equal-length `iptables-nft` → `iptables` binary patch used on stock 1.3.1. Unrelated to the IP-rotation fix.

saiyam1814:
On the long-term endpoint question in your first comment, some data from kiac, which has run multi-node kubeadm on apple/container since June and heals this whenever an address rotates on restart.

For the one-VM plugin, loopback as `controlPlaneEndpoint` is the right call and this PR looks correct to me for that topology.

For multi-node, two things that have held up in practice:

1. **A name, not an address, when one is available.** When `dns.domain` is configured, the plugin already computes the container FQDN and puts it in the SANs. Using that FQDN as `controlPlaneEndpoint` gives every kubelet and kube-proxy an identity that survives rotation, which is what k3ac relies on with its external etcd (BinHsu re-verified the A record re-registering across a restart on 1.3.1 in #1865). It does depend on container DNS, which has its own sharp edges, so it is an option rather than a silver bullet.

2. **A start-time heal when only addresses exist.** kiac's fallback on `resume` rewrites the address-bearing state in place (kubeconfigs and static pod manifests, a re-issued apiserver serving cert via `kubeadm init phase certs apiserver`, the kube-proxy ConfigMap and cluster-info, the `server:` line in each worker's `kubelet.conf`); the list lives in `pkg/cluster/persist.go` if the provisioner ever grows a multi-node start path. More moving parts than this PR needs today.

Either way, sticky addresses in the runtime (#1740) would make most of this unnecessary, so +1 to that landing first.

PR #2205, opened 2026-08-28
Avoid escaping slashes in JSON output

### LINKED ISSUE #2204: [Request]: Don't escape slashes in container JSON output.

By default, Swift's JSONEncoder escapes forward slashes. This allows JSON to be safely embedded in HTML, but makes let output less legible, and less copy/paste friendly.

`container` commands accepting `--format json` should configure JSONEncoder with `.withoutEscapingSlashes` so that output such as pathnames appear as `"/foo/bar"` and not `"\/foo\/bar".

### Code of Conduct

taljeon:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context

Fixes #2204.

Swift's default `JSONEncoder` output escapes forward slashes, which makes CLI JSON paths and URLs harder to read and copy. This change makes the shared CLI JSON renderer always use `.withoutEscapingSlashes` and routes `machine list --format json` through that renderer while preserving its ISO-8601 date encoding.

OpenAI Codex assisted with investigation, implementation, and verification. The final change is limited to JSON rendering behavior and focused regression coverage.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

- `swift test --filter RenderJSONTests` — 7 tests passed
- `swift test --filter ContainerCommandsTests` — 30 tests passed
- `make test` — 763 tests passed
- strict `swift format lint` passed for all changed files

### DIFF
--- a/Sources/ContainerCommands/Machine/MachineList.swift
+++ b/Sources/ContainerCommands/Machine/MachineList.swift
@@ -61,10 +61,8 @@ extension Application {
                 let printables = machines.map {
                     PrintableMachine($0, isDefault: $0.id == defaultMachine)
                 }
-                let encoder = JSONEncoder()
-                encoder.dateEncodingStrategy = .iso8601
-                let data = try encoder.encode(printables)
-                print(String(decoding: data, as: UTF8.self))
+                let options = JSONOptions(dateEncodingStrategy: .iso8601)
+                try Output.emit(Output.renderJSON(printables, options: options))
                 return
             }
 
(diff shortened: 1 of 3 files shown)


jglogan:
@taljeon I'm going to let this build completely before approving but it looks like it should be fine to merge.

One thing I noticed when I searched our commands is that MachineInspect doesn't offer a format option or use the shared output code. Feel free to create an issue and PR for that if you like and that should be an easy review and merge as well.

Edit: actually it looks like all our `inspect` commands are JSON only so it could be a cross-cutting change to have all of them use the common output code and support `--format` (we could exclude `table` mode for inspect).

jglogan (approved):

jglogan:
@taljeon Thanks for taking care of this! Merged.

PR #2228, opened 2026-09-02
fix: handle container clean failing on read-only named volume mounts

### LINKED ISSUE #2224: [Bug]: `container clean` fails on read-only named volume mounts.
### I have done the following

### Steps to reproduce

In one terminal:

```bash
container volume create scratch
container run -it --rm --name clean-test --cap-add ALL -v scratch:/mnt/data:ro ubuntu:noble
```

In a second terminal:

```bash
container clean clean-test
```

### Problem description

Expected behavior:

- `container clean` succeeds, cleaning the root filesystem, and no change on the read-only named volume filesystem.

Details:

- The runtime helper should check the root and named filesystem mounts and skip cleaning those that are mounted read-only.
- Should a clean fail, the runtime helper should record the target mount path in a list of failed targets, and continue cleaning the remaining targets. On return, if the list is empty, throw an error like `failed to clean mounts in {container-id}: [mount-path, ...]

Actual behavior:

```console
% container clean clean-test
Error: internalError: "failed to clean container" (cause: "internalError: "failed to clean container clean-test (cause: "unknown: "internalError: "filesystemOperation trim failed""")"")
```

### Environment

```markdown
- OS: 26
- Xcode: 26
- Container: main
```

### Code of Conduct

zdk:
## Type of Change
- [x] Bug fix

## Motivation and Context
`container clean` fails with `filesystemOperation trim failed` when the container has a named volume mounted read-only, because the runtime helper tries to trim every block mount and throws on the first error.

The fix:
- skips the rootfs when the container was created with `--read-only`, and skips block mounts whose options include `ro`.
- keeps trimming the remaining targets after a failure, then throws one error listing each failed path with its underlying error.

Fixes #2224

## Testing
- [x] Tested locally
- [x] Added/updated tests

### DIFF
--- a/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
+++ b/Sources/Services/RuntimeLinux/Server/RuntimeService.swift
@@ -860,7 +860,7 @@ public actor RuntimeService {
         self.log.info("`clean` xpc handler")
         switch self.state {
         case .running:
-            guard message.string(key: RuntimeKeys.id.rawValue) != nil else {
+            guard let id = message.string(key: RuntimeKeys.id.rawValue) else {
                 throw ContainerizationError(
                     .invalidArgument,
                     message: "no id supplied for clean"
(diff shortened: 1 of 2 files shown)


jglogan (approved):
@zdk This looks good, thank you!

PR #2130, opened 2026-08-15
fix(k8s): fall back to legacy iptables

stephenlclarke:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Closes #2120.

Probe the nftables backend during node preparation. When the node kernel cannot use it, select the available legacy iptables and ip6tables alternatives before applying the MSS rules. This also keeps subsequent Kubernetes components on the working backend.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs

Negative control with the former hard-coded nft path: the behavior fixture exited unsuccessfully and applied neither fallback nor MSS rules.

- Shell behavior regressions for nft available/unavailable paths: 2/2 passed
- Full non-integration suite: 758 tests passed
- `make fmt`, `make check`, and `git diff --check`: passed
- Commit signature: verified

PR #2244, opened 2026-09-07
docs: Add k8s plugin reference

egernst:
Add a short document that highlights basics of the k8s plugin

> [!IMPORTANT]
> All commits must be signed and verified. Pull requests containing unsigned or unverified commits cannot be built or merged. See the [GitHub documentation](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#about-commit-signature-verification) for instructions.
>
> For all but trivial fixes, make sure to first create a GitHub issue that concisely describes the bug or desired enhancement as justification for the change. Large PRs with no justifying issue will be closed.

## Type of Change
- [ ] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [x] Documentation update

## Motivation and Context

We didn't have a basic k8s doc before. Added, matching similar coverage as container machine doc.

## Testing
- [ ] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -19,5 +19,6 @@
 - [Inspecting containers and images](./container-inspection.md) — machine-readable `inspect` and `list` output for scripting.
 - [Logs](./logs.md) — container output, VM boot logs, and the `container` system's own logs.
 - [`config.toml` reference](./container-system-config.md) — every configuration key, its default, and how to view your merged configuration.
-- Container machines — persistent Linux environments built from OCI images, with your home directory mounted in and the filesystem surviving stop/start. See [container-machine.md](./container-machine.md) for the full guide.
+- [Container machines](./container-machine.md) — persistent Linux environments built from OCI images, with your home directory mounted in and the filesystem surviving stop/start.
+- [Kubernetes clusters](./kubernetes.md) — run local single-node Kubernetes clusters for development and testing, load your own images, and test deployments before production.
 - [Shell completions](./shell-completions.md) — generate and install completion scripts for `zsh`, `bash`, and `fish`.
--- a/docs/kubernetes.md
+++ b/docs/kubernetes.md
@@ -1,0 +1,206 @@
+# Kubernetes clusters
+
+Kubernetes is an open-source orchestration platform for automating deployment, scaling, and management of containerized workloads across clusters of machines. With `container`, you can run local Kubernetes clusters for development and testing.
+
+## Why Kubernetes on container
+
+Running Kubernetes locally with `container` gives you a streamlined workflow to build and test workloads before deploying to production:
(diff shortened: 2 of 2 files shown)


saiyam1814:
Read this against the plugin source and tried the version examples inside the node images. Two things to adjust before merge, one of which is a code bug the doc would otherwise document as a feature.

1. **"Run the same workload on different Kubernetes versions" does not do that today.** `--node-image` is honoured for the image that boots, but `kubernetesVersion()` in `K8sHelper+Bootstrap.swift` parses the tag of the static default `nodeImage`, so every cluster's kubeadm config says `kubernetesVersion: v1.35.5`. Inside `kindest/node:v1.34.11`, `kubeadm config images list --config` on that config returns `kube-apiserver:v1.35.5` and friends, and preflight only warns (`[WARNING KubernetesVersion]: Kubernetes version is greater than kubeadm version`), so a "1.34 cluster" would come up with a 1.35.5 control plane. Filed as #2246; the simplest fix is to omit `kubernetesVersion` and let kubeadm resolve its own version. I would hold this section until that lands, or phrase it as "the node image's version" only once it is true.

2. **Two of the example tags cannot work.** `kindest/node:v1.34.4` does not exist on Docker Hub (the 1.34 tags are v1.34.0, .2, .3, .8, .11). `kindest/node:v1.30.8` exists, but kubeadm 1.30 rejects the plugin's `kubeadm.k8s.io/v1beta4` config (`experimental API spec: "kubeadm.k8s.io/v1beta4" is not allowed`); `v1.31.14` accepts it, so 1.31 is the floor. Suggest `v1.34.11` and `v1.35.5` in the examples plus one sentence on the 1.31 minimum.

Checked and correct as written: default name `k8s-dev`, `kubectl --context <cluster-name>` (the kubeconfig context is renamed to the cluster name), the CPU and memory defaults (a quarter of the host, minimum 2 and 2g), `--rm`, and `load-image --platform`.

Non-blocking for the docs: on the shipped 1.3.1 the Quickstart stops at node prep (#2120, fixed by #2145 or #2130, neither merged yet), so the first command a reader runs fails until one of those ships. For the same reason I could not time `k8s create` myself; if "in seconds" assumes a warm image cache, it may be worth saying so.

egernst:
Great callout @saiyam1814.  I think until those issues you identified are resolved, it's best to skip that example. I have updated accordingly.  Hopefully we can add soon!

katiewasnothere (approved):

OneCricketeer:
I see the code and docs resolve to 1.35 right now, but I was curious if it was documented why 1.36 or 1.37 were not chosen, given this plugin is a new feature?

PR #2254, opened 2026-09-09
K8s plugin: Support custom CNI manifest 

katiewasnothere:
## Type of Change
- [x] New feature  

## Motivation and Context
This PR adds a new flag to the k8s create command to allow users to specify an optional path to a CNI manifest that they wish to use for the cluster instead of the bundled kindnet CNI manifest. 

I tested this with cilium by doing the following: 
```
% helm template cilium cilium/cilium \
    --version 1.20.1 \
    --namespace kube-system \
    --set ipam.mode=cluster-pool \
    --set ipam.operator.clusterPoolIPv4PodCIDRList='{<TARGETCIDR>}' \
    --set kubeProxyReplacement=false > /tmp/cilium.yaml
% container k8s create --cni /tmp/cilium.yaml
```

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

### DIFF
--- a/Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
+++ b/Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
@@ -75,9 +75,7 @@ extension K8sHelper {
 
         log.info("Applying kindnet CNI", metadata: ["node": "\(nodeID)"])
         let manifest = try await loadKindnetManifest(log: log)
-        let apply =
-            "cat > /tmp/kindnet.yaml <<'EOF'\n\(manifest)\nEOF\n"
-            + "\(kubeconfigEnv) kubectl apply -f /tmp/kindnet.yaml"
+        let apply = "\(kubeconfigEnv) kubectl apply -f - <<'EOF'\n\(manifest)\nEOF"
         r = try await execCapture(
             containerId: nodeID, executable: "/bin/sh",
             arguments: ["-c", apply], client: client)


egernst:
- Consider adding the example usage in the PR description? 
- Can you add example to the kubernetes.md?

egernst (approved):
Cool -- looks good Kathryn. Thanks for the updated docs/example

Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
```diff
@@ -73,11 +73,9 @@ extension K8sHelper {
...
         }
 
-        log.info("Applying kindnet CNI", metadata: ["node": "\(nodeID)"])
-        let manifest = try await loadKindnetManifest(log: log)
-        let apply =
-            "cat > /tmp/kindnet.yaml <<'EOF'\n\(manifest)\nEOF\n"
-            + "\(kubeconfigEnv) kubectl apply -f /tmp/kindnet.yaml"
+        log.info("Applying CNI manifest", metadata: ["node": "\(nodeID)"])
+        let manifest = try await loadCNIManifest(path: cniManifestPath, log: log)
+        let apply = "\(kubeconfigEnv) kubectl apply -f - <<'EOF'\n\(manifest)\nEOF"
```
egernst, line 78: Can things get weird here if the user submits a yaml that "accidentally" includes EOF? Should we feed it a real stdin, or base64-encoding the payload, or?

Sources/ContainerK8s/Commands/K8sCreate.swift
```diff
@@ -103,6 +106,7 @@ public struct K8sCreate: AsyncParsableCommand {
             try await K8sHelper.bootstrapControlPlane(
                 nodeID: name, apiServerSANs: sans, advertiseAddress: vmIP,
                 schedulable: provisioner.roles.contains(StandardRoles.worker),
+                cniManifestPath: cni,
```
egernst, line 115: should we check the cni path earlier? If the path doesn't exist, the whole bootstrap will fail, right? Just thinking that a fat finger wouldn't be cushy until after image pull, node preparation, kubeadm init, etc.?

Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
```diff
@@ -103,6 +101,16 @@ extension K8sHelper {
         return (token: parts[tokenIdx + 1], caCertHash: parts[hashIdx + 1])
     }
 
+    static func loadCNIManifest(path: String?, log: Logger) async throws -> String {
+        if let path {
+            guard let contents = try? String(contentsOfFile: path, encoding: .utf8) else {
```
egernst: should we do/catch so we can print out the actual error (could be a read error as well?)

egernst (changes requested):

docs/kubernetes.md
```diff
@@ -150,6 +150,39 @@ By default, clusters use `kindest/node:v1.35.5`, a Kubernetes-in-Docker image op
...
+The manifest must be a plain Kubernetes YAML file (the same shape `kubectl apply -f` expects), not a Helm chart.
+
+### Example: Cilium
+
+Cilium is distributed as a Helm chart, so render a plain manifest from it first:
+
+```bash
+helm repo add cilium https://helm.cilium.io/
+helm repo update
+helm template cilium cilium/cilium --version 1.16.5 --namespace kube-system > cilium.yaml
```
egernst: A friend of mine suggested the following here:
```
The helm template cilium … command uses chart defaults: ipam.mode=cluster-pool with clusterPoolIPv4PodCIDRList=10.0.0.0/8, which overlaps both the service CIDR 10.96.0.0/12 and kubeadm's podSubnet 10.244.0.0/16 (K8sHelper.swift:46-48). Cilium requires the pool not overlap the service CIDR. Those pod IPs also fall outside the NO_PROXY list built in K8sHelper+Networking.swift:33. Add --set ipam.mode=kubernetes to the documented command.
```

Sources/ContainerK8s/Commands/K8sCreate.swift
```diff
@@ -51,6 +51,9 @@ public struct K8sCreate: AsyncParsableCommand {
     @Option(help: "Node image reference (default: \(K8sHelper.nodeImage))")
     var nodeImage: String = K8sHelper.nodeImage
 
+    @Option(name: .long, help: "Optional path to a CNI manifest to apply. If no value is provided, the bundled kindnet CNI is used.")
```
egernst: The flag itself is optional. I would drop the "if no value is provided" since technically since "just --cni" would provide a parse error?

As defined here, it doesn't match what's in command-reference.md. I think the .md has it right?


```suggestion
    @Option(name: .long, help: "Optional path to a CNI manifest to apply. If not provided, the bundled kindnet CNI is used.")
```

Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
egernst: Also, I think a particularly large manifest may hit E2BIG - I think this would also be solved if we considered a pipe?

Tests/K8sPluginTests/K8sCreateCNITests.swift
```diff
@@ -0,0 +1,79 @@
...
+    @Test func missingPathThrowsInvalidArgument() async throws {
+        let missingPath = FileManager.default.temporaryDirectory
+            .appendingPathComponent(UUID().uuidString + "-does-not-exist.yaml").path
+
+        await #expect(throws: ContainerizationError.self) {
+            _ = try await K8sHelper.loadCNIManifest(path: missingPath, log: log)
+        }
+    }
+
+    @Test func customPathTakesPrecedenceOverBundledCNI() async throws {
```
egernst: I think this is effectively covering the same as customPathReturnsItsContents(). 

A local friend of mine had this suggestion:
```
Line 76 already asserted result == contents, and contents is the literal "custom-cni-marker". So by the time you reach line 77, result is known to be that string, which self-evidently doesn't contain "kindnet". The assertion is true as a consequence of the previous line — it has no independent power to catch anything.
```

Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
katiewasnothere: I don't know what the right answer for this is from a UX perspective. I'd like to investigate that in a separate PR if that's okay with you.

Sources/ContainerK8s/Commands/K8sCreate.swift
katiewasnothere: Added this closer to the beginning in K8sCreate.

Sources/ContainerK8s/Commands/K8sCreate.swift
katiewasnothere: Dropped in the help output.

Sources/ContainerK8s/Support/K8sHelper+Bootstrap.swift
egernst: Yeah, I think the UX would be the same, just a matter of how we handle it. I'm fine doing this as a follow up. https://github.com/apple/container/issues/2273

egernst (approved):

vincepri (approved):

PR #2267, opened 2026-09-12
Apply --ulimit in container exec and machine run

### LINKED ISSUE #2266: container exec accepts --ulimit but ignores it
`container exec --help` lists `--ulimit <limit>  Set resource limits (format: <type>=<soft>[:<hard>])`, and the flag parses without error, because `ContainerExec` embeds `Flags.Process`. Nothing on the exec path ever reads `processFlags.ulimits` though: `grep -rn '\.ulimits' Sources` hits only `Parser.swift`, which is the run/create path, so exec just inherits the init process rlimits and the value you passed is dropped silently. The guest already honors per-process rlimits on exec (containerization's `vmexec/ExecCommand.swift` calls `App.setRLimits(rlimits: process.rlimits)`), so the plumbing is there and only the CLI side is missing. `container machine run` has the same gap. A PR follows.

rishabhsai:
## Type of Change
- [x] Bug fix
- [ ] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
`container exec` and `container machine run` embed `Flags.Process`, so both print `--ulimit <limit>` in their help and accept the flag without error, but neither ever reads `processFlags.ulimits`; the only reader is `Parser.process` on the run/create path. Exec therefore keeps whatever rlimits the init process was started with and the value you passed is dropped with no warning, even though the guest already applies per-process rlimits on exec via `App.setRLimits(rlimits: process.rlimits)` in containerization's `vmexec/ExecCommand.swift`. #1129 added `--ulimit` to `Flags.Process` and wired it into run and create only, which is where the gap came from. This sets `config.rlimits` on exec when `--ulimit` is given, leaving the init process limits in place when it is not, and passes `rlimits:` into the `ProcessConfiguration` that `machine run` builds.

Fixes #2266

## Testing
- [ ] Tested locally
- [x] Added/updated tests
- [x] Added/updated docs

`swift build --product container` succeeds and `.build/debug/container exec --help | grep ulimit` still lists the flag, and `make swift-fmt` leaves the diff unchanged. The new `testExecUlimitNofile` in `TestCLIExecCommand` is compile-verified only: the integration suite needs an installed CLI and a running container system, which I could not run locally, so every test in that suite fails at fixture setup with `binaryNotFound` on my machine.

### DIFF
--- a/Sources/ContainerCommands/Container/ContainerExec.swift
+++ b/Sources/ContainerCommands/Container/ContainerExec.swift
@@ -77,6 +77,9 @@ extension Application {
                 gid: processFlags.gid, defaultUser: defaultUser)
             config.user = user
             config.supplementalGroups.append(contentsOf: additionalGroups)
+            if !self.processFlags.ulimits.isEmpty {
+                config.rlimits = try Parser.rlimits(self.processFlags.ulimits)
+            }
 
             do {
                 let io = try ProcessIO.create(tty: tty, interactive: stdin, detach: self.detach)
--- a/Sources/ContainerCommands/Machine/MachineRun.swift
+++ b/Sources/ContainerCommands/Machine/MachineRun.swift
@@ -111,7 +111,8 @@ extension Application {
                 workingDirectory: cwd,
                 terminal: tty,
                 user: user,
-                supplementalGroups: additionalGroups
+                supplementalGroups: additionalGroups,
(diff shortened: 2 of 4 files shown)


rishabhsai:
@katiewasnothere thanks for merging #2265. Would you have time to look at this one too? It makes `--ulimit` take effect in `container exec` and `machine run`, as reported in #2266.

PR #2282, opened 2026-09-16
Fix XPC server crash on fire-and-forget messages

### LINKED ISSUE #2281: [Bug]: Force-unwrapped xpc_dictionary_create_reply crashes any XPC server on fire-and-forget input
### I have done the following

### Steps to reproduce

Reproduced on `main` @ `57f0b93` (post-1.4.1).

Open an anonymous XPC connection pair, deliver one dictionary fire-and-forget (no reply expected), and call `reply()` on the received message — this mirrors what every route handler and every `XPCServer.handleMessage` error path does via `message.reply()`:

```swift
let listener = xpc_connection_create(nil, queue)
// … accept peer, capture received dictionary …
let client = xpc_connection_create_from_endpoint(xpc_endpoint_create(listener))
xpc_connection_send_message(client, xpc_dictionary_create_empty()) // fire-and-forget
// server side, on the RECEIVED dictionary:
XPCMessage(object: received).reply() // 💥 traps here
```

Observed result:

```
ContainerXPC/XPCMessage.swift:60: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

(exit 133 / SIGTRAP — the process is gone, not just the request.)

### Problem description

`XPCMessage.reply()` force-unwraps `xpc_dictionary_create_reply(object)`, which returns `NULL` for any message sent via `xpc_connection_send_message` (no reply context). Every error path in `XPCServer.handleMessage` — not-a-dictionary, uid mismatch, empty route, handler errors — funnels through `replyWithError` → `reply()`, and every success path builds its response with `reply()` too (88 call sites across 13 files, covering all `*Harness` types and `RuntimeService`).

So a single one-way dictionary delivered to `com.apple.container.apiserver` — reachable from any same-uid process through the same per-user bootstrap namespace the daemons themselves use — traps the daemon. The same applies to each helper (`container-runtime-linux`, `container-network-vmnet`, `container-core-images`, `machine-apiserver`), which share the `ContainerXPC` code.

What should happen instead: a message with no reply context should be dropped after logging, never trapped on. `reply()` itself should not be able to crash the process. Note this isn't only a hostile-input concern: the repo already sends fire-and-forget messages in normal operation (`ProgressUpdateService` uses `xpc_connection_send_message`), so "no reply context" is a legitimate case the server side must tolerate.

Suggested direction: make `reply()` non-trapping (fall back to a fresh dictionary), add a `canReply` check, and gate `handleMessage`/`replyWithError` on it.

Happy to put up a PR — I have the fix and regression tests (including the live connection-pair repro above) ready against `main`.

### Environment

- OS: macOS 27.0 (26A428)
- Xcode: CommandLineTools Swift 6.4 (no full Xcode installed)
- Container: main @ `57f0b93`

### Code of Conduct

1PoPTRoN:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
Fixes #2281.

`XPCMessage.reply()` force-unwrapped `xpc_dictionary_create_reply()`, which returns `NULL` for any message sent fire-and-forget via `xpc_connection_send_message`. Every path in `XPCServer.handleMessage` — all four error paths through `replyWithError` plus every success path built by the route handlers (88 `reply()` call sites across 13 files) — hits this, so one one-way dictionary from any same-uid process traps `container-apiserver` or any helper (`container-runtime-linux`, `container-network-vmnet`, `container-core-images`, `machine-apiserver`) with `Fatal error: Unexpectedly found nil` at `XPCMessage.swift:60`.

The fix keeps the `reply()` signature (no caller churn) but makes it non-trapping with an empty-dictionary fallback, adds a `canReply` check, and drops no-reply-context input in `handleMessage`/`replyWithError` after logging. Legitimate `_with_reply` traffic takes the identical code path as before.

## Testing
- [x] Tested locally
- [x] Added/updated tests
- [ ] Added/updated docs (no user-facing behavior change for legitimate traffic, so no doc updates)

Verification done on `main` @ `57f0b93`, macOS 27.0, CommandLineTools Swift 6.4:
- Live repro (anonymous XPC pair + one-way send, then `reply()` on the received dict): **traps pre-fix** (exit 133/SIGTRAP at `XPCMessage.swift:60`), **passes post-fix** with a usable fallback dictionary.
- New `ContainerXPCTests` suite (4 tests: no-reply-context detection, no-trap regression, round-trip, live connection-pair repro). Note: this box has CLT only, so the swift-testing suite itself was validated via equivalent standalone harnesses linked against the built modules — please let CI run `ContainerXPCTests` under Xcode.
- Full `swift build` green; `swift format lint --strict` clean on all touched files.


1PoPTRoN:
Hi @jglogan @katiewasnothere — I'm new around here and this is my first contribution, so please point me at anything I'm doing wrong. This one crashes `container-apiserver` (and every helper) on a single one-way dictionary, so I wanted eyes on it early.

I kept `reply()` non-failable with a fallback to avoid churning all 88 call sites, but I'm still learning the conventions here — happy to rework it however you'd like.

Issue #2159, opened 2026-08-26
[Request]: Add --mount option to container k8s create


janishorsts:
### Feature or enhancement request details

Currently `container k8s create` does not support mounting host directories into the node container, unlike `container create` / `container run`.

Adding `--mount` support allows mounts to the cluster node (useful for sharing data, config or other files directly into the k8s node) same way as `container create` / `container run`.

**Example:**

```bash
container k8s create --name dev --mount type=virtiofs,source=/foo/bar,target=/data,readonly
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

egernst:
What's the typical user story for this?  I'm not sure that k8s create should really map 1:1 what you'd do for a general container -- the fact that a 'node' is run within a container image is almost an implementation detail for most folks.  

I'm wondering if you'd be better suited using a custom node image?

janishorsts:
To provide persistent storage for Persistent Volume Claims in k8s.

Currently, k8s is ephemeral: when deleting the k8s container, the volumes are gone. The `--mount` option ensures these volumes persist.

Without volume persistence, "container k8s" has very narrow use cases on developer machines.

shoce:
I double the request. Having --volume/--mount would allow me to have hostPath volumes on my macbook.

shoce:
@egernst please take a look at my PR https://github.com/apple/container/pull/2283 - it adds both --volume and --mount options in a clean way

jlehman9:
> What's the typical user story for this? I'm not sure that k8s create should really map 1:1 what you'd do for a general container -- the fact that a 'node' is run within a container image is almost an implementation detail for most folks.

I can't use `container` on it's own right now due to this since my team's main Helm Chart requires a postgresql container that uses volumes like this. The [kiac](https://github.com/saiyam1814/kiac) handles this for me so I've been playing around with that on top of `container`, but having it native would be great so I wouldn't need a 3rd party tool. Excited for this PR.

PR #2283, opened 2026-09-17
K8s plugin: Add --volume and --mount options support

shoce:
## Type of Change
- [ ] Bug fix
- [x] New feature  
- [ ] Breaking change
- [ ] Documentation update

## Motivation and Context
To allow bind mounts from host machine into k8s container.

Issue https://github.com/apple/container/issues/2159

## Testing
- [x] Tested locally
- [ ] Added/updated tests
- [x] Added/updated docs


shoce:
Issue https://github.com/apple/container/issues/2159

egernst:
@shoce - thanks for the PR. I haven't reviewed in detail yet, but I have an initial concern of how we would manage this  in a multi-node configuration.

shoce:
> @shoce - thanks for the PR. I haven't reviewed in detail yet, but I have an initial concern of how we would manage this in a multi-node configuration.

sorry, i have no idea. can you please elaborate on what can go wrong in a multi-node configuration? multi-node in our case is still running on the same host machine right?

katiewasnothere:
Hi @shoce I have an open PR at https://github.com/apple/container/pull/2276 for adding multi node support. This PR is stalled right now but can give a sense of what the multi node support will look like. 

For your PR here, in the multi node case, would all of the nodes in the cluster that are created during `k8s create` also have the volume mounted to them?

shoce:
> Hi @shoce I have an open PR at #2276 for adding multi node support. This PR is stalled right now but can give a sense of what the multi node support will look like.

Thanks, i got the idea. That multi-node implementation does not look like something i would like to use - because it is a creation of few containers with only one cli command and need to provide all the configuration via arguments (and env vars possibly). A command to create a worker node and connect it to a specified control plane node would be much more handy to me.

> For your PR here, in the multi node case, would all of the nodes in the cluster that are created during `k8s create` also have the volume mounted to them?

With your current PR i would say yes, like with --cpus, --memory and --node-image options, --volume and --mount options would apply to all of them.

shoce:
@egernst @katiewasnothere 

Also now seeing that you are working on multi-node cluster creation, I am thinking if we could have a command to add/remove volumes, mounts and published ports on existing (stopped) containers? Would it be a more flexible way? 

Like:
- add --start option to `k8s create` which will be off by default
- `container k8s create` without --start
- add volumes, mounts, published ports on the created container(s)
- `container start` for the control plane and worker nodes

What do you think about this?

PR #2290, opened 2026-09-18
Remove k8s start command for now

katiewasnothere:
## Type of Change 
- [x] Breaking change

## Motivation and Context
The `container k8s start` command was added to the k8s subcommand to support starting a previously stopped cluster node. Since cluster nodes show up in the higher level `container` commands such as `container ls` and `container stop`, it made sense to have the ability to additionally start control plane nodes for the k8s clusters with `container k8s start`. However, since adding this subcommand, we've seen a number of issues around restarting (see [1](https://github.com/apple/container/issues/2156)) and we additionally have concerns about what the UX design should be for the k8s plugin moving forward. 

Since this is an experimental feature, this PR will remove `k8s start` for now while we work on improving the plugin overall. This will not only make the code easier to reason about from a developer point of view but additionally gives users less ways to potentially shoot themselves in the foot while we improve the feature.

If a control plane node is stopped in a cluster, in lieu of `k8s start`, to restart it, a user will need to delete the cluster with `container k8s delete` and re-create it with `container k8s create`. 

We're open to feedback on this decision and use cases for `k8s start`. 

## Testing
- [x] Added/updated docs


docs/command-reference.md
egernst: I think this line would be stale (cluster restart)?

egernst (commented):
Just the one stale doc bit I think. 

Suggestion, but not blocker - should we include the "how to recover" description you have in the PR header into docs/kubernetes.md?

docs/command-reference.md
katiewasnothere: Done. Also added a recovery section to the docs.

egernst (approved):

